PR #67, opened 2022-06-03
Additional endpoints

aefhm:
**Why**
Close https://github.com/oasisprotocol/oasis-indexer/issues/46

**Tasks**
- [x] Write new analyzer for querying metadata registry
- [x] get all validators
- [x] add commission schedule
- [x] get all active delegations
- [x] get all debonding delegations
- [x] Potentially fix amounts of delegation shares

**TODO**
- [x] Polish PR
- [x] Squash linty commits
- [x] Fix CI again
- [x] Run migrations locally
- [x] Run indexer locally


analyzer/util/util.go
```diff
@@ -35,3 +38,25 @@ func (b *Backoff) Reset() {
 func (b *Backoff) Timeout() time.Duration {
 	return b.currentTimeout
 }
+
+// CurrentBound returns the bound at the latest rate step that has started or nil if no step has started.
+func CurrentBound(cs staking.CommissionSchedule, now beacon.EpochTime) (currentBound staking.CommissionRateBoundStep, epochEnd uint64) {
```
aefhm, line 43: See https://pkg.go.dev/github.com/oasisprotocol/oasis-core/go@v0.2201.7/staking/api#CommissionSchedule.CurrentRate

storage/migrations/0003_oasis_3_amend_commission.up.sql
```diff
@@ -0,0 +1,164 @@
...
+-- Commissions Amendments
+
+CREATE TABLE IF NOT EXISTS oasis_3.commissions
+(
+  address TEXT PRIMARY KEY NOT NULL REFERENCES oasis_3.accounts(address),
+  schedule JSON
+);
+
+TRUNCATE oasis_3.commissions CASCADE;
+INSERT INTO oasis_3.commissions (address, schedule) VALUES
```
aefhm, line 14: We could also roll these into `0001_oasis_3_state_init.up.sql` and also move the table creation in `0000_oasis_3_init.up.sql`.

api/v1/client.go
```diff
@@ -845,6 +849,173 @@ func (c *storageClient) Account(ctx context.Context, r *http.Request) (*Account,
...
+	chainID, ok := ctx.Value(ChainIDContextKey).(string)
+	if !ok {
+		return nil, common.ErrBadChainID
+	}
+
+	qb := NewQueryBuilder(fmt.Sprintf(`
+			SELECT delegatee, shares, debond_end, escrow_balance_debonding, escrow_total_shares_debonding
+				FROM %s.debonding_delegations
+				JOIN %s.accounts ON %s.debonding_delegations.delegatee = %s.accounts.address
+				`,
```
ennsharma, line 947:
```suggestion
			SELECT delegatee, shares, debond_end, escrow_balance_debonding, escrow_total_shares_debonding
				FROM %s.debonding_delegations
				WHERE delegator = $1::text;
				`,
```
likewise here, and remove the `filters` below.

api/spec/v1.yaml
```diff
@@ -443,6 +461,46 @@ paths:
...
+    get:
+      summary: Returns an account's delegations.
+      parameters:
+        - in: path
+          name: address
+          required: true
+          schema:
+            type: string
+          description: The staking address of the account that delegated.
+          example: *staking_address_1
```
ennsharma, line 474: This shouldn't need this parameter? It's already in the path?

api/v1/client.go
```diff
@@ -845,6 +849,173 @@ func (c *storageClient) Account(ctx context.Context, r *http.Request) (*Account,
...
+	if !ok {
+		return nil, common.ErrBadChainID
+	}
+
+	qb := NewQueryBuilder(fmt.Sprintf(`
+			SELECT delegatee, shares, escrow_balance_active, escrow_total_shares_active
+				FROM %s.delegations
+				JOIN %s.accounts ON %s.delegations.delegatee = %s.accounts.address
+			`,
+		chainID, chainID, chainID, chainID), c.db)
```
ennsharma, line 864: Can you just do this? Why's do we need to `JOIN` if we're not using any columns from `accounts`?
```suggestion
			SELECT delegatee, shares, escrow_balance_active, escrow_total_shares_active
				FROM %s.delegations
				WHERE delegator = $1::text;
```

and then pass in `chi.URLParam(r, "address")` to the `db.Query` call.

api/v1/client.go
```diff
@@ -1123,3 +1294,200 @@ func (c *storageClient) ProposalVotes(ctx context.Context, r *http.Request) (*Pr
...
+				%s.accounts.escrow_balance_active AS escrow,
+				%s.commissions.schedule AS commissions_schedule,
+				case when exists(select null from %s.nodes where %s.entities.id = %s.nodes.entity_id AND voting_power > 0) then true else false end as active,
+				case when exists(select null from %s.nodes where %s.entities.id = %s.nodes.entity_id AND %s.nodes.roles like 'validator') then true else false end as status,
+				meta
+			FROM %s.entities
+			JOIN %s.accounts ON %s.entities.address = %s.accounts.address
+			JOIN %s.metadata ON %s.entities.id = %s.metadata.id
+			JOIN %s.commissions ON %s.entities.address = %s.commissions.address
+			JOIN %s.nodes ON %s.entities.id = %s.nodes.entity_id
```
ennsharma, line 1337: This is too many joins. I think the `metadata` one will go away if you take my below comment to roll that into the entities table. Overall I think the query can look like:

```
SELECT id, entity_id FROM nodes WHERE roles LIKE '%validator%';
SELECT address, metadata, voting_power > 0 AS ACTIVE FROM entities WHERE id = <entity_id_from_prev_query>;
SELECT schedule FROM commissions WHERE address = <address_from_prev_query>;
```
This will be more efficient than joining everything.

ennsharma: And something similar below for `ValidatorsList`.

analyzer/consensus/consensus.go
```diff
@@ -248,6 +253,23 @@ func (m *Main) queueBlockInserts(batch *storage.QueryBatch, data *storage.BlockD
...
 
+func (m *Main) queueEpochInserts(batch *storage.QueryBatch, data *storage.BlockData) error {
+	chainID := m.rangeCfg.ChainID
+
+	batch.Queue(fmt.Sprintf(`
+		INSERT INTO %s.epochs (id, start_height)
+			VALUES ($1, $2)
+		ON CONFLICT (id) DO
+		UPDATE SET
+			end_height = excluded.start_height;
```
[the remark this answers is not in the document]
aefhm: I see. The heuristic of suggesting that the `epoch` ends at current height probably has some complications.

analyzer/consensus/consensus.go
```diff
@@ -310,6 +332,28 @@ func (m *Main) queueEventInserts(batch *storage.QueryBatch, data *storage.BlockD
 		}
 	}
 
+	for _, signed := range data.Transactions {
```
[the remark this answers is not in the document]
aefhm: I was hoping that `core` would add that event soon 🙏 , but yeah, the processing should probably happen bundled with the other one.

api/spec/v1.yaml
```diff
@@ -598,6 +656,66 @@ components:
...
+      properties: 
+        amount:
+          type: number
+          format: float
+          description: The amount of tokens delegated.
+        shares:
+          type: integer
+          format: int64
+          description: The shares of tokens delegated.
+        validatorAddress:
```
aefhm, line 670: Address the casing differences in this spec.

analyzer/util/util.go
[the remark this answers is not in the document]
aefhm: Yup. I had a bug.

api/spec/v1.yaml
```diff
@@ -598,6 +656,66 @@ components:
       description: |
         A consensus block.
 
+    Delegation:
+      type: object
+      properties: 
+        amount:
+          type: number
+          format: float
```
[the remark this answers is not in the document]
aefhm: Gotcha. I see. Ahhh, the frontend converts base units to `10^6`.

api/v1/client.go
aefhm: Yeah, that is better.

api/spec/v1.yaml
aefhm: I think we do need to describe the `address`?

api/v1/client.go
aefhm: >  I think the metadata one will go away if you take my below comment to roll that into the entities table.

Agree.

> This will be more efficient than joining everything.

Preference for three queries? I don't have numbers, but thought default `JOIN` would be faster.

api/spec/v1.yaml
ennsharma: Oops I misread, I thought this was an additional query param. My mistake.

aefhm: Haha, I actually had one that was redundant earlier in some draft.

api/v1/client.go
```diff
@@ -845,6 +849,137 @@ func (c *storageClient) Account(ctx context.Context, r *http.Request) (*Account,
...
+func (c *storageClient) Delegations(ctx context.Context, r *http.Request) (*DelegationList, error) {
+	chainID, ok := ctx.Value(ChainIDContextKey).(string)
+	if !ok {
+		return nil, common.ErrBadChainID
+	}
+
+	qb := NewQueryBuilder(fmt.Sprintf(`
+			SELECT delegatee, shares, escrow_balance_active, escrow_total_shares_active
+				FROM %s.delegations
+				WHERE delegator = $1::text;
```
ennsharma, line 862: Ah. I think you might need the `JOIN` after all. `escrow_balance_active` and `escrow_total_shares_active` aren't fields in delegations. My mistake.

api/v1/client.go
```diff
@@ -845,6 +849,137 @@ func (c *storageClient) Account(ctx context.Context, r *http.Request) (*Account,
...
+
+// DebondingDelegations returns a list of debonding delegations.
+func (c *storageClient) DebondingDelegations(ctx context.Context, r *http.Request) (*DebondingDelegationList, error) {
+	chainID, ok := ctx.Value(ChainIDContextKey).(string)
+	if !ok {
+		return nil, common.ErrBadChainID
+	}
+
+	qb := NewQueryBuilder(fmt.Sprintf(`
+			SELECT delegatee, shares, debond_end, escrow_balance_debonding, escrow_total_shares_debonding
```
ennsharma, line 926: Same here, I think you do need a JOIN

ennsharma (approved):
Approving pending comments

Issue #43, opened 2022-05-10
Test Suite for API validation


ennsharma:
##### SUMMARY
We would like to have something that validates that the spec(s) described in `api/spec` actually match what our API supports. Some validation tests using something like [dredd](https://dredd.org/en/latest/) would be desirable.

##### ISSUE TYPE
- Feature Idea

##### COMPONENT NAME
`api/`

ennsharma:
This was done in #73

PR #73, opened 2022-06-16
tests: End-to-end tests

### LINKED ISSUE #5: Sanity Test Suite
**What is the problem that you're trying to solve?**
We would like high-reliability and fast and simple error detection.

**Describe the solution you'd like**
We'll implement the following sanity checks

1. COUNT of new blocks added to the indexed blocks database > {threshold} for each 10 minute window. {threshold} should be less than but in the ballpark of 100 based on the expected 6 second block time = 100 blocks per 10 minutes.
2. SUM of all block heights added to the indexed blocks database is the [triangular number](https://en.wikipedia.org/wiki/Triangular_number) indexed by the COUNT of blocks. This should provide an easy validation of no missed blocks.
3. Measured state to genesis, generated by [`StateToGenesis`](https://github.com/oasisprotocol/oasis-core/blob/9865f5a1cc93/go/consensus/api/api.go#L56) doesn't diverge from the current indexer state. This is a very slow operation, but it will provide an accurate picture directly from the chain of the expected state of most things that we are tracking.

ennsharma:
Tests that cover end-to-end state correctness. Will resolve #5.

**Tests**
- [x] API Validation tests
  - [x] Status
  - [x] Blocks, transactions, events tests
  - [x] Registry Backend
  - [x] Staking Backend
  - [x] Governance Backend

- [x] Checkpoint test against `StateToGenesis`
  - [x] Registry Backend
  - [x] Staking Backend
  - [x] Governance Backend

- [x] Bonus: Add support for cron analyzers

**Not in scope**
- OpenAPI validation will be handled subsequently. That is, #43 
- Having these tests run in CI. This requires `oasis-net-runner` and some additional infrastructure, which will be handled in a follow up.


storage/oasis/client.go
```diff
@@ -62,6 +70,61 @@ func (c *Client) GenesisDocument(ctx context.Context) (*genesisAPI.Document, err
 	return doc, nil
 }
 
+// GenesisDocumentAtHeight returns the genesis document at the provided height.
```
aefhm, line 73: Is this technically the "document?"

storage/migrations/0003_oasis_3_entity_nodes.up.sql
```diff
@@ -0,0 +1,15 @@
+-- Entity Node relationship migration.
+--
+-- Entities are free to claim any node, and nodes are free to associate
+-- themselves with any entity. So this relationship may not necessarily
+-- be perfectly associated as is currently assumed.
+
+ALTER TABLE oasis_3.nodes DROP CONSTRAINT nodes_entity_id_fkey;
```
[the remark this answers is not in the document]
ennsharma: Squash it is!

tests/rpc/v1/staking_test.go
```diff
@@ -0,0 +1,71 @@
...
+
+	<-tests.After(stakingEndHeight)
+
+	var list v1.AccountList
+	err := tests.GetFrom("/consensus/accounts?minAvailable=1000000000000000000", &list)
+	require.Nil(t, err)
+	require.Equal(t, 1, len(list.Accounts))
+
+	// The big kahuna (Binance Staking).
+	require.Equal(t, "oasis1qpg2xuz46g53737343r20yxeddhlvc2ldqsjh70p", list.Accounts[0].Address)
```
aefhm, line 52: Hmm. This is the same init/seed file, yeah? But below we are "mocking" the accounts below?

tests/rpc/v1/transactions_test.go
```diff
@@ -0,0 +1,194 @@
...
+		"2022-04-11T11:00:08Z",
+		"2022-04-11T11:00:27Z",
+		"2022-04-11T11:00:33Z",
+		"2022-04-11T11:00:39Z",
+		"2022-04-11T11:00:45Z",
+	}
+	require.Equal(t, len(blocks), len(timestamps))
+
+	for i := 0; i < len(blocks); i++ {
+		timestamp, err := time.Parse(time.RFC3339, timestamps[i])
```
aefhm, line 54: I guess `timestamp` is not super readable?

tests/rpc/v1/transactions_test.go
```diff
@@ -0,0 +1,194 @@
...
+	endHeight := tests.GenesisHeight + int64(len(testTransactions)-1)
+	<-tests.After(endHeight)
+
+	var list v1.TransactionList
+	err := tests.GetFrom(fmt.Sprintf("/consensus/transactions?block=%d", testTransactions[0].Height), &list)
+	require.Nil(t, err)
+	require.Equal(t, len(testTransactions), len(list.Transactions))
+
+	// Get a consistent ordering.
+	sort.Slice(list.Transactions, func(i, j int) bool {
```
aefhm, line 167: (question/non-blocking) The setup code for transactions already sorts?

analyzer/consensus/consensus.go
```diff
@@ -281,13 +281,22 @@ func (m *Main) queueTransactionInserts(batch *storage.QueryBatch, data *storage.
...
 			result.Error.Message,
 		)
+		batch.Queue(fmt.Sprintf(`
+		UPDATE %s.accounts
+		SET
+			nonce = $2
+		WHERE address = $1;
+	`, chainID),
+			sender,
+			tx.Nonce+1,
```
aefhm, line 291: I'm not that booked up on our nonces. Why `+1`?

analyzer/consensus/consensus.go
```diff
@@ -432,22 +476,31 @@ func (m *Main) queueNodeRegistrations(batch *storage.QueryBatch, data *storage.R
...
+				strings.Join(consensusAddresses, ","),
+				vrfPubkey,
+				nodeEvent.Node.Roles,
+				nodeEvent.Node.SoftwareVersion,
+				0,
+			)
+		} else {
+			// An existing node is expired.
+			batch.Queue(fmt.Sprintf(`
+				DELETE FROM %s.nodes WHERE id = $1;
```
aefhm, line 497: We don't want to keep this history (permanently forward)?

analyzer/consensus/consensus.go
```diff
@@ -572,64 +628,90 @@ func (m *Main) queueEscrows(batch *storage.QueryBatch, data *storage.StakingData
...
+						escrow_total_shares_debonding = escrow_total_shares_debonding - $3
+					WHERE address = $1;
 			`, chainID),
 				e.Reclaim.Escrow.String(),
 				e.Reclaim.Amount.ToBigInt().Uint64(),
 				e.Reclaim.Shares.ToBigInt().Uint64(),
 			)
+
+			// TODO: Delete row from `debonding_delegations` that corresponds with
+			// the reclaimed escrow. The reclaim occurs on epoch transition, so
```
aefhm, line 682: Do we want to delete or soft delete?

storage/oasis/client_test.go
```diff
@@ -0,0 +1,209 @@
...
+	require.Nil(t, err)
+
+	_, err = client.BlockData(ctx, pastHeight)
+	require.Nil(t, err)
+
+	_, err = client.BlockData(ctx, futureHeight)
+	require.NotNil(t, err)
+}
+
+func TestBeaconData(t *testing.T) {
```
aefhm, line 102: I think we are not using the `BeaconData`? Test doesn't hurt though.

tests/genesis/genesis_test.go
```diff
@@ -0,0 +1,737 @@
...
+
+	var actualHeightSum int64
+	err = postgresClient.QueryRow(ctx, fmt.Sprintf(
+		`SELECT SUM(height) FROM %s.blocks WHERE height <= $1;`,
+		chainID,
+	), latestHeight).Scan(&actualHeightSum)
+	require.Nil(t, err)
+
+	// Using formula for sum of first k natural numbers.
+	expectedHeightSum := latestHeight*(latestHeight+1)/2 - (tests.GenesisHeight-1)*tests.GenesisHeight/2
```
aefhm, line 191: K. Fine. :)

tests/genesis/genesis_test.go
```diff
@@ -0,0 +1,737 @@
...
+	"github.com/stretchr/testify/require"
+
+	"github.com/oasislabs/oasis-indexer/log"
+	"github.com/oasislabs/oasis-indexer/storage"
+	"github.com/oasislabs/oasis-indexer/storage/oasis"
+	"github.com/oasislabs/oasis-indexer/storage/postgres"
+	"github.com/oasislabs/oasis-indexer/tests"
+)
+
+type TestEntity struct {
```
aefhm, line 28: We are not able to re-use the API types?

aefhm (approved):
Mostly questions, nothing stood out to be blocking.

storage/oasis/client.go
```diff
@@ -62,6 +70,61 @@ func (c *Client) GenesisDocument(ctx context.Context) (*genesisAPI.Document, err
...
+	connection := *c.connection
+	doc, err := connection.Consensus().StateToGenesis(ctx, height)
+	if err != nil {
+		return nil, err
+	}
+
+	return doc, nil
+}
+
+// RegistryGenesis returns the registry genesis document at the provided height.
```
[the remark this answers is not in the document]
ennsharma: Yes 😅

tests/genesis/genesis_test.go
ennsharma: Technically we can. But we don't always want to validate all data, so I prefer having separate test types sometimes. And makes it easier to mock things when needed in a separate package rather than put it in the non-test one. But certainly a stylistic choice.

PR #75, opened 2022-06-27
Fix doc building command

aefhm:
**Why**

Fix doc building.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -54,7 +54,7 @@ lint: $(lint-targets)
 docs-targets := docs-api
 
 docs-api:
-	@npx --no redoc-cli build api/spec/v1.yaml -o index.html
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -54,7 +54,7 @@ lint: $(lint-targets)
 docs-targets := docs-api
 
 docs-api:
-	@npx --no redoc-cli build api/spec/v1.yaml -o index.html
```
aefhm, line 57: Unless we meant `--no-install`?

ennsharma: I may have 😅

ennsharma (approved):

Makefile
aefhm: Oh, I'm on an older version of `npx`.

aefhm: Hmm, I think we still need the install.

Issue #86, opened 2022-07-12
API instability


lukaw3d:
debonding_delegations throws storage error for some accounts

Largest validators:
- https://index.oasislabs.com/v1/consensus/accounts/oasis1qqekv2ymgzmd8j2s2u7g0hhc7e77e654kvwqtjwm/debonding_delegations
`{"msg":"internal storage error"}`

- https://index.oasislabs.com/v1/consensus/accounts/oasis1qq3xrq0urs8qcffhvmhfhz4p0mu7ewc8rscnlwxe/debonding_delegations
{"debonding_delegations":null} :heavy_check_mark: 

- https://index.oasislabs.com/v1/consensus/accounts/oasis1qpn83e8hm3gdhvpfv66xj3qsetkj3ulmkugmmxn3/debonding_delegations
{"debonding_delegations":null} :heavy_check_mark: 

- https://index.oasislabs.com/v1/consensus/accounts/oasis1qz0ea28d8p4xk8xztems60wq22f9pm2yyyd82tmt/debonding_delegations
`{"msg":"internal storage error"}`

- https://index.oasislabs.com/v1/consensus/accounts/oasis1qzv2rktycpcykhudyvg5l6u75v08lpfcw5nt7aj5/debonding_delegations
`{"msg":"internal storage error"}`

lukaw3d:
Hm, also sometimes in delegations:
https://index.oasislabs.com/v1/consensus/accounts/oasis1qz78ap0456g2rk7j6rmtvasc9v2kjhz2s58qgj90/delegations
`{"msg":"internal storage error"}`

aefhm:
https://index.oasislabs.com/v1/consensus/epochs is also erroring

lukaw3d:
https://index.oasislabs.com/v1/consensus/transactions?sender=oasis1qz78ap0456g2rk7j6rmtvasc9v2kjhz2s58qgj90
is timeouting  Error 524

aefhm:
Technically, we will need a redeploy and a re-index to clean up bad data. That should be batched with other changes.

aefhm:
Redeployed and re-indexed. The above endpoints work for me.

lukaw3d:
Seems like lots broke again now that indexer caught up :/

aefhm:
[...]

PR #89, opened 2022-07-12
Present empty arrays

aefhm:
Relates to #86


aefhm:
Unless we prefer empty array `{"delegations":[]}`

lukaw3d:
I'd prefer empty array. I'd only prefer omitting or null when information is unknown

aefhm:
> I'd prefer empty array. I'd only prefer omitting or null when information is unknown

Kewl. I prefer explicit empty as well.

ennsharma (approved):

PR #94, opened 2022-07-19
Fix negative delegation shares

aefhm:
**Why**
Fix https://github.com/oasisprotocol/oasis-indexer/issues/86

**What**
I math-ed wrong and subtracted too many shares including the number of accrued shares in a `ReclaimEscrowEvent`.


aefhm:
Also sneak in small API response fix.

aefhm:
Hmm, this seems pretty hard to test right now. Merging first.

Issue #92, opened 2022-07-19
Incorrect results for the validators query


Yawning:
`wget https://index.oasislabs.com/v1/consensus/validators\?height\=9491840\&limit\=200` returns a 115 element array.  There are more than 115 validators.

Without the limit argument, 100 validators are returned (because that's not annoying at all), which also does not correspond to the top 100 validators.

Being able to query the current validator set is presumably a useful and common operation, so this query not returning results that reflect reality, is probably a big problem.

Note: It occurs to me that this may partially have to do with how oasisscan displays this information vs the indexer (I assume oasisscan's data is stale), but since I am getting truncated data I am inclined to believe that something is also badly busted.

aefhm:
`?height` param is not available yet...

Currently (post-deploy), the default query set at max limit of `1000` returns all known entities with a `validator` roled `node` which at (still processing) height of `8328899` is `133` of which `120` are `active`.

Yawning:
> ?height param is not available yet...

:(

aefhm:
Closing. `height` param goes into https://github.com/oasisprotocol/oasis-indexer/issues/116

PR #96, opened 2022-07-20
Fix Validator query

aefhm:
**Why**
Relates to #92. See original issue https://github.com/oasisprotocol/oasis-indexer/issues/46 and reference https://github.com/oasisprotocol/oasis-indexer/issues/46#issuecomment-1129251832

**Expanded**
My assumption is that we want to make a single query that returns `active` and `inactive` validators aka `entities` based on their node voting power.

**TODO**
- [x] Remove pagination
- [x] Order by voting power

### DIFF
--- a/api/v1/client.go
+++ b/api/v1/client.go
@@ -1464,22 +1464,6 @@ func (c *storageClient) Validators(ctx context.Context, r *http.Request) (*Valid
 		}
 	}
 
-	pagination, err := common.NewPagination(r)
-	if err != nil {
-		c.logger.Info("pagination failed",
-			"request_id", ctx.Value(RequestIDContextKey),
-			"err", err.Error(),
-		)
-		return nil, common.ErrBadRequest
-	}
-	if err = qb.AddPagination(ctx, pagination); err != nil {
-		c.logger.Info("pagination add failed",
-			"request_id", ctx.Value(RequestIDContextKey),
-			"err", err.Error(),
-		)
-		return nil, common.ErrBadRequest
-	}
-
 	rows, err := c.db.Query(ctx, qb.String())
 	if err != nil {
 		c.logger.Info("query failed",
--- a/api/v1/client.go
+++ b/api/v1/client.go
@@ -1449,6 +1449,7 @@ func (c *storageClient) Validators(ctx context.Context, r *http.Request) (*Valid
 			WHERE %s.entities.id = %s.nodes.entity_id
 				AND %s.nodes.roles like '%%validator%%'
 		)
+	ORDER BY voting_power DESC
 	`, chainID, chainID, chainID, chainID, chainID, chainID, chainID, chainID, chainID, chainID, chainID, chainID, chainID, chainID, chainID, chainID, chainID, chainID, chainID, chainID, chainID, chainID, chainID, chainID, chainID, chainID, chainID, chainID), c.db)
 
 	params := r.URL.Query()
--- a/api/v1/client.go
+++ b/api/v1/client.go
@@ -1347,8 +1347,8 @@ func (c *storageClient) Validator(ctx context.Context, r *http.Request) (*Valida
 				%s.nodes.id AS node_address,
 				%s.accounts.escrow_balance_active AS escrow,
 				%s.commissions.schedule AS commissions_schedule,
-				case when exists(select null from %s.nodes where %s.entities.id = %s.nodes.entity_id AND voting_power > 0) then true else false end as active,
-				case when exists(select null from %s.nodes where %s.entities.id = %s.nodes.entity_id AND %s.nodes.roles like '%%validator%%') then true else false end as status,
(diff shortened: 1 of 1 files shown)


api/v1/client.go
```diff
@@ -1464,22 +1465,6 @@ func (c *storageClient) Validators(ctx context.Context, r *http.Request) (*Valid
...
 
-	pagination, err := common.NewPagination(r)
-	if err != nil {
-		c.logger.Info("pagination failed",
-			"request_id", ctx.Value(RequestIDContextKey),
-			"err", err.Error(),
-		)
-		return nil, common.ErrBadRequest
-	}
-	if err = qb.AddPagination(ctx, pagination); err != nil {
```
aefhm, line 1475: Maybe we still want the max limit at `1000`.

api/v1/client.go
```diff
@@ -1437,19 +1438,21 @@ func (c *storageClient) Validators(ctx context.Context, r *http.Request) (*Valid
...
+		AND %s.nodes.roles like '%%validator%%'
 		AND %s.nodes.voting_power = (
 			SELECT max(voting_power)
 			FROM %s.nodes
 			WHERE %s.entities.id = %s.nodes.entity_id
-				AND %s.nodes.roles like 'validator'
+				AND %s.nodes.roles like '%%validator%%'
 		)
-	`, chainID, chainID, chainID, chainID, chainID, chainID, chainID, chainID, chainID, chainID, chainID, chainID, chainID, chainID, chainID, chainID, chainID, chainID, chainID, chainID, chainID, chainID, chainID, chainID, chainID, chainID, chainID, chainID), c.db)
+	ORDER BY voting_power DESC
```
aefhm, line 1454: This is different from `escrow`.

aefhm: I guess either ordering should work because this is really only problematic for the ones at the bottom where `voting_power == 0`.

api/v1/client.go
```diff
@@ -1348,20 +1348,21 @@ func (c *storageClient) Validator(ctx context.Context, r *http.Request) (*Valida
...
 				case when exists(select null from %s.nodes where %s.entities.id = %s.nodes.entity_id AND voting_power > 0) then true else false end as active,
-				case when exists(select null from %s.nodes where %s.entities.id = %s.nodes.entity_id AND %s.nodes.roles like 'validator') then true else false end as status,
+				case when exists(select null from %s.nodes where %s.entities.id = %s.nodes.entity_id AND %s.nodes.roles like '%%validator%%') then true else false end as status,
 				%s.entities.meta AS meta
 			FROM %s.entities
 			JOIN %s.accounts ON %s.entities.address = %s.accounts.address
-			JOIN %s.commissions ON %s.entities.address = %s.commissions.address
-			JOIN %s.nodes ON %s.entities.id = %s.nodes.entity_id
+			LEFT JOIN %s.commissions ON %s.entities.address = %s.commissions.address
+			LEFT JOIN %s.nodes ON %s.entities.id = %s.nodes.entity_id
```
aefhm, line 1356: Need a default then.

aefhm: Probably fine to avoid all `entities` without a `validator` roled `node`.

aefhm:
I'm still thinking of this endpoint as the relevant one for the interface of our frontend, so pertaining to standup discussion, this is `indexer` + `registry`.

@lukaw3d

api/v1/client.go
```diff
@@ -1348,20 +1348,21 @@ func (c *storageClient) Validator(ctx context.Context, r *http.Request) (*Valida
 				%s.accounts.escrow_balance_active AS escrow,
 				%s.commissions.schedule AS commissions_schedule,
 				case when exists(select null from %s.nodes where %s.entities.id = %s.nodes.entity_id AND voting_power > 0) then true else false end as active,
-				case when exists(select null from %s.nodes where %s.entities.id = %s.nodes.entity_id AND %s.nodes.roles like 'validator') then true else false end as status,
+				case when exists(select null from %s.nodes where %s.entities.id = %s.nodes.entity_id AND %s.nodes.roles like '%%validator%%') then true else false end as status,
```
ennsharma, line 1351:
```suggestion
				CASE WHEN EXISTS(SELECT null FROM %s.nodes WHERE %s.entities.id = %s.nodes.entity_id AND %s.nodes.roles LIKE '%%validator%%') THEN true ELSE false END AS status,
```
Style-wise, we should be capitalizing all SQL keywords for readability. Sorry for not pointing this out earlier in the previous review.

ennsharma: Also, why do you need `%%validator%%`? Just `%validator%` should work to get anything with `validator` in any position within the string.

ennsharma (approved):

api/v1/client.go
aefhm: Ah, needed to escape the Golang parsing.

api/v1/client.go
aefhm: Went for it.

PR #97, opened 2022-07-21
storage: Add index on tx sender

ennsharma:
Hitting the `/consensus/transactions?sender=<address>` endpoint is unusably slow without this.


aefhm (approved):

storage/migrations/0006_oasis_3_tx_sender_ix.sql
```diff
@@ -0,0 +1,3 @@
+-- Creates an index on transaction sender. This is a very common
+-- query, and is unusably slow otherwise.
+CREATE INDEX ix_transactions_sender ON oasis_3.transactions (sender);
```
aefhm, line 3: (non-blocking) Do we want a foreign key constraint on this as well?

Issue #87, opened 2022-07-12
delegation amounts are weird


lukaw3d:
e.g. see amounts for validator that is self-delegation

https://index.oasislabs.com/v1/consensus/accounts/oasis1qqekv2ymgzmd8j2s2u7g0hhc7e77e654kvwqtjwm/delegations
```js
{
  "delegations": [
    {
      "amount": 8,
      "shares": 323597002076695,
      "address": "oasis1qqekv2ymgzmd8j2s2u7g0hhc7e77e654kvwqtjwm"
    }
  ]
}
```

https://api.oasisscan.com/mainnet/chain/account/delegations?address=oasis1qqekv2ymgzmd8j2s2u7g0hhc7e77e654kvwqtjwm
```js
    "list": [
      {
        "validatorAddress": "oasis1qqekv2ymgzmd8j2s2u7g0hhc7e77e654kvwqtjwm",
        "validatorName": "BinanceStaking",
        "icon": null,
        "entityAddress": null,
        "shares": "842488.881661508",
        "amount": "974852.348714934",
        "active": true
      }
    ],
```

RPC
```js
const oasisClient = new oasis.client.NodeInternal('https://grpc.oasis.dev')
await oasisClient.stakingDelegationInfosFor({
  owner: oasis.staking.addressFromBech32('oasis1qqekv2ymgzmd8j2s2u7g0hhc7e77e654kvwqtjwm'),
  height: 0,
})
```
```js
{
  "Map": [
    [
      "oasis1qqekv2ymgzmd8j2s2u7g0hhc7e77e654kvwqtjwm",
      {
        "pool": {
          "balance": "553408786.375757478",
          "total_shares": "478268068.133569286"
        },
        "shares": "842488.881661508"
        // amount = shares * pool.balance / pool.total_shares = 974852.3487149344
      }
    ]
  ]
}
```

aefhm:
I should double check after https://github.com/oasisprotocol/oasis-indexer/pull/94 deploy if incorrect deductions were the cause

aefhm:
Nope. I need to fix this.

aefhm:
Amount looks wrong too.

aefhm:
Ah, shares are right. Amount is just overflowing.

Scan
```json
{
  "code": 0,
  "data": {
    "list": [
      {
        "validatorAddress": "oasis1qqekv2ymgzmd8j2s2u7g0hhc7e77e654kvwqtjwm",
[...]

Issue #84, opened 2022-07-08
Account info is missing delegating amount and returns weird total


lukaw3d:
- [ ] There are no fields for how much this account is delegating / debonding (just how much is delegated to it = "escrow")
- [x] Total is weird. Oasisscan's calculation makes more sense
  - (maybe just omit the total and let API consumers figure out what they want for total)

https://index.oasislabs.com/v1/consensus/accounts/oasis1qqekv2ymgzmd8j2s2u7g0hhc7e77e654kvwqtjwm :

```js
  "address": "oasis1qqekv2ymgzmd8j2s2u7g0hhc7e77e654kvwqtjwm",
  "nonce": 19,
  "available": 7298938395306,
  "escrow": 551969325439709250, // validator specific (delegated to this account)
  "debonding": 5311111413978429, // validator
  "total": 557287735792083000, // = available (account) + escrow (validator) + debonding (validator)
```

https://api.oasisscan.com/mainnet/chain/account/info/oasis1qqekv2ymgzmd8j2s2u7g0hhc7e77e654kvwqtjwm :

```js
  "available": "7298.938395306",
  "escrow": "905262.706688650", // bad name
  "debonding": "0",
  "total": "912561.645083956", // = available (account) + escrow (account) + debonding (account)
```

https://monitor.oasis.dev/data/accounts/oasis1qqekv2ymgzmd8j2s2u7g0hhc7e77e654kvwqtjwm :

```js
  "liquid_balance": 7298938395306,
  "escrow_balance": 551969325439709250, // validator
  "escrow_debonding_balance": 5311111413978429, // validator
  "delegations_balance": 905262706688650,
  "debonding_delegations_balance": 0,
  "total_balance": 551976624378104600, // weird https://github.com/everstake/oasis-explorer/blob/328dbfa/services/accounts.go#L119

  // oasisindexer total = liquid_balance (account) + escrow_balance (validator) + escrow_debonding_balance (validator) 
  //    = 557287735792083000
  // oasisscan total = liquid_balance (account) + delegations_balance (account) + debonding_delegations_balance (account)
  //    = 912561645083956
```

aefhm:
> (maybe just omit the total and let API consumers figure out what they want for total)

I prefer this the most.

cc @ennsharma

aefhm:
Oops. Sorry, not deployed. 👍

aefhm:
[...]

PR #103, opened 2022-07-28
Fix debonding delegations

aefhm:
**Why**
Relates to #86, #84, #87

**TODO**
- [x] Use `coalesce` to determine `debonding_delegation` amount
- [x] Delete `debonding_delegations` on `reclaim`
- [x] Handle two debonding delegations in 1 epoch

### DIFF
--- a/analyzer/consensus/consensus.go
+++ b/analyzer/consensus/consensus.go
@@ -715,11 +715,12 @@ func (m *Main) queueEscrows(batch *storage.QueryBatch, data *storage.StakingData
 				e.Take.Amount.ToBigInt().Uint64(),
 			)
 		case e.DebondingStart != nil:
+			// TODO: drop this column, escrow_balance_debonding is not accurate
 			batch.Queue(fmt.Sprintf(`
 				UPDATE %s.accounts
 					SET
 						escrow_balance_active = escrow_balance_active - $2,
-						escrow_balance_debonding = escrow_balance_debonding + $2
+						escrow_balance_debonding = escrow_balance_debonding + $2,
 						escrow_total_shares_debonding = escrow_total_shares_debonding + $2
 					WHERE address = $1;
 			`, chainID),
--- a/analyzer/consensus/consensus.go
+++ b/analyzer/consensus/consensus.go
@@ -708,14 +708,15 @@ func (m *Main) queueEscrows(batch *storage.QueryBatch, data *storage.StakingData
 		case e.Take != nil:
 			batch.Queue(fmt.Sprintf(`
 				UPDATE %s.accounts
-				SET escrow_balance_active = escrow_balance_active - $2
+					SET
+						escrow_balance_active = escrow_balance_active - ROUND($2 * escrow_balance_active  / (escrow_balance_active + escrow_balance_debonding)),
+						escrow_balance_debonding = escrow_balance_debonding - ROUND($2 * escrow_balance_debonding  / (escrow_balance_active + escrow_balance_debonding))
 					WHERE address = $1;
 			`, chainID),
 				e.Take.Owner.String(),
 				e.Take.Amount.ToBigInt().Uint64(),
 			)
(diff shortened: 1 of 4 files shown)


analyzer/consensus/consensus.go
```diff
@@ -720,10 +720,11 @@ func (m *Main) queueEscrows(batch *storage.QueryBatch, data *storage.StakingData
 					SET
 						escrow_balance_active = escrow_balance_active - $2,
 						escrow_balance_debonding = escrow_balance_debonding + $2
+						escrow_total_shares_debonding = escrow_total_shares_debonding + $2
 					WHERE address = $1;
 			`, chainID),
 				e.DebondingStart.Escrow.String(),
-				e.DebondingStart.DebondingShares.ToBigInt().Uint64(),
```
aefhm, line 726: We want the number of shares to be [reclaimed](https://github.com/oasisprotocol/oasis-core/blob/2b35f69b5a8fa3d9b8e1dd19123d4e3da72b5e38/go/consensus/tendermint/apps/staking/transactions.go#L475).

analyzer/consensus/consensus.go
```diff
@@ -715,15 +715,17 @@ func (m *Main) queueEscrows(batch *storage.QueryBatch, data *storage.StakingData
 				e.Take.Amount.ToBigInt().Uint64(),
 			)
 		case e.DebondingStart != nil:
+			// TODO: drop this column, escrow_balance_debonding is not accurate
```
aefhm, line 718: Sighs. I need to fix this to proportionally account for slashed amount.

aefhm: Fixed.

api/v1/types.go
```diff
@@ -79,11 +79,13 @@ type AccountList struct {
...
-	Available uint64 `json:"available"`
-	Escrow    uint64 `json:"escrow"`
-	Debonding uint64 `json:"debonding"`
+	Address                     string `json:"address"`
+	Nonce                       uint64 `json:"nonce"`
+	Available                   uint64 `json:"available"`
+	Escrow                      uint64 `json:"escrow"`
+	Debonding                   uint64 `json:"debonding"`
+	DelegationsBalance          uint64 `json:"delegations_balance,omitempty"`
+	DebondingDelegationsBalance uint64 `json:"debonding_delegations_balance,omitempty"`
```
ennsharma, line 88: I don't fully understand why these are different from `Escrow` and `Debonding` fields rather than just replacing them?

ennsharma (approved):
Looks good! Nice finds :)

api/v1/client.go
```diff
@@ -1016,7 +1030,7 @@ func (c *storageClient) DebondingDelegations(ctx context.Context, r *http.Reques
 			)
 			return nil, common.ErrStorageError
 		}
-		d.Amount = d.Shares * escrowBalanceDebonding / escrowTotalSharesDebonding
+		d.Amount = uint64(float64(d.Shares) * float64(escrowBalanceDebonding) / float64(escrowTotalSharesDebonding))
```
lukaw3d, line 1033: This does not look precise :scream:

aefhm: Would you feel better knowing they are in base units? 😅

api/v1/types.go
aefhm: Hmm, I think we can remove them if we can expect `validator` information exclusively from the `/validators` endpoint?

Currently, `Escrow` returns active escrow pool balance, and `Debonding` returns debonding pool balance. `DelegationsBalance` and `DebondingDelegationsBalance` return the delegated estimates.

api/v1/client.go
lukaw3d: Not really

aefhm: Alright, let me get that in https://github.com/oasisprotocol/oasis-indexer/issues/113, and try to re-index away some problems now.

PR #110, opened 2022-08-05
oasis-indexer: Address security review concerns

### LINKED ISSUE #40: Initial cursory thoughts/review
Indexer preliminary thoughts:

- The use of fmt.Sprintf to build SQL queries is terrifying, and since
  this is done all over the place, figuring out where this is safe is
  non-trivial.  If things are going to be done this way, use Sprintf
  to construct a bunch of prepared statements for clarity.

- Under the hood this is using pgx (cockroach's fork to be specific).

  - Does the fork add anything of value?

  - If the fork isn't providing much, would using upstream pgx make this
    relatively agnostic between mainline postgresql and cockroach db?

- I would much rather not have to add a cache to this once it is available
  so someone better at least think about adding one (The most recent state
  in it's entirety should be acessible without hitting up the db).

- `analyzer/consensus/ConsensusMain.latestBlock` only needs to access the db
  once and only once per indexer startup.

- `analyzer/consensus/extractEventData` makes me sad.  Yes it's slated
  for removal.  `ty` should be an enum, not a string.

- `analyzer/consensus/chainID` should probably not be hardcoded.

### LINKED ISSUE #61: analyzer: Prepare SQL queries
##### SUMMARY
Rather than build SQL queries inline, we should factor them out to live in a single, easily auditable location. Related: #40 

##### ISSUE TYPE
- Feature Idea

##### COMPONENT NAME
`analyzer`

### LINKED ISSUE #101: nit: defaultMaxConns is not a default; number of db connections not based on number of cores, should be documented
naming nit

https://github.com/oasisprotocol/oasis-indexer/blob/5b0b5a7888df3e6db1ecae9485c2f00285af46ab/storage/cockroach/client.go#L19

is used at

https://github.com/oasisprotocol/oasis-indexer/blob/5b0b5a7888df3e6db1ecae9485c2f00285af46ab/storage/cockroach/client.go#L33

_after_ `ParseConfig` so there's no way for the connection configuration string `connString` to override the "default".

from looking at the `pgx/pgxpool` code (file `pool.go`) the configuration is specified using `pool_max_conns`.  if `NewClient`'s caller doesn't specify this, the default of max(4, numcpu) is changed to 32, which seems odd (to not be based on the number of CPU cores minus the expected number of cores that the application will keep busy) but probably tolerable; but if the caller knows pgx connection configuration setup and used `pool_max_conns` somehow, they will be surprised.  maybe currently the only caller is our own code, so there's no surprise, but it'd be nice to have a comment about this behavior.

### LINKED ISSUE #104: generic nit: add useful comments, refrain from including useless ones
for example:

https://github.com/oasisprotocol/oasis-indexer/blob/5b0b5a7888df3e6db1ecae9485c2f00285af46ab/storage/cockroach/client.go#L45

it's pretty obvious that `SendBatch` sends a transaction batch, or that `NewClient` above creates a new client.

generally, comments that do not _add_ information should be omitted -- these essentially repeats info that is readily inferred from the method (or variable) names.  _useful_ comments would be ones that explain something not obvious, e.g. preconditions and post-conditions.  

in this case, for example, maybe the rationale for why the `Exec()` results (the `pgconn.CommandTag` values) for each `batchItem` are discarded, so batched transaction execution cannot see how many rows are affected by the individual operations.  and from reading the pgx/cockroachdb-go code, it looks like the operations are not run as individual transactions (in which case they would fail independently), so to call it a transaction batch seems a little misleading -- there is a batch of operations which are done in a single transaction, vs a batch of transactions which may succeed or fail independently but sent all at once for performance reasons or whatever.  maybe this is obvious to those who are used to db lingo, but it was not obvious to me.

### LINKED ISSUE #106: tiny nits: pseudo-mechanically generated code is irregular; sentinel value
just FYI.  whenever i see code like this, it makes me wish for a macro preprocessor (that cannot be abused).  the code here:

https://github.com/oasisprotocol/oasis-indexer/blob/db5a2e10d70f0836328100d490fc1658a4924462/analyzer/api.go#L148-L149

is odd in that the enum tag is `EventRegistryNodeUnfrozenEvent` but the string is `"NodeUnfrozen"`, whereas for the other enum values the pattern is obvious: the string is just the enum tag but with the `EventRegistry` or `EventRegistry` or `EventGovernance` prefixes removed, so if this were regular this would have been `"NodeUnfrozenEvent"` instead.  the irregularity could cause friction later for engineers working with this code, since they probably would expect the correspondence to be regular.

what i mean is that it would be nice if the code could have been generated using the equivalence of C macro:

```
#define CASE(a, b) \
  case Event ## a ## b: \
    return #b
CASE(Staking, Transfer)
CASE(Staking, Burn)
 ...
CASE(Registry, AllowanceChange)
...
#undef CASE
```

of course, golang doesn't have a maco preprocessor.  however, a convenient way to do this would have been to define an editor macro (emacs has keyboard macros that are unnamed, for example), so that a master list of the enums can be turned into the mechanically generated code using only a few keystrokes, and that would ensure regularity.  doing it this way would also reduce the likelihood that an enum case is forgotten.

it appears that if the enum tag ends with `Event`, then that suffix is omitted from the string representation.  arguably this is sort of okay, except that everything in the enum tag list _are_ events so the suffix is redundant.

oh (since i'm nitpicking already), when i write code like this i prefer to put the cases like `EventUnknown` to be at the beginning of an enum range.  i'm not sure if the enum values will ever be persisted or the numeric values generated by one version of this code will ever be read by another version, but having the `EventUnknown` be the last (and highest value numerically) would mean that a backward compatible new version of this would have new event types added _after_ the Unknown case, which is awkward / weird to read.  so i make it a matter of habit to put sentinel/special case values up front, so that new values added to the end will feel more "natural".

### LINKED ISSUE #107: nit: lack of defensive programming -- precondition check
since `time.Duration` is signed (`int64`), it might be a good idea to add precondition checks at the ctor/factory

https://github.com/oasisprotocol/oasis-indexer/blob/db5a2e10d70f0836328100d490fc1658a4924462/analyzer/util/util.go#L19-L21

e.g., `initialTimeout > 0`, `initialTimeout <= maximumTimeout`, and `maximumTimeout < math.MaxInt64/2` (though using `time.MaxDuration` would be better if only it existed).  though since the last just ensures no integer overflow when `b.currentTimeut` is multiplied by 2 at
https://github.com/oasisprotocol/oasis-indexer/blob/db5a2e10d70f0836328100d490fc1658a4924462/analyzer/util/util.go#L26
insisting that `maximumTimeout` be at most a few hours might also be reasonable here.

### LINKED ISSUE #108: nit: Main object SetConfig must be called, so why isn't it part of the ctor/facotry?
the `SetConfig` is intended to be called after `NewMain` and before `Start`.  as a matter of fact, it _has_ to be called, since e.g., start uses `m.cfg.BlockRange.From` which will have zero default values and would make no sense.

https://github.com/oasisprotocol/oasis-indexer/blob/db5a2e10d70f0836328100d490fc1658a4924462/analyzer/consensus/consensus.go#L61-L66

so why isn't the config just an additional parameter to `NewMain`?  and if there has to be a way to change the config from the default zero values and have separate `NewMain` and `SetConfig`, why aren't there defensive programming precondition checks in e.g. `Start` to ensure that `SetConfig` has been called?  is it okay to invoke `SetConfig` after `Start`?  (presumably not.)  as a matter of API design, if `SetConfig`'s action must occur between `NewMain` and `Start`, it would be better to fuse `SetConfig` with `NewMain`.

### LINKED ISSUE #109: nit: database query fmt.Sprintf invariant wrt call
there are several instances of sql statements/queries where `fmt.Sprintf` is used to embed the `m.cfg.ChainID`, and the rest of the prepared statement parameters are included via the $ syntax.  assuming `m.cfg` is set once, the prepared statement would never change.  while in this case the prepared statement is not in a loop here

https://github.com/oasisprotocol/oasis-indexer/blob/db5a2e10d70f0836328100d490fc1658a4924462/analyzer/consensus/consensus.go#L137-L142

the instance in `queueTransactionInserts` is, viz

https://github.com/oasisprotocol/oasis-indexer/blob/db5a2e10d70f0836328100d490fc1658a4924462/analyzer/consensus/consensus.go#L282-L346

and while this is probably not performance critical, doing this once would be nicer.  a memoizing string w/ a ctor-like func would work well, so that the `fmt.Sprintf` can be textually close to the actual SQL parameters to continue to allow readers to eyeball that they are positionally correct, etc.

ennsharma:
Resolves #40 
Resolves #61 
Resolves #101
Resolves #104
Resolves #106 
Resolves #107 
Resolves #108
Resolves #109

General housekeeping of misc. issues.


ennsharma:
Passes local run of test against genesis state.

analyzer/consensus/queries.go
```diff
@@ -0,0 +1,308 @@
+package consensus
+
+import (
+	"fmt"
+)
+
+func makeLatestBlockQuery(chainID string) string {
```
Yawning, line 7: Errr.  So the only parameter for all of these routines is the chain ID, which also appears to be static from the config yeah?

```
type ChainID string

func NewChainID(cfg *config.Config) (ChainID, error) {
  // Blah blah blah
}

func (id ChainID) LatestBlockQuery() string { // This not being *sql.Stmt makes me sad.
  // Blah blah blah
}
```

ennsharma: For the not being `*sql.Stmt` issue, maybe we could just pass all of these to [`pgx.ConnPool.Prepare`](https://pkg.go.dev/github.com/jackc/pgx#ConnPool.Prepare) on init. And then just call named statements everywhere.

Yawning: The pgx package claims to include a cache, so in theory there isn't a need to use Stmt.  Maybe something lke:
```
type QueryFactory {
  chainId string
  conn *db.Conn // Or whatever the type is supposed to be.
}

func (qf *QueryFactory) LastBlockQuery() *sql.Stmt {
   // Blah blah blah
}
```

Yawning: Alternatively, instead of fmt.Sprintf , maybe using `text/template` is cleaner.  Dunno.

api/v1/queries.go
```diff
@@ -0,0 +1,316 @@
...
+}
+
+func makeBlocksQuery(chainID string) string {
+	return fmt.Sprintf(`
+		SELECT height, block_hash, time
+			FROM %s.blocks
+			WHERE ($1::bigint IS NULL OR height >= $1::bigint) AND
+						($2::bigint IS NULL OR height <= $2::bigint) AND
+						($3::timestamptz IS NULL OR time >= $3::timestamptz) AND
+						($4::timestampz IS NULL OR time <= $4::timestamptz)
```
lukaw3d, line 22:
```suggestion
						($3::timestamptz IS NULL OR time >= $3::timestamptz) AND
						($4::timestamptz IS NULL OR time <= $4::timestamptz)
```

api/v1/queries.go
```diff
@@ -0,0 +1,318 @@
+package v1
+
+import (
+	"fmt"
+
+	"github.com/iancoleman/strcase"
+)
+
+func makeStatusQuery(chainID string) string {
```
pro-wh, line 7: why the lint here, is this one meant to use a different chain ID than the one provided?

pro-wh: nvm resolved in a later commit

cmd/analyzer/analyzer.go
```diff
@@ -125,8 +125,7 @@ func NewService(cfg *config.AnalysisConfig) (*Service, error) {
 	// Initialize analyzers.
 	analyzers := map[string]analyzer.Analyzer{}
 	for _, analyzerCfg := range cfg.Analyzers {
-		switch analyzerCfg.Name {
-		case "consensus_main_damask":
+		if analyzerCfg.Name == "consensus_main_damask" {
```
pro-wh, line 128: ignore the lint on this one

api/v1/queries.go
ennsharma: Hrm. I think I removed that in a future commit. But I tried to keep "oasis-3" (and equivalent) as the chain ID when in places that are surfaced to the user, but it needs "oasis_3" (and equivalent) to be a schema name since "-" isn't a valid character for that.

aefhm (approved):

analyzer/api.go
```diff
@@ -119,7 +116,7 @@ const (
 	// EventGovernanceVote is a proposal vote event.
 	EventGovernanceVote
 	// EventUnknown is an unknown event type.
-	EventUnknown
+	EventUnknown = 255
```
aefhm, line 119: I guess we are going unknown at the end?

ennsharma: :+1: this will also prevent new enum values changing `EventUnknown`

PR #112, opened 2022-08-06
[api] add in-memory cache

Andrew7234:
https://github.com/oasisprotocol/oasis-indexer/issues/60

https://github.com/oasisprotocol/oasis-indexer/issues/117


storage/client/client.go
```diff
@@ -0,0 +1,1081 @@
...
+		return nil, common.ErrBadChainID
+	}
+	qf := NewQueryFactory(cid)
+
+	rows, err := c.db.Query(
+		ctx,
+		qf.TransactionsQuery(),
+		r.Block,
+		r.Method,
+		r.Sender.String(),
```
Andrew7234, line 145: Not sure whether all these args are automatically string-ified. I _think_ they are but I made it explicit just in case.

api/v1/client.go
```diff
@@ -307,174 +238,55 @@ func (c *storageClient) Entities(ctx context.Context, r *http.Request) (*EntityL
...
-		c.logger.Info("row scan failed",
-			"request_id", ctx.Value(RequestIDContextKey),
-			"err", err.Error(),
-		)
-		return nil, common.ErrStorageError
-	}
-
-	entityID, err = url.PathUnescape(chi.URLParam(r, "entity_id"))
-	if err != nil {
+	if entityID == "" {
```
Andrew7234, line 252: I haven't found a suitable way to validate the [EntityID](https://pkg.go.dev/github.com/oasisprotocol/oasis-core/go/common/entity#Entity) yet. Pointers would be appreciated :)

api/spec/v1.yaml
```diff
@@ -267,7 +269,6 @@ paths:
       parameters:
         - *limit
         - *offset
-        - *height
```
Andrew7234, line 270: The `/entities`, `/accounts`, and `/validators` endpoints have this seemingly superfluous parameter that isn't read in or used from what I can tell. Maybe a vestigial parameter? Or maybe this info is somehow included in the chainID now?

aefhm: > vestigial

More like future proofing 😁

api/v1/client.go
```diff
@@ -7,83 +7,96 @@ import (
...
+		}
+		q.To = &to
 	}
-	var after *string
 	if v := params.Get("after"); v != "" {
-		after = &v
+		after, err := time.Parse("2006-01-02T15:04:05Z-0700", v)
+		if err != nil {
+			return nil, common.ErrBadRequest
+		}
```
ennsharma, line 88: since the layout is repeated, maybe factor this (and the `ParseInt`s above) into a private `validateDatetime` and `validateInt` func respectively. doesn't really reduce LOC but will make any future changes much easier
```suggestion
		after, err := time.Parse("2006-01-02T15:04:05Z-0700", v)
		if err != nil {
			return nil, common.ErrBadRequest
		}
```

docker-compose.yml
```diff
@@ -51,6 +51,20 @@ services:
...
+  oasis-net-runner:
+    image: oasisprotocol/oasis-net-runner:dev
+    volumes:
+      - type: bind
+        source: ./
+        target: /oasis-indexer
+    command:
+      - /bin/sh
+      - -c
+      - oasis-net-runner --basedir.no_temp_dir --basedir /testnet --fixture.default.node.binary oasis-node --fixture.default.setup_runtimes=false --fixture.default.deterministic_entities=true --fixture.default.fund_entities --fixture.default.num_entities 2
```
ennsharma, line 63:
```suggestion
      - oasis-net-runner
       - --basedir.no_temp_dir
       - --basedir /testnet
       - --fixture.default.node.binary oasis-node
       - --fixture.default.setup_runtimes=false
       - --fixture.default.deterministic_entities=true
       - --fixture.default.fund_entities
       - --fixture.default.num_entities 2
```
nit but might help with readability/extensibility, the line is already getting unwieldy long

api/v1/client.go
```diff
@@ -7,83 +7,96 @@ import (
...
-	tpsWindowSizeMinutes = 5
+	blockCost = 1
+	txCost    = 1
 )
 
 // storageClient is a wrapper around a storage.TargetStorage
 // with knowledge of network semantics.
 type storageClient struct {
-	db     storage.TargetStorage
+	storage *storage.StorageClient
```
ennsharma, line 28: High level comment. Can you:

- remove this struct entirely
- move `blockCache` and `txCache` into `storage.StorageClient`

This way you have a very clean interface where `api` is really just the API and doesn't need to think about how or where the data is retrieved from, and stuff in `storage` handles that. Including if/when to cache data that needs greater availability/lower latency.

api/v1/client.go
```diff
@@ -7,83 +7,96 @@ import (
...
 	logger *log.Logger
 }
 
 // newStorageClient creates a new storage client.
-func newStorageClient(db storage.TargetStorage, l *log.Logger) *storageClient {
-	return &storageClient{db, l}
+func newStorageClient(s *storage.StorageClient, l *log.Logger) (*storageClient, error) {
+	blockCache, err := ristretto.NewCache(&ristretto.Config{
+		NumCounters:        1024 * 10,
+		MaxCost:            1024,
```
ennsharma, line 40: If I'm understanding this correctly, the `* 10` is because [ristretto suggests 10x your expected cache size](https://github.com/dgraph-io/ristretto#config) for best performance. If that's the case, we should get feedback for how many blocks/transactions worth of data is worthwhile to cache from explorer/wallets. 

Referencing oasisscan as a standard explorer experience, they seem to display the 10 most recent blocks/transactions on their [homepage](https://www.oasisscan.com/), as well as 20 blocks per page on the [blocks page](https://www.oasisscan.com/blocks).

All this to say that I think tbh 1024 is completely acceptable but also that we can probably get away with smaller, should we want it :P

cmd/common/common.go
```diff
@@ -87,5 +87,6 @@ func NewClient(cfg *config.StorageConfig, logger *log.Logger) (storage.TargetSto
 	if err != nil {
 		return nil, err
 	}
+
```
ennsharma, line 90: nit: unneeded?

api/v1/client.go
Andrew7234: Sounds good, I'll leave 1024 as a starting value for now and we can tune it once the frontends/wallets start using the indexer

cmd/common/common.go
Andrew7234: I _think_ this was just due to running `go fmt`. I actually prefer the separation for readability but happy to change if  ppl feel strongly.

docker-compose.yml
Andrew7234: Oops I didn't intend to include this file; it's a leftover from my testing setup. Removing

api/v1/client.go
Andrew7234: Updated

api/v1/client.go
Andrew7234: Moved the cache to the storage package; though I did keep the api.storageClient for convenience for now

storage/client/client.go
```diff
@@ -104,8 +131,14 @@ func (c *StorageClient) Blocks(ctx context.Context, r *BlocksRequest, p *common.
 	return &bs, nil
 }
 
-// Block returns a consensus block.
+// Block returns a consensus block. This endpoint is cached.
```
ennsharma, line 134: I'd remove this edit, or add the same to the docstring for `Transaction` below too.

ennsharma (approved):
*chef's kiss*

api/v1/client.go
ennsharma:
```
func validateEntityID(param string) (*signature.PublicKey, error) {
    var pk *signature.PublicKey
    if err := pk.UnmarshalHex(param); err != nil  || !pk.IsValid() {
        return nil, err
    }
    return pk, nil
}
```

is probably what you're looking for

storage/client/client.go
Andrew7234: Confirmed that they are automatically serialized properly. Removed the `String()` because these fields are optional and can trigger nil pointer exception errors if left in

PR #125, opened 2022-08-13
Reference default pagination

aefhm:
**Why**
I think we lost this layer of variable interpolation, but we do have a default `ORDER`.
```
{"error":"ERROR: CASE types text and integer cannot be matched (SQLSTATE 42804)","level":"error","module":"postgres","msg":"failed to query db"}
{"err":"ERROR: CASE types text and integer cannot be matched (SQLSTATE 42804)","level":"info","module":"api","msg":"query failed","request_id":"46dd81e9-3659-4d95-9b23-cbf4a2e64cf6"}
```

### DIFF
--- a/api/v1/queries.go
+++ b/api/v1/queries.go
@@ -30,10 +30,7 @@ func (qf QueryFactory) BlocksQuery() string {
 						($3::timestamptz IS NULL OR time >= $3::timestamptz) AND
 						($4::timestamptz IS NULL OR time <= $4::timestamptz)
 		ORDER BY
-			CASE
-				WHEN $5::text IS NULL THEN 1
-				ELSE $5::text
-			END
+			$5::text
 		DESC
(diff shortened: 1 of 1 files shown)


ennsharma (approved):

aefhm:
So, `ORDER` technically is not set right now, and we just go to the [default](https://github.com/oasisprotocol/oasis-indexer/blob/49b3c1799b192614237013f64172d4bc1a01bceb/api/common/pagination.go#L46) ordering by first column. I think this is fine and usable enough.

I did try reading it in from the request which did not take, but I think we would need to restructure the SQL to make it work due to this [issue](https://github.com/jackc/pgx/issues/885).

aefhm:
Gonna merge and leave ordering to https://github.com/oasisprotocol/oasis-indexer/issues/95.

Issue #120, opened 2022-08-10
Debondings aren't removed after debond_end


lukaw3d:
Current epoch: 16345

https://index.oasislabs.com/v1/consensus/accounts/oasis1qzv2rktycpcykhudyvg5l6u75v08lpfcw5nt7aj5/debonding_delegations
still returns
```json
{
  "debonding_delegations": [
    {
      "amount": 20528140536507,
      "shares": 20528140536507,
      "address": "oasis1qzv2rktycpcykhudyvg5l6u75v08lpfcw5nt7aj5",
      "debond_end": 14230
    },
    {
      "amount": 143951862299182,
      "shares": 143951862299182,
      "address": "oasis1qzv2rktycpcykhudyvg5l6u75v08lpfcw5nt7aj5",
      "debond_end": 15617
    }
  ]
}
```
sum = 164480002835689

and that includes account summary too
https://index.oasislabs.com/v1/consensus/accounts/oasis1qzv2rktycpcykhudyvg5l6u75v08lpfcw5nt7aj5
```json
{
  "address": "oasis1qzv2rktycpcykhudyvg5l6u75v08lpfcw5nt7aj5",
  "nonce": 18,
  "available": 164496118213225,
  "escrow": 199706193527015700,
  "debonding": 10196635539921422,
  "delegations_balance": 1788162093296800,
  "debonding_delegations_balance": 164480002835689,
  "allowances": []
}
```

(probably related to https://github.com/oasisprotocol/oasis-indexer/issues/84 fixes)

aefhm:
For additional context as discussed with @ennsharma, `goroutine` parallelization may mean that certain queries are getting ahead of others in the batch processing per block. #127 helps mitigate this by removing the importance of order and instead relying on the actual `epoch` reached at the `block` as per `oasis-node`.

PR #127, opened 2022-08-17
Fix debonding delegation deletion

aefhm:
**Why**
Sometimes the deletion query for debonding delegation on reclaim does not work. 😡 The deployed production Indexer has this symptom.

Close again #120. See CU-3r85g3p

I was able to sometimes repro this locally by querying the `oasis_3.epochs` table while the analyzer is running, causing the deletion to return a rows affected result of `0`. Perhaps separately, we are checking the expectations of these queries. Maybe it is important or good in some future correctness checking thing.

So there are two fixes/mitigations here:

1. Add a primary key of `id` column, so that deletion is better limited to 1.
2. Move away from a `WITH QUERY` on deleting a debonding delegation row. We can have the epoch number by querying the Consensus backend, so I added that to the `StakingData`.

**Deploy**
This will need a deploy and reindex.

### DIFF
--- a/analyzer/consensus/queries.go
+++ b/analyzer/consensus/queries.go
@@ -254,16 +254,12 @@ func (qf QueryFactory) DeleteDebondingDelegationsQuery() string {
 	// Network upgrades delays debonding by 1 epoch
 	return fmt.Sprintf(`
 		DELETE FROM %[1]s.debonding_delegations
-			WHERE (ctid) IN (
-				SELECT ctid
+			WHERE id = (
+				SELECT id
 				FROM %[1]s.debonding_delegations
 				WHERE
-					delegator = $1 AND delegatee = $2 AND shares = $3 AND debond_end IN (
-					SELECT id
-					FROM %[1]s.epochs
-					ORDER BY id DESC
-					LIMIT 2
-				) LIMIT 1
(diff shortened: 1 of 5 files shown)


kostko:
Task linked: [CU-3r85g3p Debonding delegations bug](https://app.clickup.com/t/3r85g3p)

ennsharma (approved):

analyzer/consensus/queries.go
```diff
@@ -254,16 +254,12 @@ func (qf QueryFactory) DeleteDebondingDelegationsQuery() string {
 	// Network upgrades delays debonding by 1 epoch
 	return fmt.Sprintf(`
 		DELETE FROM %[1]s.debonding_delegations
-			WHERE (ctid) IN (
-				SELECT ctid
+			WHERE id = (
```
mitjat, line 257: Why was it important to switch from `ctid` to a newly-introduced `id`?

aefhm: Supposedly, `ctid` is not a stable identifier at least [long term](https://www.postgresql.org/docs/15/ddl-system-columns.html). There was a bit of future/bullet proofing behind this decision. The main source of contention was the race condition between different goroutines (routine A inserting and updating epochs, and routine B here querying based on epoch state). This PR addressed the main problem by sourcing epoch information from the block as opposed to the current DB state.

PR #129, opened 2022-08-18
storage: Runtime storage client

ennsharma:


analyzer/consensus/consensus.go
```diff
@@ -249,7 +254,7 @@ func (m *Main) prepareBlockData(ctx context.Context, height int64, batch *storag
 		return err
 	}
 
-	for _, f := range []func(*storage.QueryBatch, *storage.BlockData) error{
+	for _, f := range []func(*storage.QueryBatch, *storage.ConsensusBlockData) error{
```
aefhm, line 257: We do this for better/scoped naming?

analyzer/emerald/emerald.go
```diff
@@ -0,0 +1,272 @@
...
+
+	client, err := factory.Runtime(id)
+	if err != nil {
+		logger.Error("error creating runtime client",
+			"err", err,
+		)
+		return nil, err
+	}
+	roundRange := analyzer.RoundRange{
+		From: uint64(cfg.From),
```
aefhm, line 75: Should the `Config`'s `From` and `To` be `uint64`?

analyzer/modules/consensusaccounts.go
```diff
@@ -0,0 +1,110 @@
...
+	qf     *analyzer.QueryFactory
+	logger *log.Logger
+}
+
+// NewConsensusAccountsHandler creates a new handler for `consensus_accounts` module data.
+func NewConsensusAccountsHandler(source storage.RuntimeSourceStorage, qf *analyzer.QueryFactory, logger *log.Logger) *ConsensusAccountsHandler {
+	return &ConsensusAccountsHandler{source, qf, logger}
+}
+
+// PrepareConsensusAccountsData prepares raw data from the `consensus_accounts` module for insertion.
```
aefhm, line 28:
```suggestion
// PrepareConsensusAccountsData prepares raw data from the `consensus_accounts` module for insertion
```

storage/oasis/runtime.go
```diff
@@ -0,0 +1,143 @@
...
+
+import (
+	"context"
+	"fmt"
+
+	config "github.com/oasisprotocol/oasis-sdk/client-sdk/go/config"
+	connection "github.com/oasisprotocol/oasis-sdk/client-sdk/go/connection"
+	runtimeSignature "github.com/oasisprotocol/oasis-sdk/client-sdk/go/crypto/signature"
+	"github.com/oasisprotocol/oasis-sdk/client-sdk/go/modules/accounts"
+	"github.com/oasisprotocol/oasis-sdk/client-sdk/go/modules/consensusaccounts"
```
aefhm, line 11: > `consensusaccounts`

Huh, interesting casing.

storage/migrations/0009_oasis_3_emerald_init.up.sql
```diff
@@ -0,0 +1,90 @@
+-- Indexer state initialization for the Emerald ParaTime, after the Damask Upgrade.
```
aefhm, line 1: We are not able to generate this yet right, but I suppose this is a fixed thing.

storage/migrations/0009_oasis_3_emerald_init.up.sql
```diff
@@ -0,0 +1,90 @@
...
+  sender TEXT NOT NULL,
+  amount NUMERIC NOT NULL
+);
+
+CREATE INDEX ix_emerald_gas_used_sender ON oasis_3.emerald_gas_used(sender);
+
+-- Accounts Module Data
+
+-- The emerald_transfers table encapsulates transfers, burns, and mints.
+-- Burns are denoted by the 0-address as the receiver and mints are
```
aefhm, line 44: Good comment :)

analyzer/consensus/consensus.go
ennsharma: Partially. More specifically because the type for [consensus blocks](https://pkg.go.dev/github.com/oasisprotocol/oasis-core/go/consensus/api#Block) differs from that for [runtime blocks](https://pkg.go.dev/github.com/oasisprotocol/oasis-core/go/roothash/api/block#Header).

Consensus blocks may contain consensus-backend specific metadata, and runtime blocks contain additional merkle roots for storage, etc.

analyzer/emerald/emerald.go
ennsharma: Ideally yes. Feel free to push back if you'd like to see this now, but the reason I didn't do it here is because then we need in our config files something that flags an analyzer as a runtime analyzer or a consensus analyzer even though that's currently intrinsic from the name of the analyzer.

It should be a safe typecast until block `2 ^ 63 - 1` so I figured it was fine to kick the can down the road for now :P

storage/oasis/runtime.go
ennsharma: Indeed. Very verbose 😛

storage/migrations/0009_oasis_3_emerald_init.up.sql
ennsharma: We aren't able to generate it, nor should we need to since runtimes don't have dump+restore upgrades, only state migrations.

I think there may be some utility in the future of writing a tool that takes in the modules a paratime supports and generates a file like this, though. But that probably doesn't save net time until we have a heck ton more runtimes to index

analyzer/api.go
```diff
@@ -166,20 +217,105 @@ func (e *Event) String() string {
...
+		return "testnet"
+	case NetworkMainnet:
+		return "mainnet"
+	default:
+		return "unknown"
+	}
+}
+
+// Runtime is an identifier for a runtime on the Oasis Network.
+type Runtime uint16
```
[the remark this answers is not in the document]
ennsharma: Yes :) Really there's really only `testnet` and `mainnet` for `Network` for the foreseeable future, and I certainly don't see there being more than 255 networks ever. 

But we can have > 255 runtimes so I leveled up to `uint16` there

storage/migrations/0009_oasis_3_emerald_init.up.sql
```diff
@@ -0,0 +1,90 @@
...
+  io_root          TEXT NOT NULL,
+  state_root       TEXT NOT NULL,
+  messages_hash    TEXT NOT NULL,
+  in_messages_hash TEXT NOT NULL,
+
+  -- Arbitrary additional data.
+  extra_data JSON
+);
+
+CREATE TABLE IF NOT EXISTS oasis_3.emerald_transactions
```
Andrew7234, line 23: should there be more data here?

storage/oasis/consensus.go
```diff
@@ -0,0 +1,357 @@
+package oasis
```
Andrew7234, line 1: er why `oasis`? do we expect to have non-oasis data sources?

analyzer/emerald/emerald.go
aefhm: Aight.

storage/api.go
```diff
@@ -147,3 +135,85 @@ type GovernanceData struct {
...
+
+	// Name returns the name of the source storage.
+	Name() string
+}
+
+// RuntimeBlockData represents data for a runtime block during a given round.
+type RuntimeBlockData struct {
+	Round uint64
+
+	BlockHeader             *block.Block
```
aefhm, line 164: Hmm, I recognize that is more consistent with the consensus analyzer, but I guess technically `BlockHeader` is a `Block`.

aefhm (approved):

analyzer/consensus/consensus.go
ennsharma: Yup, that's correct.

analyzer/emerald/emerald.go
ennsharma: The consensus layer uses `int64` everywhere, since that's what Tendermint requires. [Here](https://blog.cosmos.network/choosing-a-type-for-blockchain-height-beware-of-unsigned-integers-714804dddf1d) is their rationale for doing so.

ParaTimes, on the other hand, us `uint64` to represent height/round number. So we need something different.

analyzer/api.go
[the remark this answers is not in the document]
ennsharma: It certainly could be. But `uint16` is the smallest size that can support >= 1000 enum values, for 1000 paratimes so is ever so slightly more efficient. I think I'll change `uint` to `uint8` above.

analyzer/consensus/consensus.go
```diff
@@ -33,20 +33,10 @@ const (
...
-	// analyzer's analysis range.
-	ErrOutOfRange = errors.New("range not found. no data source available")
-
-	// ErrLatestBlockNotFound is returned if the analyzer has not indexed any
-	// blocks yet. This indicates to begin from the start of its range.
-	ErrLatestBlockNotFound = errors.New("latest block not found")
-)
-
 // Main is the main Analyzer for the consensus layer.
 type Main struct {
```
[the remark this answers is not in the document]
ennsharma: Ah! This was actually what it was called initially. But our linter complained about redundant naming, since if you import this package anywhere, you must reference `consensus.ConsensusAnalyzer`. So it felt cleaner to reference `consensus.Main` after import.

analyzer/emerald/emerald.go
```diff
@@ -0,0 +1,272 @@
...
+	qf := analyzer.NewQueryFactory(strcase.ToSnake(cfg.ChainID), emerald.String())
+
+	return &Main{
+		cfg:     ac,
+		qf:      qf,
+		target:  target,
+		logger:  logger,
+		metrics: metrics.NewDefaultDatabaseMetrics(emeraldMainDamaskName),
+
+		// module handlers
```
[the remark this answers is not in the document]
ennsharma: They can't be used for consensus, since consensus is its own beast and its storage backend architecture differs from ParaTimes and their modules architecture.

But these can are indeed intended for reuse across other ParaTimes.

config/docker-dev.yml
```diff
@@ -5,6 +5,11 @@ analysis:
       rpc: unix:/node/data/internal.sock
       chaincontext: b11b369e0da5bb230b220127f5e7b242d385ef8c6f54906243f30af63c815535
       from: 8048956
+    - name: emerald_main_damask
+      chain_id: oasis-3
```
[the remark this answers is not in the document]
ennsharma: Chain ID is used for Ethereum chains to prevent [cross-chain replay attacks](https://eips.ethereum.org/EIPS/eip-155). It's a different chain ID from the chain ID as specified by our consensus layer -- which is used to denote versions between dump&restore (breaking) upgrades at the consensus layer.

Dump&restore upgrades are _technically_ entirely independent blockchains and can have entirely different parameters and semantics, but with version `n` being given genesis state that was dumped after the last block of version `n-1`.

storage/oasis/consensus.go
ennsharma: Maybe not anymore. But initially, we thought we'd have different data sources (metadata to supplement indexer data from the [metadata registry](https://github.com/oasisprotocol/metadata-registry), for example) so hence the name.

analyzer/consensus/consensus.go
```diff
@@ -56,41 +46,54 @@ type Main struct {
 func NewMain(cfg *config.AnalyzerConfig, target storage.TargetStorage, logger *log.Logger) (*Main, error) {
 	ctx := context.Background()
 
-	var ac analyzer.Config
+	var ac analyzer.ConsensusConfig
 	if cfg.Interval == "" {
```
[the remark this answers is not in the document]
ennsharma: Adding :+1:

storage/migrations/0009_oasis_3_emerald_init.up.sql
ennsharma: I'll just remove the table. Nothing writes to it anyway, until the follow-up to this PR, which will add emerald transaction processing semantics beyond modules, is created.

Removing feels cleaner than trying to patch the migration later.

PR #132, opened 2022-08-31
storage/migrations: Initialize framework for aggregate statistics

### LINKED ISSUE #131: storage/migrations: Add aggregate statistics tables
##### SUMMARY
We would like aggregate statistics exposed via our API. This includes things such as total escrow, total delegators, transactions per second (TPS), anything else that's common that would be nice to know.

We can do this by adding some [materialized views](https://www.postgresql.org/docs/current/rules-materializedviews.html) to our storage layer.

##### ISSUE TYPE
- Feature Idea

##### COMPONENT NAME
`storage`

ennsharma:
Resolves #131 

Adds a couple of basic aggregate statistics:

- Daily transaction volume
- TPS for every 5 minute interval

This can be easily extended to add any additional aggregate stats we'd want.

Note: This shouldn't be merged in until #129 goes in, since it'll introduce a number of structural changes anyway.

### DIFF
--- a/api/v1/client.go
+++ b/api/v1/client.go
@@ -1219,9 +1219,13 @@ func (c *storageClient) Validator(ctx context.Context, r *http.Request) (*Valida
 func (c *storageClient) TransactionsPerSecond(ctx context.Context, r *http.Request) (*TpsCheckpointList, error) {
 	qf := NewQueryFactory(strcase.ToSnake(LatestChainID))
 
-	pagination := common.Pagination{
-		Limit:  100,
-		Offset: 0,
+	pagination, err := common.NewPagination(r)
+	if err != nil {
+		c.logger.Info("pagination failed",
+			"request_id", ctx.Value(RequestIDContextKey),
+			"err", err.Error(),
+		)
+		return nil, common.ErrBadRequest
 	}
 
 	rows, err := c.db.Query(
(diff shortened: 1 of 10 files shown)


api/v1/client.go
```diff
@@ -1209,3 +1214,104 @@ func (c *storageClient) Validator(ctx context.Context, r *http.Request) (*Valida
...
+	}
+
+	return &ts, nil
+}
+
+// DailyVolumes returns a list of daily transaction volumes.
+func (c *storageClient) DailyVolumes(ctx context.Context, r *http.Request) (*VolumeList, error) {
+	qf := NewQueryFactory(strcase.ToSnake(LatestChainID))
+
+	pagination := common.Pagination{
```
Andrew7234, line 1277: In the same vein, are users able to query tx volumes in the past before the prior 100 days? Seems like it might be useful for block explorers?

ennsharma: Hm. Fair enough. I'd assumed 100 was okay since block explorers typically just show a graph of recent tx volume, but you're right -- no harm in just letting them query further back if needed.

api/v1/types.go
```diff
@@ -169,7 +169,7 @@ type ProposalVote struct {
 	Vote    string `json:"vote"`
 }
 
-// Validators is the API response for GetValidators.
+// ValidatorList is the API response for GetValidators.
```
aefhm, line 172: Danke.

api/v1/types.go
```diff
@@ -207,3 +207,26 @@ type ValidatorCommissionBound struct {
 	EpochStart uint64 `json:"epoch_start"`
 	EpochEnd   uint64 `json:"epoch_end"`
 }
+
+// TpsCheckpointList is the API response for ListTransactionsPerSecond.
+type TpsCheckpointList struct {
+	IntervalMinutes int             `json:"interval_minutes"`
```
aefhm, line 213: I guess doesn't matter `uint` vs `int`.

storage/migrations/0009_agg_stats_init.up.sql
```diff
@@ -0,0 +1,25 @@
+-- Initialization of materialized views for aggregate statistics on the consensus layer.
+
+BEGIN;
+
+-- min5_tx_volume stores the transaction volume in 5 minute buckets
+-- This can be used to estimate real time TPS.
+CREATE MATERIALIZED VIEW min5_tx_volume AS
```
aefhm, line 7: Guess these are schema independent.

aefhm (approved):

PR #140, opened 2022-09-26
Update README.md

Tharkunas:
İ made some fixes.

### DIFF
--- a/README.md
+++ b/README.md
@@ -17,16 +17,35 @@ to the `docker/node/etc` directory. You will need this to run the Oasis Node con
 
 **Build**
 
-From the repository root, you can run:
-```sh
-$ make docker
+## We are doing a Linux system update.
 ```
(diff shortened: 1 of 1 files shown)


aefhm:
@Tharkunas awesome that you are looking into the Indexer, thank you for your interest!

I think we do recommend that people set up their own Docker independently. If you'd find it helpful, we could perhaps add a requirements section with a version of Docker suggested.

Tharkunas:
I'll add a detailed description of how people should do this.

Xi Zhang ***@***.***>, 28 Eyl 2022 Çar, 01:09 tarihinde şunu
yazdı:

tjanez:
@Tharkunas, thanks for your PR.

Similarly to what @aefhm said, I think environment-dependent instructions on how to set up a basic thing like Docker are not appropriate for Nexus' README.

PR #142, opened 2022-09-26
Add e2e test with oasis-net-runner

### LINKED ISSUE #119: testing: Set up test runner
##### ISSUE TYPE
- Feature Idea

aefhm:
**Why**
Close #119

Lemme beef this up in subsequent PRs.

**TODO**
- [ ] Update deploy configurations


docker-compose.e2e.yml
```diff
@@ -0,0 +1,69 @@
+services:
```
ennsharma, line 1: Organization-wise, I really think this should live under `tests/e2e`. Is it easy to just re-configure all of the paths from there, or do any problems arise?

config/e2e-dev.yml
```diff
@@ -0,0 +1,25 @@
+analysis:
+  analyzers:
+    - name: consensus_main_damask
+      chain_id: oasis-3
```
ennsharma, line 4: Does this have to be `oasis-3`? Could it be something like `oasis-e2e-test`? Really, anything that's not a chain ID we plan to index for real, so that it'll catch any errors where we're hardcoding (and not being generic enough with passing chain ID as a parameter).

Makefile
```diff
@@ -20,6 +20,10 @@ docker:
 		--tag oasislabs/oasis-indexer:dev \
 		--file docker/indexer/Dockerfile \
 		.
+  @docker build \
```
ennsharma, line 23: nit: minor indent error here?

docker-compose.e2e.yml
```diff
@@ -0,0 +1,69 @@
...
+        source: ./tests/e2e/migrations
+        target: /storage/migrations
+      - type: bind
+        source: ./config
+        target: /config
+      - type: bind
+        source: ./docker/testnet
+        target: /testnet
+  postgres:
+    image: postgres:latest
```
ennsharma, line 23:
```suggestion
    image: postgres:<some_version>
```
nit: pin this, ideally to what we use in internal deploys

metrics/metrics.go
```diff
@@ -28,7 +28,7 @@ func (s *PullService) StartInstrumentation() {
 func (s *PullService) startHandler() {
 	http.Handle("/metrics", promhttp.Handler())
 
-	if err := http.ListenAndServe(s.pullEndpoint, nil); err != nil {
+	if err := http.ListenAndServe(s.pullEndpoint, nil); err != nil { //nolint:gosec // G114: Use of net/http serve function that has no support for setting timeouts
```
ennsharma, line 31:
```suggestion
        server := &http.Server{
            Addr:                      s.pullEndpoint,
            ReadTimeout:       10 * time.Second,
            WriteTimeout:       10 * time.Second,
            MaxHeaderBytes: 1 << 20,
        }
	if err := server.ListenAndServe(); err != nil {
```
something like this should fix that lint error nicely without needing to ignore it

tests/e2e/e2e_test.go
```diff
@@ -0,0 +1,145 @@
...
+	}
+	tests.Init()
+
+	// Indexer should be up
+	var status v1.Status
+	err := tests.GetFrom("/", &status)
+	require.Nil(t, err)
+
+	connString := "unix:/testnet/net-runner/network/validator-0/internal.sock"
+	conn, _ := oasisGrpc.Dial(connString, grpc.WithTransportCredentials(insecure.NewCredentials()))
```
ennsharma, line 54:
```suggestion
	conn, err := oasisGrpc.Dial(connString, grpc.WithTransportCredentials(insecure.NewCredentials()))
	require.Nil(t, err)
```
any reason not to do this (and same for all other variables for which blank identifier is used below)?

tests/e2e/e2e_test.go
```diff
@@ -0,0 +1,145 @@
...
+	hash := crypto.SHA512
+	seed := []byte("seeeeeeeeeeeeeeeeeeeeeeeeeeeeeed")
+	name := "transfer"
+	src, _ := drbg.New(hash, seed, nil, []byte(fmt.Sprintf("txsource workload generator v1, workload %s", name)))
+
+	rng := rand.New(mathrand.New(src)) //nolint:gosec // G404: Use of weak random number generator (math/rand instead of crypto/rand)
+
+	fundingAccount, _ := memorySigner.NewFactory().Generate(signature.SignerEntity, rng)
+
+	fmt.Println("Funding account address: ", staking.NewAddress(fundingAccount.Public()))
```
ennsharma, line 70:
```suggestion
	t.Logf("Funding account address: %s", staking.NewAddress(fundingAccount.Public()))
```
I'd suggest using the `*testing.T` logging interface to be more idiomatic here and below.

ennsharma (approved):
Awesome test setup! It looks really good :) Very extensible. I left some comments, but most are nits/stylistic suggestions.

docker-compose.e2e.yml
aefhm: Sounds good.

tests/e2e/e2e_test.go
aefhm: 👌

config/e2e-dev.yml
aefhm: `oasis-test` looks okay

docker-compose.e2e.yml
aefhm: Makes sense.

config/config.go
```diff
@@ -128,6 +128,9 @@ func (cfg *AnalyzerConfig) Validate() error {
 
 // ServerConfig contains the API server configuration.
 type ServerConfig struct {
+	// ChainID is the chain ID of the chain this analyzer will process.
```
ennsharma, line 131:
```suggestion
	// ChainID is the chain ID of the default (typically latest) chain ID for the server API.
```
something like this is more accurate -- this is now in the API layer not analysis

aefhm: Oops 😄 

Hmm, I think we don't set a default? I'll say that it is usually the latest.

tests/e2e/testnet/genesis.json
```diff
@@ -0,0 +1,298 @@
...
+      },
+      {
+        "untrusted_raw_value": "omF2AmJpZFggTqUyj5Q+9vZtqu10yw6Zw7HEX3Ywe0JQA9vHyzY47TU=",
+        "signature": {
+          "public_key": "TqUyj5Q+9vZtqu10yw6Zw7HEX3Ywe0JQA9vHyzY47TU=",
+          "signature": "UiuPWSjYdXih3iEGBTPyWTWoXb4cXfPQi4xnrlgooXpin876uZe4Uy3D5tLYrqDCGpDJ+r/8r0gTJ6aIwCuvBQ=="
+        }
+      }
+    ],
+    "nodes": [
```
mitjat, line 68: Do we need to insert the genesis entities, nodes etc into the DB? During a normal run, genesis state is inserted into the DB via the manually-run migrator. Maybe we need to do it _in principle_, but the lack of that data in the DB doesn't affect the current e2e tests?

(Vaguely related: I'm no fan of the migrator-generated SQL files and would prefer to convert the genesis file to SQL on the fly during migrations. Checking autogenerated content into git is usually a bad idea. If we were to do the conversion on the fly, the whole thing would also be amenable to e2e testing. Do you or @ennsharma recall a reason for the files? If not, sidestepping the files is a low-priority issue to file.)

tests/e2e/e2e_test.go
```diff
@@ -0,0 +1,153 @@
...
+	require.Nil(t, err)
+
+	ctx := context.Background()
+	cnsc := consensus.NewConsensusClient(conn)
+	doc, err := cnsc.GetGenesisDocument(ctx)
+	require.Nil(t, err)
+	doc.SetChainContext()
+
+	hash := crypto.SHA512
+	seed := []byte("seeeeeeeeeeeeeeeeeeeeeeeeeeeeeed")
```
mitjat, line 64: :smile:

config/config.go
```diff
@@ -128,6 +128,9 @@ func (cfg *AnalyzerConfig) Validate() error {
 
 // ServerConfig contains the API server configuration.
 type ServerConfig struct {
+	// ChainID is the chain ID (normally latest) for the server API
```
mitjat, line 131: How can the user derive/determine the ID? Let's document here.

I don't see `oasis-3` in the output of `oasis-node control status`, and the old value in indexer was hardcoded with no explanation. I also don't see `oasis-3` in any of our other repos. Is it just for internal bookkeeping? If so, can we make it simply `"damask"` to be consistent with analyzer names and external naming?

tests/e2e/e2e_test.go
```diff
@@ -0,0 +1,153 @@
...
+
+const (
+	fundAccountAmount = 10000000000
+	timeout           = 1 * time.Second
+)
+
+func TestIndexer(t *testing.T) {
+	if _, ok := os.LookupEnv("OASIS_INDEXER_E2E"); !ok {
+		t.Skip("skipping test since e2e tests are not enabled")
+	}
```
mitjat, line 41: This block is copy-pasted >20 times. Can you extract it into a global helper please? `SkipTestIfEndToEndDisabled(t)` or `SkipUnlessE2E(t)` or `EnsureE2E(t)` or something.

mitjat (commented):
Excellent, I'm really looking forward to having e2e in CI! Looks pretty clean :+1:

I have a few nits, a few questions mostly to help with my understanding, and a meatier high-level concern about DB setup.

config/config.go
ennsharma: You can get it programmatically from the [genesis document](https://pkg.go.dev/github.com/oasisprotocol/oasis-core/go@v0.2201.11/genesis/api#Document), and it will change after each breaking upgrade.

The user shouldn't have to ever pass it in as a parameter, I think.

tests/e2e/migrations/0000_oasis_3_test_e2e.up.sql
```diff
@@ -0,0 +1,285 @@
+-- Indexer state initialization for the Damask Upgrade.
```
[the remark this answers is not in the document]
ennsharma: I was of the opinion that it was useful to have it's own tables to really create a sandbox environment for the tests. I like just having a different DB instance for this purpose, too.

docker/indexer/Dockerfile
```diff
@@ -7,9 +7,10 @@ COPY . ./
 RUN go mod download && \
     go build
 
-FROM golang:1.17-buster AS oasis-indexer
+FROM golang:1.18-buster AS oasis-indexer
 
 COPY --from=oasis-indexer-builder /code/go/oasis-indexer /usr/local/bin/oasis-indexer
 COPY --from=oasis-indexer-builder /code/go/storage/migrations /storage/migrations/
+COPY . /oasis-indexer
```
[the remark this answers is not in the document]
aefhm: I'm using this to copy the test files into the container.

tests/e2e/testnet/genesis.json
aefhm: > Do we need to insert the genesis entities, nodes etc into the DB?

That is the right thing to do.

> the lack of that data in the DB doesn't affect the current e2e tests?

Correct. We don't make those assertions right now.

> If we were to do the conversion on the fly, the whole thing would also be amenable to e2e testing.

I'm open to this. I did like seeing the migrations though for debugging purposes, but perhaps that is a problem to be solved there.

mitjat: Thanks for the explanation. Filed https://github.com/oasisprotocol/oasis-indexer/issues/153 to track. Mind adding a TODO comment in the code here, referencing #153?

aefhm: Cool. I don't think I can add a comment on this line, but will add a comment in the analyzer code.

config/config.go
mitjat: Gotcha. So we probably shouldn't introduce the `chainId` setting in indexer's config yml, because we do set a node endpoint, and the endpoint [can provide the genesis file](https://github.com/oasisprotocol/oasis-indexer/blob/7a473104983c33020796c8232fb777634b921b08/cmd/generator/generator.go#L187-L187) that includes the chain ID.

tests/e2e/migrations/0000_oasis_3_test_e2e.up.sql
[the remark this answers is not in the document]
mitjat: Discussed on slack: This is connected to #153. Since the contents of mainnet genesis are in our migrations files, e2e doesn't run the "normal" migrations (so they're never tested!) and has this separate copy instead. This increases the priority of #153. Once that issue is closed, we can get rid of this file.

config/config.go
mitjat: Oh I take it back. This is for the API server, who doesn't care about the node. So it makes sense to manually configure which chain it should read from the DB.

mitjat (approved):
Thanks!

PR #145, opened 2022-09-29
Fix epoch bug

aefhm:
**Why**
I was erroneously updating a non-existing epoch.

### DIFF
--- a/analyzer/consensus/consensus.go
+++ b/analyzer/consensus/consensus.go
@@ -292,11 +292,15 @@ func (m *Main) queueEpochInserts(batch *storage.QueryBatch, data *storage.Consen
 		data.Epoch,
 		data.BlockHeader.Height,
 	)
-	batch.Queue(
-		m.qf.ConsensusEpochUpdateQuery(),
-		data.Epoch-1,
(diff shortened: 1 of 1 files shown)


Andrew7234 (approved):

analyzer/consensus/consensus.go
```diff
@@ -292,11 +292,15 @@ func (m *Main) queueEpochInserts(batch *storage.QueryBatch, data *storage.Consen
...
 		data.BlockHeader.Height,
 	)
-	batch.Queue(
-		m.qf.ConsensusEpochUpdateQuery(),
-		data.Epoch-1,
-		data.BlockHeader.Height,
-	)
+
+	// Conclude previous epoch
+	if data.Epoch > 0 {
```
Andrew7234, line 297: Are epochs 0-indexed? And if they are, we don't need to close the 0th one?

aefhm: > Are epochs 0-indexed?

They are.

> we don't need to close the 0th one?

😅 this part is admittedly not super clear. Epoch 0 gets closed once we get to epoch 1. I should leave a better comment.

PR #148, opened 2022-10-05
Assorted small fixes

mitjat:
Various small improvements to readability, logging, local testing, and variable naming.

Titles of individual commits explain changes in slightly more detail.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -63,17 +63,18 @@ start-docker:
 
 # Run dockerized postgres for local development
 postgres:
-	@docker run \
-		--name postgres \
+	@docker ps --format '{{.Names}}' | grep -q indexer-postgres && docker start indexer-postgres || \
+	docker run \
+		--name indexer-postgres \
 		-p 5432:5432 \
-		-e POSTGRES_USER=indexer \
+		-e POSTGRES_USER=rwuser \
 		-e POSTGRES_PASSWORD=password \
 		-e POSTGRES_DB=indexer \
 		-d postgres
 
 # Attach to the local DB from "make postgres"
 psql:
-	@docker exec -it postgres psql -U indexer indexer
+	@docker exec -it indexer-postgres psql -U rwuser indexer
 
 shutdown-postgres:
 	@docker rm postgres --force
--- a/Makefile
+++ b/Makefile
@@ -61,6 +61,7 @@ docs: $(docs-targets)
 start-docker:
 	@docker compose up --remove-orphans	
 
+# Run dockerized postgres for local development
 postgres:
 	@docker run \
 		--name postgres \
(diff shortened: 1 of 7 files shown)


analyzer/emerald/emerald.go
```diff
@@ -218,7 +214,7 @@ func (m *Main) processRound(ctx context.Context, round uint64) error {
 		return err
 	}
 
-	opName := fmt.Sprintf("process_round_%s", emerald.String())
+	opName := fmt.Sprintf("process_round_rt%s", emerald.String())
```
Andrew7234, line 217:
```suggestion
	opName := fmt.Sprintf("process_round_rt_%s", emerald.String())
```

Andrew7234 (approved):

analyzer/emerald/emerald.go
mitjat: Warren suggested the non-underscore version [on slack](https://oasis-labs.slack.com/archives/C02UB6VEF0W/p1663697436482859), and we settled on it. I'm OK with the slight inconsistency, it makes it clear that the number is about the runtime.

ennsharma (approved):

Makefile
```diff
@@ -63,17 +63,18 @@ start-docker:
...
 # Run dockerized postgres for local development
 postgres:
-	@docker run \
-		--name postgres \
+	@docker ps --format '{{.Names}}' | grep -q indexer-postgres && docker start indexer-postgres || \
+	docker run \
+		--name indexer-postgres \
 		-p 5432:5432 \
-		-e POSTGRES_USER=indexer \
+		-e POSTGRES_USER=rwuser \
```
aefhm, line 70: Oh nice! This does both distinguish the user from the DB and brings it closer to deployment usage.

config/local-dev.yml
```diff
@@ -11,14 +11,14 @@ analysis:
       chaincontext: b11b369e0da5bb230b220127f5e7b242d385ef8c6f54906243f30af63c815535
       from: 2550000
   storage:
-    endpoint: postgresql://indexer:password@localhost:5432/indexer?sslmode=disable
+    endpoint: postgresql://rwuser:password@localhost:5432/indexer?sslmode=disable
```
aefhm, line 14: Should `config/docker-dev.yml` be changed too?

aefhm (approved):

config/local-dev.yml
aefhm: On second thought, we should just merge those two files.

aefhm: All good. Lemme just do a separate small PR for those.

aefhm: Okay, can improve `docker-dev.yml` https://github.com/oasisprotocol/oasis-indexer/pull/155, but can't de-dup the files because `migrations: file:///storage/migrations` is an absolute path due to Docker usage.

PR #149, opened 2022-10-05
[analyzer] Reduce logspam when caught up with the blockchain

### LINKED ISSUE #143: Do not log data errors when caught up with latest block
The indexer indexes blocks faster than the blockchain produces them. Once the indexer catches up with the blockchain, it starts requesting blocks that have not been processing yet, and emitting error-level logs when it fails to fetch them.
```
rpc error: code = Unknown desc = tendermint: block results query failed: height 10504358 must be less than or equal to the current blockchain height 10504357
```

Add logic that understands we've caught up, and don't emit the related error.

mitjat:
Closes #143.

Before this PR, both analyzers (consensus and emerald) over-eagerly fetched blocks. The biggest offenders:
- When the analyzers catch up with the latest block (or equivalently, round), block `latest_block+1` cannot be fetched. This was logged at the `ERROR` level, making logs full of these errors, and making it harder to find real errors. This PR demotes this kind of errors to `INFO`. It detects these errors by string matching (yikes!) the error messages generated deep inside the oasis-node and sent to the indexer via RPC any many library layers. I feel OK about the tradeoff: It would be lots of work to expose those errors in a structured way, and the only impact of potentially misparsing the error is to log it at the wrong severity level.
- Block fetching was retried with a back-off delay that was expected to settle at 6 seconds. However, the backoff timer got reset to 0.1s every time we did successfully fetch a block, which means that after every block, we retried (and logged failure) at 0.1, 0.2, 0.4, etc seconds. The PR changes this so that upon every fetched block, we decrease the sleep timer by only a little bit.
- The analyzer for each SDK module logged failures independently in addition to bubbling them up. In the prevailing case where the "error" was simply that we had caught up with the newest block, this resulted in lots of log spam. This PR removes logging at the SDK-module-analyzer level.

### DIFF
--- a/analyzer/util/util.go
+++ b/analyzer/util/util.go
@@ -43,6 +43,19 @@ func NewBackoff(initialTimeout time.Duration, maximumTimeout time.Duration) (*Ba
 // Wait waits for the appropriate backoff interval.
 func (b *Backoff) Wait() {
 	time.Sleep(b.currentTimeout)
+}
+
+// Success slightly decreases the backoff interval.
+func (b *Backoff) Success() {
+	b.currentTimeout *= 9
+	b.currentTimeout /= 10
+	if b.currentTimeout < b.initialTimeout {
+		b.currentTimeout = b.initialTimeout
+	}
+}
+
+// Failure increases the backoff interval.
+func (b *Backoff) Failure() {
 	b.currentTimeout *= 2
 	if b.currentTimeout > b.maximumTimeout {
 		b.currentTimeout = b.maximumTimeout
--- a/analyzer/consensus/consensus.go
+++ b/analyzer/consensus/consensus.go
@@ -138,22 +138,25 @@ func (m *Main) Start() {
 		return
 	}
 	for m.cfg.Range.To == 0 || height <= m.cfg.Range.To {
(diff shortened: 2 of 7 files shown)


analyzer/util/util.go
```diff
@@ -43,6 +43,19 @@ func NewBackoff(initialTimeout time.Duration, maximumTimeout time.Duration) (*Ba
 // Wait waits for the appropriate backoff interval.
 func (b *Backoff) Wait() {
 	time.Sleep(b.currentTimeout)
+}
+
+// Success slightly decreases the backoff interval.
+func (b *Backoff) Success() {
+	b.currentTimeout *= 9
+	b.currentTimeout /= 10
```
ennsharma, line 51: Maybe parametrize the percentage to decrement?

ennsharma (approved):
🎉

analyzer/util/util.go
mitjat: I think it's not worth it yet. The backoff is only used in one place, so parametrization would only really make a difference if I then exposed it all the way up through Helm. Well, it would also make for slightly cleaner code, but I don't think as-is is too bad.

PR #154, opened 2022-10-06
[db] Log failed queries

mitjat:
When a SQL query fails, the existing DB interface did not allow us to log the failing query. This PR adds additional tracking and adds the failing query to the error log line.

### DIFF
--- a/storage/api.go
+++ b/storage/api.go
@@ -22,14 +22,48 @@ import (
 	"github.com/oasisprotocol/oasis-sdk/client-sdk/go/types"
 )
 
+type batchItem struct {
+	cmd  string
+	args []interface{}
+}
+
 // QueryBatch represents a batch of queries to be executed atomically.
-type QueryBatch = pgx.Batch
+// We use a custom type that mirrors pgx.Batch but also allows introspection for debugging.
+type QueryBatch struct{ items []*batchItem }
 
 // QueryResults represents the results from a read query.
 type QueryResults = pgx.Rows
 
 // QueryResult represents the result from a read query.
 type QueryResult = pgx.Row
+
+// Queue adds query to a batch.
+func (b *QueryBatch) Queue(cmd string, args ...interface{}) {
+	b.items = append(b.items, &batchItem{
+		cmd:  cmd,
+		args: args,
+	})
+}
+
+// AsPgxBatch converts a QueryBatch to a pgx.Batch.
+func (b *QueryBatch) AsPgxBatch() pgx.Batch {
+	pgxBatch := pgx.Batch{}
+	for _, item := range b.items {
+		pgxBatch.Queue(item.cmd, item.args...)
+	}
+	return pgxBatch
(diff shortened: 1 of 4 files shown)


ennsharma (approved):

storage/postgres/client.go
```diff
@@ -21,13 +22,54 @@ type Client struct {
...
+	case pgx.LogLevelError, pgx.LogLevelNone:
+		return l.logger.Error
+	default:
+		l.logger.Warn("Unknown log level", "unknown_level", level)
+		return l.logger.Info
+	}
+}
+
+// Implements pgx.Logger interface.
+func (l *pgxLogger) Log(ctx context.Context, level pgx.LogLevel, msg string, data map[string]interface{}) {
```
mitjat, line 48: I was going to set up logging just through this: If a query with an error comes through, log it. That doesn't work -- the error message doesn't contain the query. (Which makes sense -- we bubble up those errors anyway)

Then I beefed this up: Every time a query is logged, remember it but don't print it (to reduce logspam). Then when an error is logged, attach the last remembered query to the logline. This didn't work either -- if a query fails, it only produces an error log; the other informational/tracing logs only happen once a query succeeds.

So this part of integration now only takes care of surfacing warnings and errors in our logging system. And the latter sort-of already worked: When pgx generates an error log, it usually (always?) also returns an `error` value from the offending func, and then our code usually (always?) logs based on that error value.

storage/api.go
```diff
@@ -22,15 +22,49 @@ import (
...
+	pgxBatch := pgx.Batch{}
+	for _, item := range b.items {
+		pgxBatch.Queue(item.cmd, item.args...)
+	}
+	return pgxBatch
+}
+
+// Queries returns the queries in the batch. Each item of the returned slice
+// is composed of the SQL command and its arguments.
+func (b *QueryBatch) Queries() [][]interface{} {
```
aefhm, line 59: > `[][]interface{}`

Ah, we are just printing them.

Andrew7234 (approved):

storage/postgres/client.go
aefhm: > if a query fails, it only produces an error log

Hmm, what is the problem with the remembered query?

mitjat: the error log does not contain the query. I was hoping that for a bad query, we'd see two emitted logs:
1) An info log saying "about to execute query `SELECT foo FROM bar`"; we're memorize the query here
2) An error log saying "bad query; error around `foo`"; we'd display the memorize query at this point

Alas, only step 2 ever happens for bad queries. The info log, when produced, means "query XXX was _successfully_ executed"

pro-wh:
why 13 hours ago of all times :sob:

storage/postgres/client.go
aefhm: Gotcha. This makes sense.

PR #155, opened 2022-10-06
Improve Docker dev and e2e test

aefhm:
**Why**

Make Docker setup and configuration more consistent with non-Docker local development.

Increase e2e test timeout to reduce flakes.

### DIFF
--- a/config/docker-dev.yml
+++ b/config/docker-dev.yml
@@ -11,7 +11,7 @@ analysis:
       chaincontext: b11b369e0da5bb230b220127f5e7b242d385ef8c6f54906243f30af63c815535
       from: 2550000
   storage:
-    endpoint: postgresql://indexer:password@postgres:5432/indexer?sslmode=disable
+    endpoint: postgresql://rwuser:password@postgres:5432/indexer?sslmode=disable
     backend: postgres
   migrations: file:///storage/migrations
 
@@ -19,7 +19,7 @@ server:
   chain_id: oasis-3
   endpoint: 0.0.0.0:8008
   storage:
-    endpoint: postgresql://indexer:password@postgres:5432/indexer?sslmode=disable
+    endpoint: postgresql://rwuser:password@postgres:5432/indexer?sslmode=disable
     backend: postgres
 
 log:
--- a/tests/e2e/e2e_test.go
+++ b/tests/e2e/e2e_test.go
@@ -30,7 +30,9 @@ import (
 
 const (
 	fundAccountAmount = 10000000000
-	timeout           = 1 * time.Second
+	// Net runner seems to generate a block per second.
+	// 2 seconds appears to be enough for now.
+	timeout = 2 * time.Second
 )
 
 func TestIndexer(t *testing.T) {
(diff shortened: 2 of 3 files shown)


config/docker-dev.yml
```diff
@@ -11,15 +11,15 @@ analysis:
       chaincontext: b11b369e0da5bb230b220127f5e7b242d385ef8c6f54906243f30af63c815535
       from: 2550000
   storage:
-    endpoint: postgresql://indexer:password@postgres:5432/indexer?sslmode=disable
+    endpoint: postgresql://rwuser:password@postgres:5432/indexer?sslmode=disable
```
mitjat, line 14: My guess is this needs to be `@oasis-postgres` now that you've changed the container name.

FWIW, I had `indexer-postgres` in the Makefile and `local-dev.yml`. Might be worth consolidating. I was thinking `indexer-postgres` is more appropriate because the username in that postgres instance is "indexer", so it's not really intended for other oasis projects.

tests/e2e/e2e_test.go
```diff
@@ -31,7 +31,7 @@ import (
 
 const (
 	fundAccountAmount = 10000000000
-	timeout           = 1 * time.Second
+	timeout           = 2 * time.Second
```
aefhm, line 34: Gross. Apparently this makes a difference.

aefhm: https://github.com/oasisprotocol/oasis-indexer/actions/runs/3201060501/attempts/1

mitjat: ahh I was going to comment on that line in the e2e PR! :D
Is 2s definitely enough? At what interval does the testnet produce blocks?

aefhm: > At what interval does the testnet produce blocks?

I believe it was 1 second.

> Is 2s definitely enough?

🙏 ? I think the `oasis-node` orchestration under the hood takes time and the indexer is pretty fast. We could bump it to 3s? 🤔

config/docker-dev.yml
aefhm: > My guess is this needs to be @oasis-postgres now that you've changed the container name.

I think the url is by service name not container name.

> indexer-postgres

Hmm, I can go either way. You are suggesting we rename the user to `indexer-postgres`?

mitjat: Ah I stand corrected. If it works, it works.
Nah, I was not suggesting we rename the user. In the Makefile, I intentionally named the user `rwuser` so it's very different from variations on `indexer` -- makes for easier grepping of the username, and easier grokking of lines such as `postgresql://indexer:indexer@indexer-postgres:5432/indexer` :)

mitjat (approved):

config/docker-dev.yml
aefhm: Hmm, actually, yeah, I see the `indexer-postgres` naming now.

aefhm: I suppose then `oasis_indexer` is really just `indexer`. Uh, gonna merge first to unflake tests. Open to renaming.

PR #156, opened 2022-10-07
docs: clarify balance update

mitjat:


ennsharma (approved):

aefhm:
I think e2e is still flaky (based on old timeout). Yeah, def is a bug. Looks like we did release recently `oasislabs/oasis-indexer:v0.0.2`, so def need to clean that later.

aefhm (approved):

aefhm:
Uhhh, nvm. I misread this. I don't believe there to be a bug.

The particular [event](https://github.com/oasisprotocol/oasis-core/blob/go/v0.2201.11/go/staking/api/api.go#L266-L273):
```
is the event emitted when stake is transferred into an escrow
```

Sooo, the e2e is indeed working correctly as you described. I thought this line
https://github.com/oasisprotocol/oasis-indexer/blob/db39d20f069b933d387be3ef888452c20f01d70f/analyzer/consensus/consensus.go#L711-L724

[changed](https://github.com/oasisprotocol/oasis-indexer/blame/db39d20f069b933d387be3ef888452c20f01d70f/analyzer/consensus/consensus.go#L719) during a [refactor](https://github.com/oasisprotocol/oasis-indexer/pull/110), but the refactor was done correctly, and the code was fine before that.

aefhm (commented):
I think we don't need this change.

mitjat:
Thank you for looking into this, @aefhm . I submitted the draft just to run e2e more easily :grimacing: 

It's comforting to know there was no bug; we'd have a much bigger quality control problem on our hands otherwise :). I reverted the original change and instead documented and renamed the function that misled us all into thinking there might be a bug.

aefhm:
> It's comforting to know there was no bug; we'd have a much bigger quality control problem on our hands otherwise :)

My feelings are the same.

analyzer/consensus/consensus.go
```diff
@@ -621,7 +621,7 @@ func (m *Main) queueBurns(batch *storage.QueryBatch, data *storage.StakingData)
 }
 
 func (m *Main) queueEscrows(batch *storage.QueryBatch, data *storage.StakingData) error {
-	addGeneralBalanceUpdateQuery := m.qf.ConsensusAddGeneralBalanceUpdateQuery()
+	stakeGeneralBalanceUpdateQuery := m.qf.ConsensusDecreaseGeneralBalanceUpdateQuery()
```
aefhm, line 624: > `stakeGeneralBalanceUpdateQuery`

(nit) I dunno if we want to say it is sending or adding to escrow or something.

mitjat: sigh ... I was thinking the same. Updated to a long name.

On a related note, I dislike that we have `ConsensusBurnUpdateQuery` and `ConsensusDecreaseGeneralBalanceForEscrowUpdateQuery` that return _identical_ sql.

Either
1. The context from which they are called doesn't matter, and they should be a single method (`DecreaseGeneralBalance`) that happens to be called in multiple places (here, from escrow and from burn).
2. The context (here, putting funds into escrow) is important, and it doesn't make sense to have `ConsensusDecreaseGeneralBalanceForEscrowUpdateQuery` as a standalone function; instead, one function should generate all three SQL statements that need to happen to keep track of moving funds into escrow.

I like 2 because 1 results in these fragmented sql statements that are harder to make sense of (and write without bugs) without reading the callsites. What we have is mostly (there aren't many duplicated queries), but not entirely, 1.

aefhm: Agree that we mostly tried to follow 1 but not all the way. I'm also okay with 2.

PR #163, opened 2022-10-11
API: Improve latest Indexer status documentation

### LINKED ISSUE #102: api: Add current Oasis Node block status
##### SUMMARY

We want to be able to see the latest height reached by an Oasis Node.

##### ISSUE TYPE
- Feature Idea

##### ADDITIONAL INFORMATION

This can be probably cached because this should be a commonly hit endpoint.

aefhm:
## Why
Close #102 and update documentation around status endpoint.

Move away from the fractional time displayed.

### Question
Status endpoint `https://index.oasislabs.com/v1/` returns:
```json
{"latest_chain_id":"oasis-3","latest_block":10639663,"latest_update":"2022-10-06T06:25:34.233355Z"}
```
while the block endpoint `https://index.oasislabs.com/v1/consensus/blocks/10639663` returns:
```json
{"height":10639663,"hash":"9a6b77dee42acace88583ea02ca6db2e15a38c5ce07c7a08b63ad179abaf1752","timestamp":"2022-10-06T06:25:24Z"}
```

I think we probably want to standardize the time format.

### DIFF
--- a/api/spec/v1.yaml
+++ b/api/spec/v1.yaml
@@ -767,17 +767,17 @@ components:
       properties:
         latest_chain_id:
           type: string
-          description: The latest chain ID at the head of indexing.
+          description: The most recently indexed chain ID.
           example: *chain_id_1
         latest_block:
           type: integer
           format: int64
-          description: The latest indexed block at the head of indexing.
+          description: The height of the most recent indexed block. Query a synced Oasis node for the latest block produced.
           example: *block_height_1
         latest_update:
           type: string
           format: date-time
-          description: The RFC 3339 formatted time of latest indexing update.
+          description: The RFC 3339 formatted time when the Indexer processed the latest block. Compare with current time for approximate indexing progress with the Oasis Network.
           example: *iso_timestamp_1
 
     BlockList:
--- a/storage/client/client.go
+++ b/storage/client/client.go
@@ -82,6 +82,10 @@ func (c *StorageClient) Status(ctx context.Context) (*Status, error) {
 		)
 		return nil, common.ErrStorageError
 	}
+	// oasis-node control status returns time truncated to the second
+	// https://github.com/oasisprotocol/oasis-core/blob/5985dc5c2844de28241b7b16b19d91a86e5cbeda/docs/oasis-node/cli.md?plain=1#L41
+	s.LatestUpdate = s.LatestUpdate.Truncate(time.Second)
+
 	return &s, nil
 }
 


storage/client/client.go
```diff
@@ -82,6 +82,8 @@ func (c *StorageClient) Status(ctx context.Context) (*Status, error) {
 		)
 		return nil, common.ErrStorageError
 	}
+	s.LatestUpdate = s.LatestUpdate.UTC()
```
mitjat, line 85: Does this clip the precision to one second? The docs don't mention it, the name doesn't imply it.

Edit: Ah I misread the PR description. Yes, +1 to having the same precision in both endpoints. I'm actually surprised that the node truncates the sub-second part in `oasis-node control status`, but I guess we best follow their lead. So let's truncate here, and comment why.

api/spec/v1.yaml
```diff
@@ -767,17 +767,17 @@ components:
...
           type: integer
           format: int64
-          description: The latest indexed block at the head of indexing.
+          description: The most recent indexed block. Query a synced Oasis node for the latest block produced.
           example: *block_height_1
         latest_update:
           type: string
           format: date-time
-          description: The RFC 3339 formatted time of latest indexing update.
+          description: The RFC 3339 formatted time of the most recently indexed block. Check for drift with a synced Oasis node.
```
mitjat, line 780: Nit: I wouldn't recommend comparing with a live node here. If you compare this timestamp with the timestamp of the latest block, you're getting the exact same info as from `latest_block`. I see the utility of this field more so in the fact that the value is (most of the time) meaningful without node access.

Possible wording:

```suggestion
          description: The RFC 3339 formatted time of the most recently indexed block. To see how much the index lags behind the live blockchain, compare this timestamp to the current time for a rough estimate, or use `latest_block` do determine the lag exactly.
```

mitjat (approved):
Thank you!

storage/client/client.go
aefhm: > I'm actually surprised that the node truncates the sub-second part in oasis-node control status, but I guess we best follow their lead.

Nice. I didn't even check that.

aefhm: Ugh, there is additional goofiness. We had mentioned this was an [RFC 3339](https://github.com/tendermint/tendermint/issues/6233) timestamp. https://github.com/oasisprotocol/oasis-indexer/blob/e17cca44637fe654934cd4d482fb780b9ba3d2f2/api/spec/v1.yaml#L568

However, we aren't super consistent. I'm gonna stay with UTC and amend the spec. I think staying with `oasis-node control status` is best which I believe is on `UTC`.

api/spec/v1.yaml
```diff
@@ -767,17 +767,17 @@ components:
...
         latest_chain_id:
           type: string
-          description: The latest chain ID at the head of indexing.
+          description: The most recently indexed chain ID.
           example: *chain_id_1
         latest_block:
           type: integer
           format: int64
-          description: The latest indexed block at the head of indexing.
+          description: The most recent indexed block. Query a synced Oasis node for the latest block produced.
```
Andrew7234, line 775:
```suggestion
          description: The height of the most recent indexed block. Query a synced Oasis node for the latest block produced.
```

api/spec/v1.yaml
```diff
@@ -767,17 +767,17 @@ components:
...
           type: integer
           format: int64
-          description: The latest indexed block at the head of indexing.
+          description: The height of the most recent indexed block. Query a synced Oasis node for the latest block produced.
           example: *block_height_1
         latest_update:
           type: string
           format: date-time
-          description: The RFC 3339 formatted time of latest indexing update.
+          description: The RFC 3339 formatted time of the most recently indexed block. Compare with current time for approximate indexing progress with the Oasis Network.
```
Andrew7234, line 780:
```suggestion
          description: The RFC 3339 formatted time of the latest indexing update. Compare with current time for approximate indexing progress with the Oasis Network.
```

It's unclear to me whether "time of the recent block" refers to the time it was indexed, or the time the block was produced. I think the original phrasing is more precise.

aefhm: Opted for "latest indexing." 🤔

Andrew7234: Hrm 'latest indexing' isn't very clear to me, mainly bc I'm not sure what an 'indexing' (noun) is. I think I still prefer 'latest indexing update', wdyt?

aefhm: We settled offline on "latest Indexer update."

api/spec/v1.yaml
```diff
@@ -767,17 +767,17 @@ components:
...
           type: integer
           format: int64
-          description: The latest indexed block at the head of indexing.
+          description: The height of the most recent indexed block. Query a synced Oasis node for the latest block produced.
           example: *block_height_1
         latest_update:
           type: string
           format: date-time
-          description: The RFC 3339 formatted time of latest indexing update.
+          description: The RFC 3339 formatted time of the last Indexer update. Compare with current time for approximate indexing progress with the Oasis Network.
```
pro-wh, line 780: wait was this to track indexer updates? (and is an indexer update what I think it is?)

Andrew7234 (approved):

api/spec/v1.yaml
aefhm: > indexer updates

This is the time of the processing of the most recent ingested block by the Indexer. 😓 

> indexer update

Oh boy. Uhh, maybe this is not clear. An indexer update is not related to deploy/versioning but ingest or block processing. The source of this information is:

https://github.com/oasisprotocol/oasis-indexer/blob/2cb4ddda6b521c98d59ef0b1746eee881ef7f462/storage/client/queries.go#L16-L22

We then infer that at `latest_update` timestamp. The Indexer fully processed its most recent block (or greatest height block).

aefhm:
```suggestion
          description: The RFC 3339 formatted time at which the Indexer processed the latest block. Compare with current time for approximate indexing progress with the Oasis Network.
```

aefhm: Open to further future changes, but merging for now.

Issue #116, opened 2022-08-09
api+analyzer: Support height based queries


aefhm:
##### SUMMARY

We initially planned on using Cockroach DB's [feature](https://www.cockroachlabs.com/docs/stable/as-of-system-time.html) to query the DB for the state at a historical block height (inferred from the processing time of said block).

We likely want to view past balances.

##### ISSUE TYPE
- Feature Idea

##### COMPONENT NAME

API/Analyzer

##### ADDITIONAL INFORMATION

Would this conflict with parallel indexing?

mitjat:
Major caveats on using cockroachDB time-travel for this:
- limited in how far back we can go; the [default is 25h](https://www.cockroachlabs.com/docs/stable/as-of-system-time.html), which is orders of magnitude different from what we need
- cockrach allows querying at any given _system_ time, which does not linearly correlate to block height -- if we index blocks out of order, or if we restore the rows from a DB dump (which groups statements by sql table, not by blockchain height), the time-travel results will be trash

PR #167, opened 2022-10-11
storage: Remove support for CockroachDB

mitjat:
The implementation of the cockroach interface is a copy-paste of the postgres interface, with only cosmetic changes ("postgres"->"cockroach" in logs, etc), and a differently library for communicating with the db (but API-compatible with `pgx`. 

Therefore, maintaining the "fork" presents only a maintenance burden; any changes on the postgres side need to be implemented on the cockroach side too. We can re-fork the postgres handler again should we need cockroach. With https://github.com/oasisprotocol/oasis-indexer/issues/116#issuecomment-1275106474, this seems a little less likely than before.

The PR does not remove the architectural support for multiple DBs (= parsing configs, having switchable backends, etc); it just removes/disables this one particular backend. This does leave postgres as the only supported backend for now.


aefhm (approved):
I'm onboard with removing this given the maintenance burden vs tradeoff.

Is there a way to reference this in our issues or otherwise note down for our future selves that the CockroachDB client should just mimic Postgres? Granted, this should be pretty obvious later.

Andrew7234 (approved):
Given the limitations of cockroachdb I'd be in favor of closing #116 until we have an ask from product/users for it

mitjat:
> Given the limitations of cockroachdb I'd be in favor of closing #116 until we have an ask from product/users for it

Height-based queries can be implemented in non-cockroach ways too (especially if we only end up needing time travel for a limited subset of data, e.g. balances), so I'm inclined to leave it open but low priority.

PR #170, opened 2022-10-12
infra: update github actions

mitjat:
Our docker build CI action is [suddenly failing](https://github.com/oasisprotocol/oasis-indexer/actions/runs/3237922106) with `buildx failed with: ERROR: failed to solve: server message: insufficient_scope: authorization failed`.

The error is accompanied with some deprecation warnings that this PR fixes by bumping versions. My bet is that the PR will not actually solve the error, but better safe than sorry, and we need to upgrade eventually anyway.

### DIFF
--- a/.github/workflows/docker-indexer.yaml
+++ b/.github/workflows/docker-indexer.yaml
@@ -5,6 +5,7 @@ on:
   push:
     branches:
       - main
+  workflow_dispatch:  # manual; for debugging workflow before merging branch into `main`
 
 jobs:
   build-docker:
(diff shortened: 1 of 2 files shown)


.github/workflows/docker-indexer.yaml
```diff
@@ -5,6 +5,7 @@ on:
   push:
     branches:
       - main
+  workflow_dispatch:  # manual; for debugging workflow before merging branch into `main`
```
mitjat, line 8: Hah. Looks like we cannot use manual triggering until this PR is merged in :)

Andrew7234 (approved):

Issue #166, opened 2022-10-11
Emerald: nexus incoming code Go changes


pro-wh:
split off from #158

- awkward runtime id https://github.com/oasisprotocol/oasis-indexer/pull/141#discussion_r990572194 -> #158
- awkward chain context https://github.com/oasisprotocol/oasis-indexer/pull/141#discussion_r990572448 -> #158
- logging https://github.com/oasisprotocol/oasis-indexer/pull/141#discussion_r990593708 -> #174
- longer name for this loop var https://github.com/oasisprotocol/oasis-indexer/pull/141#discussion_r990594050, although we might be able to refactor that loop to be shorter and call a function instead -> #171
- review error details https://github.com/oasisprotocol/oasis-indexer/pull/141#discussion_r990596086 -> #171
- tests beautification https://github.com/oasisprotocol/oasis-indexer/pull/141#discussion_r990597040 -> #172
- integer conversions https://github.com/oasisprotocol/oasis-indexer/pull/147#discussion_r988269229, although we might start using uint64 in our structs instead
- query string templating https://github.com/oasisprotocol/oasis-indexer/pull/147#discussion_r991422582 -> #182

pro-wh:
everything is checked off. gonna close this

PR #171, opened 2022-10-12
Emerald: rename loop var in long loop

pro-wh:
for #166 

also a missing error wrap


mitjat (approved):
Now there's an impactful PR :)

PR #174, opened 2022-10-13
Emerald: printf -> log

pro-wh:
image


analyzer/uncategorized/helpers.go
```diff
@@ -1,18 +1,19 @@
...
 
 import (
-	"fmt"
 	"io"
+
+	"github.com/oasisprotocol/oasis-indexer/log"
 )
 
-func CloseOrLog(c io.Closer) {
+func CloseOrLog(c io.Closer, logger *log.Logger) {
```
mitjat, line 9: This and `WriteOrLog` don't seem to be used anywhere. Can we remove them? I don't think the indexer will be writing to files much, or at all. It's all the DB.

analyzer/emerald/incoming.go
```diff
@@ -189,8 +189,7 @@ func extractRound(sigContext signature.Context, b *block.Block, txrs []*sdkClien
 		blockTransactionData.RelatedAccountAddresses = map[string]bool{}
 		tx, err := common.VerifyUtx(sigContext, &txr.Tx)
 		if err != nil {
-			err = fmt.Errorf("tx %d: %w", txIndex, err)
-			fmt.Println(err)
+			logger.Error("error verifying tx", "tx_index", txIndex, "err", err)
 			tx = nil
```
mitjat, line 193: Not strictly related to logging, but: In my understanding, tx verification should "never" fail because we trust the node. So if verification does fail, it's a big deal and we should probably `return nil, fmt.Errorf(...)` so that the main emerald loop will keep retrying the block. If the error was due to a bit flip :roll_eyes:, the retry will succeed. Otherwise, eventually, healthchecks and grafana etc will catch up.

Alternatively, we could panic the whole program and alert externally sooner, but this is the wrong layer to be making a business decision like that.

Alternatively alternatively, I'm misunderstanding and there are more pedestrian reasons for which tx verification might fail :)

analyzer/uncategorized/helpers.go
pro-wh: they're currently not. it would be nice though. we have some places where we `defer xxx.Close()` and ignore errors, e.g. SendBatch with `batchResults`.

in nexus I ended up using these in the API server. WriteOrLog in particular appeared a lot in the handlers, and indexer is similar

analyzer/emerald/incoming.go
pro-wh: it should be that, but I'm not sure if the Go codebase supports all the authentication schemes that the runtime sdk supports. and we have not done something like fuzz test for discrepancies between the Go and Rust implementations.

so the decision to logging here and proceeding without the tx comes from these considerations:

1. there might be odd cases where it fails
2. we'd like to make progress on building a working-enough-that-it's-useful explorer without doing a deep dive to seek out (1)
3. the existence of (1) is of interest to our team
4. protect functionality for people using very normal transactions from people using less well supported transactions
5. protect on-call persons from the latter of (4) as well :laughing:

pro-wh: also asking on slack to confirm about runtime blocks only having txs with valid authentication :see_no_evil:

mitjat: Thank you. Let's see what slack says. If we it's possible to encounter blocks here that are genuinely invalid and should not be processed (or processed minimally, in that we log them somewhere, but don't let their txs impact account balances etc), we need to take that into account in the genesis analyzer also.

If the block is valid for sure (assuming an honest node), then we should process it even if validation fails. Otherwise our indexed state (balances etc) will silently become out of sync with reality. That probably means we need not `tx = nil`, but some variation of `tx = txr.Tx` that makes the type checker happy?

pro-wh: confirmed on slack: it won't have txs with wrong authentication. let's switch to not checking the authentication

pro-wh: oh fyi for maintaining indexed state, prefer to use events. there are various effects like out of gas that prevent an authenticated transaction from taking effect as written

pro-wh: #185 :confetti_ball:

analyzer/uncategorized/helpers.go
pro-wh: #184 :confetti_ball:

analyzer/emerald/incoming.go
mitjat: Thank you. I have some more clarifying questions around failure modes for txs, but I'll continue on Slack and in #185.

mitjat (approved):

PR #175, opened 2022-10-13
Emerald: eliminate uint64 -> int64 conversions

pro-wh:
for #166 

from earlier discussion:

1. keep data in uint64
2. let the database complain about out of range numbers going into BIGINT

also included: convert a now-uint64 addition to use checked arithmetic. should we? or is it too ugly and unlikely

### DIFF
--- a/analyzer/emerald/incoming.go
+++ b/analyzer/emerald/incoming.go
@@ -340,6 +340,10 @@ func extractRound(sigContext signature.Context, b *block.Block, txrs []*sdkClien
 			}
 		}
 		blockData.TransactionData = append(blockData.TransactionData, &blockTransactionData)
+		// If this overflows, it will do so silently. However, supported
+		// runtimes internally use u64 checked math to impose a batch gas,
+		// which will prevent it from emitting blocks that use enough gas to
+		// do that.
 		blockData.GasUsed += txGasUsed
 		// Inaccurate: Re-serialize signed tx to estimate original size.
 		txSize := len(cbor.Marshal(txr.Tx))
--- a/analyzer/emerald/incoming.go
+++ b/analyzer/emerald/incoming.go
@@ -47,7 +47,7 @@ type AddressPreimageData struct {
 type BlockData struct {
 	Hash             string
 	NumTransactions  int
-	GasUsed          int64
+	GasUsed          uint64
 	Size             int
 	TransactionData  []*BlockTransactionData
 	AddressPreimages map[string]*AddressPreimageData
@@ -242,7 +242,7 @@ func extractRound(sigContext signature.Context, b *block.Block, txrs []*sdkClien
 				return nil, fmt.Errorf("tx %d: %w", txIndex, err)
 			}
 		}
-		var txGasUsed int64
+		var txGasUsed uint64
 		foundGasUsedEvent := false
 		if err = common.VisitSdkEvents(txr.Events, &common.SdkEventHandler{
 			Core: func(event *core.Event) error {
@@ -251,7 +251,7 @@ func extractRound(sigContext signature.Context, b *block.Block, txrs []*sdkClien
 						return fmt.Errorf("multiple gas used events")
 					}
 					foundGasUsedEvent = true
-					txGasUsed = int64(event.GasUsed.Amount)
+					txGasUsed = event.GasUsed.Amount
 				}
 				return nil
 			},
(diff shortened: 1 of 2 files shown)


analyzer/emerald/incoming.go
```diff
@@ -334,21 +335,25 @@ func extractRound(sigContext signature.Context, b *block.Block, txrs []*sdkClien
...
 				// Inaccurate: Treat as not using any gas.
 			}
 		}
 		blockData.TransactionData = append(blockData.TransactionData, &blockTransactionData)
-		blockData.GasUsed += txGasUsed
+		if newGasUsed, carry := bits.Add64(blockData.GasUsed, txGasUsed, 0); carry == 0 {
+			blockData.GasUsed = newGasUsed
+		} else {
+			return nil, fmt.Errorf("gas used overflow")
+		}
```
mitjat, line 348: Nit: Given that we error out, this can be just
```suggestion
		if blockData.GasUsed, carry := bits.Add64(blockData.GasUsed, txGasUsed, 0); carry != 0 {
			return nil, fmt.Errorf("gas used overflow")
		}
```
probably?

I don't know how (im)possible this overflow scenario is. Do you have ballpark credible intervals for gas per tx and txs per block?
If we think it's at all possible, erroring out (and blocking the indexer from ever progressing past this block) will be very unpleasant to firefight when it happens. Instead, we could
- use bigint for this field (in db and in go), or
- cap the sum at `math.MaxUint64`. We could even note it in the API docs for the pedants.
- ignore the problem, as you suggest between the lines. Depends on how impossible the overflow is.

I expect the second option is more ergonomic for almost every client. If anybody _really_ cares about the real sum (I don't see why they would, "lotsa gas" is probably close enough), they can detect the MAX_UINT value and fetch txs for that block one by one to derive the real total gas cost.

pro-wh: re: 3-line version, I was thinking that too but

```
$ go build .
# github.com/oasisprotocol/oasis-indexer/analyzer/emerald
analyzer/emerald/incoming.go:344:6: non-name blockData.GasUsed on left side of :=
```

pro-wh: it's in the realm of "this won't be the only thing that goes wrong if it happens." the runtime sdk uses u64 for max batch gas

mitjat: Oh that's very useful! I didn't think of that parameter. Then I suggest we simply not check for overflow here, and just note `// guaranteed to not overflow; max total batch gas is defined as u64 in oasis-sdk`

pro-wh: removed!

mitjat: Thank you!

mitjat (approved):

analyzer/emerald/incoming.go
```diff
@@ -348,7 +352,7 @@ func extractRound(sigContext signature.Context, b *block.Block, txrs []*sdkClien
 	return &blockData, nil
 }
 
-func emitRoundBatch(batch *storage.QueryBatch, round int64, blockData *BlockData) {
+func emitRoundBatch(batch *storage.QueryBatch, round uint64, blockData *BlockData) {
```
Andrew7234, line 355: Apologies if this was asked already, but what's the reason for switching to `uint64` here? 

Intuitively it makes sense since the round can't be negative. However, the oasis consensus type for block height is `int64`, so the indexer `consensus` types thus use `int64` as well. Eth's ethclient [block number](https://pkg.go.dev/github.com/ethereum/go-ethereum@v1.10.25/core/types#Header) is `big.Int`, so I'm wondering if it'd be better for us to use `big.Int` instead?

pro-wh: 1. Go's conversions from uint64 to int64 is silent about stuff going out of range, and we don't want to write our own checked conversions if we don't have to.
2. The Oasis runtime SDK uses uint64 for round, so there has to be some uint64 somewhere in the system.
3. The database is using bigint (equivalent to int64) for round, so there has to be some int64 somewhere in the system.
4. The database errors out if a value is out of range.

so the minimal solution to all this is for the Go code to use uint64 on the analysis side and let the database detect out of range errors

Andrew7234: Got it, thanks for the detailed explanation!

aefhm: > Go's conversions from uint64 to int64 is silent about stuff going out of range

Already got bit by this.

pro-wh: what happened

aefhm: Technically underflow fixed in #145. I had an epoch of `(0 - 1)` being too big.

Issue #153, opened 2022-10-06
Parse genesis as part of analyzer


mitjat:
_Discussed in https://github.com/oasisprotocol/oasis-indexer/pull/142#discussion_r988200197_

**Now**:
Currently, the genesis file needs to be converted to a SQL file by manually running a command. The SQL file is stored as a "special" migration. 

**Desired**:
Parse the genesis block (if it hasn't been parsed yet) as the very first step of each analyzer, before listening to the blockchain. Insert parsed values into the DB directly.

**Why**:
- Conceptually, genesis is "block number -1", so it makes sense to parse it along with other blocks
- Checking autogenerated content into git is usually a bad match. (For linting, reviewing, versioning, etc)
- Allows for faster iteration on the genesis-parsing code (largely moot now because that codebase is stable)
- Allows e2e tests to include or rely on genesis state (see [discussion](https://github.com/oasisprotocol/oasis-indexer/pull/142#discussion_r988200197))

Benefits are minor, so priority is low.

Generated SQL files are nice for debugging, so consider providing a debugging switch (env variable?) to dump the SQL.

mitjat:
Increasing priority because it blocks us from e2e-testing our SQL migration files.

PR #176, opened 2022-10-14
Refactor migration generator

mitjat:
This is a functional no-op in preparation for #153. The goal is to make the genesis migration generator more reusable; it currently can only spit out a giant string containing all SQL queries derived from the genesis document.

The PR makes the following changes:
- Moves the generator to the analyzer directory (in preparation for the generator becoming the first step of the consensus analyzer)
- Makes the generator's output deterministic; this was not the case because of various `map`s. Deterministic output is important for being able to diff the output as we make changes to the generator, here and in the future
- Fixes some minor bugs in the generator. Nothing that impacts proper parsing of genesis; the bugs just caused invalid SQL statements to be produced for certain minimalist genesis documents, such as the one used with our e2e test and oasis-test-runner.
- Refactors the generator so it no longer produces a single giant string (though that interface still exists), but rather a list of single SQL statements.

**Note:** In the last bullet point, I could have gone further. The generator continues to do string interpolation very manually, instead of using SQL templating or other helper functions. I don't think it's worth cleaning up more at this point -- it's not very pretty but it gets the job done, and it's not likely we'll have to modify it much in the future because the genesis doc ain't gonna change.

**Reviewers:** I strongly recommend reviewing commit by commit.


analyzer/consensus/genesis.go
```diff
@@ -1,6 +1,6 @@
-// Package generator generates migrations for the Oasis Indexer
-// from the genesis file at a particular height.
-package generator
+// Conversion of the genesis state into SQL statements.
+
+package consensus
```
aefhm, line 3: 🤔 but this file is `genesis.go` not say `migrator.go`?

aefhm:
> Makes the generator's output deterministic

Nice. I didn't know that `Golang` `map` was not deterministic.

analyzer/consensus/genesis.go
```diff
@@ -0,0 +1,553 @@
...
+}
+
+// NewMigrationGenerator creates a new migration generator.
+func NewMigrationGenerator(logger *log.Logger) *MigrationGenerator {
+	return &MigrationGenerator{logger}
+}
+
+// WriteGenesisDocumentMigrationOasis3 creates a new migration that re-initializes all
+// height-dependent state as per the provided genesis document.
+func (mg *MigrationGenerator) WriteGenesisDocumentMigrationOasis3(w io.Writer, document *genesis.Document) error {
```
aefhm, line 37: This will only be `oasis3`?

analyzer/consensus/genesis.go
```diff
@@ -0,0 +1,553 @@
...
+}
+
+func (mg *MigrationGenerator) addGovernanceBackendMigrations(document *genesis.Document) (queries []string, err error) {
+	// Populate proposals.
+	queries = append(queries, `-- Governance Backend Data
+TRUNCATE {{ChainId}}.proposals CASCADE;`)
+
+	if len(document.Governance.Proposals) > 0 {
+
+		// TODO(ennsharma): Extract `executed` for proposal.
```
aefhm, line 444: (nit) Maybe just `TODO` now?

Andrew7234 (approved):
Thanks! Looks much cleaner now

analyzer/consensus/genesis.go
mitjat: The package name comes from the folder name. The filename comes from the fact that within the `analyzer/consensus` package, this file provides analysis of the genesis part of the consensus layer.

analyzer/consensus/genesis.go
mitjat: We'll see. Theoretically, oasis4 could be entirely different. In practice, I expect we'll be able to slightly refactor this function and reuse most of it for oasis4 too. For now, it is true that it only ever parses oasis3.

analyzer/consensus/genesis.go
```diff
@@ -635,3 +646,25 @@ VALUES
...
+	i := 0
+	for k := range m {
+		keys[i] = k
+		i++
+	}
+	sort.Slice(keys, func(i, j int) bool { return keys[i] < keys[j] })
+	return keys
+}
+
+func sortedKeys[V any](m map[staking.Address]V) []staking.Address {
```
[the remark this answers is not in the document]
mitjat: Our docker images specify 1.18 (which is the first version with generics support), so I figured it's fine to use them :)

mitjat:
Merging now to minimize rebase conflicts (I've already dealt with them a fair amount). Happy to address any comments in a small followup PR.

pro-wh (commented):
:+1: 

:information_source:  sortedKeys specifically for when the keys are staking addresses

mitjat:
> +1
> 
> information_source sortedKeys specifically for when the keys are staking addresses

Fair, should be `sortedAddressKeys`, especially because I also introduced `sortedIntKeys`. Will send it out at some point; I made the change locally for now. Thanks.

analyzer/consensus/genesis.go
aefhm: I thought you went for a more exact refactor/renaming from the `generator.go`. Cool.

PR #178, opened 2022-10-14
db: Squash migrations

mitjat:
This PR groups migrations files into a smaller, easier-to-read set of files.

This a housekeeping, no-op PR. It does not change the tables that are produced after all the migrations run. It does break backward compatibility in that it cannot be deployed to an existing DB without some manual surgery on the `schema_migrations` table.


Andrew7234 (approved):
I too hope there are fewer than 100 migrations

storage/migrations/README.md
```diff
@@ -6,10 +6,10 @@ each respective runtime.
 
 ## Naming Convention
 
-Migration filenames should follow the [convention](https://github.com/golang-migrate/migrate/blob/master/MIGRATIONS.md) `<id>_<name>.up.sql`, where `<id>` is the 4-digit index of the migration and `<name>` is a concise descriptor for the migration. For example, `0000_init.up.sql`.
+Migration filenames should follow the [convention](https://github.com/golang-migrate/migrate/blob/master/MIGRATIONS.md) `<id>_<name>.up.sql`, where `<id>` is the 2-digit index of the migration and `<name>` is a concise descriptor for the migration. For example, `00_init.up.sql`.
 
 We add the extra convention that migrations tied to major network upgrades specify `<name>` as `<chain_id>_init.sql`.
-For example, `0001_oasis_3_init.sql` might encode initialization for the [Damask upgrade](https://github.com/oasisprotocol/mainnet-artifacts/releases/tag/2022-04-11).
+For example, `01_oasis_3_init.sql` might encode initialization for the [Damask upgrade](https://github.com/oasisprotocol/mainnet-artifacts/releases/tag/2022-04-11).
```
pro-wh, line 12: this sentence being no longer true of the damask migration, should we keep it as an example?

aefhm, line 12:
```suggestion
For example, `01_oasis_3_init.up.sql` might encode initialization for the [Damask upgrade](https://github.com/oasisprotocol/mainnet-artifacts/releases/tag/2022-04-11).
```

(nit) This is unrelated to your change, but would like to fix.

mitjat: To be honest, I missed the "_init" convention. I like the suggestion, let's simplify. Removed these two sentences.

pro-wh (approved):

PR #184, opened 2022-10-15
log errors when closing things

pro-wh:
eh at least we won't be throwing away the errors

- closing the output stream in the migration generator: this'll probably obsolete from #177. we should really exit nonzero in this case :shrug: 
- closing the batch in the analyzer: not sure if an error in closing the batch can even happen, given that we try to process every response. nor if something goes wrong in the connection after reading all the responses whether that means anything bad regarding the durability of the transactions :shrug: 
- (unaddressed here) closing http response writers: being conservative here, not logging in case an adversary can open a bunch of connections and reset them to produce log spam. but we might want to know if errors happen otherwise :shrug: 
- (unaddressed here) something in tests/util.go: :shrug: :shrug: 

do we want these?

### DIFF
--- a/cmd/generator/generator.go
+++ b/cmd/generator/generator.go
@@ -16,6 +16,7 @@ import (
 
 	oasisConfig "github.com/oasisprotocol/oasis-sdk/client-sdk/go/config"
 
+	aCommon "github.com/oasisprotocol/oasis-indexer/analyzer/uncategorized"
 	"github.com/oasisprotocol/oasis-indexer/cmd/common"
 	"github.com/oasisprotocol/oasis-indexer/config"
 	"github.com/oasisprotocol/oasis-indexer/log"
@@ -125,7 +126,7 @@ func (g *Generator) WriteMigration() error {
 		if err != nil {
 			return err
 		}
-		defer w.Close()
+		defer aCommon.CloseOrLog(w, g.logger)
 	}
 
 	// Generate migration.


mitjat:
Good call!
* The first one is going away soon, yes
* Apparently yes, closing the batch can error out if the connection with the DB was lost. I think that this will almost certainly be reflected at the outer layer too (in the [call to `pool.BeginTxFunc()`](https://github.com/oasisprotocol/oasis-indexer/blob/4aa5a529147586a1316c2c5f25dd7ccc56e1ac58/storage/postgres/client.go#L90-L90)), but it certainly doesn't hurt to log here.
* HTTP writers: Let's add `logOrClose()` there too. [This article](https://medium.com/@simonfrey/manual-flush-golang-http-responsewriter-55675bb1415a) implies that writing HTTP responses uses a buffer. [This other article](https://www.joeshaw.org/dont-defer-close-on-writable-files/) explains that closing a buffered output should be monitored for errors because (unsurprisingly) that's when the buffer gets flushed. It's true that with HTTP responses, we can't do much when we learn about the error. But it'd still be nice to know when it happens. I don't worry too much about log spam; the logging pipeline is hugely scaleable, plus we have cloudflare to protect us from the worst.
* No strong opinion on the tests; the stakes are low there, and either way is fine IMO.

mitjat (approved):

Andrew7234 (approved):

pro-wh:
let's build out the HTTP writing+closing thing with this idea for raising the abstraction for writing responses https://github.com/oasisprotocol/oasis-indexer/pull/190#discussion_r997626670

PR #185, opened 2022-10-15
Emerald: don't authenticate transactions

pro-wh:
the runtime will authenticate them for us.

the function is still fallible though... :thinking: 

not sure if there are cases where unforseen differences in Go and Rust serialization and non-authentication validation can cause us to error here

### DIFF
--- a/analyzer/emerald/incoming.go
+++ b/analyzer/emerald/incoming.go
@@ -9,7 +9,6 @@ import (
 	"github.com/oasisprotocol/oasis-core/go/common/crypto/address"
 	"github.com/oasisprotocol/oasis-core/go/roothash/api/block"
 	sdkClient "github.com/oasisprotocol/oasis-sdk/client-sdk/go/client"
-	"github.com/oasisprotocol/oasis-sdk/client-sdk/go/crypto/signature"
 	"github.com/oasisprotocol/oasis-sdk/client-sdk/go/modules/accounts"
 	"github.com/oasisprotocol/oasis-sdk/client-sdk/go/modules/consensusaccounts"
 	"github.com/oasisprotocol/oasis-sdk/client-sdk/go/modules/core"
@@ -172,7 +171,7 @@ func registerRelatedEthAddress(addressPreimages map[string]*AddressPreimageData,
 }
 
 //nolint:gocyclo
-func extractRound(sigContext signature.Context, b *block.Block, txrs []*sdkClient.TransactionWithResults, logger *log.Logger) (*BlockData, error) {
+func extractRound(b *block.Block, txrs []*sdkClient.TransactionWithResults, logger *log.Logger) (*BlockData, error) {
 	var blockData BlockData
 	blockData.Hash = b.Header.EncodedHash().String()
 	blockData.NumTransactions = len(txrs)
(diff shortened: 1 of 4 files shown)


analyzer/emerald/incoming.go
```diff
@@ -188,7 +187,7 @@ func extractRound(sigContext signature.Context, b *block.Block, txrs []*sdkClien
 		}
 		blockTransactionData.Raw = cbor.Marshal(txr.Tx)
 		blockTransactionData.RelatedAccountAddresses = map[string]bool{}
-		tx, err := common.VerifyUtx(sigContext, &txr.Tx)
+		tx, err := common.OpenUtxNoVerify(&txr.Tx)
 		if err != nil {
 			logger.Error("error verifying tx", "tx_index", txIndex, "err", err)
```
mitjat, line 192: Nit: Might be useful to also have `txr.Tx.Hash()`; if we're handling this incorrectly and an external complaint comes in about data from tx such-and-such missing, they'll probably reference it by hash.

analyzer/emerald/incoming.go
```diff
@@ -188,7 +187,7 @@ func extractRound(sigContext signature.Context, b *block.Block, txrs []*sdkClien
...
 			tx = nil
```
mitjat, line 193: Is this enough, or should we also skip event processing (line 248)? Skipping events feels more consistent.

analyzer/emerald/incoming.go
mitjat, line 192:
```suggestion
			logger.Error("skipping malformed tx", "tx_index", txIndex, "err", err)
```

or similar. "error verifying" is outdated for sure. "malformed" might be a little too narrow, but it does seem to be the primary reason why `UpenUtxNoVerify` might error out.

pro-wh: oh yeah good idea

analyzer/emerald/incoming.go
pro-wh: events are an independent take on what happened in this transaction. and it's the runtime itself's word vs our different-language, different-sdk word. thus, the events list is a more reliable piece of information than us being able to decode the tx or not. so we should not skip event processing under this condition where we know our decoding has diverged from the runtime's decoding

analyzer/emerald/incoming.go
pro-wh: it could be malformed, e.g. if enough of the committee dishonestly allows it, or if there's a bug in the runtime that treats it as well formed.

or it could be well formed, but we have a bug in the indexer or some other mismatch with how we decode it.

analyzer/emerald/incoming.go
aefhm: >  thus, the events list is a more reliable piece of information than us being able to decode the tx or not

This is also where I am with regards to `Consensus` events.

analyzer/emerald/incoming.go
pro-wh: ugh and I just realized this omits the round too

analyzer/emerald/incoming.go
mitjat: Nice explanation, thanks! Makes sense.

mitjat (approved):

Issue #126, opened 2022-08-15
analyzer: Investigate accounts with negative balances


aefhm:
##### SUMMARY
`oasis1qzepsdrcfxq0lrc3pnmxwyr9arrl402jrg54qufn` and certain other accounts

##### ISSUE TYPE
- Bug Report

##### ACTUAL RESULTS

```paste below

```

aefhm:
Also `oasis1qzk6qlmgnq40cq2n3jfkw3307feqngt4gvksfml6`

mitjat:
These balances were discovered by manually querying the db for balances <0. 

TODO: double check if the `tests/genesis` test/tool catches these accounts. It should.

mitjat:
See also https://github.com/oasisprotocol/oasis-indexer/issues/159 with a similar issue, but wrt delegated amounts

aefhm:
As of Oct 27, we still have some bugs.

```
                    address                     |   general_balance   | nonce | escrow_balance_active | escrow_total_shares_active | escrow_balance_debonding | escrow_total_shares_debonding | extra_data
------------------------------------------------+---------------------+-------+-----------------------+----------------------------+--------------------------+-------------------------------+------------
 oasis1qzv97k6splzd67l3t3nv49vkqdjdd665qgjzxv7j |       -419100000000 |     4 |                     0 |                          0 |                        0 |                             0 |
 oasis1qrgtextryhkmtn2c3f0qerky8qk6a8asqyxu9v09 |        -99200000000 |     2 |                     0 |                          0 |                        0 |                             0 |
 oasis1qqcmsam6em636qqytxua70xss6ezwf55u57zut62 |     -36412900000000 |     9 |                     0 |                          0 |                        0 |                             0 |
 oasis1qzmd5khd6w8w3cnsy8yyj52usrc5nsg6lgattxn5 |       -700000000000 |    26 |                     0 |                          0 |                        0 |                             0 |
 oasis1qqqn56k79st0zy4060m0wvmec76vd3pl6czmg90a |         -5693113094 |     8 |     42822008381952793 |          39489726798050888 |             508269208727 |                  508269208727 |
 oasis1qrnj40xg65qge7k9qd5ag82ufsnvpsc0jcaup3qg |       -399140111427 |     8 |                     0 |                          0 |                        0 |                             0 |
[...]

PR #188, opened 2022-10-17
analyzer: Index all services backends' events

### LINKED ISSUE #123: analyzer: Capture non-transaction events
##### SUMMARY

Certain events are not related to a transaction and are only picked up by the relevant backend. For example, [`DebondingStartEscrowEvent`](https://github.com/oasisprotocol/oasis-core/blob/3fa509673fcf9de8db239db2fdef0081b1abb8b2/go/consensus/tendermint/apps/staking/transactions.go#L471-L478) is not associated with a particular `transaction`.

##### ISSUE TYPE
- Feature Idea

##### COMPONENT NAME

Analyzer. We probably also want to expose events per block height.

aefhm:
## Why

Close #123 

### TODO
- [x] document and handle `c672b8d1ef56ed28ab87c3622c5114069bdd3ad7b8f9737498d0c01ecef0967a` the canonical null address
- [x] check format for events API representation
- [x] check if special address transfer events need to be indexed here

### DIFF
--- a/analyzer/consensus/consensus.go
+++ b/analyzer/consensus/consensus.go
@@ -627,7 +627,7 @@ func (m *Main) queueRootHashEvents(batch *storage.QueryBatch, data *storage.Root
 			ty = analyzer.EventRoothashExecutorCommitted
 			body, err = json.Marshal(event.ExecutorCommitted)
 		case event.ExecutionDiscrepancyDetected != nil:
-			ty = analyzer.EventRoothashDiscrepancyDetected
+			ty = analyzer.EventRoothashExecutionDiscrepancyDetected
 			body, err = json.Marshal(event.ExecutionDiscrepancyDetected)
 		case event.Finalized != nil:
 			ty = analyzer.EventRoothashFinalized
--- a/analyzer/consensus/consensus.go
+++ b/analyzer/consensus/consensus.go
@@ -16,6 +16,9 @@ import (
 	"github.com/oasisprotocol/oasis-core/go/common/cbor"
 	"github.com/oasisprotocol/oasis-core/go/consensus/api/transaction"
 	"github.com/oasisprotocol/oasis-core/go/consensus/api/transaction/results"
+	governance "github.com/oasisprotocol/oasis-core/go/governance/api"
+	registry "github.com/oasisprotocol/oasis-core/go/registry/api"
+	roothash "github.com/oasisprotocol/oasis-core/go/roothash/api"
 	staking "github.com/oasisprotocol/oasis-core/go/staking/api"
 	oasisConfig "github.com/oasisprotocol/oasis-sdk/client-sdk/go/config"
 
@@ -268,7 +271,7 @@ func (m *Main) processBlock(ctx context.Context, height int64) error {
 		m.queueBlockInserts,
 		m.queueEpochInserts,
 		m.queueTransactionInserts,
-		m.queueEventInserts,
+		m.queueTxEventInserts,
 	} {
 		if err := f(batch, data.BlockData); err != nil {
 			if strings.Contains(err.Error(), "must be less than or equal to the current blockchain height") {
@@ -440,12 +443,13 @@ func (m *Main) queueTransactionInserts(batch *storage.QueryBatch, data *storage.
 	return nil
 }
 
-func (m *Main) queueEventInserts(batch *storage.QueryBatch, data *storage.ConsensusBlockData) error {
+// Enqueue DB statements to store events that were generated as the result of a TX execution.
+func (m *Main) queueTxEventInserts(batch *storage.QueryBatch, data *storage.ConsensusBlockData) error {
 	eventInsertQuery := m.qf.ConsensusEventInsertQuery()
 
 	for i := 0; i < len(data.Results); i++ {
 		for j := 0; j < len(data.Results[i].Events); j++ {
-			_, ty, body, err := extractEventData(data.Results[i].Events[j])
+			ty, body, err := extractEventData(data.Results[i].Events[j])
 			if err != nil {
 				return err
 			}
@@ -564,39 +568,19 @@ func (m *Main) queueNodeEvents(batch *storage.QueryBatch, data *storage.Registry
 
 func (m *Main) queueRegistryEvents(batch *storage.QueryBatch, data *storage.RegistryData) error {
 	eventInsertQuery := m.qf.ConsensusEventInsertQuery()
-	backend := analyzer.BackendRegistry
 
 	for _, event := range data.Events {
 		hash := util.SanitizeTxHash(event.TxHash.Hex())
 		if hash != nil {
-			continue
+			continue // Events associated with a tx are processed in queueTxEventInserts
 		}
 
-		var ty analyzer.Event
-		var body []byte
-		var err error
-
-		switch {
-		case event.RuntimeEvent != nil:
-			ty = analyzer.EventRegistryRuntime
-			body, err = json.Marshal(event.RuntimeEvent)
-		case event.EntityEvent != nil:
-			ty = analyzer.EventRegistryEntity
-			body, err = json.Marshal(event.EntityEvent)
-		case event.NodeEvent != nil:
-			ty = analyzer.EventRegistryNode
-			body, err = json.Marshal(event.NodeEvent)
-		case event.NodeUnfrozenEvent != nil:
-			ty = analyzer.EventRegistryNodeUnfrozen
-			body, err = json.Marshal(event.NodeUnfrozenEvent)
-		}
-
+		ty, body, err := extractRegistryEvent(event)
 		if err != nil {
 			return err
 		}
 
 		batch.Queue(eventInsertQuery,
-			backend.String(),
 			ty.String(),
 			string(body),
 			data.Height,
@@ -610,36 +594,19 @@ func (m *Main) queueRegistryEvents(batch *storage.QueryBatch, data *storage.Regi
 
 func (m *Main) queueRootHashEvents(batch *storage.QueryBatch, data *storage.RootHashData) error {
 	eventInsertQuery := m.qf.ConsensusEventInsertQuery()
-	backend := analyzer.BackendRoothash
 
 	for _, event := range data.Events {
 		hash := util.SanitizeTxHash(event.TxHash.Hex())
 		if hash != nil {
-			continue
+			continue // Events associated with a tx are processed in queueTxEventInserts
 		}
 
-		var ty analyzer.Event
-		var body []byte
-		var err error
-
-		switch {
-		case event.ExecutorCommitted != nil:
-			ty = analyzer.EventRoothashExecutorCommitted
-			body, err = json.Marshal(event.ExecutorCommitted)
-		case event.ExecutionDiscrepancyDetected != nil:
-			ty = analyzer.EventRoothashExecutionDiscrepancyDetected
-			body, err = json.Marshal(event.ExecutionDiscrepancyDetected)
-		case event.Finalized != nil:
-			ty = analyzer.EventRoothashFinalized
-			body, err = json.Marshal(event.Finalized)
-		}
-
+		ty, body, err := extractRootHashEvent(event)
 		if err != nil {
 			return err
 		}
 
 		batch.Queue(eventInsertQuery,
-			backend.String(),
 			ty.String(),
 			string(body),
 			data.Height,
@@ -784,51 +751,19 @@ func (m *Main) queueAllowanceChanges(batch *storage.QueryBatch, data *storage.St
 
 func (m *Main) queueStakingEvents(batch *storage.QueryBatch, data *storage.StakingData) error {
 	eventInsertQuery := m.qf.ConsensusEventInsertQuery()
-	backend := analyzer.BackendStaking
 
 	for _, event := range data.Events {
 		hash := util.SanitizeTxHash(event.TxHash.Hex())
 		if hash != nil {
-			continue
+			continue // Events associated with a tx are processed in queueTxEventInserts
 		}
 
-		var ty analyzer.Event
-		var body []byte
-		var err error
-
-		switch {
-		case event.Transfer != nil:
-			ty = analyzer.EventStakingTransfer
-			body, err = json.Marshal(event.Transfer)
-		case event.Burn != nil:
-			ty = analyzer.EventStakingBurn
-			body, err = json.Marshal(event.Burn)
-		case event.Escrow != nil:
-			switch t := event.Escrow; {
-			case t.Add != nil:
-				ty = analyzer.EventStakingAddEscrow
-				body, err = json.Marshal(event.Escrow.Add)
-			case t.Take != nil:
-				ty = analyzer.EventStakingTakeEscrow
-				body, err = json.Marshal(event.Escrow.Take)
-			case t.DebondingStart != nil:
-				ty = analyzer.EventStakingDebondingStart
-				body, err = json.Marshal(event.Escrow.DebondingStart)
-			case t.Reclaim != nil:
-				ty = analyzer.EventStakingReclaimEscrow
-				body, err = json.Marshal(event.Escrow.Reclaim)
-			}
-		case event.AllowanceChange != nil:
-			ty = analyzer.EventStakingAllowanceChange
-			body, err = json.Marshal(event.AllowanceChange)
-		}
-
+		ty, body, err := extractStakingEvent(event)
 		if err != nil {
 			return err
 		}
 
 		batch.Queue(eventInsertQuery,
-			backend.String(),
 			ty.String(),
 			string(body),
 			data.Height,
@@ -957,39 +892,19 @@ func (m *Main) queueVotes(batch *storage.QueryBatch, data *storage.GovernanceDat
 
 func (m *Main) queueGovernanceEvents(batch *storage.QueryBatch, data *storage.GovernanceData) error {
 	eventInsertQuery := m.qf.ConsensusEventInsertQuery()
-	backend := analyzer.BackendGovernance
 
 	for _, event := range data.Events {
 		hash := util.SanitizeTxHash(event.TxHash.Hex())
 		if hash != nil {
-			continue
+			continue // Events associated with a tx are processed in queueTxEventInserts
 		}
 
-		var ty analyzer.Event
-		var body []byte
-		var err error
-
-		switch {
-		case event.ProposalSubmitted != nil:
-			ty = analyzer.EventGovernanceProposalSubmitted
-			body, err = json.Marshal(event.ProposalSubmitted)
-		case event.ProposalExecuted != nil:
-			ty = analyzer.EventGovernanceProposalExecuted
-			body, err = json.Marshal(event.ProposalExecuted)
-		case event.ProposalFinalized != nil:
-			ty = analyzer.EventGovernanceProposalExecuted
-			body, err = json.Marshal(event.ProposalFinalized)
-		case event.Vote != nil:
-			ty = analyzer.EventGovernanceVote
-			body, err = json.Marshal(event.Vote)
-		}
-
+		ty, body, err := extractGovernanceEvent(event)
 		if err != nil {
 			return err
 		}
 
 		batch.Queue(eventInsertQuery,
-			backend.String(),
 			ty.String(),
 			string(body),
 			data.Height,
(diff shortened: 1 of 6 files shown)


analyzer/consensus/consensus.go
```diff
@@ -536,6 +511,124 @@ func (m *Main) queueNodeEvents(batch *storage.QueryBatch, data *storage.Registry
...
+			}
+		}
+
+		batch.Queue(eventInsertQuery,
+			backend.String(),
+			ty.String(),
+			string(body),
+			data.Height,
+			event.TxHash.Hex(),
+			0,
```
aefhm, line 560: Probably comment that this is gonna be `0` because event should be only emitted once per transaction.

storage/migrations/0000_oasis_3_init.up.sql
```diff
@@ -56,9 +56,7 @@ CREATE TABLE IF NOT EXISTS oasis_3.events
 
   txn_block  BIGINT NOT NULL,
   txn_hash   TEXT NOT NULL,
-  txn_index  INTEGER,
-
-  FOREIGN KEY (txn_block, txn_hash, txn_index) REFERENCES oasis_3.transactions(block, txn_hash, txn_index)
+  txn_index  INTEGER
```
aefhm, line 59: I promise to resolve this conflict.

analyzer/consensus/consensus.go
```diff
@@ -611,6 +586,124 @@ func (m *Main) queueNodeEvents(batch *storage.QueryBatch, data *storage.Registry
...
+		return err
+	}
+
+	for _, f := range []func(*storage.QueryBatch, *storage.RootHashData) error{
+		m.queueRootHashEvents,
+	} {
+		if err := f(batch, data); err != nil {
+			return err
+		}
+	}
```
mitjat, line 659: Optional:
```suggestion
	if err := m.queueRootHashEvents(batch, data); err != nil {
			return err
		}
```

though I can see preferring to follow the format of the other similar funcs.

analyzer/consensus/consensus.go
```diff
@@ -611,6 +586,124 @@ func (m *Main) queueNodeEvents(batch *storage.QueryBatch, data *storage.Registry
...
+		var err error
+
+		switch {
+		case event.RuntimeEvent != nil:
+			ty = analyzer.EventRegistryRuntime
+			body, err = json.Marshal(event.RuntimeEvent)
+
+			if err != nil {
+				return err
+			}
```
mitjat, line 605: Here and in other newly added `switch/case` blocks: Can we handle `err != nil` after the `switch/case` block? It will increase code compactness+readability. We don't do any event-specific error handling, and I think it's unlikely that we will do so in the future.

storage/api.go
```diff
@@ -97,6 +98,8 @@ type ConsensusSourceStorage interface {
 	// includes all proposals, their respective statuses and voting responses.
 	GovernanceData(ctx context.Context, height int64) (*GovernanceData, error)
 
+	RootHashData(ctx context.Context, height int64) (*RootHashData, error)
```
mitjat, line 101: Docstring appreciated

mitjat:
This PR does two independent things, as far as I can tell:
1) a fairly sizable refactor of `queueEventInserts()/extractEventData()` into smaller funcs,
2) adds Events to several internally-consumed structs that represent blockchain data. This lets us fetch all events from `.Events`, rather than just the tx-associated ones from `.Results[i].Events`.

Is that right?

~Also, is it pretty easy by any chance to check the `general_balance` of the `CommonPoolAddress` in an e2e test? If it requires a whole new test, let's not.~
Edit: I see the TODOs now. Are those for this PR, or for later? If the latter, let's open an issue to track.

mitjat (approved):
Thank you!

storage/migrations/01_oasis_3_consensus.up.sql
```diff
@@ -58,9 +58,7 @@ CREATE TABLE oasis_3.events
 
   txn_block  BIGINT NOT NULL,
   txn_hash   TEXT NOT NULL,
-  txn_index  INTEGER,
-
-  FOREIGN KEY (txn_block, txn_hash, txn_index) REFERENCES oasis_3.transactions(block, txn_hash, txn_index)
```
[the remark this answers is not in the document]
aefhm: > It would be preferable to instead just add DEFERRABLE INITIALLY DEFERRED at the end of this line, causing the constraint to be enforced at the end of each tx, rather than immediately when the row is inserted.

Agree with this principle. So, the problem is we are not currently indexing certain "transactions", and really neither are others. For example, this [event](https://www.oasisscan.com/events/10738156_2) has a transaction hash not [indexed](https://www.oasisscan.com/transactions/c672b8d1ef56ed28ab87c3622c5114069bdd3ad7b8f9737498d0c01ecef0967a).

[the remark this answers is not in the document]
mitjat: Discussed offline: The special transaction mentioned here (c672b8d1ef56ed28ab87c3622c5114069bdd3ad7b8f9737498d0c01ecef0967a) is the [SHA256 of the empty string](https://www.mathcelebrity.com/encryption.php?num=&pl=Encrypt), which is the event API's incredibly ugly way of saying "there is no associated tx". So we'll just map this to null, and make the db fields `txn_{block,hash,index}` nullable.
And it's OK to have null values even in composite FK ([example](https://dba.stackexchange.com/questions/265382/null-in-foreign-key-columns)).

aefhm:
This PR does not meaningfully change `accounts.general_balance`. I did query and find accounts with negative balances. Leaving comments on #126

storage/migrations/01_oasis_3_consensus.up.sql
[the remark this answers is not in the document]
aefhm: > make the db fields txn_{block,hash,index} nullable.

I think we only have to make `txn_hash` nullable. We still want the `block` `height` and the default `0` `index`.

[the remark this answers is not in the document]
mitjat: ah my bad, makes sense for the `height` to never be `null`.
doesn't make sense for the `index` to default to 0, unless I misunderstand it. `index==i` means "this event was produced in the `i`-th tx of block `height`", right?

analyzer/consensus/consensus.go
aefhm: Actually, let's remove this. I think we can still achieve a reasonable amount of reference just from `txn_block` and `txn_hash`.

storage/migrations/01_oasis_3_consensus.up.sql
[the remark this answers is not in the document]
aefhm: > index==i means "this event was produced in the i-th tx of block height", right?

Yeah, but I think it makes sense to remove that as well now. I don't think that is a meaningful reference on this table.

analyzer/consensus/consensus.go
```diff
@@ -611,6 +586,124 @@ func (m *Main) queueNodeEvents(batch *storage.QueryBatch, data *storage.Registry
...
+		}
+
+		batch.Queue(eventInsertQuery,
+			backend.String(),
+			ty.String(),
+			string(body),
+			data.Height,
+			event.TxHash.Hex(),
+			"NULL",
```
aefhm, line 700: Alternative is that we make the foreign key constraint an index.

analyzer/consensus/consensus.go
aefhm: SG. Dunno if this is more kosher `Golang` but looks nicer to me as well!

aefhm:
> Is that right?

Yep.

> I see the TODOs now. Are those for this PR, or for later? If the latter, let's open an issue to track.

Yes. Will convert open TODOs to issues.

aefhm:
Routing and presentation in https://github.com/oasisprotocol/oasis-indexer/issues/208

storage/migrations/01_oasis_3_consensus.up.sql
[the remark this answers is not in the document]
mitjat: Are you saying the entire foreign key is not meaningful? If so, I don't understand. Some events are still associated with a tx, right? And for those, we want to record the association.

Or are you saying only the index is not meaningful? If so, we do need it just as a technicality of the FK, which points to [this composite PK](https://github.com/oasisprotocol/oasis-indexer/blob/305095e8b1c04916924b517eeca4405d97586a90/storage/migrations/01_oasis_3_consensus.up.sql#L50-L50).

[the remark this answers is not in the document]
aefhm: > Are you saying the entire foreign key is not meaningful?

I think the foreign key is meaningful. Alternatively, we could do two indices (one to `blocks` table and another to `transactions`) but all that information is captured here in the composite FK.

> Or are you saying only the index is not meaningful? If so, we do need it just as a technicality of the FK, which points to [this composite PK](https://github.com/oasisprotocol/oasis-indexer/blob/305095e8b1c04916924b517eeca4405d97586a90/storage/migrations/01_oasis_3_consensus.up.sql#L50-L50).

Yeah, I don't think knowing which `txn_index` the particular `transaction` associated with the `event` is all that meaningful, and you could always do a subsequent lookup via the `transactions` table. My take is that we keep the foreign key and insert `NULL`.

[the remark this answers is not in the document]
mitjat: The definition of the composite PK (linked above) mentions why the index matters: the hash might not be unique.

Also, note that for composite FKs, if any single FK column is `NULL`, the entire FK constraint is lifted for that row.

So I think we do need to include the `txn_index`.

aefhm:
Resolution:

We will use `GetTransactionsWithResults` and include the `txn_index` with all available events.

analyzer/util/util.go
```diff
@@ -93,3 +95,11 @@ func CurrentBound(cs staking.CommissionSchedule, now beacon.EpochTime) (currentB
 
 	return latestStartedStep, uint64(cs.Bounds[i].Start - 1)
 }
+
+func SanitizeTxHash(hash string) string {
+	if hash == EmptyTxHash {
+		return "NULL"
```
mitjat, line 101: I would be surprised if this eventually does anything other than insert the actual string `"NULL"` into the db as the `tx_hash`. Not sure how pgx works, but my guess is you can consume and return `*string` in this func, and set it to `nil` when suitable.

Edit: [yup](https://stackoverflow.com/questions/63785376/inserting-empty-string-or-null-into-postgres-as-null-using-jackc-pgx)

analyzer/consensus/consensus.go
aefhm: we are including it.

analyzer/util/util.go
aefhm: Ah, `nil` is the right thing then.

analyzer/consensus/consensus.go
aefhm: Not doing anymore

analyzer/consensus/consensus.go
```diff
@@ -636,7 +743,7 @@ func (m *Main) queueMetadataRegistry(ctx context.Context, batch *storage.QueryBa
 	return nil
 }
 
-func (m *Main) prepareStakingData(ctx context.Context, height int64, batch *storage.QueryBatch) error {
+func (m *Main) prepareStakingData(ctx context.Context, height int64, batch *storage.QueryBatch, txHashIndexLookup map[string]int) error {
```
aefhm, line 746: There's gotta be a better way than pass this data all the way down right? 💭

analyzer/consensus/consensus.go
```diff
@@ -323,8 +328,33 @@ func (m *Main) processBlock(ctx context.Context, height int64) error {
...
+	}
+
+	data, err := source.BlockData(ctx, height)
+	if err != nil {
+		return nil, err
+	}
+
+	for i := range data.Transactions {
+		_, exists := txHashIndexLookup[data.Transactions[i].Hash().Hex()]
+		// Event should associated with the first transaction of this hash.
```
pro-wh, line 347: not so much "should," but it's what we're stuck with unless we change oasis-core

analyzer/consensus/consensus.go
mitjat: There's ... using the data that was inserted into SQL with previous statements? :)

I'll re-post this suggestion here to save it from Slack evisceration:
```sql
INSERT INTO oasis_3.events (backend, type, body, txn_block, txn_hash, txn_index)
SELECT ($1, $2, $3, $4, $5, txn_index) FROM oasis_3.transactions
WHERE txn_block=$4 AND txn_hash=$5 -- AND code=0 (code 0 means success; but a non-successful tx could also generate events)
UNION
SELECT ($1, $2, $3, $4, $5, -1)  -- dummy value (that will fail FK constraint)
ORDER BY txn_index DESC LIMIT 1; -- prefer non-dummy value if SELECT succeeds
```

If we use this version of the insert statement, we don't have to pass anything around in golang.

That said, I now see that we have multiple insert statements of this kind (= where we need the tx_idx). In which case, the code duplication above is not great (as we have modify multiple `INSERT` statements with the same heuristic code), and it'd be better to have a function. We'd define this in migrations:
```sql
-- Returns the likely index of transaction `_tx_hash` in block at `_height`.
CREATE FUNCTION guess_tx_index(_height BIGINT, _tx_hash TEXT) RETURNS INT AS $$
  SELECT txn_index FROM oasis_3.transactions
  WHERE txn_block=_height AND txn_hash=_tx_hash
  ORDER BY code ASC  -- code==0 means success, >0 is for failures; prefer successful txs
  LIMIT 1
$$ LANGUAGE SQL
```
then when inserting data, we'd call the function:
```sql
INSERT INTO oasis_3.events (backend, type, body, txn_block, txn_hash, txn_index)
VALUES ($1, $2, $3, $4, $5, COALESCE(guess_tx_index($4, $5), -1))
```
The `COALESCE` is optional and is just there to explicitly violate the FK with `-1` if `guess_tx_index` returns NULL -- that should never happen if input data is clean and our code is bug-free.

analyzer/consensus/consensus.go
```diff
@@ -453,14 +484,14 @@ func (m *Main) prepareRegistryData(ctx context.Context, height int64, batch *sto
 		return err
 	}
 
-	for _, f := range []func(*storage.QueryBatch, *storage.RegistryData) error{
+	for _, f := range []func(*storage.QueryBatch, *storage.RegistryData, map[string]int) error{
 		m.queueRuntimeRegistrations,
 		m.queueRuntimeStatusUpdates,
 		m.queueEntityEvents,
 		m.queueNodeEvents,
 		m.queueRegistryEvents,
```
mitjat, line 492: We don't need to feed `txHashIndexLookup` to all of these, right? It's misleading for them to consume the map but then not use it. Options:
- Have two for loops, one for `[]func(*storage.QueryBatch, *storage.RegistryData, map[string]int) error` and one for `[]func(*storage.QueryBatch, *storage.RegistryData)`
- If not that, let's at least prepend the arg name of `txHashIndexLoop` with `_` as a soft indication that it's unused. For example, `func (m *Main) queueRuntimeRegistrations(batch *storage.QueryBatch, data *storage.RegistryData, _txHashIndexLookup map[string]int)`

analyzer/util/util.go
```diff
@@ -103,3 +103,13 @@ func SanitizeTxHash(hash string) *string {
 
 	return &hash
 }
+
+func TxHashToIndex(hash string, lookup map[string]int) *int {
```
mitjat, line 107: Nit: Since this fn knows nothing about hashes or txs, can we call it `GetOrNil(key string, kvs map[string][int])` or similar?

analyzer/util/util.go
```diff
@@ -103,3 +103,13 @@ func SanitizeTxHash(hash string) *string {
...
+	index, exists := lookup[hash]
+
+	if exists {
+		return &index
+	}
+
+	return nil
```
mitjat, line 114: This should never happen, right? We should log some big fat errors here (though it's awkward that we don't have much context available ... the caller has it, really).

pro-wh:
idea for a different algorithm:

1. get transactions with results to get the events that are associated with a transaction
2. each of these events is associated with the transaction that it appears with in this way
3. get events from the consensus apps
4. remove events that have non-"empty" hash
5. each of these remaining events is associated with no transaction

the problem with the current txhash to index idea is that we're forced to break "ties" by associating with the first tx of any given hash, which might be the wrong one. this one preserves the event--tx association from tendermint

mitjat:
> idea for a different algorithm: get transactions with results to get the events that are associated with a transaction

Thank you, yes, that is definitely best. I remembered we discussed that, then did an (overly) quick search and only found the `GetTransactionsWithResults` for the runtimes, not [the one for consensus](https://github.com/oasisprotocol/oasis-core/blob/c16c3af7d5f8e41bdcc74ee05d576ccee7df1b34/go/consensus/api/api.go#L172-L172).

storage/migrations/01_oasis_3_consensus.up.sql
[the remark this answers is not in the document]
aefhm: Recap:

We agree to insert/include index when possible (aka for events associated with a transaction). For other events, they'll be null.

analyzer/consensus/consensus.go
aefhm: Agree.

analyzer/util/util.go
aefhm: Removed.

analyzer/consensus/consensus.go
aefhm: Using two separate loops of events processing to bypass this problem.

analyzer/util/util.go
```diff
@@ -12,6 +12,8 @@ import (
 const (
 	InitialTimeoutLowerBoundSeconds = 0
 	MaximumTimeoutUpperBoundSeconds = 60
+	// SHA256 of empty string.
```
mitjat, line 15:
```suggestion
	// SHA256 of empty string. oasis-core uses this to tags event that are not associated with any real transactions. For example, the DebondingStart escrow event.
```

analyzer/consensus/consensus.go
```diff
@@ -588,6 +595,95 @@ func (m *Main) queueNodeEvents(batch *storage.QueryBatch, data *storage.Registry
 	return nil
 }
 
+func (m *Main) queueRegistryEvents(batch *storage.QueryBatch, data *storage.RegistryData) error {
```
mitjat, line 598: With all the new `queue*Events()` methods, can we delete `queueEventInserts()` and `extractEventData()` in this file? IIUC their functionality was refactored into these new methods.

config/docker-dev.yml
```diff
@@ -16,15 +16,15 @@ analysis:
       chaincontext: b11b369e0da5bb230b220127f5e7b242d385ef8c6f54906243f30af63c815535
       from: 2550000
   storage:
-    endpoint: postgresql://rwuser:password@postgres:5432/indexer?sslmode=disable
+    endpoint: postgresql://rwuser:password@indexer-postgres:5432/indexer?sslmode=disable
```
mitjat, line 19: Sorry I missed these (and more) :bow:

storage/api.go
```diff
@@ -166,12 +173,21 @@ type StakingData struct {
...
 	Escrows          []*staking.EscrowEvent
 	AllowanceChanges []*staking.AllowanceChangeEvent
 }
 
+// RootHashData represents data for runtime processing at a given height.
+type RootHashData struct {
+	Height int64
+
+	Events []*roothash.Event
+	// TODO: Extend this interface to include runtime blocks.
```
mitjat, line 188: Maybe. What would we do with them? Just trying to understand the TODO.

Right now, it's a conscious decision to have each runtime have its own analyzer, so there's no reason for the consensus analyzer to fetch the runtime blocks. I think this design might be helpful once we start dealing with multiple versions of e.g. emerald, for which we'll want to run two slightly different emerald analyzers, each on a separate subset of heights. Meanwhile, the same consensus analyzer will process all of those blocks, assuming there's no breaking consensus version change. And vice versa, consensus might change and emerald might not.

analyzer/consensus/consensus.go
mitjat: Oh I missed all of these blocks:
```
if hash != nil {
			continue
		}
```

So the new methods only process events that are not associated with a tx? Then `queueEventInserts()` should only process events that _are_ associated with a tx, I suppose, though I don't see why we wouldn't just process both kinds of events here.

mitjat:
From the checkbox list in PR description:

>  Bring back the nullable foreign key constraint

I don't see an associated diff. Do I understand correctly that this was meanwhile done in Andy's #234 and is no longer in scope for this PR?

analyzer/consensus/consensus.go
aefhm: Good record keep, but we no longer will pass this index data down. Whew.

analyzer/consensus/consensus.go
aefhm: > Then queueEventInserts() should only process events that are associated with a tx

Yes. This is the case before and after this PR.

> why we wouldn't just process both kinds of events here.

Unfortunately, we are doing two passes. First, we obtain the `tx_index` information based on the position of the associated `tx` in the array from `GetTransactionsWithResults`. Next, we get the events without a `tx` from the individual service backends.

I believe this is the improved algorithm mentioned [here](https://github.com/oasisprotocol/oasis-indexer/pull/188#issuecomment-1307863567).

Hmm, perhaps I can make this clearer by extrapolating the switch case statements into helpers.

storage/api.go
aefhm: History was https://github.com/oasisprotocol/oasis-indexer/blob/699a599d14f8566cc27575a8c90f816721093e81/storage/api.go#L62-L64

Hmm, yeah, I think this can be removed now.

aefhm:
> I don't see an associated diff. Do I understand correctly that this was meanwhile done in Andy's https://github.com/oasisprotocol/oasis-indexer/pull/234 and is no longer in scope for this PR?

Sorry. I didn't update the PR description. Yes, #234 correctly removes the constraint, so this PR should not change the migration file.

aefhm:
> The underlying ticket https://github.com/oasisprotocol/oasis-indexer/issues/123#issuecomment-1218514752 that capturing these non-tx events will let us process certain transfers that we haven't been capturing so far.

> This PR, from what I can tell, only inserts the events themselves into the DB, and does not insert anything else. So we need to

> expand this PR, or
not close https://github.com/oasisprotocol/oasis-indexer/issues/123 yet, or
update https://github.com/oasisprotocol/oasis-indexer/issues/123 to say we no longer care about the updates that these newly-captured events bring about.
I'm not sure which of these is best. For example, I think it's fine to ignore the common pool addr and not track its balance as a special case. (Or, if we're going to track it's balance, we should do it [here](https://github.com/oasisprotocol/oasis-indexer/blob/68a9e329571c35b63aa97b4484077ee3d52d85b8/analyzer/consensus/consensus.go#L628-L628) where all the effects of DebondingStart are considered; we should not present it as a transfer event.)

I think I'd prefer capping off this PR and either not closing #123 or splitting off the separate point about common address into a separate issue.

analyzer/consensus/consensus.go
```diff
@@ -801,103 +889,139 @@ func (m *Main) queueVotes(batch *storage.QueryBatch, data *storage.GovernanceDat
...
+	case event.EntityEvent != nil:
+		ty = analyzer.EventRegistryEntity
+		body, err = json.Marshal(event.EntityEvent)
+	case event.NodeEvent != nil:
+		ty = analyzer.EventRegistryNode
+		body, err = json.Marshal(event.NodeEvent)
+	case event.NodeUnfrozenEvent != nil:
+		ty = analyzer.EventRegistryNodeUnfrozen
+		body, err = json.Marshal(event.NodeUnfrozenEvent)
+	}
```
mitjat, line 992: Let's be defensive here and in all the other `extract*Event` functions:

```suggestion
  default:
    return fmt.Errorf("unsupported registry event type: %#v", event)
	}
```

mitjat (approved):
Thank you!

storage/api.go
mitjat: Thanks for the explanation. Yes, sounds like we can remove this now.

mitjat:
> I think I'd prefer capping off this PR and either not closing #123 or splitting off the separate point about common address into a separate issue.

+1.
The only remaining issue in #123 is accurately tracking the balance of the common pool. Like I said, I'm fine to special case that account and not track its balance accurately.
If, however, you're aware of accounts that we're _not_ happy to ignore, _and_ their balance was impossible to track correctly before this PR (because we didn't have the relevant events) ... then let's note those addresses in #123 and take care of updating their balances in a separate PR.
OTOH if you're not aware of any such addresses, let's just say as much in #123 and close the issue.

aefhm:
Since the data flow is somewhat subtle, can we please try to make it a little more obvious? I suggest:

> Rename queueEventInserts to queueTxEventInserts or queueTxGeneratedEventInserts or similar

Yes. Let's go to `queueTxEventInserts`.
> Add docstring // Enqueues DB statements for storing events that were generated as the result of executing a tx

Done.
> Add a comment to all hash != nil blocks:
if hash != nil {
  continue  // Events associated with a tx are already handled in queueTxEventInserts
}

Cool.

Thanks, these are helpful!

analyzer/consensus/consensus.go
aefhm: Danke. Revised as suggested.

aefhm:
> If, however, you're aware of accounts that we're not happy to ignore, and their balance was impossible to track correctly before this PR (because we didn't have the relevant events) ... then let's note those addresses in https://github.com/oasisprotocol/oasis-indexer/issues/123 and take care of updating their balances in a separate PR.

I find #123 to be more events based now 😬, so perhaps we can just relegate the separate PR to separate [issue](https://github.com/oasisprotocol/oasis-indexer/issues/261)? I do recall us caring about this at one point (albeit not high priority).

aefhm:
Closed https://app.clickup.com/t/3u9y7xn

kostko:
Task linked: [CU-3u9y7xn analyzer: Capture non-tx events](https://app.clickup.com/t/3u9y7xn)

PR #191, opened 2022-10-19
Update e2e tests documentation

aefhm:
## Why

Recommend running e2e tests in container.


tests/README.md
```diff
@@ -7,6 +7,13 @@ This directory contains, primarily, e2e tests for the Oasis Indexer. They have b
 
 ## Setup for E2E Tests
 
-To ensure that tests behave as expected, you should have either the local environment or Docker environment configured as per the [top-level README](../README.md#docker-development).
+To ensure that tests behave as expected, you should have either Docker environment configured as per the [top-level README](../README.md#docker-development).
```
mitjat, line 10: It would be nice to encourage a set of containers (I guess just the db, at this point) that's separate for e2e, as opposed to regular development. The e2e set always needs to start clean and can be killed right after the tests.

We could package the whole operation (build fresh docker image, bring up docker compose, exec into the indexer and run tests, tear down docker compose) into a single Makefile operation.

mitjat (commented):
I don't know that we _have_ to run this in the docker, but I agree it will be almost never worth the effort to set up non-dockerized dependencies for each e2e run. +1 to simplifying the instructions to dockerized-only.

tests/README.md
```diff
@@ -1,12 +1,38 @@
...
 - `genesis/` contains tests that validate our database state is correct against `oasis-node` genesis state.
 - `http/` contains tests that validate API endpoints behave as expected and return correct output.
 
-## Setup for E2E Tests
+## E2E Tests
 
-To ensure that tests behave as expected, you should have either the local environment or Docker environment configured as per the [top-level README](../README.md#docker-development).
+To ensure that tests behave as expected, you should have Docker environment configured as per the [top-level README](../README.md#docker-development).
 
-Then you can run these tests from the repository root with `make test-ci` or directly from this directory with `go test ./... -v`.
```
aefhm, line 12: Okay, fair. I see the `env HOST_UID=$(id -u) HOST_GID=$(id -g)` set up now to run e2e tests outside of Docker.

Makefile
```diff
@@ -28,6 +28,9 @@ docker:
 clean:
 	@$(GO) clean
 
+clean-e2e:
+	rm -r tests/e2e/testnet/net-runner
```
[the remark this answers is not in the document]
aefhm: > Do we want to also clean the db?

Yeah, for sure. I think `docker compose down -v` should do the trick?

```
stop-e2e:
	@docker compose -f tests/e2e/docker-compose.e2e.yml down -v -t 0
```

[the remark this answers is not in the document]
mitjat: LGTM from the docs. Do please test though :)

tests/README.md
```diff
@@ -1,12 +1,38 @@
...
+### Running e2e tests locally
+
+For the canonical way to run these tests, see the [CI config](../../.github/workflows/ci-test.yaml).
+Here are some useful one-liners derived from those commands, and adapted for running locally and repeatedly.
+
+### Get docker containers with current code up&running
+```sh
+make docker &&
+{ docker compose -f tests/e2e/docker-compose.e2e.yml down -t 0 || true; } &&
```
mitjat, line 21: This should probably be a part of `clean-e2e`? With the addition of the `-v` switch you proposed.

tests/README.md
```diff
@@ -1,12 +1,38 @@
...
+make clean-e2e &&
+env HOST_UID=$(id -u) HOST_GID=$(id -g) docker compose -f tests/e2e/docker-compose.e2e.yml up -d
```
mitjat, line 23: Optional: Now that you've taken on creating higher-level makefile recipes, I would suggest renaming the old `test-e2e` to `test-e2e-local`, and creating a new `test-e2e` that consists of this line and line 32.

Then the steps for running e2e become pretty digestible: `make docker && make clean-e2e && make test-e2e`

aefhm: Used `start-e2e` to describe what you intend by `test-e2e` here I think.

tests/README.md
aefhm: I ack this nice to have.

aefhm: https://github.com/oasisprotocol/oasis-indexer/issues/202

tests/README.md
mitjat: I had a hard time explaining ... I pushed a commit with what I had in mind. Your call on whether you want to include it or go with what you prepared. I'm OK with both; the reason I prefer my edit is that the commands for running tests locally are then concise. (That's the commands I care about; it shouldn't take massive history/docs hunting and long copy-paste to run tests. In CI, everything is in code and hands-off so it's less of a problem if commands are a little longer there.)

mitjat (approved):
Thanks :)

Makefile
```diff
@@ -69,6 +75,14 @@ docs: $(docs-targets)
 start-docker:
 	@docker compose up --remove-orphans
 
+start-docker-e2e:
+	@docker compose -f tests/e2e/docker-compose.e2e.yml up -d
+
+start-e2e: export HOST_UID=$(id -u)
+start-e2e: export HOST_GID=$(id -g)
+start-e2e: start-docker-e2e
+	docker exec oasis-indexer sh -c "cd /oasis-indexer && make test-e2e"
```
Andrew7234, line 84: Multiple targets for `start-e2e`; typo? The last rule will overwrite the others.

Also nit: I'm planning to update the dockerfile with `WORKDIR /oasis-indexer` so I'll hopefully be able to simplify the command

Makefile
```diff
@@ -28,13 +28,19 @@ docker:
...
 test:
 	@$(GO) test -short -v ./...
 
 test-ci:
 	@$(GO) test -race -coverpkg=./... -coverprofile=coverage.txt -covermode=atomic -v ./...
 
-test-e2e: export OASIS_INDEXER_E2E = true 
+# Run the e2e tests locally, assuming the environment is already set up.
+# The recommended usage is via the `start-e2e` target which sets up the environment first.
+test-e2e: export OASIS_INDEXER_E2E = true
```
Andrew7234, line 43: is this a duplicate rule?

tests/README.md
```diff
@@ -1,12 +1,38 @@
...
+For the canonical way to run these tests, see the [CI config](../../.github/workflows/ci-test.yaml).
+Here are some useful one-liners derived from those commands, and adapted for running locally and repeatedly.
+
+#### Build docker containers with current code
+```sh
+make docker
+```
+
+#### Run the e2e tests
+You can run tests from the `oasis-indexer` container:
```
Andrew7234, line 24:
```suggestion
This line will clean up any prior runs and start a new run of tests inside the `oasis-indexer` container:
```

Makefile
Andrew7234: Oh line 44 also defines a `test-e2e` rule, which will overwrite this one. You probably want something like 
```
OASIS_INDEXER_E2E=true @$(GO) test -race -coverpkg=./... -coverprofile=coverage.txt -covermode=atomic -v ./tests/e2e
```

aefhm: Uhhh, somehow the line 44 version works and the below doesn't?

```
OASIS_INDEXER_E2E=true @$(GO) test -race -coverpkg=./... -coverprofile=coverage.txt -covermode=atomic -v ./tests/e2e
```

mitjat: Line 44 doesn't redefine/overwrite the rule because line 43 is not a rule, it's more like an annotation of the rule. This is how you export env variables to make targets. See [official docs](https://www.gnu.org/software/make/manual/html_node/Target_002dspecific.html) and [a more readable discussion](https://stackoverflow.com/questions/23843106/how-to-set-child-process-environment-variable-in-makefile)

The line in the comment just above mine (assuming you tried it verbatim) doesn't work because the `@` is Makefile syntax for "don't log this command just before you execute it"; you want `@OASIS_INDEXER_E2E=true $(GO) test [...]`

I'm fine with either syntax :)

Makefile
aefhm: Removed as unused.

Makefile
aefhm: Thanks. I learned something new.

PR #193, opened 2022-10-19
(no merge) consensus: use wrong permissions

pro-wh:
investigating gosec issue

### DIFF
--- a/analyzer/consensus/consensus.go
+++ b/analyzer/consensus/consensus.go
@@ -242,7 +242,7 @@ func (m *Main) processGenesis(ctx context.Context) error {
 	debugPath := os.Getenv("CONSENSUS_DAMASK_GENESIS_DUMP")
 	if debugPath != "" {
 		sql := strings.Join(queries, "\n")
-		if err := os.WriteFile(debugPath, []byte(sql), 0o600 /* Permissions: rw------- */); err != nil {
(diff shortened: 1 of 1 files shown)


mitjat:
FYI, you can `make lint-go` locally to run the linter.
Alas, you have to manually make sure you have the "right" versions of go, golangci-lint, and gofumpt (= same versions as in our `.github/workflows`) to get the same results as in CI.

analyzer/consensus/consensus.go
```diff
@@ -242,7 +242,7 @@ func (m *Main) processGenesis(ctx context.Context) error {
 	debugPath := os.Getenv("CONSENSUS_DAMASK_GENESIS_DUMP")
 	if debugPath != "" {
 		sql := strings.Join(queries, "\n")
-		if err := os.WriteFile(debugPath, []byte(sql), 0o600 /* Permissions: rw------- */); err != nil {
+		if err := os.WriteFile(debugPath, []byte(sql), 0o466 /* lmao let everyone else write but me */); err != nil {
```
aefhm, line 245: 🤣

pro-wh:
https://github.com/securego/gosec/pull/883 now that'll gradually make its way through a gosec release and then a version bump in golangci-lint

pro-wh:
it'll be a while before gosec makes a new release and a while before golangci-lint picks that up and a while before we update our golangci-lint version. but it'll all happen eventually and we don't especially need to hold our breaths

PR #201, opened 2022-10-21
 Emerald: store transaction result 

pro-wh:
we're about to add the transactions api, where we'll want to look at the tx results. I'm proposing that we store a copy in the db so that we don't have to bother the grpc server when answering indexer clients. we do this for the transactions already, i.e. we have the inputs but not the outputs.

### DIFF
--- a/storage/migrations/02_oasis_3_emerald.up.sql
+++ b/storage/migrations/02_oasis_3_emerald.up.sql
@@ -35,6 +35,8 @@ CREATE TABLE oasis_3.emerald_transactions
   -- later store sufficiently detailed data in other columns or if we turn out
   -- to be able to get a copy of the transaction elsewhere.
   raw         BYTEA NOT NULL,
+  -- result_raw is cbor(CallResult).
+  result_raw  BYTEA NOT NULL,
   PRIMARY KEY (round, tx_index)
 );
 
--- a/storage/postgres/client.go
+++ b/storage/postgres/client.go
@@ -4,6 +4,7 @@ package postgres
 
 import (
 	"context"
+	"fmt"
 
 	"github.com/jackc/pgx/v4"
 	"github.com/jackc/pgx/v4/pgxpool"
@@ -93,7 +94,7 @@ func (c *Client) SendBatch(ctx context.Context, batch *storage.QueryBatch) error
 		defer common.CloseOrLog(batchResults, c.logger)
 		for i := 0; i < pgxBatch.Len(); i++ {
 			if _, err := batchResults.Exec(); err != nil {
-				return err
+				return fmt.Errorf("query %d %v: %w", i, batch.Queries()[i], err)
 			}
 		}
 
--- a/analyzer/emerald/incoming.go
+++ b/analyzer/emerald/incoming.go
@@ -34,6 +34,7 @@ type BlockTransactionData struct {
 	Hash                    string
 	EthHash                 *string
 	Raw                     []byte
+	RawResult               []byte
 	SignerData              []*BlockTransactionSignerData
 	RelatedAccountAddresses map[string]bool
 }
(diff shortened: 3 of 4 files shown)


storage/migrations/02_oasis_3_emerald.up.sql
```diff
@@ -35,6 +35,8 @@ CREATE TABLE oasis_3.emerald_transactions
   -- later store sufficiently detailed data in other columns or if we turn out
   -- to be able to get a copy of the transaction elsewhere.
   raw         BYTEA NOT NULL,
+  -- result_raw is cbor(CallResult).
+  result_raw  BYTEA NOT NULL,
```
mitjat, line 39: (Just a thought, not a suggestion to change anything in this PR)

Might be nice to have both of these be the native `JSON` type in postgres. We'd gain human-readable column values, and the ability to perform ad-hoc queries/analyses over "raw" data in psql. OTOH we risk losing some esoteric cbor nuance that gets lost when converting to json.

mitjat (approved):
:+1:

storage/postgres/client.go
```diff
@@ -93,7 +94,7 @@ func (c *Client) SendBatch(ctx context.Context, batch *storage.QueryBatch) error
 		defer common.CloseOrLog(batchResults, c.logger)
 		for i := 0; i < pgxBatch.Len(); i++ {
 			if _, err := batchResults.Exec(); err != nil {
-				return err
+				return fmt.Errorf("query %d: %w", i, err)
```
mitjat, line 97: If you want, you could log `batch.Queries()[i]`. On one hand, too verbose to be the actual err message, but OTOH very helpful when we have batches of 200 queries and the error message says effectively "refer to query 132 in the above huge list"

Andrew7234 (approved):

aefhm (commented):
Looks like you are 2/1 for code conspirators :)

storage/migrations/02_oasis_3_emerald.up.sql
pro-wh: that might be nice. also if someone could come up with a way to deep parse RawMessage innards, e.g. the tx in the utx, and the body in the tx and get it all into json. similarly for the result

storage/postgres/client.go
pro-wh: [i] would be good

storage/migrations/02_oasis_3_emerald.up.sql
mitjat: Agreed. Not in this PR though, right? Filed #207 

> the tx in the utx
That what your [`OpenUtxNoVerify()`](https://github.com/oasisprotocol/oasis-indexer/blob/09dd2b5dd4a74d381ead68151ac0d42ec6aa5c59/analyzer/uncategorized/runtimes.go#L17-L17), I thought? Are there other special cases besides evm that we need to cover?

> the body in the tx
that's just cbor, so it's easy to decode generically, right?

pro-wh: oh shoot, ethereum format transactions :thinking:

storage/postgres/client.go
pro-wh: formatting with %v

storage/migrations/02_oasis_3_emerald.up.sql
mitjat: Yeah for eth format we'd have to either store the JSON version of [`decodeEthRawTx()`](https://github.com/oasisprotocol/oasis-indexer/blob/09dd2b5dd4a74d381ead68151ac0d42ec6aa5c59/analyzer/uncategorized/evm_raw_tx.go#L14), or else store both the raw byte[] _and_ the JSON ...

Issue #192, opened 2022-10-19
Fix tracking of escrow slashing


mitjat:
(This investigation into how the indexer tracks escrowed tokens came from https://github.com/oasisprotocol/oasis-indexer/issues/179#issuecomment-1279686551.)

The indexer tracks escrow balances (and shares) by observing events. Most of these events can be processed in an order-independent manner, but one update query currently cannot. The query is executed when we observe the `TakeEscrowEvent`, i.e. slashing of escrowed tokens:
https://github.com/oasisprotocol/oasis-indexer/blob/4aa5a529147586a1316c2c5f25dd7ccc56e1ac58/analyzer/queries.go#L207-L211

It appears that the query has several issues. Note that the `$2` in that query is the [`Amount` in `TakeEscrowEvent`](https://github.com/oasislabs/oasis-core/blob/531421a928dc3450abd721c832d376549626a473/docs/consensus/services/staking.md#L660-L660) documented as "the amount (in base units) taken (= slashed)".

Questions to make sure we fix this correctly:

1) [The query we're discussing](https://github.com/oasisprotocol/oasis-indexer/blob/4aa5a529147586a1316c2c5f25dd7ccc56e1ac58/analyzer/queries.go#L209-L209) grabs this amount field, performs shares-derivation math on it, and then updates the `balance` (!) fields. Is this buggy, and it should be updating `escrow_total_shares_*` to the value that it currently computes?
2) Should that same query additionally be updating the balances by setting `UPDATE ... SET escrow_balance_active -= {{.Take.Amount}}, escrow_balance_debonding -= {{.Take.Amount}}`?
3) Is that same query buggy in a separate way in that it subtracts `$2` from both `*_active` _and_ `*_debonding`? The [oasis-core implementation](https://github.com/oasislabs/oasis-core/blob/531421a928dc3450abd721c832d376549626a473/go/consensus/tendermint/apps/staking/state/state.go#L746-L752) seems to take/slash as much as possible from the active escrow, and if there's not enough money there, then it takes the remainder from the debonding escrow. If so, I guess that `TakeEscrowEvent` should be reporting `ActiveAmount` and `DebondingAmount` separately, rather than reporting their sum in a single `Amount` field?
[...]

PR #210, opened 2022-11-03
Change delegations rounding->floor

Andrew7234:
#192 

The validator delegations balances are slightly too low. Current hypothesis is that the rounding in `ConsensusTakeEscrowUpdateQuery` is the root cause. Since postgres `numeric` types support arbitrary precision, we should instead do the rounding at the API layer. Testing still in progress.

### DIFF
--- a/analyzer/queries.go
+++ b/analyzer/queries.go
@@ -221,8 +221,8 @@ func (qf QueryFactory) ConsensusTakeEscrowUpdateQuery() string {
 	return fmt.Sprintf(`
 		UPDATE %s.accounts
 			SET
-				escrow_balance_active = escrow_balance_active - ROUND($2 * escrow_balance_active / (escrow_balance_active + escrow_balance_debonding)),
-				escrow_balance_debonding = escrow_balance_debonding - ROUND($2 * escrow_balance_debonding / (escrow_balance_active + escrow_balance_debonding))
+				escrow_balance_active = escrow_balance_active - ($2 * escrow_balance_active / (escrow_balance_active + escrow_balance_debonding)),
+				escrow_balance_debonding = escrow_balance_debonding - ($2 * escrow_balance_debonding / (escrow_balance_active + escrow_balance_debonding))
 			WHERE address = $1`, qf.chainID)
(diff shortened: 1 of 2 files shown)


mitjat:
This code in SQL emulates the slashing math in oasis-core. Slashing starts [here](https://github.com/oasisprotocol/oasis-core/blob/c16c3af7d5f8e41bdcc74ee05d576ccee7df1b34/go/consensus/tendermint/apps/staking/state/state.go#L723-L723), then a couple layers in, it [calls `.Quo()`](https://github.com/oasisprotocol/oasis-core/blob/c16c3af7d5f8e41bdcc74ee05d576ccee7df1b34/go/consensus/tendermint/apps/staking/state/state.go#L706-L706) for whole-integer division. So integer ~rounding~ truncation does happen in oasis-core also.

In indexer SQL, the `ROUND` was probably an unsuccessful attempt at emulating this behavior. `FLOOR` would emulate it more faithfully. I first thought even `FLOOR` was not needed because in SQL, division of integers already returns an integer. However the columns here are `NUMERIC`, which can include decimals.

Looks like we should
- replace `ROUND` with `FLOOR`
- change the type of `escrow_*` fields [here](https://github.com/oasisprotocol/oasis-indexer/blob/305095e8b1c04916924b517eeca4405d97586a90/storage/migrations/01_oasis_3_consensus.up.sql#L151-L151) to be `NUMERIC(1000,0)`, i.e. an integer (0 [means no decimal places](https://www.postgresql.org/docs/current/datatype-numeric.html)).

mitjat:
In addition to that, my (unconfirmed) working theory from https://github.com/oasisprotocol/oasis-indexer/issues/192 is that the bookkeeping here might be completely wrong; we shouldn't be (re)calculating anything, and instead just decreasing `escrow_balance_{active,debonding}` by the `Amount` field from the event. But I'm not sure if that amount is relevant to only `escrow_balance_active` or also to `escrow_balance_debonding`; let me spelunk again.

mitjat (approved):
Thanks!

PR #212, opened 2022-11-10
Andrew7234/genesis test

Andrew7234:
Adds a dockerfile for the genesis-test indexer healthcheck. 

Edit*: Adapts the existing indexer dockerfile for use in the indexer statecheck

Related: https://github.com/oasislabs/private-ops/pull/2596


docker/indexer/healthcheck.Dockerfile
```diff
@@ -0,0 +1,25 @@
+# This image is currently pushed to the oasislabs/oasis-indexer:healthcheck tag
```
mitjat, line 1: Can we reuse the "regular" `oasislabs/oasis-indexer` image? Looks like the only differences are
- the apt, which we could easily have in the regular image too
- the env tag, which we can hopefully set outside the image, in the k8s `CronJob`

If we cannot reuse the same image, I'd strongly prefer to push to something other than `oasislabs/oasis-indexer:<whatever>`; it feels messy to have two disparate families of images under the same name.

Andrew7234: Sounds good! I initially assumed that k8s couldn't override the `ENTRYPOINT` but turns out that's [not the case](https://kubernetes.io/docs/tasks/inject-data-application/define-command-argument-container/#define-a-command-and-arguments-when-you-create-a-pod)

aefhm (approved):

PR #213, opened 2022-11-12
Emerald: track tokens

pro-wh:
we already have the plumbing to run something on each erc20 Transfer event. this gathers them into a token_balances table.

details:
- transfers from the zero ethereum address count as mints
- transfers to the zero ethereum address count as burns
- addresses in the table are oasis addresses. look up ethereum addresses as needed in the address preimages table

other notes:
- this doesn't count holders. I'll ask about how to implement this
- there's no enforcement about starting from genesis-----you can start whenever, and you may have negative "balance" where it only shows the total of all in-minus-out during the time scanned
- no nothin done about contracts using erc20-compatible Transfer events in any wrong way

see design https://app.clickup.com/t/3u9y034?comment=1272147909

### DIFF
--- a/analyzer/emerald/incoming.go
+++ b/analyzer/emerald/incoming.go
@@ -170,7 +170,6 @@ func registerRelatedAddressSpec(addressPreimages map[string]*AddressPreimageData
 	return addr, nil
 }
 
-//nolint:unparam
 func registerRelatedEthAddress(addressPreimages map[string]*AddressPreimageData, relatedAddresses map[string]bool, ethAddr []byte) (string, error) {
 	addr, err := registerEthAddress(addressPreimages, ethAddr)
 	if err != nil {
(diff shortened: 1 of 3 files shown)


storage/migrations/02_oasis_3_emerald.up.sql
```diff
@@ -91,6 +91,14 @@ CREATE TABLE oasis_3.address_preimages
     address_data       BYTEA NOT NULL
 );
 
+CREATE TABLE oasis_3.emerald_token_balances
+(
+  token_address TEXT NOT NULL,
+  account_address TEXT NOT NULL,
+  balance NUMERIC NOT NULL,
```
mitjat, line 98: Nit: `NUMERIC(1000,0)` to enforce integers; the 0 means no decimals.

analyzer/emerald/incoming.go
```diff
@@ -45,13 +46,19 @@ type AddressPreimageData struct {
 	Data              []byte
 }
 
+type TokenChangeKey struct {
```
mitjat, line 49: Nit: `tokenChangeKey`, i.e. private?

mitjat (approved):
Nice!

analyzer/emerald/incoming.go
```diff
@@ -45,13 +46,19 @@ type AddressPreimageData struct {
...
+}
+
 type BlockData struct {
 	Hash             string
 	NumTransactions  int
 	GasUsed          uint64
 	Size             int
 	TransactionData  []*BlockTransactionData
 	AddressPreimages map[string]*AddressPreimageData
+	TokenChanges     map[TokenChangeKey]*big.Int
```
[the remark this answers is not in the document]
pro-wh: negative numbers needed in this use case. can't use Quantity for that reason

analyzer/emerald/incoming.go
```diff
@@ -45,13 +46,19 @@ type AddressPreimageData struct {
 	Data              []byte
 }
 
+type TokenChangeKey struct {
+	ContractAddress string
+	AccountAddress  string
```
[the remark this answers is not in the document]
pro-wh: line 50 :+1: but ERC-20 will have to be in an "e.g."

line 51 :+1:

storage/migrations/02_oasis_3_emerald.up.sql
pro-wh: ok

analyzer/emerald/incoming.go
pro-wh: it's part of public BlockData

pro-wh:
oh shoot I used "contract address" in the struct but "token address" in the db

storage/migrations/02_oasis_3_emerald.up.sql
pro-wh: changed, will need further change since we since moved to types like uint_numeric

pro-wh: #223 to track that

analyzer/emerald/incoming.go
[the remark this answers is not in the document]
pro-wh: added

analyzer/emerald/incoming.go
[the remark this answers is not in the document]
pro-wh: declining suggestion

analyzer/emerald/incoming.go
pro-wh: declining suggestion

PR #214, opened 2022-11-15
cmd: flag for wiping the db on startup

mitjat:
This PR allows one to wipe the DB with a config flag:
```
analysis:
  analyzers: [...]
  storage:
    backend: postgres
    DANGER__wipe_storage_on_startup: true
```

I found it handy in my testing. It's also useful for hypothetical scenarios in staging/prod where we know we have to reindex, and a flag is better than doing manual surgery.

The PR also includes some very minor unrelated logging improvements.

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -202,6 +202,10 @@ type StorageConfig struct {
 
 	// Backend is the storage backend to select.
 	Backend string `koanf:"backend"`
+
+	// If true, we'll first delete all tables in the DB to
+	// force a full re-index of the blockchain.
+	WipeStorage bool `koanf:"DANGER__WIPE_STORAGE_ON_STARTUP"`
 }
 
 // Validate validates the storage configuration.
--- a/storage/postgres/client.go
+++ b/storage/postgres/client.go
@@ -138,3 +138,26 @@ func (c *Client) Shutdown() {
 func (c *Client) Name() string {
 	return moduleName
 }
+
+// Wipe removes all contents of the database.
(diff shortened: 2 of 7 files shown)


config/config.go
```diff
@@ -202,6 +202,10 @@ type StorageConfig struct {
 
 	// Backend is the storage backend to select.
 	Backend string `koanf:"backend"`
+
+	// If true, we'll first delete all tables in the DB to
+	// force a full re-index of the blockchain.
+	WipeStorage bool `koanf:"DANGER__wipe_storage_on_startup"`
```
aefhm, line 208: I'd be cool with all caps too.

storage/postgres/client.go
```diff
@@ -138,3 +138,26 @@ func (c *Client) Shutdown() {
 func (c *Client) Name() string {
 	return moduleName
 }
+
+// Wipe removes all contents of the database.
+func (c *Client) Wipe(ctx context.Context) error {
+	rows, err := c.Query(ctx, `
+		SELECT schemaname, tablename
+		FROM pg_tables
+		WHERE schemaname IN ('oasis_3', 'indexer', 'public')
```
aefhm, line 147: Not just `oasis_3`?

aefhm (approved):

config/config.go
mitjat: I take it that's your preference? Sure, I'm fine either way. Uppercased.

aefhm: 😬 soft preference, but yes, you called it.

PR #215, opened 2022-11-16
Emerald: add index for token holders count

pro-wh:
here's the index, but I haven't made a huge table to see if the EXPLAIN is good

```
indexer=# analyze oasis_3.emerald_token_balances;
ANALYZE
indexer=# explain select token_address, count(*) as num_holders from oasis_3.emerald_token_balances where balance != 0 group by token_address;
                                  QUERY PLAN                                   
-------------------------------------------------------------------------------
 HashAggregate  (cost=6.95..7.22 rows=27 width=55)
   Group Key: token_address
   ->  Seq Scan on emerald_token_balances  (cost=0.00..6.25 rows=140 width=47)
         Filter: (balance <> '0'::numeric)
(4 rows)
```

reviewers: will this work?

### DIFF
--- a/storage/migrations/02_oasis_3_emerald.up.sql
+++ b/storage/migrations/02_oasis_3_emerald.up.sql
@@ -106,6 +106,8 @@ CREATE TABLE oasis_3.emerald_token_balances
   PRIMARY KEY (token_address, account_address)
 );
 
+CREATE INDEX ix_emerald_token_address ON oasis_3.emerald_token_balances (token_address) WHERE balance != 0;
+
 -- Core Module Data
 CREATE TABLE oasis_3.emerald_gas_used
 (


storage/migrations/02_oasis_3_emerald.up.sql
```diff
@@ -99,6 +99,8 @@ CREATE TABLE oasis_3.emerald_token_balances
   PRIMARY KEY (token_address, account_address)
 );
 
+CREATE INDEX ix_emerald_token_balances_token_address ON oasis_3.emerald_token_balances (token_address, account_address) WHERE balance != 0;
```
mitjat, line 102: We can actually simplify to
```suggestion
CREATE INDEX ix_emerald_token_address ON oasis_3.emerald_token_balances (token_address) WHERE balance != 0;
```

I didn't notice before. For some reason I thought we'd doing `SELECT count(account_address)` (which means "count rows with non-null account_address"), and then we'd probably want the index to span `account_address` too. But because `account_address` should never be `NULL`, it's enough to just count the number of rows overall with `count(*)`, as you implemented it :+1: 

I think this should work, but postgres is complex, so no promises. I suggest you psql into the prod indexer, create the index, and run the query.

mitjat: Argh you pointed out that the tokens table is not yet merged, much less populated in prod.

You can create reasonably realistic fake data with
```
insert into oasis_3.emerald_token_balances(token_address, account_address, balance) select md5(round(random_exp()*5)::text), md5(random()::text), round(random()*50) from generate_series(1,10) s(i);
```
where `random_exp` comes from [here](https://bugfactory.io/blog/generating-random-numbers-according-to-a-continuous-probability-distribution-with-postgresql/)

We want token addrs exponentially distributed because in reality, it's highly likely they will be too. Holder addrs don't matter. Balances are selected to be non-zero most of the time, but not always.

mitjat: It works :+1: 
```sql
indexer=# CREATE OR REPLACE FUNCTION random_exp( [...]
CREATE FUNCTION
indexer=# CREATE TABLE oasis_3.emerald_token_balances [...]
CREATE TABLE
indexer=# CREATE INDEX ix_emerald_token_address ON oasis_3.emerald_token_balances (token_address) WHERE balance != 0;
CREATE INDEX
indexer=# insert into oasis_3.emerald_token_balances(token_address, account_address, balance) select md5(round(random_exp()*5)::text), md5(random()::text), round(random()*50) from generate_series(1,100000);
INSERT 0 100000
indexer=# explain select token_address, count(*) as num_holders from oasis_3.emerald_token_balances where balance != 0 group by token_address;
                                     QUERY PLAN                                     
------------------------------------------------------------------------------------
 HashAggregate  (cost=1709.80..1711.80 rows=200 width=40)
   Group Key: token_address
   ->  Seq Scan on emerald_token_balances  (cost=0.00..1574.62 rows=27034 width=32)
         Filter: (balance <> '0'::numeric)
(4 rows)

indexer=# analyze oasis_3.emerald_token_balances ;
ANALYZE
indexer=# explain select token_address, count(*) as num_holders from oasis_3.emerald_token_balances where balance != 0 group by token_address;
                                                        QUERY PLAN                                                        
--------------------------------------------------------------------------------------------------------------------------
 GroupAggregate  (cost=0.29..2365.36 rows=47 width=41)
   Group Key: token_address
   ->  Index Only Scan using ix_emerald_token_address on emerald_token_balances  (cost=0.29..1869.74 rows=99030 width=33)
```

mitjat (approved):

storage/migrations/02_oasis_3_emerald.up.sql
pro-wh: thanks, I'll go ahead and remove the `, account_address`

PR #216, opened 2022-11-16
[api] check for divide-by-zero

Andrew7234:
If a validator's stake (ie stake delegated to it) drops to zero, trying to fetch the account info of any account that ever delegated stake to that validator will trigger a divide by 0 error.

Example: 
https://index.oasislabs.com/v1/consensus/accounts/oasis1qr8znzva3h28hausy7n3fput0fn80sajeywmwp5y

### DIFF
--- a/storage/client/queries.go
+++ b/storage/client/queries.go
@@ -121,18 +121,18 @@ func (qf QueryFactory) AccountsQuery() string {
 func (qf QueryFactory) AccountQuery() string {
 	return fmt.Sprintf(`
 		SELECT address, nonce, general_balance, escrow_balance_active, escrow_balance_debonding,
-			(
-				SELECT COALESCE(ROUND(SUM(shares * escrow_balance_active / escrow_total_shares_active)), 0) AS delegations_balance
+			COALESCE (
+				(SELECT COALESCE(ROUND(SUM(shares * escrow_balance_active / escrow_total_shares_active)), 0) AS delegations_balance
 				FROM %[1]s.delegations
 				JOIN %[1]s.accounts ON %[1]s.accounts.address = %[1]s.delegations.delegatee
-				WHERE delegator = $1::text
-			) AS delegations_balance,
-			(
-				SELECT COALESCE(ROUND(SUM(shares * escrow_balance_debonding / escrow_total_shares_debonding)), 0) AS debonding_delegations_balance
+				WHERE delegator = $1::text AND escrow_total_shares_active != 0)
+			, 0) AS delegations_balance,
+			COALESCE (
+				(SELECT COALESCE(ROUND(SUM(shares * escrow_balance_debonding / escrow_total_shares_debonding)), 0) AS debonding_delegations_balance
 				FROM %[1]s.debonding_delegations
 				JOIN %[1]s.accounts ON %[1]s.accounts.address = %[1]s.debonding_delegations.delegatee
-				WHERE delegator = $1::text
-			) AS debonding_delegations_balance
+				WHERE delegator = $1::text AND escrow_total_shares_debonding != 0)
+			, 0) AS debonding_delegations_balance
 			FROM %[1]s.accounts
 			WHERE address = $1::text`, qf.chainID)
 }


storage/client/queries.go
```diff
@@ -125,13 +125,13 @@ func (qf QueryFactory) AccountQuery() string {
 				SELECT COALESCE(ROUND(SUM(shares * escrow_balance_active / escrow_total_shares_active)), 0) AS delegations_balance
 				FROM %[1]s.delegations
 				JOIN %[1]s.accounts ON %[1]s.accounts.address = %[1]s.delegations.delegatee
-				WHERE delegator = $1::text
+				WHERE delegator = $1::text AND escrow_total_shares_active != 0
```
mitjat, line 128: If the subselect returns 0 rows, `delegations_balance` will be `NULL`, which is probably not a good output. I think we want another `COALESCE(..., 0)` around each subselect, WDYT?

This was already a bug before this PR, but now it's slightly more likely to happen because of the additional `WHERE`.

mitjat (approved):
LGTM for the actual fix, thank you! I have another small related fix suggestion.

storage/client/queries.go
Andrew7234: Good call! Looks like the current behavior causes those `NULL`s to become missing fields in the returned `Account`, which makes sense since the row scan returns `null` to the struct fields. Example: https://index.oasislabs.com/v1/consensus/accounts/oasis1qzzwkrjmyrzey46hyste5xszue3ggy90ag7k3687
(missing debonding_delegations_balance)

Will add and update

mitjat: @pro-wh you can use CASE but that still leaves the cases where the inner select returns 0 rows.

This is all nitpicking really, and there will be no externally observable change after what I propose.

From  `Account` [definition](https://github.com/oasisprotocol/oasis-indexer/blob/6c5738c679fbd03113c2e3e5a0a68e37f16e7a38/storage/client/types.go#L88-L88):
```
DebondingDelegationsBalance uint64 `json:"debonding_delegations_balance,omitempty"`
```

Two things to notice:
- The field is an `uint64`, not `*uint64`. So IMO it's somewhat ugly that pgx is even willing to map the `NULL` onto an int. At best, it's a reasonable but not obvious (= the _only_ reasonble) choice. Erroring would be another choice. By always returning non-null here, we avoid the ambiguity.
- `omitempty` will presumably continue to cause the field to be omitted in the JSON response when the value is 0.

pro-wh: o true

why are we declaring ours omitempty btw? I think in oasis-core it was a hack to save some storage space, but our storage client types aren't persisted

> So IMO it's somewhat ugly that pgx is even willing to map the NULL onto an int. At best, it's a reasonable but not obvious (= the only reasonble) choice. Erroring would be another choice. By always returning non-null here, we avoid the ambiguity.

you mean through ScanRow? bleh that is kinda ugly

mitjat: > why are we declaring ours omitempty btw?

Seems wrong to me too. Let's remove the `omitempty`.

@Andrew7234 wanna do it in this PR? I was going to create a separate `omitempty` cleanup PR. But then I just went thru `omitempty` annotations in https://github.com/oasisprotocol/oasis-indexer/blob/790cc87862a54c91ab30634daa29ded409b0fa1c/storage/client/types.go (there's only 9 of them), and the others seem justified. The least obvious to me was the `Proposal` struct which has several of them; they come from the db where [this comment](https://github.com/oasisprotocol/oasis-indexer/blob/790cc87862a54c91ab30634daa29ded409b0fa1c/storage/migrations/01_oasis_3_consensus.up.sql#L230-L230) explains why they're nullable.

pro-wh: > If this proposal is a new proposal.

> new

?

aefhm: > Also, why is the [Address field in Entity](https://github.com/oasisprotocol/oasis-indexer/blob/790cc87862a54c91ab30634daa29ded409b0fa1c/storage/client/types.go#L52) is optional?

I think the address may not be known at time of insertion.

aefhm (approved):

PR #218, opened 2022-11-18
Emerald: add tokens api

pro-wh:
here's an endpoint to do an index scan for most-number-of-holders tokens

all we have right now is the token contract address (oasis address) and that number of holders

### DIFF
--- a/api/v1/client.go
+++ b/api/v1/client.go
@@ -670,6 +670,21 @@ func (c *storageClient) RuntimeTransactions(ctx context.Context, r *http.Request
 	return &apiTransactions, err
 }
 
+func (c *storageClient) RuntimeTokens(ctx context.Context, r *http.Request) (*storage.RuntimeTokenList, error) {
+	var q storage.RuntimeTokensRequest
+
+	p, err := common.NewPagination(r)
+	if err != nil {
+		c.logger.Info("pagination failed",
+			"request_id", ctx.Value(storage.RequestIDContextKey),
+			"err", err.Error(),
+		)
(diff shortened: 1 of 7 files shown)


mitjat (approved):
Thanks!

> all we have right now is the token contract address (oasis address) and that number of holders

We also have enough data to show the total token supply. You could choose to implement it here without an index (and add one later), or not expose it until we're super efficient about it.
I did some testing:
- An index actually won't help; you can't instruct it to keep aggregates like SUMs in index nodes (assuming it's a tree). We need a table (token_addr, total_supply) and a trigger to update it automatically.
- Even without indexes/triggers, postgres knows that SUM is associative, so it parallelizes the crap out of it and works fast enough for up to ~1M token holders on my laptop (0.5s for 10M holders). So it's not too bad to kick the can down the road wrt optimizations.

pro-wh:
erc-20 has a method for total supply. I'm hoping not to have to do our own summing

mitjat:
> erc-20 has a method for total supply. I'm hoping not to have to do our own summing

That similar to how oasis-node has a method for an account's balance. We still do summing on the indexer side (based on transfers/mints/burns) because it saves us many RPCs.
For wrapped tokens, the total supply can change very frequently, so if we just query the ERC-20 contract once an hour or something, it can be pretty off.

pro-wh:
ya but I'm afraid of us being too optimistic about Transfer events working well

pro-wh:
we can run the request in the analyzer on each block where there are any transfers

PR #221, opened 2022-11-18
bugfix: stringify big.Int and uint64 before sending to pgx

mitjat:
This is a forward fix for the e2e tests broken in #217.

The underlying problem was a bad uint64 cast from go to sql. The e2e test uncovered one such cast, but then I found many more. Take the following as an example:
```
batch.Queue(transactionInsertQuery, [...],
			tx.Fee.Amount.ToBigInt().Uint64(),
			tx.Fee.Gas, [...]
)
```
This represents both types of errors fixed by the PR:
- Amount is an arbitrary-precision int, both in core and in the db. We simply ignore any values over uint64 here. That's pretty bad, though the e2e test didn't actually catch this. 
- Gas is `uint64` in core, and `uint_numeric` (custom type, defined as `numeric(1000,0) CHECK(VALUE>=0)` ) in the db. Postgres doesn't have uint64, it only has int64. Before #217, pgx somehow recognized that the go `uint64` was being used to fill a `numeric` db field, and presented it to postgres as a string (?). Now that the db field is a custom type, pgx doesn't want to force stringification, so it instead tries to represent the `uint64` to postgres as `Int8` (pgx's internal name for `int64`, as in "8 bytes"). For values above 1<<63, this fails.

The **fix in this PR** for both cases is to manually serialize the go value to string before sending it to pgx. This works, but it's brittle, because we have to remember to do it every time we write sql queries. Possible other fixes:
- Add explicit casts in sql statements (`... $1::numeric ...`); maybe that will help pgx. I only thought of this now, haven't tried it, also haven't tried to understand how pgx handles types internally.
- Introduce stronger typing to `queries.go`, so that every query-constructing fn would have a signature like `AddTransactionQuery(AddTransactionQueryParams params)`. We'd still have to get all the stringifying done manually and correctly, but it would be in a central spot, in `queries.go`.
- Use something other than pgx(?)

### DIFF
--- a/analyzer/consensus/consensus.go
+++ b/analyzer/consensus/consensus.go
@@ -403,8 +403,8 @@ func (m *Main) queueTransactionInserts(batch *storage.QueryBatch, data *storage.
 			signedTx.Hash().Hex(),
 			i,
 			tx.Nonce,
-			tx.Fee.Amount.ToBigInt().Uint64(),
-			tx.Fee.Gas,
+			tx.Fee.Amount.String(),
+			fmt.Sprintf("%d", tx.Fee.Gas),
 			tx.Method,
 			sender,
 			tx.Body,
@@ -664,16 +664,13 @@ func (m *Main) queueTransfers(batch *storage.QueryBatch, data *storage.StakingDa
 	receiverUpsertQuery := m.qf.ConsensusReceiverUpdateQuery()
 
 	for _, transfer := range data.Transfers {
-		from := transfer.From.String()
-		to := transfer.To.String()
-		amount := transfer.Amount.ToBigInt().Uint64()
 		batch.Queue(senderUpdateQuery,
-			from,
-			amount,
+			transfer.From.String(),
+			transfer.Amount.String(),
 		)
 		batch.Queue(receiverUpsertQuery,
-			to,
-			amount,
+			transfer.To.String(),
+			transfer.Amount.String(),
 		)
 	}
 
@@ -686,7 +683,7 @@ func (m *Main) queueBurns(batch *storage.QueryBatch, data *storage.StakingData) 
 	for _, burn := range data.Burns {
 		batch.Queue(burnUpdateQuery,
 			burn.Owner.String(),
-			burn.Amount.ToBigInt().Uint64(),
+			burn.Amount.String(),
 		)
 	}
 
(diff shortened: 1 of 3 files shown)


mitjat:
There is a large category of uint64 values where I didn't bother to stringify: `height`, `round`, `epoch` values. If any of those ever reaches 1<<63, I'll be happy even if the indexer then crashes :)

pro-wh (approved):
oh does gas really go above 1<<63?

analyzer/consensus/consensus.go
```diff
@@ -823,7 +819,7 @@ func (m *Main) queueValidatorUpdates(batch *storage.QueryBatch, data *storage.Sc
 	for _, validator := range data.Validators {
 		batch.Queue(validatorNodeUpdateQuery,
 			validator.ID.String(),
-			validator.VotingPower,
+			fmt.Sprintf("%d", validator.VotingPower),
```
pro-wh, line 822: what type was VotingPower?

mitjat: VotingPower is `int64` in go. I thought it was `uint64`. I think we can safely assume it's non-negative ... So I'm switching it in the db to `UINT31` (from `UINT_NUMERIC`), and removing the stringification here. Thanks!

mitjat:
> oh does gas really go above 1<<63?

I don't expect it to. But I'm still inclined to play it safe, i.e. represent it as `NUMERIC` in the db and stringify here. By contrast, all the `uint64`  values where I was OK with just assuming they won't really go past 1<<63 are either sequential in nature (height, round, epoch, nonce) or are really enums in disguise where it's unlikely that the most significant bit will be used (error codes).

storage/migrations/01_oasis_3_consensus.up.sql
```diff
@@ -123,7 +123,7 @@ CREATE TABLE oasis_3.nodes
   software_version TEXT,
 
   -- Voting power should only be nonzero for consensus validator nodes.
-  voting_power     BIGINT DEFAULT 0
+  voting_power     UINT31 DEFAULT 0
```
aefhm, line 126: (non-blocking) I could use some help understanding the type conversion. 😬🤔

aefhm: Edit. I see the referenced PR's comments.

PR #227, opened 2022-11-24
genesis_test: Numerous improvements

mitjat:
Speed improvements:
- Optionally use genesis dump from local file (crucial for faster testing) (see **NOTE**)
- Memoize chainId (speeds things up considerably)

Features and bugfixes:
- check that all node-tracked accounts are present in indexer (we only checked in the other direction before)
- bugfix: genesis_test: ignore reserved addresses (previously, the test complained because indexer had them and the node didn't)
- bugfix: interpret NULL values in runtimes table (bug introduced by #217 which stopped storing literal `"none"` from the node in the db, storing it instead as `NULL`. genesis_test wasn't expecting the NULL.)
- bugfix: fix TestBlocksSanityCheck (just needed to know the genesis round)
- bugfix: Makefile: start a stopped postgres if present (unrelated to the rest of PR, too small to warrant its own PR)

**NOTE:** To create a genesis dump at a given height, you have to use a locally compiled oasis-node. The released binary has a bug that ignores the `--height` flag and snapshots the most recent height instead. Example:
```
height=8050156; cd ~/oasis/oasis-core && make build-go && time ./go/oasis-node/oasis-node genesis dump --height $height --genesis.file ~/oasis/oasis-indexer/genesis@${height}.json -a unix:/tmp/node.sock
```

**NOTE 2:** genesis_test is still **BROKEN**. 1200 blocks in, the nodes are not tracked correctly. If I run the test at only one block after genesis, the test does pass. To be debugged later.

This is in preparation for parallelization in #179 -- we'll need to sanity-check any changes to processing.


Andrew7234 (approved):
Looks good! This should make debugging easier

Issue #211, opened 2022-11-04
investigate inconsistent account balances


Andrew7234:
Indexer:
https://index.oasislabs.com/v1/consensus/accounts/oasis1qzzwkrjmyrzey46hyste5xszue3ggy90ag7k3687

Oasisscan:
https://api.oasisscan.com/mainnet/chain/account/info/oasis1qzzwkrjmyrzey46hyste5xszue3ggy90ag7k3687

Notes:
@aefhm found this account while searching for validators with few delegators.

aefhm:
I think the `escrow_balance_debonding` math is probably wrong as seen on this account.

pro-wh:
wrong oasisscan link. that one shows outgoing delegations, indexer shows incoming delegations

mitjat:
We're focusing on incorrect balance tracking, but AFAIK the issue of incorrect node registry state also hasn't been fixed yet.
I ran bisection (over different heights) to figure out when indexer starts diverging from `StateToGenesis` output. The discrepancy is large-ish and happens all at once: Height 8050155 is GOOD (no discrepancy), 8050156 is BAD
```
                    Error:          Not equal:                                                                                                                                                                [4/45732]
                                expected: 130                                                                                                                                                 
                                actual  : 108                                                                                                                                                 
                Test:           TestGenesisFull                                                                                                                                               
                Messages:       wrong number of nodes                                                                                                                                         
    genesis_test.go:447: node g/gNUJ3L8NbJilg+ivA0JHXFoCpzXwedqSLO3Xd0FSM= expected, but not found                                                                                            
    genesis_test.go:447: node bsq05GjUhJU7CdNOGrI2satPmpAUjRmT0gAGbWyfhLw= expected, but not found                                                                                            
    genesis_test.go:447: node o+KFxCWnNyUstx2iCbMDvTsf+P/pDCC4lLxlljtOUeM= expected, but not found                                                                                            
    genesis_test.go:447: node bbkPZRl8KA//0joD1R9HPnucJ4GklCoAtUoJrincsoo= expected, but not found                                                                                            
    genesis_test.go:447: node DudLYGDVKMgh/OZ0rOsoxeCNzSxalm4bvQsrsE/H9XA= expected, but not found                                                                                            
    genesis_test.go:447: node J2vFhYaht7l1B++dBUTzqubT72WfOeeBkI8rOamrgGc= expected, but not found                                                                                            
[...]

Issue #159, opened 2022-10-07
Investigate account with inaccurate delegations balance


lukaw3d:
##### ACTUAL RESULTS

- https://index.oasislabs.com/v1/consensus/accounts/oasis1qz78ap0456g2rk7j6rmtvasc9v2kjhz2s58qgj90
  `"delegations_balance": 97397247798`

##### EXPECTED RESULTS

- https://api.oasisscan.com/mainnet/chain/account/info/oasis1qz78ap0456g2rk7j6rmtvasc9v2kjhz2s58qgj90
  `"escrow": "97.439537965"`

- https://grpc.oasis.dev/oasis-core.Staking/DelegationInfosFor
  ```sh
  curl 'https://grpc.oasis.dev/oasis-core.Staking/DelegationInfosFor' \
    -H 'accept: application/grpc-web-text' \
    -H 'content-type: application/grpc-web-text' \
    --data-raw 'AAAAACWiZW93bmVyVQC8foX1ppCh29LQ9rZ2GCsVaVxKhWZoZWlnaHQA'

  # https://lukaw3d.github.io/oasis-parse-grpc/#AAAAAFKhVQBmeOT33FDbsClmtGlEEMrtKPP7t6JkcG9vbKJnYmFsYW5jZUgELqJv4Ln132x0b3RhbF9zaGFyZXNIAzYcrNOsHV9mc2hhcmVzRRFrsbGHgAAAAB5ncnBjLXN0YXR1czowDQpncnBjLW1lc3NhZ2U6DQo%3D
  {
    "shares": "74.821251463",
    "pool": {
      "balance": "301,356,826.475754975",
      "total_shares": "231,403,959.467384159"
    }
  }
  ```
  `74.821251463 * 301356826.475754975 / 231403959.467384159 = 97.43953796526196`

lukaw3d:
oh, this looks like indexer hasn't indexed latest blocks yet (so staking rewards aren't caught up)
https://index.oasislabs.com/v1/ "latest_block":10639663
https://www.oasisscan.com/mainnet/dashboard/network "curHeight":10704286

ptrus:
The balance for the account is now correct compared to response via `DelegationInfosFor` query from the node. (`70.537114364 * 109164046.050101134 / 76083889.684323639` at the time of writing this).

PR #228, opened 2022-11-28
Update concurrency of process block goroutines

Andrew7234:
Warren pointed out that our `QueryBatch.Queue` was not thread-safe. This MR aims to remedy that. Part of a broader goal of tracking down inconsistent balances/db state in #211  and #159 

I've been running a version of this locally to hunt for negative balances. No issues as of yet

### DIFF
--- a/analyzer/consensus/consensus.go
+++ b/analyzer/consensus/consensus.go
@@ -275,7 +275,7 @@ func (m *Main) processBlock(ctx context.Context, height int64) error {
 
 	// Prepare and perform updates.
 	batch := &storage.QueryBatch{}
-	queries := make([]storage.QueryBatch, 6)
+	queries := make([]storage.QueryBatch, 5)
 
 	type prepareFunc = func(context.Context, int64, *storage.QueryBatch) error
 	for i, f := range []prepareFunc{
@@ -285,9 +285,9 @@ func (m *Main) processBlock(ctx context.Context, height int64) error {
 		m.prepareSchedulerData,
 		m.prepareGovernanceData,
 	} {
-		func(f prepareFunc, idx int) {
+		func(f prepareFunc, i int) {
 			group.Go(func() error {
-				return f(groupCtx, height, &queries[idx])
+				return f(groupCtx, height, &queries[i])
 			})
 		}(f, i)
 	}
@@ -311,7 +311,7 @@ func (m *Main) processBlock(ctx context.Context, height int64) error {
 
 	for i, b := range queries {
 		if b.Len() == 0 {
-			m.logger.Info(fmt.Sprintf("Block %d missing %d data", height, i))
+			m.logger.Info(fmt.Sprintf("Block %d goroutine %d emitted zero queries", height, i))
 			continue
 		}
 		batch.Append(&b)
--- a/analyzer/consensus/consensus.go
+++ b/analyzer/consensus/consensus.go
@@ -316,7 +316,7 @@ func (m *Main) processBlock(ctx context.Context, height int64) error {
 			m.logger.Debug(fmt.Sprintf("Block %d goroutine %d emitted zero queries", height, i))
 			continue
 		}
-		batch.Extend(&b)
+		batch.Extend(b)
 	}
 
 	opName := "process_block_consensus"
--- a/analyzer/consensus/consensus.go
+++ b/analyzer/consensus/consensus.go
@@ -275,7 +275,7 @@ func (m *Main) processBlock(ctx context.Context, height int64) error {
 
 	// Prepare and perform updates.
 	batch := &storage.QueryBatch{}
-	queries := make([]storage.QueryBatch, 5)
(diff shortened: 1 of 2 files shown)


analyzer/consensus/consensus.go
```diff
@@ -275,20 +275,21 @@ func (m *Main) processBlock(ctx context.Context, height int64) error {
...
 		m.prepareRegistryData,
 		m.prepareStakingData,
 		m.prepareSchedulerData,
 		m.prepareGovernanceData,
 	} {
-		func(f prepareFunc) {
+		func(f prepareFunc, i int) {
 			group.Go(func() error {
-				return f(groupCtx, height, batch)
+				return f(groupCtx, i, queries, height)
```
pro-wh, line 290: looks like these routines don't interact with other indices, maybe we could do something like

```go
queries := make([]storage.QueryBatch, 6)
...
f(groupCtx, height, &queries[i])
```

pro-wh: so that we don't have to change the function signatures

Andrew7234: Updated; thanks!

analyzer/consensus/consensus.go
```diff
@@ -275,20 +275,21 @@ func (m *Main) processBlock(ctx context.Context, height int64) error {
 
 	// Prepare and perform updates.
 	batch := &storage.QueryBatch{}
+	queries := make([]storage.QueryBatch, 6)
```
pro-wh, line 278: now that I look, aren't there only five

analyzer/consensus/consensus.go
```diff
@@ -308,6 +309,14 @@ func (m *Main) processBlock(ctx context.Context, height int64) error {
 		return err
 	}
 
+	for i, b := range queries {
+		if b.Len() == 0 {
+			m.logger.Info(fmt.Sprintf("Block %d missing %d data", height, i))
```
pro-wh, line 314: `missing 3 data` might be hard to understand

`Block %d goroutine %d emitted zero queries` :tipping_hand_person:

pro-wh (approved):
thx for fixing this

analyzer/consensus/consensus.go
Andrew7234: Nice, updated!

storage/api.go
```diff
@@ -46,6 +47,16 @@ func (b *QueryBatch) Queue(cmd string, args ...interface{}) {
 	})
 }
 
+// Append merges another batch into the current batch.
+func (b *QueryBatch) Append(qb *QueryBatch) {
```
mitjat, line 51: Nit: I would prefer to follow py nomenclature here, or at least not clash with it. In py, `.append(el)` is for appending a single element, and `.extend(lst)` is for appending all the elements of `lst`. Similarly in go, `append()` expects each element-to-be-appended as a a separate argument. So maybe `Extend` or `ExtendWith` here?

mitjat (approved):

analyzer/consensus/consensus.go
```diff
@@ -275,20 +275,21 @@ func (m *Main) processBlock(ctx context.Context, height int64) error {
 
 	// Prepare and perform updates.
 	batch := &storage.QueryBatch{}
+	queries := make([]storage.QueryBatch, 5)
```
mitjat, line 278: Instead of hardcoding the `5`, you can probably extend the array as you go:
```go
	queries := make([]storage.QueryBatch, 0)
...
		func(f prepareFunc) {
			batch := storage.QueryBatch{}
			queries = append(queries, &batch)
			group.Go(func() error {
				return f(groupCtx, height, &batch)
			})
		}(f)
```

Go has non-straightforward rules around references, loops, and goroutines, but I believe this is fine -- `batch` is created locally inside the loop, so each iteration has its own `batch`. As long as the garbage collector doesn't free it prematurely.

pro-wh: `append` may reallocate the array that backs `queries` and we'll lose writes

mitjat: Hm, `queries` only contains _pointers_ to batches (I see I specified the type incorrectly in `make` above; fixed now). So reallocations should move the pointers around but leave their values and the batches untouched, right?

pro-wh: o ok

Andrew7234: Updated! Thanks for the discussion

analyzer/consensus/consensus.go
```diff
@@ -285,8 +286,10 @@ func (m *Main) processBlock(ctx context.Context, height int64) error {
 		m.prepareGovernanceData,
 	} {
 		func(f prepareFunc) {
+			batch := storage.QueryBatch{}
```
pro-wh, line 289: wow the linter didn't complain about shadowing

PR #230, opened 2022-11-30
Allow partial configs, optionally run just API or just analyzers

mitjat:
This PR allows one to provide a "partial" config.yaml, e.g. without an `server:` or `analysis:` section (which configure the API server and the analyzers, respectively), and the binary will simply not run the components without a config. Before this PR, the binary would crash.

I later discovered that it is already possible to run the binary as `./analyzer analyze` or `analyzer serve`, which will make it read only one specific part of the config and thus also work with partial config files. Even so, it doesn't hurt to have the additional robustness provided in this PR.

### DIFF
--- a/cmd/root.go
+++ b/cmd/root.go
@@ -55,28 +55,31 @@ func rootMain(cmd *cobra.Command, args []string) {
 	logger := common.Logger()
 
 	// Initialize services.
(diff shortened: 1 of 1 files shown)


pro-wh (approved):
I'm okay with that

PR #231, opened 2022-11-30
bugfix: genesis_test: Ignore expired nodes in registry

mitjat:
oasis-node keeps reporting some nodes (in the `registry` part of its StateToGenesis output) even after they've already expired. Their expiration epoch is included in the output, so the output is not technically wrong; it's just not useful to insist in genesis_test that the indexer's internal state and the node's StateToGenesis agree on which _expired_ nodes should be listed.
This PR ignores expired nodes on both sides before doing a comparison.

With this fix, genesis_test passes at height 8050156 (which used to be problematic) and also at height 8055536.

The PR is motivated by #211 (so genesis_test is more reliable for finding bugs) and #179 (so genesis_test is usable for finding would-be regressions introduced by that refactor).

### Remaining flakiness

Vaguely related: At 8055536, genesis_test passes the registry part but shows two new errors: Wrong nonce for oasis1qrn99n0hz4ka6vdj8vkmm8y04de5mwhtrckrgsrj, and wrong balance for oasis1qzv7a0gkxwpfelv985fvkl24k7jh3arfwy84zw7q.
Re-running the indexer to that height from scratch made the two problems go away. Just another confirmation that #211 is not resolved yet.

### DIFF
--- a/tests/genesis/genesis_test.go
+++ b/tests/genesis/genesis_test.go
@@ -244,7 +244,7 @@ func TestGenesisFull(t *testing.T) {
 		stakingGenesis = &genesis.Staking
 		governanceGenesis = &genesis.Governance
 	} else {
-		t.Log("Fetching genesis from node", genesisPath)
+		t.Log("Fetching state dump at height", height, "from node")
 		registryGenesis, err = oasisClient.RegistryGenesis(ctx, height)
 		require.Nil(t, err)
 		stakingGenesis, err = oasisClient.StakingGenesis(ctx, height)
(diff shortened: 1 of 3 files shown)


tests/genesis/genesis_test.go
```diff
@@ -437,6 +449,12 @@ func validateNodes(t *testing.T, genesis *registryAPI.Genesis, source *oasis.Con
 		)
 		assert.Nil(t, err)
 
+		if (&node.Node{Expiration: n.Expiration}).IsExpired(uint64(epoch)) {
+			// The indexer DB stores some nodes that are expired because
+			// an expiration event was never produced for them. Ignore them.
```
aefhm, line 454: > an expiration event was never produced for them. Ignore them.

Yikes.

aefhm (approved):

tests/genesis/genesis_test.go
```diff
@@ -416,8 +425,11 @@ func validateNodes(t *testing.T, genesis *registryAPI.Genesis, source *oasis.Con
 			id, entity_id, expiration,
 			tls_pubkey, tls_next_pubkey, p2p_pubkey,
 			vrf_pubkey, roles, software_version
-		FROM %s.nodes_checkpoint
-		WHERE roles LIKE '%s'`, chainID, "%validator%"),
+		FROM
+			%s.nodes_checkpoint
+		WHERE
+			roles LIKE '%%validator%%'
```
Andrew7234, line 431: Does the double percent sign add anything? or should it be 
```suggestion
			roles LIKE '%validator%'
```

Andrew7234 (approved):
Thanks! Looks good with one minor comment

tests/genesis/genesis_test.go
mitjat: To be fair, I think those nodes' expirations are in the deep-ish past, before damask. (Not 100% sure because expiration is expressed in epochs, and I didn't check the epoch of damask genesis.) They might be special in some regard ...
I didn't investigate further, but it didn't make me think "oasis-node is randomly neglecting to emit or expose events".

tests/genesis/genesis_test.go
mitjat: The `%%` inside a string just resolves to a single literal %. The special escape sequence is needed to differentiate the literal % from those % that are used for formatting, e.g. %s.
The previous code used `"%validator%"` without `%%`, and I guess because that string was not passed into `Sprintf` or similar, go decided not to complain. So both solutions work, this edit is just a tad cleaner.

PR #232, opened 2022-12-01
consensus: Do not track runtime (un)suspensions

mitjat:
This PR removes functionality: For a given runtime, we are now no longer able to say if it is suspended.

**Motivation:** Suspension tracking was implemented with lots of RPCs -- roughly one third of all RPCs was dedicated to this functionality. At the same time, in the short/med term, it is not really important if a runtime is suspended; we expect to hand-pick the supported runtimes, and we expect them to never be suspended. Even if we later decide that this is important real-time information to have, we could implement tracking more efficiently, but it would take refactoring (and, for the most efficient solution, an additional field in the relevant oasis-node RPC response)

**Result:** On my machine (with lots of lag), the time to process a consensus block goes from 2.93s to 2.10s. Haven't tested in k8s but I expect a similar ratio since RPCs dominate processing time.

### DIFF
--- a/analyzer/consensus/consensus.go
+++ b/analyzer/consensus/consensus.go
@@ -257,7 +257,7 @@ func (m *Main) processBlock(ctx context.Context, height int64) error {
 		return err
 	}
 
-	data, err := source.AtHeightData(ctx, height)
+	data, err := source.AllData(ctx, height)
 	if err != nil {
 		return err
 	}
--- a/analyzer/consensus/consensus.go
+++ b/analyzer/consensus/consensus.go
@@ -280,7 +280,6 @@ func (m *Main) processBlock(ctx context.Context, height int64) error {
 
 	for _, f := range []func(*storage.QueryBatch, *storage.RegistryData) error{
 		m.queueRuntimeRegistrations,
-		m.queueRuntimeStatusUpdates,
 		m.queueEntityEvents,
 		m.queueNodeEvents,
 	} {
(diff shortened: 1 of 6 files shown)


analyzer/consensus/consensus.go
```diff
@@ -500,20 +499,6 @@ func (m *Main) queueRuntimeRegistrations(batch *storage.QueryBatch, data *storag
 	return nil
 }
 
-func (m *Main) queueRuntimeStatusUpdates(batch *storage.QueryBatch, data *storage.RegistryData) error {
-	runtimeSuspensionQuery := m.qf.ConsensusRuntimeSuspensionQuery()
-	runtimeUnsuspensionQuery := m.qf.ConsensusRuntimeUnsuspensionQuery()
-
-	for _, runtime := range data.RuntimeSuspensions {
-		batch.Queue(runtimeSuspensionQuery, runtime)
```
Andrew7234, line 508: should we remove the `runtime[Un]SuspensionQuery` as well? Or is it used when processing genesis?

Andrew7234 (approved):

analyzer/consensus/consensus.go
mitjat: Good catch, I missed those! Removed now.

storage/api.go
```diff
@@ -83,6 +83,9 @@ type ConsensusSourceStorage interface {
 	// GenesisDocument returns the genesis document for the chain.
 	GenesisDocument(ctx context.Context) (*genesisAPI.Document, error)
 
+	// AtHeightData returns all data tied to a specific height.
+	AtHeightData(ctx context.Context, height int64) (*ConsensusAtHeightData, error)
```
aefhm, line 87: (totally non-blocking nit/thought) I was thinking `DataAtHeight` rolls off the tongue or `AllData`.

aefhm (approved):
Nice speed up.

storage/api.go
mitjat: It does sound weird. Changed to `AllData`.

PR #233, opened 2022-12-02
Do not fetch ChainContext when known

mitjat:
This PR speeds up startup by avoiding 1-2 queries per analyzer (= consensus/emerald/...).
Affects startup only; these are not recurring queries. They are, however, on the heftier side.

Also includes a not-really-related commit (but in the config area, where I was mucking around): There's an `Interval` param that we were not using, except in one place, when constructing the consensus analyzer. And in that one place, I'm pretty sure it was forgotten about, and actually specifying an `Interval` would crash the program.

### DIFF
--- a/storage/oasis/client.go
+++ b/storage/oasis/client.go
@@ -23,7 +23,7 @@ type ClientFactory struct {
 
 // NewClientFactory creates a new oasis-node client factory.
 func NewClientFactory(ctx context.Context, network *config.Network) (*ClientFactory, error) {
-	connection, err := connection.Connect(ctx, network)
+	connection, err := connection.ConnectNoVerify(ctx, network)
 	if err != nil {
 		return nil, err
 	}
@@ -41,13 +41,8 @@ func (cf *ClientFactory) Consensus() (*ConsensusClient, error) {
 	connection := *cf.connection
 	client := connection.Consensus()
 
-	consensusChainContext, err := client.GetChainContext(ctx)
-	if err != nil {
-		return nil, err
-	}
-
 	// Configure chain context for all signatures using chain domain separation.
-	signature.SetChainContext(consensusChainContext)
+	signature.SetChainContext(cf.network.ChainContext)
 
 	c := &ConsensusClient{
 		client:  client,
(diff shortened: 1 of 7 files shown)


storage/oasis/client.go
```diff
@@ -22,14 +22,23 @@ type ClientFactory struct {
 }
 
 // NewClientFactory creates a new oasis-node client factory.
-func NewClientFactory(ctx context.Context, network *config.Network) (*ClientFactory, error) {
-	connection, err := connection.Connect(ctx, network)
```
mitjat, line 26: `Connect` (as opposed to `ConnectNoVerify`) fetches the chain context from the node and compares it with the one specified (manually, via config) in `network`. The fetching is "not needed" because we specify the chain context in the config. But there is a downside to not checking: If the chain context specified in the config does not match the chain that's actually provided by the node in the config, that will result in errors way later, in the form of violated FK constraints. Not sure why/how that even happens; it probably has to do with the chain context being used to [check the signatures](https://github.com/oasisprotocol/oasis-indexer/blob/f9d135cb05cbf08e682859318a02770f3c9bc51e/storage/oasis/client.go#L54-L54) of some of the node payloads.

Alternatively, we could always fetch the context automatically. But that also increases the danger of accidentally connecting to the wrong node and polluting the DB with completely unrelated information. I was on the fence for what to do here, and eventually went with the safer solution.

Andrew7234 (approved):
Nice!

storage/oasis/client.go
aefhm: Do we need to check again that chain context from the node matches our config?

aefhm:
> There's an Interval param that we were not using, except in one place, when constructing the consensus analyzer. And in that one place, I'm pretty sure it was forgotten about, and actually specifying an Interval would crash the program.

Yes.

aefhm (approved):

mitjat:
I ended up restoring the `Interval` cfg option because the two new minor analyzers introduced in https://github.com/oasisprotocol/oasis-indexer/pull/243 require it. That's only the config option though; its use in the consensus and runtime analyzers (which was the bulk of `Interval`-related code) is still removed.

PR #234, opened 2022-12-02
Update fkey constraint

Andrew7234:
Per discussion in Slack, the `transactions` table will never have a null txn_index so we tighten it. Also, the `events` table can have  a null `txn_hash`, so we adjust the fkey constraint there

### DIFF
--- a/storage/migrations/01_oasis_3_consensus.up.sql
+++ b/storage/migrations/01_oasis_3_consensus.up.sql
@@ -35,9 +35,9 @@ CREATE TABLE oasis_3.blocks
 CREATE TABLE oasis_3.transactions
 (
   block BIGINT NOT NULL REFERENCES oasis_3.blocks(height) DEFERRABLE INITIALLY DEFERRED,
-
-  txn_hash   HEX64,
   txn_index  UINT31 NOT NULL,
+  
+  txn_hash   HEX64 NOT NULL,
   nonce      UINT63 NOT NULL,
   fee_amount UINT_NUMERIC,
   max_gas    UINT_NUMERIC,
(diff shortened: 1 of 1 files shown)


storage/migrations/01_oasis_3_consensus.up.sql
```diff
@@ -36,8 +36,8 @@ CREATE TABLE oasis_3.transactions
 (
   block BIGINT NOT NULL REFERENCES oasis_3.blocks(height) DEFERRABLE INITIALLY DEFERRED,
 
-  txn_hash   HEX64 NOT NULL,
-  txn_index  UINT31,
+  txn_hash   HEX64,
```
Andrew7234, line 39: We could keep this as `txn_hash HEX64 NOT NULL`?

mitjat: yes please

storage/migrations/01_oasis_3_consensus.up.sql
```diff
@@ -36,8 +36,8 @@ CREATE TABLE oasis_3.transactions
...
+  txn_index  UINT31 NOT NULL,
```
mitjat, line 40: Nit: WDYT about moving this up and above the blank line, so we have
```
block
txn_index

txn_hash
nonce
...
```
to visually separate/highlight the PK?

Same in `events`, I'd slightly prefer to have txn_block and txn_index together, because they form a unit of sorts.

I checked our SQL statements and they're resilient to us reordering the columns.

mitjat (approved):
Thanks!

storage/migrations/01_oasis_3_consensus.up.sql
Andrew7234: Sure, updated! 

Oops I just saw #235 , I went ahead and added your inline comments from there to this PR as well. Thanks for that

Andrew7234:
Ran this locally; this no longer violates fkey constraints on the events table.

storage/migrations/01_oasis_3_consensus.up.sql
```diff
@@ -51,24 +51,24 @@ CREATE TABLE oasis_3.transactions
   code    UINT31,  -- From https://github.com/oasisprotocol/oasis-core/blob/f95186e3f15ec64bdd36493cde90be359bd17da8/go/consensus/api/transaction/results/results.go#L20-L20
   message TEXT,
 
-  -- We require a composite primary key since duplicate transactions can
+  -- We require a composite primary key since duplicate transactions (with identical hashes) can
   -- be included within blocks for this chain.
-  PRIMARY KEY (block, txn_hash, txn_index)
```
aefhm, line 56: 👍 I do wonder if we'll have to index this later 🤔, but probably better we do those more systematically.

PR #235, opened 2022-12-02
db: smaller FK for events

mitjat:
Also fixes some NULLability issues; an event is not necessarily associated with a tx.


mitjat:
Duplicate of https://github.com/oasisprotocol/oasis-indexer/pull/234/files

PR #237, opened 2022-12-05
Update db types

Andrew7234:
Updates the db fields to better match oasis-core types. 

For `account.general_balance` and `proposals.deposit`, I made them `UINT_NUMERIC` to better match the other token balance-y fields; but maybe there was a reason for not doing so originally?

### DIFF
--- a/storage/migrations/01_oasis_3_consensus.up.sql
+++ b/storage/migrations/01_oasis_3_consensus.up.sql
@@ -225,7 +225,7 @@ CREATE TABLE oasis_3.proposals
   submitter     oasis_addr NOT NULL,
   state         TEXT NOT NULL DEFAULT 'active',  -- "active" | "passed" | "rejected" | "failed"; see https://github.com/oasisprotocol/oasis-core/blob/f95186e3f15ec64bdd36493cde90be359bd17da8/go/governance/api/proposal.go#L29-L29
   executed      BOOLEAN NOT NULL DEFAULT false,
-  deposit       UINT_NUMERIC NOT NULL, 
+  deposit       UINT_NUMERIC NOT NULL,
 
   -- If this proposal is a new proposal.
   handler            TEXT,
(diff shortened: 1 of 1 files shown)


storage/migrations/01_oasis_3_consensus.up.sql
```diff
@@ -239,7 +239,7 @@ CREATE TABLE oasis_3.proposals
 
   created_at    UINT63 NOT NULL,  -- EpochTime, i.e. number of epochs since base epoch
   closes_at     UINT63 NOT NULL,  -- EpochTime, i.e. number of epochs since base epoch
-  invalid_votes UINT_NUMERIC NOT NULL DEFAULT 0
+  invalid_votes UINT63 NOT NULL DEFAULT 0
```
Andrew7234, line 242: https://github.com/oasisprotocol/oasis-core/blob/go/v0.2201.11/go/governance/api/proposal.go#L102

mitjat: Well spotted.

When changing the DB types, I wasn't consistent in what I did with values that are uint64 in go. I mapped them to:
- `UINT63` for values that are sequential (nonces, IDs, etc) or have a known range that is unlikely to be >2^63 even in the future (some error codes somewhere IIRC; or maybe that was about uint32)
- `UINT_NUMERIC` where the value is less constrained and I was afraid it might realistically reach the 2^63..2^64 range. This field, and to a lesser extent `max_gas` below, fall into this category.

Perhaps we should leave a comment, something like
```
-- uint64 in go; because the value might conceivably be >2^63, we use UINT_NUMERIC over UINT63 here.
```

mitjat (approved):
Thank you! The two remaining plain `NUMERIC` types were an oversight.

storage/migrations/01_oasis_3_consensus.up.sql
Andrew7234: Got it, updated w/ comment!

PR #239, opened 2022-12-06
docker: per-user dev tags

mitjat:
Makes it harder for devs to inadvertently step on each other's toes when pushing locally-built images. (Which is naughty in the first place and shouldn't be encouraged, but ...)

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -17,15 +17,15 @@ oasis-indexer:
 
 docker:
 	@docker build \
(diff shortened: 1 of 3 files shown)


Andrew7234 (approved):

Andrew7234:
> 😅 who is pushing those images?

🙋 (Though I think i only pushed to `andy-dev`)

PR #241, opened 2022-12-06
openAPI: simplify stats stack, publish html version

mitjat:
This is a collection of loosely connected commits (so reviewing commit by commit is helpful) that I did as prework for writing more HTTP APIs (#242).

The three big changes are:
- Use a template in the openapi .yaml to represent possible HTTP error codes. Functionally a no-op, and a quick change, but a large red diff.
- Render the openAPI into a .html and make it accessible through the regular HTTP API, making it self-documenting. The URL is pretty much impossible to guess though: https://index.oasislabs.com/v1/spec/v1.html
- Simplify the existing tx_volume statistic. The two granularities (stats once per 5 minutes, or stats once per hour) had wildly different names and separate implementations throughout the stack. This PR reduces duplication a lot, makes the API consistent, and introduces a new `bucket_size_seconds` param in anticipation of future stats queries. (**Note:** The tx_volume query and tables also still need to be expanded to include paratimes. I did not want to cram that into this PR.)

### DIFF
--- a/.github/workflows/ci-test.yaml
+++ b/.github/workflows/ci-test.yaml
@@ -57,10 +57,17 @@ jobs:
     steps:
       - name: Checkout code
         uses: actions/checkout@v3
+      - name: Normalize YAML
+        run: |
+          VERSION=v4.30.5 BINARY=yq_linux_amd64
+          wget https://github.com/mikefarah/yq/releases/download/${VERSION}/${BINARY} -O ./yq
+          chmod +x ./yq
+          # OpenAPI parser fails to parse YAML templates (*, <<:). Expand them out here.
+          <api/spec/v1.yaml ./yq --output-format json | ./yq --output-format yaml --input-format json >api/spec/v1-normalized.yaml
       - name: Validate OpenAPI definition
         uses: char0n/swagger-editor-validate@v1
         with:
-          definition-file: api/spec/v1.yaml
+          definition-file: api/spec/v1-normalized.yaml
 
   validate-migrations:
     runs-on: ubuntu-20.04
--- a/api/common/pagination.go
+++ b/api/common/pagination.go
@@ -17,6 +17,8 @@ const (
 	DefaultOffset = uint64(0)
 
 	MaximumLimit = uint64(1000)
+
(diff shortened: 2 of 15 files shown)


api/v1/v1.go
```diff
@@ -88,13 +91,11 @@ func (h *Handler) RegisterRoutes(r chi.Router) {
 
 			// Aggregate Statistics.
 			r.Route("/stats", func(r chi.Router) {
-				r.Get("/tps", h.ListTransactionsPerSecond)
-				r.Get("/daily_volume", h.ListDailyVolume)
```
mitjat, line 92: This was the 5-minute tx volume and hourly tx volume, respectively.

api/common/pagination.go
```diff
@@ -52,3 +60,18 @@ func NewPagination(r *http.Request) (p Pagination, err error) {
 	}
 	return
 }
+
+// BucketedStatsParams extracts bucket size parameters from an http request.
+func NewBucketedStatsParams(r *http.Request) (b BucketedStatsParams, err error) {
```
mitjat, line 65: The idea is that future stats queries will want this parameter also.

pro-wh (approved):
rendered specs look fine

no objections to the stats changes, other reviewers feel free to review

.github/workflows/ci-test.yaml
```diff
@@ -57,10 +57,17 @@ jobs:
     steps:
       - name: Checkout code
         uses: actions/checkout@v3
+      - name: Normalize YAML
+        run: |
+          VERSION=v4.30.5 BINARY=yq_linux_amd64
+          wget https://github.com/mikefarah/yq/releases/download/${VERSION}/${BINARY} -O ./yq
+          chmod +x ./yq
+          # OpenAPI parser fails to parse YAML templates (*, <<:). Expand them out here.
```
aefhm, line 65: Oh nice. I totally didn't know about this fix/solution when I un-templated.

aefhm (approved):

.github/workflows/ci-test.yaml
mitjat: It ain't pretty. But the bug is in the linter, not in our spec.

PR #243, opened 2022-12-06
consensus: refactor: fetch all round data at once, then process

mitjat:
This is a step towards parallelizing the indexer ([#179](https://github.com/oasisprotocol/oasis-indexer/issues/179#issuecomment-1325827884))

The PR refactors the consensus so that it proceeds in three distinct phases:
1) fetch all data into a large-ish struct
2) compute DB updates
3) apply DB updates

It also factors out **two new analyzers**:
- one that computes aggregate stats (tx volume per time bucket, e.g.)
- one that pulls data from the metadata registry

Both of these were awkwardly folded into the `consensus` analyzer even though they have nothing to do with the regular per-block work in that analyzer.

**Note:** All node requests for a given height now happen sequentially. So sequential processing of heights (which is for now still the only option) is a little slower. I _could_ make those requests parallel, but it makes the code a little more complex, and once we have the full reindex running in parallel, it shouldn't matter because processing multiple heights in parallel will mean that we'll also have multiple parallel node accesses. So I think it's better to keep it simple at this level.


pro-wh (commented):
looking pretty good so far

I see we're getting rid of the 'prepare*' functions (:

analyzer/metadata_registry.go
```diff
@@ -0,0 +1,117 @@
...
+		}
+	}
+	if cfg.ChainID == "" {
+		return nil, fmt.Errorf("metadata_registry analyzer: `ChainID` must be specified in the config")
+	}
+
+	logger.Info("Starting metadata_registry analyzer")
+	return &MetadataRegistryAnalyzer{
+		interval: interval,
+		qf:       NewQueryFactory(strcase.ToSnake(cfg.ChainID), "" /*runtime*/),
```
pro-wh, line 54: do our queries use the chain ID?

pro-wh (approved):
these look nice

analyzer/metadata_registry.go
pro-wh: hm they do

analyzer/consensus/consensus.go
```diff
@@ -271,54 +268,83 @@ func (m *Main) processGenesis(ctx context.Context) error {
...
-		m.prepareBlockData,
-		m.prepareRegistryData,
-		m.prepareStakingData,
-		m.prepareSchedulerData,
-		m.prepareGovernanceData,
+	for _, f := range []func(*storage.QueryBatch, *storage.ConsensusBlockData) error{
+		m.queueBlockInserts,
+		m.queueEpochInserts,
+		m.queueTransactionInserts,
+		m.queueEventInserts,
```
Andrew7234, line 288: Just checking - this will be parallelized again in the upcoming PRs, right?

config/config.go
```diff
@@ -111,15 +111,6 @@ func (cfg *AnalyzerConfig) Validate() error {
...
 	}
-	if cfg.ChainID == "" {
-		return fmt.Errorf("malformed chain id '%s'", cfg.ChainID)
-	}
-	if cfg.RPC == "" {
-		return fmt.Errorf("malformed RPC endpoint '%s'", cfg.RPC)
-	}
-	if cfg.ChainContext == "" {
-		return fmt.Errorf("malformed chain context '%s'", cfg.ChainContext)
-	}
```
Andrew7234, line 122: Looks like these are still needed for the consensus/emerald analyzers; do we want to move this logic to `[consensus/emerald].NewMain`? Or maybe the returned error when trying to connect w/ an empty context is descriptive enough?

analyzer/metadata_registry.go
mitjat: I was surprised too :)

config/config.go
mitjat: Yea good catch. This is not a great change. The config struct is a bit tech-debty now.

It used to describe analyzers that had to do with scanning a chain block by block. With the new analyzers, the set of analyzers is more diverse. Really we'd need 2+ different config objects, and then we'd need to refactor the whole config file to accommodate that. It would be cleaner and better, but I didn't want to spend time on it. My reasoning was that 99%-100% of the time, we're tweaking existing configs, not creating them from scratch. And those "templates" provide stronger guidance than these removed checks will.

analyzer/metadata_registry.go
aefhm: Yeah, this was both for "future proofing" chain ID context separated tables as well trying to make the current scheme somewhat reusable.

PR #244, opened 2022-12-06
db: fix permissions, search paths

mitjat:
After this PR, the API server should be able to access tables more robustly across deployments and schema migrations. 

PR fixes two problems:
- After a migration (or a wipe+regeneration of all tables), access was not granted to the new tables.
- API server was using the wrong schema.table path to some tables

Underlying issues:
- Some of our queries don't use explicit `<schema>.<table>` references to tables, using just `<table>` instead. 
 The API server and the analyzer use different users (read-only vs read-write), so the API server was unable to find those tables. The default `search_path` in postgres (= schemas in order of attempted use) is `"$user", public`, so those tables ended up in different schemas for the two users.
- `GRANT SELECT ON ALL TABLES IN SCHEMA` doesn't affect the tables creaetd after this command.

Changes in the PR:
 - Move all tables (except schema_migrations, used by an opaque library, and only used by the readwrite user anyway) out of the `public` schema, into explicit schemas.
 - Add `GRANT` statements to the end of each of our migrations files.

Tested locally with a separate read-only user.

**TODO**: When deploying to prod and staging, we should delete the now-outdated `indexer` and `api` schemas.

### DIFF
--- a/analyzer/queries.go
+++ b/analyzer/queries.go
@@ -25,7 +25,7 @@ func (qf QueryFactory) LatestBlockQuery() string {
 func (qf QueryFactory) IsGenesisProcessedQuery() string {
 	return `
 		SELECT EXISTS (
-			SELECT 1 FROM processed_geneses
+			SELECT 1 FROM multichain.processed_geneses
 			WHERE chain_id = $1 AND analyzer = $2
 		)`
 }
@@ -39,7 +39,7 @@ func (qf QueryFactory) IndexingProgressQuery() string {
 
 func (qf QueryFactory) GenesisIndexingProgressQuery() string {
 	return `
-		INSERT INTO processed_geneses (chain_id, analyzer, processed_time)
+		INSERT INTO multichain.processed_geneses (chain_id, analyzer, processed_time)
 			VALUES
 				($1, $2, CURRENT_TIMESTAMP)`
 }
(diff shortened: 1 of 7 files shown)


analyzer/queries.go
```diff
@@ -25,7 +25,7 @@ func (qf QueryFactory) LatestBlockQuery() string {
 func (qf QueryFactory) IsGenesisProcessedQuery() string {
 	return `
 		SELECT EXISTS (
-			SELECT 1 FROM processed_geneses
+			SELECT 1 FROM multichain.processed_geneses
```
pro-wh, line 28: why is this outside of oasis_3?

analyzer/queries.go
```diff
@@ -439,12 +439,12 @@ func (qf QueryFactory) RuntimeTokenChangeUpdateQuery() string {
 
 func (qf QueryFactory) RefreshDailyTxVolumeQuery() string {
 	return `
-		REFRESH MATERIALIZED VIEW daily_tx_volume
+		REFRESH MATERIALIZED VIEW stats.daily_tx_volume
```
pro-wh, line 442: hm and stats too. I'd better ask in slack

mitjat: Same as `multichain`. Could be in `multichain`, but we'll continue to have more stats tables so I feel it's nice to carve out a schema for them now.

analyzer/queries.go
mitjat: Because contents are chain-independent. See also the comment at `CREATE DOMAIN`.
Discussed in Slack: We can't be 100% certain that we won't want to change any of the `multichain` tables once we actually have multiple chains, but it's a fair guess, and easy to undo/fix later.

storage/migrations/02_oasis_3_emerald.up.sql
```diff
@@ -195,4 +195,9 @@ CREATE TABLE oasis_3.emerald_withdraws
 CREATE INDEX ix_emerald_withdraws_sender ON oasis_3.emerald_withdraws(sender);
 CREATE INDEX ix_emerald_withdraws_receiver ON oasis_3.emerald_withdraws(receiver);
 
+-- Grant others read-only use.
+-- (We granted already in 01_oasis_3_consensus.up.sql, but the grant does not apply to new tables.)
```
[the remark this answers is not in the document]
aefhm: Aye

aefhm (approved):

Andrew7234 (approved):

Issue #242, opened 2022-12-06
Specify openAPI for MVP


mitjat:
Fill out the v1.yaml openapi spec enough to cover the use cases of the internally-defined MVP.

ptrus:
This was done in the referenced PRs.

Issue #88, opened 2022-07-12
api: Get all activity related to an account


aefhm:
Independent endpoints for each ParaTime /consensus

1. Start with transactions related to an account
2. Later extend with events related to an account

aefhm:
We would need events as well.

Staking rewards progression would be nice.

mitjat:
https://app.clickup.com/t/3t22md6

mitjat:
I would argue that largely, activity == events. This will include things not covered by txs, like end of debonding, or the payout of staking rewards at the end of an epoch.
Well, we'll also want to show failed txs, I guess? But events are the main star. 

#246 is the first step towards adding events; it covers just the openAPI. The remaining work is (copied from #246 description for visibility):
- "Just" piping through what we already have in the DB. Most of the fields fall in this category. Still quite a bit of work because of all of our boilerplate.
- Implementing the "related addresses" field for events will require a new DB table (columns `event_id`, `related_address`; the former is a FK into `events`, which still needs a synthetic, internal-only PK id) and an extension to [this large nested `switch` statement](https://github.com/oasisprotocol/oasis-indexer/blob/d48de379c258fc467e09f33ecd614c03bfb8847d/analyzer/consensus/consensus.go#L840-L840) in the analyzer so that we'll extract whatever addresses are involved in the event, regardless of the type.

pro-wh:
events related addresses might be nice to have. currently we have separate tables for different kinds of events, each with dedicated columns, so maybe we could get by with a great big union. although I think we currently lose the ordering due to events being in different tables

farther out idea: display non-transaction events among transactions. as a special non-transaction transaction/transactions per block. we do this in the rosetta gateway. and we'll possibly treating the consensus chain as the non-paratime paratime :face_with_spiral_eyes: speaking from the explorer perspective, where a lot of existing explorers are focused on events. imo this does a better job of meeting users where they are :shrug:

Andrew7234:
I like the idea of having a `related_events` table in addition to the `related_transactions` table Mitja described. It should make the events query simpler(=faster), and it also nicely handles events that don't have a corresponding transaction. 

> switch-analyze the tx itself to find any associated addresses
I know we can extract the `sender` from the tx signature public key; are there any other addresses that can be extracted besides the ones from the tx.Results.Events?

csillag:
Copying discussion from elsewhere:

@Andrew7234:

[...]

PR #246, opened 2022-12-09
openapi: Consensus events endpoint

mitjat:
Required by #88 

Part of #242

None of this is implemented yet, but I wanted us to agree on the API first. I think there are two large parts to the implementation:
- Just piping through what we already have in the DB. Most of the fields fall in this category. Still quite a bit of work because of all of our boilerplate.
- Implementing the "related addresses" field for events will require a new DB table (columns `event_id`, `related_address`; the former is a FK into `events`, which still needs a synthetic, internal-only PK id) and an extension to [this large nested `switch` statement](https://github.com/oasisprotocol/oasis-indexer/blob/d48de379c258fc467e09f33ecd614c03bfb8847d/analyzer/consensus/consensus.go#L840-L840) in the analyzer so that we'll extract whatever addresses are involved in the event, regardless of the type.

### DIFF
--- a/api/spec/v1.yaml
+++ b/api/spec/v1.yaml
@@ -101,6 +101,25 @@ x-common-types:
     - beacon.PVSSCommit
     - beacon.PVSSReveal
     - beacon.VRFProve
+  consensus-event-types: &consensus_event_types
+    - AddEscrow
+    - AllowanceChange
+    - Burn
+    - DebondingStart
+    - DiscrepancyDetected
+    - Entity
+    - ExecutorCommitted
+    - Finalized
+    - Node
+    - NodeUnfrozen
+    - ProposalExecuted
+    - ProposalFinalized
+    - ProposalSubmitted
+    - ReclaimEscrow
+    - Runtime
+    - TakeEscrow
+    - Transfer
+    - Vote
 
 x-err-responses:
   base-error: &base_error_response
@@ -269,6 +288,66 @@ paths:
             application/json:
               schema: 
                 $ref: '#/components/schemas/Transaction'
+        <<: *common_error_responses
+
+  /consensus/events:
+    get:
+      summary: Returns a list of consensus events.
+      parameters:
+        - *limit
+        - *offset
+        - in: query
+          name: block
+          schema:
+            type: integer
+            format: int64
+          description: A filter on block height.
+          example: *block_height_1
+        - in: query
+          name: tx_index
+          schema:
+            type: integer
+            format: int32
+          description: |
+            A filter on transaction index. The returned events all need to originate
+            from a transaction that appeared in `tx_index`-th position in the block.
+            It is invalid to specify this filter without also specifying a `block`.
+            Specifying `tx_index` and `block` is an alternative to specifying `tx_hash`;
+            either works to fetch events from a specific transaction.
+          example: 3
+        - in: query
+          name: tx_hash
+          schema:
+            type: string
+          description: |
+            A filter on the hash of the transaction that originated the events.
+            Specifying `tx_index` and `block` is an alternative to specifying `tx_hash`;
+            either works to fetch events from a specific transaction.
+          example: *tx_hash_1
+        - in: query
+          name: rel
+          schema:
+            type: string
+          description: |
+            A filter on related accounts. Every returned event will refer to
+            this account. For example, for a `Transfer` event, this will be the
+            the sender or the recipient of tokens.
+          example: *staking_address_1
+        - in: query
+          name: type
+          schema:
+            type: string
+            enum: *consensus_event_types
+          description: A filter on the event type.
+          example: Transfer
+      responses:
+        '200':
+          description: |
+            Consensus events matching the filters, sorted by most recent first.
+          content:
+            application/json:
+              schema: 
+                $ref: '#/components/schemas/ConsensusEventList'
         <<: *common_error_responses
 
   /consensus/entities:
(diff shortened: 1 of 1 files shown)


api/spec/v1.yaml
```diff
@@ -910,7 +990,50 @@ components:
...
+
+    ConsensusEvent:
+      type: object
+      properties:
+        block:
+          type: integer
+          format: int64
+          description: The block height at which this event was generated.
+          example: *block_height_1
+        tx_index:
```
aefhm, line 1012: Not introduced here, but now I can't unsee that we name `tx_index` here and `txn_index` in the DB. https://github.com/oasisprotocol/oasis-indexer/blob/d48de379c258fc467e09f33ecd614c03bfb8847d/storage/migrations/01_oasis_3_consensus.up.sql#L39

api/spec/v1.yaml
```diff
@@ -101,6 +101,26 @@ x-common-types:
...
+    - NodeUnfrozen
+    - ProposalExecuted
+    - ProposalFinalized
+    - ProposalSubmitted
+    - ReclaimEscrow
+    - Runtime
+    - TakeEscrow
+    - Transfer
+    - Vote
+    - Unknown
```
Andrew7234, line 123: Were these event types taken from [/analyzer/api.go](https://github.com/oasisprotocol/oasis-indexer/blob/main/analyzer/api.go#L129)? Wondering if there's an official list of consensus event types from our documentation. 

Some observations from comparing the indexer event types (`/analyzer/api.go`) to the ones I could find from [oasis-core](https://pkg.go.dev/github.com/oasisprotocol/oasis-core/go/consensus/api#ClientBackend):

- The DiscrepancyDetected event is named `ExecutionDiscrepancyDetected` ([ref](https://pkg.go.dev/github.com/oasisprotocol/oasis-core/go@v0.2202.3/roothash/api#ExecutionDiscrepancyDetectedEvent)). Assuming the oasis-core docs are the ground truth and since we expose event types in the `type` param in the indexer events endpoint, what do you think of renaming this field to `ExecutionDiscrepancyDetected` for consistency here, and we can rename the field in a follow-up PR?
- (minor) The `roothash.InMsgProcessed` event ([ref](https://pkg.go.dev/github.com/oasisprotocol/oasis-core/go@v0.2202.3/roothash/api#InMsgProcessedEvent)) isn't indexed; I assume it was unnecessary when Nikhil first crafted the indexer event types, but I just wanted to confirm that this is still true now that we're adding support for Emerald/Sapphire.
- It seems like we're not using the `Unknown` event type currently (prod db has 0 events with that type). What do you think of removing `Unknown` from the openapi spec? I can do more digging to see if we can fully remove it from the indexer codebase in a follow-up PR too.

api/spec/v1.yaml
```diff
@@ -910,7 +990,50 @@ components:
           description: Whether this transaction successfully executed.
       description: |
         A consensus transaction.
-    
+
+    ConsensusEventList:
+      type: object
+      properties:
+        transactions:
```
Andrew7234, line 997:
```suggestion
        events:
```

api/spec/v1.yaml
mitjat: I noticed that too :) and was unsure which one to go with. But eventually I decided it makes more sense to use `tx_` and be consistent with just about everything except those weirdly-named DB fields. I now sent out https://github.com/oasisprotocol/oasis-indexer/pull/252 to rename the DB field also.

Andrew7234 (approved):

lukaw3d:
Very likely wasn't closed intentionally. Bad github, bad!

aefhm:
> Very likely wasn't closed intentionally. Bad github, bad!

Probably one of those "relates" :)

mitjat:
There's only one explicit LGTM on this and I'd prefer more, but at the same time most reviewers have commented and seem to have no further comments. Merging to not have this in flux for too long, but happy to follow up with changes later if needed.

Andrew7234:
Discussed the 3-endpoint vs single endpoint with @pro-wh; 

- The single endpoint matches the existing api pattern. If we split up the filters into endpoints, in theory the other endpoints like transactions should also be `/consensus/account/transactions`. 
- Do we want to explicitly disallow using more than one of the `blocks/txs/address` filters? The 3-endpoint option makes it explicit that only one of the 3 filters (blocks/txs/address) can be used per request. 
-

aefhm:
> Discussed the 3-endpoint vs single endpoint with @pro-wh;
> 
> * The single endpoint matches the existing api pattern. If we split up the filters into endpoints, in theory the other endpoints like transactions should also be `/consensus/account/transactions`.

I'm down.
> * Do we want to explicitly disallow using more than one of the `blocks/txs/address` filters? The 3-endpoint option makes it explicit that only one of the 3 filters (blocks/txs/address) can be used per request.

I vote explicit.
> * With the 3-endpoint approach, we can write separate simpler queries for each endpoint instead of a big omni-query. Though we can still use separate queries for a single `/consensus/events` if we disallow using multiple filters

Soft preference for having frontend deal with the merging unless we are cool with supporting a `/search` type endpoint (which I'm not opposed to).

api/spec/v1.yaml
```diff
@@ -910,7 +989,50 @@ components:
...
+          type: integer
+          format: int64
+          description: The block height at which this event was generated.
+          example: *block_height_1
+        tx_index:
+          type: integer
+          format: int32
+          nullable: true
+          description: |
+            0-based index of this event's originating transaction within its block
```
aefhm, line 1016: (nit)

```suggestion
            0-based index of this event's originating transaction within its block.
```

aefhm (approved):

mitjat:
Thank you for the additional discussion around single API endpoint vs multiple endpoints.

I am also leaning towards single endpoint. The only advantages of multiple endpoints are:
- Slightly more aesthetic, in that it's more RESTful. However, Andy raised a valid counterpoint: "If we split up the filters into endpoints, in theory the other endpoints like transactions should also be `/consensus/account/transactions`". So there's an aesthetics/consistency argument for single endpoint too.
- Possible performance optimizations (specialized queries). But we can also do that by manually limiting the possible filter combos with a single endpoint.

I already started writing docstrings about how we only allow a single filter, but then I realized that efficiency-wise it should not be a problem to combine them. As soon as the block or the tx is specified, DB can use the index  to filter on that, and then beyond that it's a small set of events so a linear scan is fast enough. To keep the API/docs simple, let's wait and limit this only if the DB doesn't figure out how to do the filtering efficiently.

PR #247, opened 2022-12-09
openapi: Add assorted fields required by MVP

mitjat:
This PR adds various fields to our API that are required for an explorer MVP.

On the openapi side, some of the additions are commented out. That's because their underlying data is not something we have in the DB yet; they will therefore be more work to implement. But I want to float them here, and even merge them in their commented-out state, so that frontend folks can code against something.

**Note on account activity:**
For the "account activity" explorer page, this PR takes a double approach. In consensus, I was more ambitious and prepared #246 that sets the stage for displaying activity based on _events_. This is the preferred approach, but takes more work. But then for emerald I didn't go through with it (yet). As a lower-effort approach, I added a filter (by "related accounts") to the txs endpoints for now, on both the emerald and consensus side.

**WIP:** openapi only for now; implementation is TODO in this PR. Except for the parts that are commented out in openapi; those will take more work and will be a separate PR.

### DIFF
--- a/api/spec/v1.yaml
+++ b/api/spec/v1.yaml
@@ -876,6 +876,22 @@ components:
           format: date-time
           description: The second-granular consensus time.
           example: *iso_timestamp_1
+        num_transactions:
+          type: integer
+          format: int32
+          description: Number of transactions in the block.
+          example: 17
+        # TODO: Not available on backend
+        # size:
+        #   type: integer
+        #   format: int64
+        #   description: Size of the block, in bytes. TODO: Be more exact. Should this include results? By far the easiest is to give the total bytesize of txs here.
+        #   example: 123456
+        # gas_used:
+        #   type: integer
+        #   format: int64
+        #   description: Total gas used by the transactions in the block.
+        #   example: 123456
       description: |
         A consensus block.
 
@@ -954,15 +970,30 @@ components:
     Transaction:
       type: object
       properties: 
-        height:
+        block:
           type: integer
           format: int64
           description: The block height at which this transaction was executed.
           example: *block_height_1
+        index_in_block:
+          type: integer
+          format: int32
+          description: 0-based index of this transaction in its block
+          example: *block_height_1
+        timestamp:
+          type: string
+          format: date-time
+          # TODO REVIEW: Is this the time the block was _proposed_, or agreed on, or some other time?
+          description: The second-granular consensus time when this tx's block was proposed.
+          example: *iso_timestamp_3
         hash:
           type: string
           description: The cryptographic hash of this transaction's encoding.
           example: *tx_hash_1
+        sender:
+          type: string
+          description: The address of who sent this transaction.
+          example: *staking_address_1
         nonce:
           type: integer
           format: int64
@@ -982,6 +1013,9 @@ components:
           example: *tx_method_1
         body:
           type: string
+          # TODO: This is a serialzied CBOR blob. Expose as a JSON subtree instead.
+          # Front-end can pull the "recipient" field from here for transactions of a suitable type.
+          format: binary
           description: The method call body.
           example: *tx_body_1
         success:
--- a/api/spec/v1.yaml
+++ b/api/spec/v1.yaml
@@ -120,6 +120,10 @@ x-common-types:
     - TakeEscrow
     - Transfer
     - Vote
+  emv-token-types: &evm_token_types
+    - ERC20
+    - ERC721
+    - ERC1155
 
 x-err-responses:
   base-error: &base_error_response
@@ -778,6 +782,18 @@ paths:
             format: int64
           description: A filter on block round.
           example: *runtime_block_round_1
+        - in: query
+          name: rel
+          schema:
+            type: string
+          # TODO: Implement autodetection of Eth addresses, get rid of last sentence.
+          description: |
+            A filter on related accounts. Every returned transaction will refer to
+            this account in a way. For example, for an `accounts.Transfer` tx, this will be the
+            the sender or the recipient of tokens.
+            We detect related accounts inside EVM transactions and events on a best-effort basis.
+            However, you must provide the oasis-style derived address here, not the Eth address.
+          example: *staking_address_1
       responses:
         '200':
           description: |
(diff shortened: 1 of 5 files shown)


api/spec/v1.yaml
```diff
@@ -797,6 +797,22 @@ components:
           format: date-time
           description: The second-granular consensus time.
           example: *iso_timestamp_1
+        tx_count:
```
pro-wh, line 800: aaah it's `num_transactions` on the emerald side :laughing:

pro-wh: what were they?

api/spec/v1.yaml
```diff
@@ -903,6 +950,9 @@ components:
           example: *tx_method_1
         body:
           type: string
+          # TODO: This is a serialzied CBOR blob. Expose as a JSON subtree instead.
```
aefhm, line 953: ➕

api/spec/v1.yaml
```diff
@@ -1073,18 +1123,92 @@ components:
...
           description: The staking address for this account.
           example: *staking_address_1
+        address_preimage:
+          $ref: '#/components/schemas/AddressPreimage'
         nonce:
           type: integer
           format: int64
           description: A nonce used to prevent replay.
           example: 0
+        # TODO: limit this to 1000 entries. If folks have more,
```
aefhm, line 1206: 👍 I think we should also expose our default limit which I believe is 100.

api/spec/v1.yaml
```diff
@@ -875,15 +907,30 @@ components:
     Transaction:
       type: object
       properties: 
-        height:
+        block:
           type: integer
           format: int64
           description: The block height at which this transaction was executed.
           example: *block_height_1
+        index_in_block:
```
aefhm, line 915: I guess we want to make these more frontend friendly?

api/spec/v1.yaml
```diff
@@ -875,15 +907,30 @@ components:
...
+          type: integer
+          format: int32
+          description: 0-based index of this transaction in its block
+          example: *block_height_1
+        timestamp:
+          type: string
+          format: date-time
+          # TODO REVIEW: Is this the time the block was _proposed_, or agreed on, or some other time?
```
aefhm, line 923: consensus or agreed time?

api/spec/v1.yaml
```diff
@@ -1397,10 +1531,37 @@ components:
...
+          enum: *evm_token_types
+          nullable: true
+          description: |
+            The heuristically determined interface that the token contract implements.
+            A less specialized variant of the token might be detected; for example, an
+            ERC-1363 token might be labeled as ERC-20 here. If the type cannot be
+            detected or is not supported, this field will be null/absent.
+        total_supply:
+          type: string
+          description: The total number of base units available.
```
Andrew7234, line 1564:
```suggestion
        total_supply:
          type: string
          description: The total number of base units available, as provided by the token contract's `totalSupply()` method.
```

Andrew7234 (approved):
Looks good overall, left mostly minor comments

api/spec/v1.yaml
```diff
@@ -101,6 +101,10 @@ x-common-types:
     - beacon.PVSSCommit
     - beacon.PVSSReveal
     - beacon.VRFProve
+  emv-token-types: &evm_token_types
```
lukaw3d, line 104: typo

api/spec/v1.yaml
mitjat: Are you suggesting a different field name? I had that discussion with @pro-wh once on a related timestamp field (I was advocating for a rename), and the conclusion was to leave it as `timestamp` to be consistent with oasis-core.

api/spec/v1.yaml
```diff
@@ -1073,18 +1123,92 @@ components:
...
+            - "oasis-runtime-sdk/address: multisig"
+            - "oasis-runtime-sdk/address: module"
+          description: |
+            The method by which the oasis address was derived from `address_data`.
+          example: "oasis-runtime-sdk/address: secp256k1eth"
+        context_version:
+          type: integer
+          nullable: true
+          default: 0
+          description: Version of the `context`.
```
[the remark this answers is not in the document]
mitjat: Nah Explorer would ignore it. It's for ... people who might want to do the derivation in the other direction themselves for some reason? IDK. It's here for completeness only. I imagine we will tag it with `json:omitempty`, and since eevrything is v0, the field will not explicitly appear on the wire anyway.
If it fills you with rage against bureaucracy and boilerplate, LMK and let's remove it.

api/spec/v1.yaml
mitjat: Added to the spec:
```
      default: 100
      minimum: 1
      maximum: 1000
```
Default is [indeed 100](https://github.com/oasisprotocol/oasis-indexer/blob/d56599667bfd4d0d5318f7102a5c493cd5770686/api/common/pagination.go#L16-L16)

api/spec/v1.yaml
mitjat: Not all the time. We'll query it from time to time but we can't promise to be up to date. (Unlike with symbol() and name(), which are supposed to be static.) So I imagine we'll do a combo of calling totalSupply() and doing dead reckoning via events emitted by [_mint()](https://docs.openzeppelin.com/contracts/2.x/api/token/erc20#ERC20-_mint-address-uint256-) and _burn.

api/spec/v1.yaml
pro-wh: https://github.com/tendermint/tendermint/blob/v0.34.x/spec/core/data_structures.md#header

> weighted median of validators present in the last commit

api/spec/v1.yaml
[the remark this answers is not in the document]
pro-wh: https://github.com/oasisprotocol/oasis-core/blob/v22.2.3/go/common/crypto/address/context.go#L43

the context is made of `identifier string, version uint8`. these two values get serialized into a binary string, and that binary string gets prepended to the data when hashing it into an address

https://github.com/oasisprotocol/oasis-core/blob/v22.2.3/go/common/crypto/address/address.go#LL99C21-L99C21

I think the front end ought to see a context (an identifier-version pair) such as oasis-runtime-sdk/address: secp256k1eth v0, recognize "hey that's the derivation from an ethereum compatible address," and display something like "from Ethereum-compatible address 0x..."

api/spec/v1.yaml
```diff
@@ -1073,18 +1130,93 @@ components:
...
+        Runtimes support all types. This means that every consensus address is also
+        valid in every runtime. For example, in EVM runtimes, you can use staking
+        addresses, but only with oasis tools (e.g. a wallet); EVM contracts such as
+        ERC20 tokens or tools such as Metamask cannot interact with staking addresses.
+      type: object
+      properties:
+        context:
+          type: string
+          enum:
+            - "oasis-core/address: staking"
```
pro-wh, line 1154: one more from oasis-core:

```go
	// AddressRuntimeV0Context is the unique context for v0 runtime account addresses.
	AddressRuntimeV0Context = address.NewContext("oasis-core/address: runtime", 0)
```

api/spec/v1.yaml
```diff
@@ -1073,18 +1130,93 @@ components:
...
+          default: 0
+          description: Version of the `context`.
+        address_data:
+          type: string
+          format: hex
+          description: |
+            The hex-encoded data from which the oasis address was derived. 
+            When `context = "oasis-runtime-sdk/address: secp256k1eth"`, this
+            is the Ethereum address (without the leading `0x`).
+          example: '9907A0cF64Ec9Fbf6Ed8FD4971090DE88222a9aC'
```
pro-wh, line 1174: it will lack the letter case checksum. plain hex encoding of the binary column only. not sure if that's uppercase or lowercase

api/spec/v1.yaml
```diff
@@ -1337,20 +1475,24 @@ components:
           example: 9e6a5837c6366d4a7e477c71ffe32d40915cdef7ef209792259e5ee70caf2705
         sender_0:
           type: string
-          description: The Oasis address of this transaction's 0th signer.
+          description: |
+            The Oasis address of this transaction's 0th signer.
+            Unlike Ethereum, Oasis natively supports multiple-signature transactions.
+            However, the great majority of transactions only have a single signer in practice.
+          # TODO: To access the other signers, ...
```
pro-wh, line 1482: join with the transaction_signers table :skull:

api/spec/v1.yaml
```diff
@@ -1369,7 +1511,12 @@ components:
           type: string
           description: |
             A reasonable "to" Oasis address associated with this transaction,
-            if applicable. The meaning varies based on the transaction method.
+            if applicable. The meaning varies based on the transaction method. Some notable examples:
+              - For `method = "accounts.Transfer"`, this is the paratime account receiving the funds.
+              - For `method = "consensus.Deposit"`, this is the paratime account receiving the funds.
+              - For `method = "consensus.Withdraw"`, this is a consensus (!) account receiving the funds.
+              - For `method = "evm.Create"`, this is the address of the newly created smart contract.
+              - For `method = "evm.Call"`, this is the address of the called smart contract
```
pro-wh, line 1519: :+1:  thanks

api/spec/v1.yaml
pro-wh: on a tangent,

> symbol() and name(), which are supposed to be static

is that stated? like we won't really support letting them change, but we can commence feeling good if spec says they won't change

pro-wh (approved):
:+1: on the struct field naming consistency

api/spec/v1.yaml
aefhm: > Are you suggesting a different field name? 

No. I was giving a stab at inferring the correct description, but @pro-wh gave it conclusively.

api/spec/v1.yaml
aefhm: I thought this particular name/change `index_in_block` was to there to make it easier for frontend devs to consume. My soft preference is to stay as close to the backend as possible unless it introduces a lot of confusion.

Also, I guess this is resolved by going to `index`.

api/spec/v1.yaml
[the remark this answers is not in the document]
mitjat: I agree with everything you said @pro-wh, including the last sentence. But for that, the version is not needed. And it's slightly easier to parse just an enum-like context identifier, rather than a string that glues together the identifier and the version.
Anyway, no strong feelings here, I see, so resolving this and leaving as-is at least for now.

api/spec/v1.yaml
mitjat: It doesn't look like it's promised/enforced in any way: https://eips.ethereum.org/EIPS/eip-20
It's still reasonable for us to behave unreasonably towards a hypothetica tokenwith an un unresonable `symbol()` impl, IMO.

api/spec/v1.yaml
mitjat: Thanks both. I had little doubt that it was consensus time; not much else makes sense in a distributed trustless setting. But with tendermint's two-round approach (in my understanding, one round to agree that a block was proposed, another round to agree on the block's validity and results), I was not 100% sure which of the rounds counts. Warren's link says it's the former, IIUC. Updated.

api/spec/v1.yaml
mitjat: Good point! It's lowercase. Updated.

api/spec/v1.yaml
mitjat: Thanks. I left it as an internal note to ourselves, and documented for now that the API does not support retrieving other signers. It would be nice to have for sure, but let's wait for external demand first; we have plenty of TODOs.

mitjat:
Thank you everyone for reviews. I believe my last set of changes resolves all outstanding comment threads, and all of them seemed simple enough, so I'm going ahead and merging this.

PR #251, opened 2022-12-15
api: Show large numbers as strings

### LINKED ISSUE #85: api: Losing precision in JSON
#### What

https://index.oasislabs.com/v1/consensus/accounts/oasis1qz0ea28d8p4xk8xztems60wq22f9pm2yyyd82tmt returns
```js
{
    "address": "oasis1qz0ea28d8p4xk8xztems60wq22f9pm2yyyd82tmt",
    "nonce": 26,
    "available": 458433997081289,
    "escrow": 218143027902957902,
    "debonding": 169116656872914,
    "total": 218770578556912105
}
```

but it is JSON.parsed into
```js
{
    "address": "oasis1qz0ea28d8p4xk8xztems60wq22f9pm2yyyd82tmt",
    "nonce": 26,
    "available": 458433997081289,
    "escrow": 218143027902957900,  // !!
    "debonding": 169116656872914,
    "total": 218770578556912100  // !!
}
```

#### Fix

We agreed to present numbers as strings.

aefhm:
Close #85 

#### TODO
- [x] Add a test for formatting -- technically covered by e2e
- [x] Add unmarshalling

### DIFF
--- a/storage/client/client.go
+++ b/storage/client/client.go
@@ -37,11 +37,11 @@ type StorageClient struct {
 	logger *log.Logger
 }
 
-// numericToBigInt converts a pgtype.Numeric to a big.int using the private method found at
-// https://github.com/jackc/pgtype/blob/master/numeric.go#L398
-func (c *StorageClient) numericToBigInt(ctx context.Context, n *pgtype.Numeric) (big.Int, error) {
+// numericToBigInt converts a pgtype.Numeric to a BigInt similar to the
+// private method found at https://github.com/jackc/pgtype/blob/master/numeric.go#L398
+func (c *StorageClient) numericToBigInt(ctx context.Context, n *pgtype.Numeric) (BigInt, error) {
 	if n.Exp == 0 {
-		return *n.Int, nil
+		return BigInt{*n.Int}, nil
 	}
 
 	big0 := big.NewInt(0)
@@ -52,7 +52,7 @@ func (c *StorageClient) numericToBigInt(ctx context.Context, n *pgtype.Numeric) 
 		mul := &big.Int{}
 		mul.Exp(big10, big.NewInt(int64(n.Exp)), nil)
 		bi.Mul(bi, mul)
-		return *bi, nil
+		return BigInt{*bi}, nil
 	}
 
 	div := &big.Int{}
@@ -65,9 +65,9 @@ func (c *StorageClient) numericToBigInt(ctx context.Context, n *pgtype.Numeric) 
 			"request_id", ctx.Value(RequestIDContextKey),
 			"err", err,
 		)
-		return *big0, err
+		return BigInt{*big0}, err
 	}
-	return *big0, nil
+	return BigInt{*big0}, nil
 }
 
 // NewStorageClient creates a new storage client.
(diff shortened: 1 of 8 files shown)


storage/client/types.go
```diff
@@ -42,6 +43,17 @@ type Transaction struct {
 	Success bool    `json:"success"`
 }
 
+func (t *Transaction) MarshalJSON() ([]byte, error) {
```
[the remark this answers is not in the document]
aefhm: So I do like the idea of not breaking codegen, albeit this was not quite as nice as I hoped. The type inference is just like a lil lacking.

storage/client/client.go
```diff
@@ -802,9 +809,11 @@ func (c *StorageClient) DebondingDelegations(ctx context.Context, r *DebondingDe
 		if err != nil {
 			return nil, common.ErrStorageError
 		}
-		amount := new(big.Int).Mul(&shares, &escrowBalanceDebonding)
-		amount.Quo(amount, &escrowTotalSharesDebonding)
-		d.Amount = *amount
+		amount := new(big.Int).Mul(big.NewInt(shares.Int64()), big.NewInt(escrowBalanceDebonding.Int64()))
```
mitjat, line 812: Oy. If we cast our BigInt to int64, there's not much point in using arbitrary precision initially.
The wrapped types are hairy though, agreed :/.

Here's what I found:
According to https://go.dev/ref/spec#Struct_types, you can get the inner `big.Int` from a `BigInt` as `myBigInt.Int`.
So we can probably do math in `big.Int` space and then convert to `BigInt` at the end:
```go
amount := new(big.Int).Mul(&shares.Int, &escrowBalanceDebonding.Int)
amount.Quo(amount, &escrowTotalSharesDebonding.Int)
d.Amount = BigInt{*amount}
```

Ditto in line 740.

storage/client/types.go
```diff
@@ -2,10 +2,25 @@
...
 )
 
+type BigInt struct {
+	big.Int
+}
+
+func NewBigInt(v int64) BigInt {
+	var b BigInt
+	b.Set(big.NewInt(v))
+	return b
```
mitjat, line 17: I realized my original code was needlessly convoluted here; the entire function can be simply 
`return BigInt{*big.NewInt(v)}`

You can also use this syntax inside `validateBigInt()` and `numericToBigInt()`.

mitjat (commented):
Hey I was fearing much worse! I think this is quite elegant, nice job!

storage/client/types.go
```diff
@@ -2,10 +2,29 @@
...
 	"math/big"
+	"strings"
 	"time"
 )
 
+type BigInt struct {
+	big.Int
+}
+
+func NewBigInt(v int64) BigInt {
```
aefhm, line 15: I'm a medium face on how much this helps. 🤔

mitjat: I find the implementation syntax to be mildly convoluted, so having this does help readability a tiny bit. But I wouldn't be too sorry to see it go, if you prefer.

mitjat (approved):
Nice!

You did you test this manually, right? I feel tests are not 100% needed here because real-world usage will uncover problems fast enough and obviously enough, but if you are inclined to do it more properly, I think this is a good place for an e2e test: https://github.com/oasisprotocol/oasis-indexer/blob/4ea5341a760d5769ec8ecdc3198a5f0ef0d2d885/tests/http/v1/staking_test.go#L47-L48
We don't have access to the raw JSON here, but even just checking that binance's balance is some known expected number (whatever it turns out to be) would test JSON encoding and decoding. Assuming the balance is >uint64, that is; otherwise, the test is less meaningful.

aefhm:
> I find the implementation syntax to be mildly convoluted, so having this does help readability a tiny bit. But I wouldn't be too sorry to see it go, if you prefer.

Keeping :)

aefhm:
> You did you test this manually, right?

I confirmed that the return result is string enclosed. I think frontend will still have to use some lib to properly resolve them.
> I feel tests are not 100% needed here because real-world usage will uncover problems fast enough and obviously enough, but if you are inclined to do it more properly, I think this is a good place for an e2e test

Oooh, good spot. Let's try it. But yeah, missing the unmarshalling/decoding did break the single e2e test.

aefhm:
> checking that binance's balance is some known expected number

Unfortunately, I think I need to fix the current indexing deploy issue before I can see the results this e2e test.

> otherwise, the test is less meaningful.

Yeah, I don't think that line (as I coded) actually gives more coverage than the current e2e.

PR #252, opened 2022-12-15
db: Rename fields txn_* -> tx_*

mitjat:
Resolves https://github.com/oasisprotocol/oasis-indexer/pull/246#discussion_r1047909347

The `txn_*` naming style is used almost nowhere in the indexer or oasis-core; `tx_` is pretty much the convention. This is a breaking change, but let's live fast (and let the db die young) while we can.


Andrew7234 (approved):

aefhm (approved):
💨 🏃‍♂️

pro-wh (approved):
no objections {:

PR #253, opened 2022-12-15
event enums: Rename to conform with oasis-core

mitjat:
This came up in https://github.com/oasisprotocol/oasis-indexer/pull/246/files#r1048033749

Unlike the indexer, oasis-core does not have an enum of all event types. Instead, events are represented in a [nested struct that is used as a union type](https://github.com/oasisprotocol/oasis-core/blob/7f38a068c21160094ce11be1644b08cfb82b53a1/go/consensus/api/transaction/results/results.go#L13-L13). So to represent a Transfer event, for example, oasis-core use an object like `Event { Staking: { Transfer: { ... } } }`. There's no explicit enumeration/names for event types. The indexer, OTOH, needs an enum so that clients can somehow say "give me only events of type X".

Enums were named somewhat arbitrarily. They were named after the last element of the "path" in the above composite object. In addition, the first element of the path (in practice, the originating tendermint app) was saved as a separate `backend` column in the DB, but it only ever makes sense to expose both `backend` and `type` at the same time, or neither. 

This PR changes the enum value scheme as follows:
- `backend` and `type` are merged into a single field, simplifying handling and understanding
-  JSON/CBOR names of fields are used, rather than Go-internal struct field names
- When the "path" from the top-level `Event` to the leaf event struct has more than two elements, all elements are included in the enum name (e.g. `staking.escrow.take`)

### DIFF
--- a/analyzer/api.go
+++ b/analyzer/api.go
@@ -162,43 +162,43 @@ const (
 func (e *Event) String() string {
 	switch *e {
 	case EventStakingTransfer:
-		return "Transfer"
+		return "staking.transfer"
 	case EventStakingBurn:
-		return "Burn"
+		return "staking.burn"
 	case EventStakingAddEscrow:
-		return "AddEscrow"
+		return "staking.escrow.add"
 	case EventStakingTakeEscrow:
-		return "TakeEscrow"
+		return "staking.escrow.take"
 	case EventStakingDebondingStart:
-		return "DebondingStart"
+		return "staking.escrow.debonding_start"
 	case EventStakingReclaimEscrow:
-		return "ReclaimEscrow"
+		return "staking.escrow.reclaim"
 	case EventStakingAllowanceChange:
-		return "AllowanceChange"
+		return "staking.allowance_change"
 	case EventRegistryRuntime:
-		return "Runtime"
+		return "registry.runtime"
 	case EventRegistryEntity:
-		return "Entity"
+		return "registry.entity"
 	case EventRegistryNode:
-		return "Node"
+		return "registry.node"
 	case EventRegistryNodeUnfrozen:
-		return "NodeUnfrozen"
+		return "registry.node_unfrozen"
 	case EventRoothashExecutorCommitted:
-		return "ExecutorCommitted"
+		return "roothash.executor_committed"
 	case EventRoothashExecutionDiscrepancyDetected:
-		return "ExecutionDiscrepancyDetected"
+		return "roothash.execution_discrepancy_detected"
 	case EventRoothashFinalized:
-		return "Finalized"
+		return "roothash.finalized"
 	case EventGovernanceProposalSubmitted:
-		return "ProposalSubmitted"
+		return "governance.proposal_submitted"
 	case EventGovernanceProposalExecuted:
-		return "ProposalExecuted"
+		return "governance.proposal_executed"
 	case EventGovernanceProposalFinalized:
-		return "ProposalFinalized"
+		return "governance.proposal_finalized"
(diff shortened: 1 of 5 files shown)


analyzer/consensus/consensus.go
```diff
@@ -438,13 +438,12 @@ func (m *Main) queueEventInserts(batch *storage.QueryBatch, data *storage.Consen
 
 	for i := 0; i < len(data.Results); i++ {
 		for j := 0; j < len(data.Results[i].Events); j++ {
-			backend, ty, body, err := extractEventData(data.Results[i].Events[j])
+			_, ty, body, err := extractEventData(data.Results[i].Events[j])
```
mitjat, line 441: I did not change `extractEventData()` because it will be refactored in https://github.com/oasisprotocol/oasis-indexer/pull/188/files (heads up @aefhm) and I wanted to minimize merge conflicts.

analyzer/api.go
```diff
@@ -162,43 +162,43 @@ const (
...
+		return "staking.burn"
 	case EventStakingAddEscrow:
-		return "AddEscrow"
+		return "staking.escrow.add"
 	case EventStakingTakeEscrow:
-		return "TakeEscrow"
+		return "staking.escrow.take"
 	case EventStakingDebondingStart:
-		return "DebondingStart"
+		return "staking.escrow.debonding_start"
```
mitjat, line 173: The current naming style (with dots) is probably the most readable. A downside is that some openapi autogenerators might be unhappy with dots in enum values. I could go with a double underscore (`staking__escrow__debonding_start`) or similar if anybody has concerns/suggestions. 

I'd say it's fine as-is though. The openapi spec doesn't limit the character set, and when I saw somebody complaining online about a generator not accepting a special char (`%` in their case), it was in the context of a github issue, and it was considered a genuine bug and fixed.

Andrew7234 (approved):
Thanks!

analyzer/api.go
```diff
@@ -162,43 +162,43 @@ const (
...
 	case EventRegistryNodeUnfrozen:
-		return "NodeUnfrozen"
+		return "registry.node_unfrozen"
 	case EventRoothashExecutorCommitted:
-		return "ExecutorCommitted"
-	case EventRoothashDiscrepancyDetected:
-		return "DiscrepancyDetected"
+		return "roothash.executor_committed"
+	case EventRoothashExecutionDiscrepancyDetected:
+		return "roothash.execution_discrepancy_detected"
```
pro-wh, line 189: https://github.com/oasisprotocol/oasis-core/blob/v22.2.3/go/roothash/api/api.go#L514

you know what's weird, oasis-core actually has an underscore style name for it for json serialization, and that name leaves out the word "detected."

pro-wh (approved):
k

analyzer/api.go
mitjat: Huh, good eye. Omitting "detected" in the indexer too for consistency, thanks.

Fun fact: there's also ["roothash/execution_discrepancy_detected"](https://github.com/oasisprotocol/oasis-core/blob/7f38a068c21160094ce11be1644b08cfb82b53a1/go/roothash/api/api.go#L35-L35), authored by one @pro-wh :)

mitjat:
> (non-blocking) We don't foresee searching by `backend`?

I can't think of a use case, no. And if we need to for some niche reason, we can still do prefix matching on `type`.

analyzer/api.go
pro-wh: how can he do this to us :laughing:

PR #254, opened 2022-12-20
[db] minor type fix

Andrew7234:
`oasis_3.blocks height` is a `UINT63`, so `oasis_3.transactions block` should be `UINT63 as well. Also updated other `BIGINT` fields that can be tightened (confirmed via oasis-core types)

### DIFF
--- a/storage/migrations/01_oasis_3_consensus.up.sql
+++ b/storage/migrations/01_oasis_3_consensus.up.sql
@@ -35,7 +35,7 @@ CREATE TABLE oasis_3.blocks
 
 CREATE TABLE oasis_3.transactions
 (
-  block BIGINT NOT NULL REFERENCES oasis_3.blocks(height) DEFERRABLE INITIALLY DEFERRED,
+  block UINT63 NOT NULL REFERENCES oasis_3.blocks(height) DEFERRABLE INITIALLY DEFERRED,
   tx_index  UINT31 NOT NULL,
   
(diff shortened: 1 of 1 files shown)


storage/migrations/01_oasis_3_consensus.up.sql
```diff
@@ -101,7 +101,7 @@ CREATE TABLE oasis_3.nodes
   -- For those cases, (id, entity_id) is not a foreign key into oasis_3.claimed_nodes.
   -- Similarly, an entity can un-claim a node after the node registered, but the node can remain be registered for a while.
   entity_id  base64_ed25519_pubkey NOT NULL REFERENCES oasis_3.entities(id),
-  expiration BIGINT NOT NULL,
+  expiration UINT63 NOT NULL,
```
mitjat, line 104: Optional:
```suggestion
  expiration UINT63 NOT NULL,  -- The epoch in which this node expires.
```
(I was wondering if it was a timestamp, or a round, or something else)

mitjat (approved):
Thank you!
When merging, please rebase+squash or use other ways to keep the commit history on `main` readable.

aefhm (approved):

storage/migrations/01_oasis_3_consensus.up.sql
Andrew7234: Added!

PR #265, opened 2022-12-24
Emerald: token queries, handle deterministic errors

pro-wh:
okay so I'm treating three errors as deterministic (i.e. don't retry them, move on)

1. EVM "execution failed" error (code 2)
2. EVM "reverted" error (code 8)
3. if decoding the the output fails

### DIFF
--- a/analyzer/modules/evm.go
+++ b/analyzer/modules/evm.go
@@ -8,6 +8,7 @@ import (
 
 	"github.com/ethereum/go-ethereum/accounts/abi"
 	"github.com/ethereum/go-ethereum/common"
+	"github.com/oasisprotocol/oasis-core/go/common/errors"
 
 	"github.com/oasisprotocol/oasis-indexer/analyzer/evmabi"
 	"github.com/oasisprotocol/oasis-indexer/log"
@@ -36,6 +37,65 @@ type EVMBlockTokenData struct {
 	TokenData map[string]*EVMTokenData
 }
 
+type EVMDeterministicError struct {
+	// Note: .error is the implementation of .Error, .Unwrap etc. It is not
+	// in the Unwrap chain. Use something like
+	// `EVMDeterministicError{fmt.Errorf("...: %w", err)}` to set up an
+	// instance with `err` in the Unwrap chain.
+	error
+}
+
+func (err EVMDeterministicError) Is(target error) bool {
+	if _, ok := target.(EVMDeterministicError); ok {
+		return true
+	}
+	return false
+}
+
+// TODO: can we move this to oasis-sdk/client-sdk/go/modules/evm?
+const EVMModuleName = "evm"
+
+var (
+	// https://github.com/oasisprotocol/oasis-sdk/blob/runtime-sdk/v0.2.0/runtime-sdk/modules/evm/src/lib.rs#L123
+	ErrEVMExecutionFailed = errors.New(EVMModuleName, 2, "execution failed")
+	// https://github.com/oasisprotocol/oasis-sdk/blob/runtime-sdk/v0.2.0/runtime-sdk/modules/evm/src/lib.rs#L147
+	ErrEVMReverted = errors.New(EVMModuleName, 8, "reverted")
+)
+
+func evmCallWithABICustom(
+	ctx context.Context,
+	source storage.RuntimeSourceStorage,
+	round uint64,
+	gasPrice []byte,
+	gasLimit uint64,
+	caller []byte,
+	contractEthAddr []byte,
+	value []byte,
+	contractABI *abi.ABI,
+	result interface{},
+	method string,
+	params ...interface{},
+) error {
+	inPacked, err := contractABI.Pack(method, params...)
+	if err != nil {
+		return fmt.Errorf("packing evm simulate call data: %w", err)
+	}
+	outPacked, err := source.EVMSimulateCall(ctx, round, gasPrice, gasLimit, caller, contractEthAddr, value, inPacked)
+	if err != nil {
+		err = fmt.Errorf("runtime client evm simulate call: %w", err)
+		if errors.Is(err, ErrEVMExecutionFailed) || errors.Is(err, ErrEVMReverted) {
+			err = EVMDeterministicError{err}
+		}
+		return err
+	}
+	if err = contractABI.UnpackIntoInterface(result, method, outPacked); err != nil {
+		err = fmt.Errorf("unpacking evm simulate call output: %w", err)
+		err = EVMDeterministicError{err}
+		return err
+	}
+	return nil
+}
+
 // evmCallWithABI: Given a runtime `source` and `round`, and given an EVM
 // smart contract (at `contractEthAddr`, with `contractABI`) deployed in that
 // runtime, invokes `method(params...)` in that smart contract. The method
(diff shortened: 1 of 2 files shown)


pro-wh:
current problems:

- I don't have a test case for execution failure. https://github.com/oasisprotocol/oasis-sdk/blob/398c7891716b7d315a62ff426cebca9d008210eb/runtime-sdk/modules/evm/src/lib.rs#L187 looks like it should be sufficient to run a query with insufficient gas
- the debug printout for these errors doesn't work well `modules.EVMDeterministicError{error:(*fmt.wrapError)(0xc00007e020)}`

pro-wh:
oh shoot and I have the logic flipped

pro-wh:
added evmCallWithABICustom to allow setting low gas. this successfully triggers an execution failed error. logic un-flipped. still need to do something about the debug print.

pro-wh:
proceeding not to do anything about `%#v` formatting

analyzer/modules/evm.go
```diff
@@ -36,6 +37,63 @@ type EVMBlockTokenData struct {
...
+	if _, ok := target.(EVMDeterministicError); ok {
+		return true
+	}
+	return false
+}
+
+// TODO: can we move this to oasis-sdk/client-sdk/go/modules/evm?
+const EVMModuleName = "evm"
+
+var (
```
mitjat, line 58: Let's add a link to where these come from.

Edit: Searching for the error message `"execution failed"` (with quotes) returns no results in our codebase. I found our EVM SDK module [here](https://github.com/oasisprotocol/oasis-sdk/blob/b37e6da699df331f5a2ac62793f8be099c68469c/runtime-sdk/modules/evm/src/lib.rs#L121-L147), but I'm not clear on what magic removes the `": "` at the end.

analyzer/modules/evm.go
```diff
@@ -36,6 +37,63 @@ type EVMBlockTokenData struct {
...
+	ErrEVMExecutionFailed = errors.New(EVMModuleName, 2, "execution failed")
```
mitjat, line 59: I would prefer to encode just the error codes (2, 8) as constants, and then check something like `if err.Code() == ErrCodeEVMExecutionFailed` or similar. I feel the EVM API is obligated to keep the error codes unchanged across versions, but not so for the exact wording of the error messages. Also, if my code pointer in the previous comment is correct, it seems very possible that the error message might sometimes be of the form `"execution failed: foo bar"`.

analyzer/modules/evm.go
```diff
@@ -36,6 +37,63 @@ type EVMBlockTokenData struct {
...
+	contractEthAddr []byte,
+	value []byte,
+	contractABI *abi.ABI,
+	result interface{},
+	method string,
+	params ...interface{},
+) error {
+	inPacked, err := contractABI.Pack(method, params...)
+	if err != nil {
+		return fmt.Errorf("packing evm simulate call data: %w", err)
```
mitjat, line 79: This is a deterministic error too, though one could argue it happens before the EVM. But I'd still say that if you try to call a hypothetical method `balance_for_user_id(int)` with no params, that's the same class of error as calling `total_balance()` on a contract that doesn't implement it. We label the latter as `EVMDeterministicError`, so we should do it with this one too.

That said, in today's world where we tightly control the allowable ABIs, it mostly doesn't matter how we mark this error because we'll only ever call `Pack()` with a fixed and very limited set of params, so manual debugging suffices. So ultimately, your call :)

mitjat (approved):
:ok_hand:

analyzer/modules/evm.go
```diff
@@ -83,19 +130,30 @@ func evmDownloadTokenERC20(ctx context.Context, logger *log.Logger, source stora
 	}
 	// These mandatory methods must succeed, or we do not count this as an ERC-20 token.
 	if err := evmCallWithABI(ctx, source, round, tokenEthAddr, evmabi.ERC20, &tokenData.TotalSupply, "totalSupply"); err != nil {
-		// todo: propagate nondeterministic errors
 		logError("calling totalSupply", err)
+		if !errors.Is(err, EVMDeterministicError{}) {
+			return nil, fmt.Errorf("calling name: %w", err)
```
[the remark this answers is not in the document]
pro-wh: oh shoot, and that method field shouldn't even have the "calling " prefix :facepalm:

analyzer/modules/evm.go
```diff
@@ -83,19 +130,30 @@ func evmDownloadTokenERC20(ctx context.Context, logger *log.Logger, source stora
...
+		}
 		return nil, nil
 	}
 	// These optional methods may fail.
 	if err := evmCallWithABI(ctx, source, round, tokenEthAddr, evmabi.ERC20, &tokenData.Name, "name"); err != nil {
 		logError("calling name", err)
+		if !errors.Is(err, EVMDeterministicError{}) {
```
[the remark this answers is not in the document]
pro-wh: thanks, adding

analyzer/modules/evm_test.go
```diff
@@ -27,3 +33,74 @@ func TestEVMTokenDataERC20(t *testing.T) {
...
+	require.NoError(t, err)
+	source, err := cf.Runtime(config.DefaultNetworks.All["mainnet"].ParaTimes.All["emerald"].ID)
+	require.NoError(t, err)
+	// Wormhole bridged USDT on Emerald mainnet.
+	tokenEthAddr, err := hex.DecodeString("5555555555555555555555555555555555555555")
+	require.NoError(t, err)
+	var name string
+	err = evmCallWithABI(ctx, source, api.RoundLatest, tokenEthAddr, evmabi.ERC20, &name, "name")
+	require.Error(t, err)
+	fmt.Printf("err %+v\n", err)
```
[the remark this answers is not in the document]
pro-wh: yeah

analyzer/modules/evm_test.go
```diff
@@ -27,3 +33,74 @@ func TestEVMTokenDataERC20(t *testing.T) {
...
 }
+
+func TestEVMFailDeterministicUnoccupied(t *testing.T) {
+	// TODO: Would be nice to have an offline test.
+	ctx := context.Background()
+	cf, err := oasis.NewClientFactory(ctx, config.DefaultNetworks.All["mainnet"], true)
+	require.NoError(t, err)
+	source, err := cf.Runtime(config.DefaultNetworks.All["mainnet"].ParaTimes.All["emerald"].ID)
+	require.NoError(t, err)
+	// Wormhole bridged USDT on Emerald mainnet.
```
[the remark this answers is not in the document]
pro-wh: :+1: thanks

analyzer/modules/evm.go
pro-wh: that horrible magic: https://github.com/oasisprotocol/oasis-sdk/blob/runtime-sdk/v0.2.0/runtime-sdk/modules/evm/src/lib.rs#L147

:arrow_left: that horrible person who wrote it

analyzer/modules/evm.go
pro-wh: more details on this: the oasis coded error system looks at module name and code only. see `errors.FromCode`. the grpc system calls into that to get the actual, registered error object (wrapped if it contains extra context info)

analyzer/modules/evm.go
pro-wh: agreed on this being deterministic. but it's different from finding out that a contract on chain doesn't support a given method. this one is not the contract's fault, and it's likely not the fault of any transient condition either. it would be a problem in the indexer's own programming, so let's not have it go down the "mark this as properly-determined-not-to-work" codepath. unfortunate about the semantic shortcoming of it really being deterministic though

analyzer/modules/evm.go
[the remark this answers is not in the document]
pro-wh: fixed

analyzer/modules/evm.go
pro-wh: also, indeed they do come in that form. from our tests:

> query that runs out of gas should fail: runtime client evm simulate call: execution failed: out of gas

> querying an unsupported method should fail: runtime client evm simulate call: reverted: no revert reason

analyzer/modules/evm.go
pro-wh: links to sdk added

PR #266, opened 2023-01-02
Add account on new allowance owner

aefhm:
#### Description
Fix a foreign key constraint violation on the `allowances` table from the helpful change https://github.com/oasisprotocol/oasis-indexer/pull/217.

Our logs show:

```
["\n\t\tINSERT INTO oasis_3.allowances (owner, beneficiary, allowance)\n\t\t\tVALUES ($1, $2, $3)\n\t\tON CONFLICT (owner, beneficiary) DO\n\t\t\tUPDATE SET allowance = excluded.allowance","oasis1qq2n4alcafedvrp76sr9h8cw9y6kacxz3gtyyump","oasis1qzvlg0grjxwgjj58tx2xvmv26era6t2csqn22pte","100000000000"],
```
#### Related
Hopefully this doesn't bother us here as well. I don't think so.
https://github.com/oasisprotocol/oasis-indexer/blob/825bf5046e73456cf87f95d24dc9d6433c1998d4/storage/migrations/01_oasis_3_consensus.up.sql#L191-L192

#### Question
Potentially we may want to better document this behavior.

### DIFF
--- a/analyzer/consensus/consensus.go
+++ b/analyzer/consensus/consensus.go
@@ -730,6 +730,7 @@ func (m *Main) queueEscrows(batch *storage.QueryBatch, data *storage.StakingData
 func (m *Main) queueAllowanceChanges(batch *storage.QueryBatch, data *storage.StakingData) error {
 	allowanceChangeDeleteQuery := m.qf.ConsensusAllowanceChangeDeleteQuery()
 	allowanceChangeUpdateQuery := m.qf.ConsensusAllowanceChangeUpdateQuery()
+	allowanceOwnerUpsertQuery := m.qf.ConsensusAllowanceOwnerUpsertQuery()
 
 	for _, allowanceChange := range data.AllowanceChanges {
 		if allowanceChange.Allowance.IsZero() {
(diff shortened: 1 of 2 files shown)


analyzer/consensus/consensus.go
```diff
@@ -738,6 +739,7 @@ func (m *Main) queueAllowanceChanges(batch *storage.QueryBatch, data *storage.St
 				allowanceChange.Beneficiary.String(),
 			)
 		} else {
+			batch.Queue(allowanceOwnerUpsertQuery, allowanceChange.Owner.String())
```
mitjat, line 742: Since this was a little sneaky, let's document it?
```suggestion
			// It is possible to create an allowance even for a brand new account with no funds.  
			// Make sure the account exists in indexer DB so that the `allowances->accounts` foreign key is satisfied.
			batch.Queue(allowanceOwnerUpsertQuery, allowanceChange.Owner.String())
```

mitjat (approved):
Thank you!

analyzer/consensus/consensus.go
aefhm: Added a version of this.

Issue #255, opened 2022-12-20
Autogenerate Go types for web API from openapi spec


mitjat:
Our HTTP API types are currently encoded twice:
- [in openapi](https://github.com/oasisprotocol/oasis-indexer/blob/4ea5341a760d5769ec8ecdc3198a5f0ef0d2d885/api/spec/v1.yaml)
- [in Go](https://github.com/oasisprotocol/oasis-indexer/blob/4ea5341a760d5769ec8ecdc3198a5f0ef0d2d885/storage/client/types.go) 

This has a number of downsides, the biggest being:
- nothing enforces that the openapi types match the actual implementation
- parsing of HTTP requests is done in a very manual, boilerplaty, copy-pasty, error-prone way for every request endpoint

We can implement the request and response types, and the parsing of HTTP requests into those types, with autogenerated code from the openapi spec.
Depending on the library used, we can probably autogenerate stub endpoint handlers and routing rules too, but I see that as less important.

I've done some early experimenting with https://github.com/deepmap/oapi-codegen and it looks promising so far.

PR #268, opened 2023-01-03
refactor: Autogenerate Go types from openapi

mitjat:
This PR is the first step towards using autogenerated Go types (#255). It
- tweaks existing Go code and openapi specs so that they conform to each other. (= describe the same types, which was only mostly true before)
- modfiies the build process to autogeenerate Go types from openapi spec
- modifies the Go code to use these new types instead of the old hand-written ones.

The PR does **not** introduce autogenerated parsing of HTTP requests. That is planned for a follow-up PR.  

Changes at a more technical level:
- `BigInt` and the `ContextKey` type (related to HTTP requests) are moved from `storage` package to a new `common` package to break an import cycle.
- Openapi spec is extended with Go-specific annotations. In particular, something to use the custom `BigInt` type for big ints, and something to deal with the (stupid) CamelCase exceptions enforced by the Go linter (`ID` instead of `Id`, etc).
- In openapi, explicit enum types are introduced; they used to be anonymous types, with permissible values listed inline (and using YAML `*anchor` syntax).
- Some small field renames for consistency; **some of them in openapi** (impacts frontend).


storage/client/client.go
```diff
@@ -41,7 +41,7 @@ type StorageClient struct {
 // private method found at https://github.com/jackc/pgtype/blob/master/numeric.go#L398
 func (c *StorageClient) numericToBigInt(ctx context.Context, n *pgtype.Numeric) (BigInt, error) {
 	if n.Exp == 0 {
-		return BigInt{*n.Int}, nil
+		return BigInt{Int: *n.Int}, nil
```
mitjat, line 46: The linter started complaining about these. I don't know why; maybe it has to do with moving it to a different package (and the previous package had some check disabled), or maybe we use a non-pinned linter version and it happened to change while I was working on this PR. I was lazy and just made the linter happy this rather than finding the root cause.
cc @pro-wh since you asked.

storage/client/client.go
```diff
@@ -41,7 +43,7 @@ type StorageClient struct {
...
```
Andrew7234, line 46: Just curious - what necessitated the change?

storage/client/types.go
```diff
@@ -2,253 +2,99 @@
...
-	NumTransactions int       `json:"num_transactions"`
-	Size            int       `json:"size_bytes"`
-	GasUsed         int64     `json:"gas_used"`
-}
+type RuntimeBlock = api.RuntimeBlock
 
 // RuntimeTransactionList is the storage response for RuntimeTransactions.
 type RuntimeTransactionList struct {
 	Transactions []RuntimeTransaction `json:"transactions"`
 }
```
Andrew7234, line 92:
```suggestion
type RuntimeTransactionList = api.RuntimeTransactionList 
```

storage/client/client.go
Andrew7234: Oh I see the above comment

storage/client/client.go
pro-wh: oh interesting

api/spec/v1.yaml
```diff
@@ -1646,10 +1646,9 @@ components:
           description: The method that was called.
           example: "evm.Call"
         body:
-          type: string
-          format: byte
+          type: object
           description: The method call body.
-          example: "o2RkYXRhQGV2YWx1ZUgBY0V4wN0lc2dhZGRyZXNzVDlzcMsG8d3D2LY5FTn+1svi/l+V"
+          # example: "o2RkYXRhQGV2YWx1ZUgBY0V4wN0lc2dhZGRyZXNzVDlzcMsG8d3D2LY5FTn+1svi/l+V"
```
aefhm, line 1673: Do we still want an example here?

Andrew7234 (approved):

aefhm (approved):

PR #271, opened 2023-01-04
bugfix: Process fee accumulator disbursements last

mitjat:
The special fee accumulator account can temporarily attain negative balance due to the order in which we process events. When that happens, the indexer grinds to a halt (first at height 8049946) because negative balances violate DB constraints.

This PR fixes the order of event processing to avoid that situation. 

An alternative solution discussed in Slack is to special-case the fee accumulator account, and not track it at all.
Another alternative (slow) is to try and change oasis-core so it returns events in the "right" order (see comment in the code diff).

*Testing:* Ran locally to height 8050256 without problems.

### DIFF
--- a/analyzer/consensus/consensus.go
+++ b/analyzer/consensus/consensus.go
@@ -293,11 +293,12 @@ func (m *Main) processBlock(ctx context.Context, height int64) error {
 	}
 
 	for _, f := range []func(*storage.QueryBatch, *storage.StakingData) error{
-		m.queueTransfers,
+		m.queueRegularTransfers,
 		m.queueBurns,
 		m.queueEscrows,
 		m.queueAllowanceChanges,
 		m.queueStakingEvents,
+		m.queueDisbursementTransfers,
 	} {
 		if err := f(batch, data.StakingData); err != nil {
 			return err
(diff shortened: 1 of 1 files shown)


Andrew7234:
Just to add another option - when I was debugging the negative balances, I just special-cased the constraint: 
```
ALTER TABLE oasis_3.accounts 
    ADD CONSTRAINT positive_balance CHECK (address = 'oasis1qqnv3peudzvekhulf8v3ht29z4cthkhy7gkxmph5' OR general_balance >= 0);
```

It's not the cleanest, but given that we only have a single fee accumulator address it worked well enough.

mitjat:
> Just to add another option - when I was debugging the negative balances, I just special-cased the constraint

I thought of that approach too, and am not opposed to it. A downside of the custom types that we're now using (like `UINT_NUMERIC`) is that we cannot encode such an exception, unless we stopped using the custom type for that particular column. That is also _fine_, but it adds a little to the ugliness.

I do prefer the solution from this PR because it doesn't throw away any data or safety checks, and feels (to me) like it introduces a less-special special case, despite the extra code. But if it turns out that it does not suffice, I think this SQL-based exception is also a good way to go :+1:

analyzer/consensus/consensus.go
```diff
@@ -293,11 +293,12 @@ func (m *Main) processBlock(ctx context.Context, height int64) error {
 	}
 
 	for _, f := range []func(*storage.QueryBatch, *storage.StakingData) error{
-		m.queueTransfers,
+		m.queueRegularTransfers,
 		m.queueBurns,
 		m.queueEscrows,
 		m.queueAllowanceChanges,
 		m.queueStakingEvents,
```
pro-wh, line 300: does anything weird happen in here?

pro-wh (approved):
how did this transfers system work? I didn't know it was separate from the staking events processing

aefhm (approved):
➕ I do prefer showing the special accounts.

I guess we can't `defer` check the constraint.

Andrew7234 (approved):
Ah I forgot about the custom types. This makes sense to me

analyzer/consensus/consensus.go
mitjat: No, this just inserts (= logs) the events that come from the `staking` app (_and_ are not associated with a tx) as-is into the `events` table.
We do have a similar, more aptly (IMO) named fn `queueTxEventInserts`. I went ahead and renamed this one in the same vein, to `queueStakingEventInserts`, so it sounds less like there's some processing of events going on.

I also renamed another 4 or 5 similarly named fns (one per tendermint app) in the same way.

mitjat:
> how did this transfers system work? I didn't know it was separate from the staking events processing

That's the same thing. "transfers" = actual processing of (most) staking events, i.e. tracking of balances. "queueStakingEvents" = no processing/balancing, just logging. See also my other comment; hope that clarifies.

mitjat:
> I guess we can't `defer` check the constraint.

Ah yes we only discussed this on slack. Correct, we can defer FK constraints but not CHECK constraints until the end of the db tx. It's an unfortunate postgres limitation.

PR #272, opened 2023-01-05
bugfix: Convert bigint params to string before passing to DB

mitjat:
Fixes two bugs around BigInts and the HTTP interface that I discovered accidentally:
- When parsing BigInt params, we rejected all well-formed params, and accepted others
- When we try to pass a BigInt into a DB query as a param, it fails to convert automatically and causes the query to fail. So we just have to know to convert the bigints to strings manually before passing them to the DB. Go's type system enforces/checks nothing at that layer.

While these fixes work, they are a bandaid. Proper solutions:
- (1st bug) Have a regression test that hits our HTTP API with a reasonable number of different requests. (filed #273)
- (2nd bug) Implement custom postgres serialization for BigInt (filed #274)
- (2nd bug) Refactor DB queries so that they cannot be called with arbitrarily-typed params; enforce param types.

### DIFF
--- a/api/v1/client.go
+++ b/api/v1/client.go
@@ -45,9 +45,9 @@ func validateUint64(param string) (uint64, error) {
 // validateBigInt parses a big.Int url parameter.
 func validateBigInt(param string) (*storage.BigInt, error) {
 	i := big.NewInt(0)
-	i, err := i.SetString(param, 10)
-	if err {
-		return nil, common.ErrBadRequest
+	i, ok := i.SetString(param, 10)
+	if !ok {
+		return nil, fmt.Errorf("invalid big.Int: %+v", param)
 	}
 	return &storage.BigInt{Int: *i}, nil
 }
--- a/storage/client/client.go
+++ b/storage/client/client.go
@@ -70,6 +70,14 @@ func (c *StorageClient) numericToBigInt(ctx context.Context, n *pgtype.Numeric) 
 	return BigInt{*big0}, nil
 }
 
+func toString(b *BigInt) *string {
+	if b == nil {
+		return nil
+	}
+	s := b.String()
+	return &s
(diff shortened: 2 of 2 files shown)


api/v1/client.go
```diff
@@ -45,9 +45,9 @@ func validateUint64(param string) (uint64, error) {
 // validateBigInt parses a big.Int url parameter.
 func validateBigInt(param string) (*storage.BigInt, error) {
 	i := big.NewInt(0)
-	i, err := i.SetString(param, 10)
-	if err {
-		return nil, common.ErrBadRequest
+	i, ok := i.SetString(param, 10)
```
pro-wh, line 48: :information_source: https://pkg.go.dev/math/big#Int.SetString ah

storage/client/client.go
```diff
@@ -395,7 +395,7 @@ func (c *StorageClient) Entity(ctx context.Context, r *EntityRequest) (*Entity,
 	nodeRows, err := c.db.Query(
 		ctx,
 		qf.EntityNodeIdsQuery(),
-		r.EntityID,
+		r.EntityID.String(),
```
pro-wh, line 398: what's entity ID?

pro-wh (approved):

storage/client/client.go
mitjat: It's a pubkey. Or rather, a pointer to a pubkey, though it seems it can never be `nil`, from my cursory reading of the code. IDK if it actually has to be converted to a string, but the same type is converted in several other places.

api/v1/client.go
Andrew7234: 🤯 my bad; thanks for the fix

Andrew7234 (approved):

PR #275, opened 2023-01-06
api: extract replyJSON method

### LINKED ISSUE #196: helper for JSON API responses
Can this be abstracted away? So we call it from here with something like
```go
h.writeJsonResponse(w, blocks, r.URL.Path, ctx)
```

You can backport it to non-emerald handlers if you're feeling the least bit OCD, but even just using it for Emerald handlers would set a good precedent.

_Originally posted by @mitjat in https://github.com/oasisprotocol/oasis-indexer/pull/190#discussion_r997626670_

pro-wh:
fixes #196 

actually some of these had incorrectly logged a serialization error as a database error :eyes:

### DIFF
--- a/api/v1/handlers.go
+++ b/api/v1/handlers.go
@@ -20,23 +20,7 @@ func (h *Handler) GetStatus(w http.ResponseWriter, r *http.Request) {
 		return
 	}
 
-	resp, err := json.Marshal(status)
-	if err != nil {
-		h.logAndReply(ctx, "failed to marshal status", w, err)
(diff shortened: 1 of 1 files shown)


pro-wh:
oops gotta rebase. storage.RequestIDContextKey -> common.

mitjat (approved):
Very nice. I'm actually working on autogenerating the red parts of your PR in #255. So I hope to make this PR obsolete soon, but it's good insurance policy to merge it in, in case my more ambitious refactor doesn't pan out. 

> actually some of these had incorrectly logged a serialization error as a database error

Having massive copy-paste all over the codebase was a bad idea? Color me surprised /s

Issue #267, opened 2023-01-03
Emerald analyzer stops silently


buberdds:
https://index.oasislabs.com/v1/emerald/transactions?limit=10&offset=0 the latest indexed block is 3934708

mitjat:
Thanks for reporting.

I had to dig pretty far into log history to find any entries pertaining to emerald. The most recent entry reads `{..., "msg":"processing round", "round":3934709}`. Followed by no errors, just radio silence. I also checked for errors, core dumps etc not associated with the emerald_damask analyzer, but there's nothing.

Very weird. :confused: The analyzer runs an infinite main loop that should print out an error or else move on to the next block ...

Restarting the analyzer fixed it. Slightly unsettling, but I don't think we should dig further for now, given this is the first time it happened. Also, I have no idea how to try and repro, other than running a bunch of stacks and waiting. The next version of analyzer (merged in git, but not yet deployed) has a little more logging, hopefully that will help, should the problem recur.

mitjat:
Reopening because we just saw two more cases of this within a day (one in staging, one in prod; both at different heights; both running v0.0.5 release).

Early debugging. Produced stacktrace dump with
```
kubectl -n mainnet exec -i -t deployments/mainnet-oasis-indexer -c analyzer -- bash -c 'kill -QUIT 1'
```
(This dumps stack traces AND kills the program; k8s restarts it immediately. Btoh staging and prod recovered nicely.)

Looked at stderr in GCP or with
```
kubectl -n mainnet logs deployments/mainnet-oasis-indexer -c analyzer --previous --since 1h| less
```

Here are all the goroutines that contain "emerald" in the stack trace ...
<details>
<summary>Goroutines</summary>

```
goroutine 259858 [chan receive, 1153 minutes]:
google.golang.org/grpc/internal/transport.(*http2Client).closeStream(0x16c6c60?, 0xc00019a088?, {0x16c6c60?, 0xc00019a088?}, 0xb0?, 0xc0?, 0xb857a8?, 0x16c6c60?, 0x88?)
        /go/pkg/mod/google.golang.org/grpc@v1.49.0/internal/transport/http2_client.go:822 +0x22d
google.golang.org/grpc/internal/transport.(*http2Client).CloseStream(0x3e?, 0xc000c75548?, {0x16c6c60, 0xc00019a088})
        /go/pkg/mod/google.golang.org/grpc@v1.49.0/internal/transport/http2_client.go:814 +0x6b
[...]

PR #276, opened 2023-01-07
storage: source storage timeout

pro-wh:
adding a timeout to 'routine' grpc requests

- https://grpc.io/blog/deadlines/#go grpc recommends using the usual go context timeout/deadline system
- #267 reports that requests freeze (thousands of minutes)
- selected timeout in this PR is 61 seconds, a little longer than the usual minute-long timeout used in common proxies
- only grpc requests are subject to the timeout. other analysis stuff including writing to the db are still able to hang \\: feel free to open an alternate PR that puts an entire loop iteration under a timeout
- not exposed in configuration :face_with_spiral_eyes: 
- 'routine' grpc requests only, exempting methods that get entire genesis documents. consensus `AllData` gives each subcomponent a timeout

### DIFF
--- a/analyzer/consensus/consensus.go
+++ b/analyzer/consensus/consensus.go
@@ -34,6 +34,8 @@ import (
 
 const (
 	ConsensusDamaskAnalyzerName = "consensus_damask"
+
+	ProcessBlockTimeout = 61 * time.Second
 )
 
 type EventType = apiTypes.ConsensusEventType // alias for brevity
@@ -266,13 +268,16 @@ func (m *Main) processGenesis(ctx context.Context) error {
 // from source storage and committing an atomically-executed batch of queries
 // to target storage.
 func (m *Main) processBlock(ctx context.Context, height int64) error {
+	ctxWithTimeout, cancel := context.WithTimeout(ctx, ProcessBlockTimeout)
+	defer cancel()
+
 	// Fetch all data.
 	source, err := m.source(height)
 	if err != nil {
 		return err
 	}
 
-	data, err := source.AllData(ctx, height)
+	data, err := source.AllData(ctxWithTimeout, height)
 	if err != nil {
 		return err
 	}
--- a/analyzer/emerald/emerald.go
+++ b/analyzer/emerald/emerald.go
@@ -27,6 +27,8 @@ const (
 	emerald = analyzer.RuntimeEmerald
 
 	EmeraldDamaskAnalyzerName = "emerald_damask"
+
+	ProcessRoundTimeout = 61 * time.Second
 )
 
 // Main is the main Analyzer for the Emerald Runtime.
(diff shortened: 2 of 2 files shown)


storage/oasis/runtime.go
```diff
@@ -27,6 +28,9 @@ type RuntimeClient struct {
 
 // BlockData gets block data in the specified round.
 func (rc *RuntimeClient) BlockData(ctx context.Context, round uint64) (*storage.RuntimeBlockData, error) {
+	ctx, cancel := context.WithTimeout(ctx, 61*time.Second)
```
mitjat, line 31: I'm okay with not overcomplicating this with external config flags, but let's at least extract the 61 into a top-level constant please.

mitjat (commented):
Thank you! While this works, I dislike having the same setup repeated in many places, and having to remember to add it for any future RPCs.
Would you be opposed to setting a time-constrained context a level higher, i.e. time-constraining the whole `processRound()` or `processBlock()`? In [267](https://github.com/oasisprotocol/oasis-indexer/issues/267#issuecomment-1374284220), you mention that you're worried about how that would interact with db queries. Any other concerns? IMO it might actually eb nice to have a 61s timeout for that too, as a failsafe (because I don't know for sure that `pgx` handles a dropped connection robustly, for example; it should, though).

But regardless, we can send a time-constrained context to just the data-fetching (and possibly data-processing) parts of each round, in a single place:
- Emerald: derive `ctxWithTimeout` from `ctx` and pass it into errgroup creation [here](https://github.com/oasisprotocol/oasis-indexer/blob/2d1dae31c661df4cdc72f85a7d26dd5f49494e51/analyzer/emerald/emerald.go#L221-L221). We can cancel the context after `group.Wait()`. The DB stuff happens outside the errgroup, so it will not be impacted by `groupCtx` (which derives from `ctxWithTimeout`).
- Incidentally, looks like we have a small bug and [`prepareBlockData()` here](https://github.com/oasisprotocol/oasis-indexer/blob/2d1dae31c661df4cdc72f85a7d26dd5f49494e51/analyzer/emerald/emerald.go#L227-L227) should be using `groupCtx`, not `ctx`.
- Consensus: There's no async processing so it's even easier, we can just pass a `ctxWithTimeout` to the AllData() call [here](https://github.com/oasisprotocol/oasis-indexer/blob/2d1dae31c661df4cdc72f85a7d26dd5f49494e51/analyzer/consensus/consensus.go#L266-L266) and cancel the context right after the call returns.

storage/oasis/runtime.go
pro-wh: extracted

pro-wh:
ok pulled up to 'main' analyzer level

analyzer/consensus/consensus.go
```diff
@@ -354,7 +359,7 @@ func (m *Main) processBlock(ctx context.Context, height int64) error {
 	timer := m.metrics.DatabaseTimer(m.target.Name(), opName)
 	defer timer.ObserveDuration()
 
-	if err := m.target.SendBatch(ctx, batch); err != nil {
+	if err := m.target.SendBatch(ctxWithTimeout, batch); err != nil {
```
pro-wh, line 362: actually I'm going to revert this hunk

mitjat (approved):
Thank you!

PR #283, opened 2023-01-12
bugfix: openapi, Go: Various type problems 

mitjat:
This PR fixes various small problems with types. See individual commits.
I hit on these while working on #280, but I'm carving them out into its own PR because they're trivial, and not directly related to the crux of the already-oversized #280.

Note: This introduces the use of our custom Address and PublicKey types into openapi. I use them for param parsing; this lets us weed out bad requests eariler, and work with native types (= more safety) in Go. I do NOT use native types for reponses; it's too much trouble to convert the text from the DB into an Address. It's either a good amount of custom-type-integration code, or a lot of brittle manual conversion at every SQL query. (Also, the Address we pull from the DB would then get converted back to text immediately.)


api/spec/v1.yaml
```diff
@@ -121,6 +121,13 @@ x-type-annotations:
     x-go-type: staking.Address
     x-go-type-import: { name: staking, path: "github.com/oasisprotocol/oasis-core/go/staking/api" }
     description: An oasis-style (bech32) address.
+  pubkey: &Ed25519PubKeyType
+    type: string
+    format: byte  # means base64-encoded arw bytes
```
Andrew7234, line 126: what are `arw` bytes?

Andrew7234 (approved):

api/spec/v1.yaml
mitjat: raw :)
fixed.

pro-wh (approved):
ok cool

Issue #225, opened 2022-11-23
Emerald: additional token standards


pro-wh:
currently we only have ERC-20

mitjat:
Example of where we want to expose this: https://github.com/oasisprotocol/oasis-indexer/pull/224#discussion_r1030922644

pro-wh:
discussion related to querying other standards' metadata:

https://github.com/oasisprotocol/oasis-indexer/pull/229#discussion_r1043387086

https://github.com/oasisprotocol/oasis-indexer/pull/229#discussion_r1043390224

pro-wh:
https://github.com/oasisprotocol/oasis-indexer/pull/284#discussion_r1083006875

pro-wh:
https://app.clickup.com/t/866age7bv ticket for ERC-721 support

ptrus:
ERC-721 support was added. For other token standards new issues can be opened.

PR #284, opened 2023-01-13
Emerald: tokens analyzer

pro-wh:
here's a new tokens analyzer. notes from the demo:

> ~70% speedup in block scanning

**before**

image

**after**

image

---

image

---

image

### DIFF
--- a/analyzer/emerald/incoming.go
+++ b/analyzer/emerald/incoming.go
@@ -344,23 +344,31 @@ func extractRound(b *block.Block, txrs []*sdkClient.TransactionWithResults, logg
 					ERC20Transfer: func(fromEthAddr []byte, toEthAddr []byte, amountU256 []byte) error {
 						amount := &big.Int{}
 						amount.SetBytes(amountU256)
-						if !bytes.Equal(fromEthAddr, common.ZeroEthAddr) {
+						fromZero := bytes.Equal(fromEthAddr, common.ZeroEthAddr)
+						toZero := bytes.Equal(toEthAddr, common.ZeroEthAddr)
+						if !fromZero {
 							fromAddr, err2 := registerRelatedEthAddress(blockData.AddressPreimages, blockTransactionData.RelatedAccountAddresses, fromEthAddr)
 							if err2 != nil {
 								return fmt.Errorf("from: %w", err2)
 							}
 							registerTokenDecrease(blockData.TokenBalanceChanges, eventAddr, fromAddr, amount)
 						}
-						if !bytes.Equal(toEthAddr, common.ZeroEthAddr) {
+						if !toZero {
 							toAddr, err2 := registerRelatedEthAddress(blockData.AddressPreimages, blockTransactionData.RelatedAccountAddresses, toEthAddr)
 							if err2 != nil {
 								return fmt.Errorf("to: %w", err2)
 							}
 							registerTokenIncrease(blockData.TokenBalanceChanges, eventAddr, toAddr, amount)
 						}
-						blockData.PossibleTokens[eventAddr] = &modules.EVMPossibleToken{
-							PossibleTypes: []modules.EVMTokenType{modules.EVMTokenTypeERC20},
-							EthAddr:       event.Address,
+						if _, ok := blockData.PossibleTokens[eventAddr]; !ok {
+							blockData.PossibleTokens[eventAddr] = &modules.EVMPossibleToken{}
+						}
+						// Mark as mutated if transfer is between zero address
+						// and nonzero address (either direction) and nonzero
+						// amount. These will change the total supply as mint/
+						// burn.
+						if fromZero != toZero && amount.Cmp(&big.Int{}) != 0 {
+							blockData.PossibleTokens[eventAddr].Mutated = true
 						}
 						return nil
 					},
@@ -377,9 +385,8 @@ func extractRound(b *block.Block, txrs []*sdkClient.TransactionWithResults, logg
 								return fmt.Errorf("spender: %w", err2)
 							}
 						}
-						blockData.PossibleTokens[eventAddr] = &modules.EVMPossibleToken{
-							PossibleTypes: []modules.EVMTokenType{modules.EVMTokenTypeERC20},
-							EthAddr:       event.Address,
+						if _, ok := blockData.PossibleTokens[eventAddr]; !ok {
+							blockData.PossibleTokens[eventAddr] = &modules.EVMPossibleToken{}
 						}
 						return nil
 					},
@@ -412,7 +419,7 @@ func extractRound(b *block.Block, txrs []*sdkClient.TransactionWithResults, logg
 	return &blockData, nil
 }
 
-func emitRoundBatch(batch *storage.QueryBatch, qf *analyzer.QueryFactory, round uint64, blockData *BlockData, blockTokenData *modules.EVMBlockTokenData) {
+func emitRoundBatch(batch *storage.QueryBatch, qf *analyzer.QueryFactory, round uint64, blockData *BlockData) {
 	for _, transactionData := range blockData.TransactionData {
 		for _, signerData := range transactionData.SignerData {
 			batch.Queue(
@@ -435,14 +442,11 @@ func emitRoundBatch(batch *storage.QueryBatch, qf *analyzer.QueryFactory, round 
 	for key, change := range blockData.TokenBalanceChanges {
 		batch.Queue(qf.RuntimeEvmBalanceUpdateQuery(), key.TokenAddress, key.AccountAddress, change.String())
 	}
-	for tokenAddr, tokenData := range blockTokenData.TokenData {
-		batch.Queue(
-			qf.RuntimeTokenUpdateQuery(),
-			tokenAddr,
-			tokenData.Name,
-			tokenData.Symbol,
-			tokenData.Decimals,
-			tokenData.TotalSupply.String(),
-		)
+	for addr, possibleToken := range blockData.PossibleTokens {
+		if possibleToken.Mutated {
+			batch.Queue(qf.RuntimeEVMTokenAnalysisMutateInsertQuery(), addr, round)
+		} else {
+			batch.Queue(qf.RuntimeEVMTokenAnalysisInsertQuery(), addr, round)
+		}
 	}
 }
--- a/analyzer/evmtokens/evm_tokens.go
+++ b/analyzer/evmtokens/evm_tokens.go
@@ -119,10 +119,10 @@ func (m Main) getStaleTokens(ctx context.Context, limit int) ([]*StaleToken, err
 	return staleTokens, nil
 }
 
-func (m Main) processBatch(ctx context.Context) error {
+func (m Main) processBatch(ctx context.Context) (int, error) {
 	staleTokens, err := m.getStaleTokens(ctx, MaxDownloadBatch)
 	if err != nil {
-		return fmt.Errorf("getting discovered tokens: %w", err)
+		return 0, fmt.Errorf("getting discovered tokens: %w", err)
 	}
 	m.logger.Info("processing", "num_stale_tokens", len(staleTokens))
 	if len(staleTokens) == 0 {
@@ -185,13 +185,13 @@ func (m Main) processBatch(ctx context.Context) error {
 	}
 
 	if err := group.Wait(); err != nil {
-		return err
+		return 0, err
 	}
 
 	if err := m.target.SendBatch(ctx, batch); err != nil {
-		return fmt.Errorf("sending batch: %w", err)
+		return 0, fmt.Errorf("sending batch: %w", err)
 	}
-	return nil
+	return len(staleTokens), nil
 }
 
 func (m Main) Start() {
@@ -213,8 +213,16 @@ func (m Main) Start() {
 	for {
 		backoff.Wait()
 
-		if err := m.processBatch(ctx); err != nil {
+		numProcessed, err := m.processBatch(ctx)
+		if err != nil {
 			m.logger.Error("error processing batch", "err", err)
+			backoff.Failure()
+			continue
+		}
+
+		if numProcessed == 0 {
+			// Count this as a failure to reduce the polling when we are
+			// running faster than the block analyzer can find new tokens.
 			backoff.Failure()
 			continue
 		}
--- a/analyzer/evmtokens/evm_tokens.go
+++ b/analyzer/evmtokens/evm_tokens.go
@@ -1,0 +1,228 @@
+package evmtokens
+
+import (
+	"context"
+	"fmt"
+	"time"
+
+	oasisConfig "github.com/oasisprotocol/oasis-sdk/client-sdk/go/config"
+	"golang.org/x/sync/errgroup"
+
+	"github.com/oasisprotocol/oasis-indexer/analyzer"
+	"github.com/oasisprotocol/oasis-indexer/analyzer/modules"
+	"github.com/oasisprotocol/oasis-indexer/analyzer/util"
+	"github.com/oasisprotocol/oasis-indexer/config"
+	"github.com/oasisprotocol/oasis-indexer/log"
+	"github.com/oasisprotocol/oasis-indexer/storage"
+	"github.com/oasisprotocol/oasis-indexer/storage/oasis"
+)
+
+// The token analyzer (1) gets a list from the database of tokens to download
+// the info for, (2) downloads that info, and (3) saves the info in the
+// database. Note that steps (1) and (3) don't happen in a single transaction,
+// so that we don't hold the tables locked and disrupt the block scanner. So
+// be careful of block analyzer potentially making further updates in between.
+// That's why, for example, there are separate last_mutate_round and
+// last_download_round columns. The block analyzer updates last_mutate_round,
+// this token analyzer updates last_download_round.
+
+const (
+	EmeraldDamaskTokensAnalyzerName = "emerald_damask_tokens"
+	MaxDownloadBatch                = 20
+	DownloadTimeout                 = 61 * time.Second
+)
+
+type Main struct {
+	source storage.RuntimeSourceStorage
+	target storage.TargetStorage
+	qf     analyzer.QueryFactory
+	logger *log.Logger
+}
+
+var _ analyzer.Analyzer = (*Main)(nil)
+
+func NewMain(cfg *config.AnalyzerConfig, target storage.TargetStorage, logger *log.Logger) (*Main, error) {
+	ctx := context.Background()
+
+	// Initialize source storage.
+	networkCfg := oasisConfig.Network{
+		ChainContext: cfg.ChainContext,
+		RPC:          cfg.RPC,
+	}
+	factory, err := oasis.NewClientFactory(ctx, &networkCfg, cfg.FastStartup)
+	if err != nil {
+		logger.Error("error creating client factory",
+			"err", err,
+		)
+		return nil, err
+	}
+
+	network, err := analyzer.FromChainContext(cfg.ChainContext)
+	if err != nil {
+		return nil, err
+	}
+
+	id, err := analyzer.RuntimeEmerald.ID(network)
+	if err != nil {
+		return nil, err
+	}
+	logger.Info("Emerald runtime ID determined", "runtime_id", id)
+
+	client, err := factory.Runtime(id)
+	if err != nil {
+		logger.Error("error creating runtime client",
+			"err", err,
+		)
+		return nil, err
+	}
+
+	return &Main{
+		source: client,
+		target: target,
+		qf:     analyzer.NewQueryFactory("oasis_3", analyzer.RuntimeEmerald.String()),
+		logger: logger.With("analyzer", EmeraldDamaskTokensAnalyzerName),
+	}, nil
+}
+
+type StaleToken struct {
+	Addr                  string
+	LastMutateRound       uint64
+	LastDownloadRound     *uint64
+	Type                  *modules.EVMTokenType
+	AddrContextIdentifier string
+	AddrContextVersion    int
+	AddrData              []byte
+}
+
+func (m Main) getStaleTokens(ctx context.Context, limit int) ([]*StaleToken, error) {
+	var staleTokens []*StaleToken
+	rows, err := m.target.Query(ctx, m.qf.RuntimeEVMTokensAnalysisStaleQuery(), limit)
+	if err != nil {
+		return nil, fmt.Errorf("querying discovered tokens: %w", err)
+	}
+	defer rows.Close()
+	for rows.Next() {
+		var staleToken StaleToken
+		if err = rows.Scan(
+			&staleToken.Addr,
+			&staleToken.LastMutateRound,
+			&staleToken.LastDownloadRound,
+			&staleToken.Type,
+			&staleToken.AddrContextIdentifier,
+			&staleToken.AddrContextVersion,
+			&staleToken.AddrData,
+		); err != nil {
+			return nil, fmt.Errorf("scanning discovered token: %w", err)
+		}
+		staleTokens = append(staleTokens, &staleToken)
+	}
+	return staleTokens, nil
+}
+
+func (m Main) processBatch(ctx context.Context) error {
+	staleTokens, err := m.getStaleTokens(ctx, MaxDownloadBatch)
+	if err != nil {
+		return fmt.Errorf("getting discovered tokens: %w", err)
+	}
+	m.logger.Info("processing", "num_stale_tokens", len(staleTokens))
+	if len(staleTokens) == 0 {
+		return 0, nil
+	}
+
+	ctxWithTimeout, cancel := context.WithTimeout(ctx, DownloadTimeout)
+	defer cancel()
+	group, groupCtx := errgroup.WithContext(ctxWithTimeout)
+
+	batch := &storage.QueryBatch{}
+
+	for _, st := range staleTokens {
+		// Redeclare `st` for unclobbered use within goroutine.
+		staleToken := st
+		group.Go(func() error {
+			m.logger.Info("downloading", "stale_token", staleToken)
+			// todo: assert that addr context is secp256k1
+			if staleToken.LastDownloadRound == nil {
+				tokenData, err := modules.EVMDownloadNewToken(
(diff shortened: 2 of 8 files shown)


pro-wh:
current state:
- trying this thing where instead of a dedicated tokens-to-query table, I'm adding a "discovery state" column to the tokens table. we can come up with a state machine like (row nonexistent) -> discovered -> info populated <-> need total supply update etc

pro-wh:
readying this for review, there are some issues with this though

analyzer/emeraldtokens/emerald_tokens.go
```diff
@@ -0,0 +1,194 @@
+package emeraldtokens
```
mitjat, line 1: We're planning to make runtime analyzers and tables more agnostic wrt the specific runtime, and focus them around sdk modules instead. Can we make this into `evm_tokens`? And in general do a search for `[Ee]merald` and see what can be meaningfully and easily renamed or generalized. Otherwise, we'll have to come back and do it in a couple weeks for Sapphire.

Renaming the emerald_tokens DB table (and adding a `runtime` column) is probably out of scope for this PR.

mitjat (commented):
Nice! It's a good amount of boilerplate to figure out; thank you!

I think this is overall good; I have a few small style/docs suggestions, and one higher-level one about focusing the analyzer on the EVM module, not Emerald.

storage/migrations/02_oasis_3_emerald.up.sql
```diff
@@ -111,12 +111,16 @@ CREATE INDEX ix_emerald_token_address ON oasis_3.emerald_token_balances (token_a
 CREATE TABLE oasis_3.emerald_tokens
 (
   token_address oasis_addr PRIMARY KEY,
+  discovery_state TEXT NOT NULL,
+  last_seen_round UINT63 NOT NULL,
```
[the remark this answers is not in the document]
mitjat: Oh now I see last week's update where it seems you intentinoally migrated this into a single table. Why did you prefer that? The state machine as you describe it can be implemented with a one table or two.

I think it's  cleaner in a separate table. Having shorter rows in the `emerald_tokens` table will also be helpful for faster joins, and we'll be joining this table to others a lot.
Ultimately, not a dealbreaker though.

storage/migrations/02_oasis_3_emerald.up.sql
```diff
@@ -111,12 +111,16 @@ CREATE INDEX ix_emerald_token_address ON oasis_3.emerald_token_balances (token_a
 CREATE TABLE oasis_3.emerald_tokens
 (
   token_address oasis_addr PRIMARY KEY,
+  discovery_state TEXT NOT NULL,
```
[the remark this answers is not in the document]
mitjat: Also, just something to consider as an alternative to `discovery_state`:
Two `TIMESTAMP` columns, `last_attempted_processing_at` (= when an analyzer grabbed this row) and `last_processed_at` (= when an analyzer finished processing the token). Analyzer(s) grab new tokens with
```
UPDATE mytable
SET last_attempted_processing_at=now()  -- mark token as being under processing
WHERE token_address IN (
  SELECT token_address FROM mytable
  WHERE
  last_processed_at IS NULL AND  -- token hasn't been processed yet
  last_attempted_processing_at < now() - '5 minutes'::interval  -- no other analyzer has grabbed this token recently
  LIMIT 20  -- a limited batch size
  FOR UPDATE  -- lock the rows until UPDATE finishes 
)
RETURNING token_address;
```

I think both solutions are fine.
The one I described has the benefit of supporting multiple analyzers running in parallel (which shouldn't really be a concern/priority for this analyzer, though it's nice if we're resilient to multiple analyzers accidentally running at once), and it easily supports periodic re-querying of contracts (e.g. for new totalSupply) based on `last_processed_at`.
The nice thing about `discovery_state` is that it's easier to add new states.

analyzer/emerald/incoming.go
```diff
@@ -435,14 +435,12 @@ func emitRoundBatch(batch *storage.QueryBatch, qf *analyzer.QueryFactory, round
...
-			tokenData.Name,
-			tokenData.Symbol,
-			tokenData.Decimals,
-			tokenData.TotalSupply.String(),
-		)
+	for addr, possibleToken := range blockData.PossibleTokens {
+		if len(possibleToken.PossibleTypes) != 1 || possibleToken.PossibleTypes[0] != modules.EVMTokenTypeERC20 {
+			// We'll need to expand the discovery state to cover multiple
+			// possible token types, as we determine during scanning.
+			panic(fmt.Sprintf("discarding token %s type info %v", addr, possibleToken.PossibleTypes))
```
mitjat, line 442: We can probably simply dump the possible types into a SQL column? (Columns can be [array-typed](https://github.com/oasisprotocol/oasis-indexer/blob/77852199760eb6b7ae4a2ed7443950fbe9981868/storage/migrations/01_oasis_3_consensus.up.sql#L72-L72))

Regardless of whether you track this in the discover-state table or the tokens metadata table (I'm suggesting separating the two tables in another comment thread) or if you have a single table: At the end of the process, the tokens metadata table should contain the final determined type so we can present it to the frontend.

storage/migrations/02_oasis_3_emerald.up.sql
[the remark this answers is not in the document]
pro-wh: yeah two tables is fine. let's go with that

pro-wh:
designing new table to use three columns: first seen, last round when mutable info changed, last round when we analyzed it. diagram of the latter two columns:

image

<details><summary>source</summary>

```dot
digraph {
    start [shape=none]
    node [shape=box]
    start -> no_row
    no_row [peripheries=2]
    null_round [peripheries=2]
    round_round [peripheries=2]
    // block analyzer
    edge [color=blue]
    no_row -> null_null [label="see transfer"]
    no_row -> round_null [label="see mint"]
    null_null -> round_null [label="see mint"]
    null_round -> high_low [label="see mint"]
    round_round -> high_low [label="see mint"]
    // evm token analyzer
    edge [color=green]
    null_null -> null_round [label="download"]
    null_null -> high_low [label="stale download"]
    round_null -> high_low [label="stale download"]
    round_null -> round_round [label="download"]
    high_low -> round_round [label="download mutable"]
}
```

</details>

blue is what happens in the block analyzer. green is what happens in the evm token analyzer. self edges omitted.

double boxes are consistent states. following only green edges should allow the system to settle into one of these states.

"stale download" is what happens when the evm token analyzer reads one "last round when mutable info changed" into its goroutine, then the block analyzer detects something that would further change the mutable info while the evm token analyzer is downloading info for an out of date round, then the evm token analyzer saves that out of date info.

analyzer/queries.go
```diff
@@ -442,16 +442,40 @@ func (qf QueryFactory) RuntimeTokenBalanceUpdateQuery() string {
...
 
+func (qf QueryFactory) RuntimeTokensDiscoveredQuery() string {
+	return fmt.Sprintf(`
+		SELECT
+			token_address,
+			discovery_state,
+			last_seen_round,
+			context_identifier,
+			context_version,
+			address_data
```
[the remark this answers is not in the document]
pro-wh: yup

storage/migrations/02_oasis_3_emerald.up.sql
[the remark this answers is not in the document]
pro-wh: going with the last-mutated-at + last-processed-at. seems that's expressive enough without having an explicit discovery state column. block analyzer bumps one, token analyzer bumps the other

storage/migrations/02_oasis_3_emerald.up.sql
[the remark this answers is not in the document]
pro-wh: went with separate table

analyzer/emeraldtokens/emerald_tokens.go
```diff
@@ -0,0 +1,194 @@
...
+		dts = append(dts, &discoveredToken)
+	}
+	return dts, nil
+}
+
+func (m Main) processBatch(ctx context.Context) error {
+	// This part is not in the update transaction. We don't want to hold
+	// up the tokens table while we download the token info. Make sure the
+	// Emerald analyzer and this Emerald tokens analyzer can cooperate
+	// correctly without stopping each other.
```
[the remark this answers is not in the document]
pro-wh: ok

analyzer/emeraldtokens/emerald_tokens.go
pro-wh: changing, also changing in #296

analyzer/emerald/incoming.go
pro-wh: trying without this info. may reconsider in future work on #225

analyzer/emeraldtokens/emerald_tokens.go
```diff
@@ -0,0 +1,214 @@
...
+
+	batch := &storage.QueryBatch{}
+
+	for _, st := range staleTokens {
+		// Redeclare `st` for unclobbered use within goroutine.
+		staleToken := st
+		group.Go(func() error {
+			m.logger.Info("downloading", "stale_token", staleToken)
+			// todo: assert that addr context is secp256k1
+			if staleToken.Type == nil {
```
pro-wh, line 132: this condition isn't right. if the block scanner finds a possible token and we scan and find it's not a recognized token type, we won't create a row in evm_tokens. thus staleToken.Type will still be nil. should change this to query in the last_download_round and check if that's nil instead.

analyzer/emeraldtokens/emerald_tokens.go
```diff
@@ -0,0 +1,214 @@
...
+	for {
+		backoff.Wait()
+
+		if err := m.processBatch(ctx); err != nil {
+			m.logger.Error("error processing batch", "err", err)
+			backoff.Failure()
+			continue
+		}
+
+		backoff.Success()
```
pro-wh, line 208: should probably use backoff.Failure if we didn't find any to process, so that we reduce the polling frequency when there's nothing to do.

analyzer/emerald/incoming.go
pro-wh: closing this out for now. added a cross reference from #225

analyzer/queries.go
```diff
@@ -438,16 +439,76 @@ func (qf QueryFactory) RuntimeEvmBalanceUpdateQuery() string {
...
+	template = strings.ReplaceAll(template, "emerald", "%[2]s")
+	return fmt.Sprintf(template, qf.chainID, qf.runtime)
+}
+
+func (qf QueryFactory) RuntimeEVMTokensAnalysisStaleQuery() string {
+	return qf.templatize(`
+		SELECT
+			evm_token_analysis.token_address,
+			evm_token_analysis.last_mutate_round,
+		    evm_tokens.token_type,
```
pro-wh, line 454: oh my god the spaces

analyzer/emeraldtokens/emerald_tokens.go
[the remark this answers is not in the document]
pro-wh: moved, reworded

analyzer/emeraldtokens/emerald_tokens.go
pro-wh: renamed

analyzer/queries.go
pro-wh: fixed

analyzer/emeraldtokens/emerald_tokens.go
pro-wh: trying backoff.Failure

analyzer/emeraldtokens/emerald_tokens.go
pro-wh: fixed

pro-wh:
description updated. re-opening for review

analyzer/emerald/incoming.go
```diff
@@ -344,23 +344,31 @@ func extractRound(b *block.Block, txrs []*sdkClient.TransactionWithResults, logg
...
 							}
 							registerTokenDecrease(blockData.TokenBalanceChanges, eventAddr, fromAddr, amount)
 						}
-						if !bytes.Equal(toEthAddr, common.ZeroEthAddr) {
+						if !toZero {
 							toAddr, err2 := registerRelatedEthAddress(blockData.AddressPreimages, blockTransactionData.RelatedAccountAddresses, toEthAddr)
 							if err2 != nil {
 								return fmt.Errorf("to: %w", err2)
 							}
 							registerTokenIncrease(blockData.TokenBalanceChanges, eventAddr, toAddr, amount)
```
mitjat, line 361: In the case of mints/burns, will we `registerTotalSupplyIncrease/Decrease()`?

mitjat (commented):
Looking good!
Consider this pretty much approved from my end, I'm just holding off because I'm not sure if we are on the same page re/ dead reckoning for mints and burns.

analyzer/evmtokens/evm_tokens.go
```diff
@@ -0,0 +1,219 @@
+package evmtokens
```
mitjat, line 1: nit: would be nice to have the file and the folder have the same name; I slightly prefer `evm_tokens` over `evmtokens`, but either is fine.

analyzer/evmtokens/evm_tokens.go
```diff
@@ -0,0 +1,219 @@
...
+	}
+	return staleTokens, nil
+}
+
+func (m Main) processBatch(ctx context.Context) error {
+	staleTokens, err := m.getStaleTokens(ctx, MaxDownloadBatch)
+	if err != nil {
+		return fmt.Errorf("getting discovered tokens: %w", err)
+	}
+	m.logger.Info("processing", "num_stale_tokens", len(staleTokens))
```
mitjat, line 123: If `len(staleTokens)==0`, you can early exit here and avoid sending an empty batch to the DB

analyzer/evmtokens/evm_tokens.go
```diff
@@ -0,0 +1,219 @@
...
+		m.logger.Error("error configuring indexer backoff policy",
+			"err", err,
+		)
+		return
+	}
+
+	for {
+		backoff.Wait()
+
+		if err := m.processBatch(ctx); err != nil {
```
mitjat, line 207: `processBatch` should return the number of updated tokens. If 0, that means we're all caught up, and we should also call `backoff.Failure()`.
Currently, once the analyzer catches up, it does a series of `backoff.Success()` and keeps processing 0 tokens at max speed.

analyzer/modules/evm.go
```diff
@@ -22,17 +22,21 @@ const (
...
 
 type EVMTokenData struct {
+	Type        EVMTokenType
 	Name        string
 	Symbol      string
 	Decimals    uint8
 	TotalSupply *big.Int
 }
 
+type EVMTokenMutateData struct {
```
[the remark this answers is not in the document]
pro-wh: aah shoot I meant to switch over one of the methods to use this

analyzer/modules/evm.go
```diff
@@ -164,29 +168,21 @@ func evmDownloadTokenERC20(ctx context.Context, logger *log.Logger, source stora
...
+}
+
+func EVMDownloadMutatedToken(ctx context.Context, logger *log.Logger, source storage.RuntimeSourceStorage, round uint64, tokenEthAddr []byte, tokenType EVMTokenType) (*EVMTokenData, error) {
+	// todo: simplify to download mutable fields only
+	tokenData, err := evmDownloadTokenERC20(ctx, logger, source, round, tokenEthAddr)
+	if err != nil {
+		return nil, fmt.Errorf("download token ERC-20: %w", err)
 	}
-	return &blockTokenData, nil
+	return tokenData, nil
```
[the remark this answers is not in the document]
pro-wh: gonna reimplement this

storage/migrations/02_oasis_3_emerald.up.sql
```diff
@@ -117,15 +117,26 @@ CREATE INDEX ix_emerald_token_address ON oasis_3.evm_token_balances (token_addre
...
+  total_supply uint_numeric,
+  PRIMARY KEY (runtime, token_address)
 );
 
+CREATE TABLE oasis_3.evm_token_analysis
+(
+  runtime runtime NOT NULL,
+  token_address oasis_addr NOT NULL,
+  last_mutate_round UINT63 NOT NULL,
+  last_download_round UINT63,
```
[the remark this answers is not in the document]
pro-wh: adding a comment

storage/migrations/02_oasis_3_emerald.up.sql
```diff
@@ -117,15 +117,26 @@ CREATE INDEX ix_emerald_token_address ON oasis_3.evm_token_balances (token_addre
...
+  decimals INTEGER,
+  total_supply uint_numeric,
+  PRIMARY KEY (runtime, token_address)
 );
 
+CREATE TABLE oasis_3.evm_token_analysis
+(
+  runtime runtime NOT NULL,
+  token_address oasis_addr NOT NULL,
+  last_mutate_round UINT63 NOT NULL,
```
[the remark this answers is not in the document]
pro-wh: adding a comment

analyzer/emerald/incoming.go
```diff
@@ -344,23 +344,31 @@ func extractRound(b *block.Block, txrs []*sdkClient.TransactionWithResults, logg
...
-							EthAddr:       event.Address,
+						if _, ok := blockData.PossibleTokens[eventAddr]; !ok {
+							blockData.PossibleTokens[eventAddr] = &modules.EVMPossibleToken{}
+						}
+						// Mark as mutated if transfer is between zero address
+						// and nonzero address (either direction) and nonzero
+						// amount. These will change the total supply as mint/
+						// burn.
+						if fromZero != toZero && amount.Cmp(&big.Int{}) != 0 {
+							blockData.PossibleTokens[eventAddr].Mutated = true
```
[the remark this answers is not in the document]
pro-wh: it's done as separate branches so that different ones can trigger in different iterations of the outer loop. one iteration can find a non-mutating event and create a default EVMPossibleToken{} and a later iteration can find a mutating event and turn on the Mutated flag

analyzer/modules/evm.go
```diff
@@ -22,17 +22,21 @@ const (
 )
 
 type EVMPossibleToken struct {
-	PossibleTypes []EVMTokenType
-	EthAddr       []byte
+	Mutated bool
 }
 
 type EVMTokenData struct {
+	Type        EVMTokenType
```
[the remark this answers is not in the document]
pro-wh: we end up not creating a record in the tokens table, and we bump the token analysis table to indicate that we looked. in later encounters, the token analyzer sees that it's already previously analyzed, and the type is NULL from the LEFT JOIN

analyzer/modules/evm.go
```diff
@@ -164,29 +168,21 @@ func evmDownloadTokenERC20(ctx context.Context, logger *log.Logger, source stora
...
-				logger.Info("EVMDownloadRoundTokens missing handler for a possible token type",
-					"token_oasis_addr", tokenAddr,
-					"token_eth_addr", possibleToken.EthAddr,
-					"possible_type", possibleType,
-				)
-			}
-		}
+func EVMDownloadNewToken(ctx context.Context, logger *log.Logger, source storage.RuntimeSourceStorage, round uint64, tokenEthAddr []byte) (*EVMTokenData, error) {
+	// todo: check ERC-165 0xffffffff compliance
+	// todo: try other token standards based on ERC-165
```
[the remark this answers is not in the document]
pro-wh: #225

analyzer/queries.go
```diff
@@ -438,16 +438,69 @@ func (qf QueryFactory) RuntimeEvmBalanceUpdateQuery() string {
...
+			evm_token_analysis.token_address,
+			evm_token_analysis.last_mutate_round,
+			evm_tokens.token_type,
+			address_preimages.context_identifier,
+			address_preimages.context_version,
+			address_preimages.address_data
+		FROM %[1]s.evm_token_analysis
+		LEFT JOIN %[1]s.evm_tokens ON
+			evm_tokens.runtime = evm_token_analysis.runtime AND
+			evm_tokens.token_address = evm_token_analysis.token_address
```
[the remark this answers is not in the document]
pro-wh: oo I'll try that

analyzer/evmtokens/evm_tokens.go
pro-wh: our org hasn't been using underscores in package names \:

analyzer/evmtokens/evm_tokens.go
pro-wh: let's add that

analyzer/evmtokens/evm_tokens.go
pro-wh: yeah I was thinking that too

analyzer/emerald/incoming.go
pro-wh: it's not settled, and the total supply dead reckoning isn't implemented yet. if we do, then I suppose this is where we'll put it. current issue is how do we coordinate updates to the total supply in the db? will it be two separate numbers (one dead reckoned and one read from the contract)? how will the API know which one to use?

analyzer/modules/evm.go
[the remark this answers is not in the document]
pro-wh: oh there had been no logic for that :facepalm:

analyzer/modules/evm.go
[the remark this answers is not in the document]
pro-wh: should be used now

analyzer/modules/evm.go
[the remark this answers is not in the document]
pro-wh: reimplemented. now:

- "mutate" data is now called "mutable" to be a little less weird
- mutable data is a subset of the token data, so it's an embedded field in it now
- mutable data is a subset of the token data, so evmDownloadTokenERC20 calls evmDownloadTokenERC20Mutable

storage/migrations/02_oasis_3_emerald.up.sql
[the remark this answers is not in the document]
pro-wh: so this actually has a different meaning. we're not doing the dead reckoning yet. it's only a number used to see if the token analyzer needs to process this token.

storage/migrations/02_oasis_3_emerald.up.sql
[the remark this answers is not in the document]
pro-wh: added

analyzer/modules/evm.go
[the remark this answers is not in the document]
pro-wh: added that logic :skull:

analyzer/modules/evm.go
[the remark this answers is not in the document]
pro-wh: link added

analyzer/evmtokens/evm_tokens.go
pro-wh: added

analyzer/emerald/incoming.go
mitjat: ok let's shelve dead reckoning for now. I expect that the token analyzer will catch up fast, and also that hardly anyone cares about total supply being exactly right.

For the other concerns: I was thinking the two analyzers would use the same output column, and the evm_token analyzer would only overwrite the data if it was obtained at round that is >= the round of dead reckoning. Maybe I'm missing a corner case?

analyzer/emerald/incoming.go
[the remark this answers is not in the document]
mitjat: ooh tricksy hobbitses
:+1:

analyzer/modules/evm.go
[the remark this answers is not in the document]
mitjat: nice!

mitjat (approved):
:sheep: it!

analyzer/evmtokens/evm_tokens.go
pro-wh: hm, package names in oasis core:

- gen_vectors
- example_signer_plugin
- gen_account_vectors

- pluginsigner
- remotesigner
- fixgenesis
- consensusrpc
- configparser
- testdata
- policywatcher
- enclaverpc
- keymanager
- txsource
- supplementarysanity
- roothash
- testvectors
- prettyprint
- workerpool
- accessctl
- tuplehash
- mathrand
- keyformat

- extract-metrics

packages names in runtime sdk:

- consensusaccounts

pro-wh: https://google.github.io/styleguide/go/decisions#package-names oh I found this, maybe we've been following some stuff from Google's style guide for things where we had been indifferent

analyzer/emerald/incoming.go
pro-wh: sure, let's shelve.

> evm_token analyzer would only overwrite the data if it was obtained at round that is >= the round of dead reckoning

hm yeah, i.e. if the token analyzer is fully caught up. possibly a little fiddly in SQL. we need to join with the round progress table or something. in an UPDATE query :thinking:

pro-wh: resolving for merge. added a link from #293

analyzer/evmtokens/evm_tokens.go
pro-wh: resolving for merge

analyzer/evmtokens/evm_tokens.go
pro-wh: added

analyzer/emerald/incoming.go
mitjat: > we need to join with the round progress table or something. in an UPDATE query.

I played around with that before I realized we don't have dead reckoning yet, actually :)
You can do something like
```sql
UPDATE tokens
SET ...
WHERE contract_address = ... AND
current_round >= (
	SELECT last_updated_round
	FROM token_analysis_progress
	WHERE contract_address=tokens.contract_address
)
```

analyzer/evmtokens/evm_tokens.go
mitjat: Thanks for the pointer. Interesting; the styleguide just asserts "no underscores", and links to a Go(ogle) blog post that also just asserts "no underscores" without further justification.

PR #285, opened 2023-01-17
analyzer/metadata_registry: Support for test backend

mitjat:
This PR adds a script that lets us check the HTTP output of various endpoints against old versions of the indexer.

The script is *NOT* integrated into CI. Its major caveat is that it's most informative when you run the indexer against prod data for a good number of rounds, and then check what it outputs. I don't think the script is useful enough to do it on every CI run -- it takes time, and it increases brittleness (because the test needs an external, synced oasis-node).

Changes in the PR:
- `tests/e2e_regression/run.sh` - the main script
- `tests/e2e_regression/expected/*` - files with the output as produced currently. I'm not checking these in because of the one-off nature of the script's intended usage.
-  analyzer/metadata_registry.go - a patch that lets one set up an analyzer with deterministic outputs: Use a config that runs the consensus+emerald analyzer only up to a certain height (this was already possible before this PR), and use the metadata_registry analyzer against a static intended-for-test branch of the registry.


tests/e2e_regression/run.sh
```diff
@@ -0,0 +1,98 @@
...
+    echo "ERROR: test case $name is not unique"
+    exit 1
+  fi
+  seen+=("$name")
+
+  echo "Running test case: $name"
+  # Fetch the server response for $url
+  curl --silent --show-error --dump-header "$outDir/$name.headers" "$url" > "$outDir/$name.body"
+  # Try to pretty-print the output
+  <"$outDir/$name.body" jq >/tmp/pretty 2>/dev/null && cp /tmp/pretty "$outDir/$name.body" || true
```
pro-wh, line 84: is this separation to handle non-json responses?

pro-wh (approved):
alright, but this might rot

tests/e2e_regression/run.sh
```diff
@@ -0,0 +1,98 @@
+#!/bin/bash
+
+# This script is a simple e2e regression test for the indexer.
+# 
+# It fetches a fixed set of URLs from the indexer and saves the responses to files,
+# then check that the responses match expected outputs (from a previous run).
+# different from the previous run.
```
aefhm, line 7:
```suggestion
```

tests/e2e_regression/run.sh
mitjat: Yes. Expanded the comment to clarify.

mitjat:
> alright, but this might rot

Agreed. I'd say it's mostly valuable for refactoring PRs. I so happened to be working on these more, and found this test useful (and it found a bug!) several times already, and we're likely to do some more refactoring to accommodate multiple runtimes. So I think it has short-term potential :)

PR #286, opened 2023-01-17
Implement `/emerald/transactions/{tx_hash}`

### LINKED ISSUE #264: Add endpoint to fetch Emerald tx by hash
URL: `/emerald/transactions/{tx_hash}`, analogous to what [consensus has](https://github.com/oasisprotocol/oasis-indexer/blob/3d9514fa6a2b7599c59689e6d736dd1fab24cc31/api/spec/v1.yaml#L282)

The backend should match on both kinds of hashes: "oasis" hash and eth hash.

mitjat:
Closes #264

### DIFF
--- a/api/spec/v1.yaml
+++ b/api/spec/v1.yaml
@@ -768,6 +768,26 @@ paths:
                 $ref: '#/components/schemas/RuntimeTransactionList'
         <<: *common_error_responses
 
+  /emerald/transactions/{tx_hash}:
+    get:
+      summary: Returns an Emerald transaction with the given transaction hash.
+      parameters:
+        - in: path
+          name: tx_hash
+          required: true
+          schema:
+            type: string
+          description: The transaction hash of the transaction to return.
+          example: *tx_hash_1
+      responses:
+        '200':
+          description: The requested Emerald transactions.
+          content:
+            application/json:
+              schema:
+                $ref: '#/components/schemas/RuntimeTransaction'
+        <<: *common_error_responses
+
   /emerald/tokens:
     get:
       summary: Returns a list of ERC-20 tokens on Emerald.
--- a/storage/client/client.go
+++ b/storage/client/client.go
@@ -1328,6 +1328,7 @@ func (c *StorageClient) RuntimeTransactions(ctx context.Context, p apiTypes.GetE
 		ctx,
 		qf.RuntimeTransactionsQuery(),
 		p.Block,
+		nil, // tx_hash; filter not supported by this endpoint
 		p.Limit,
 		p.Offset,
 	)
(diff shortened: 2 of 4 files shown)


storage/client/client.go
```diff
@@ -1315,6 +1315,42 @@ func (c *StorageClient) RuntimeTransactions(ctx context.Context, p apiTypes.GetE
...
+		return nil, apiCommon.ErrBadChainID
+	}
+	runtime, ok := ctx.Value(common.RuntimeContextKey).(string)
+	if !ok {
+		return nil, apiCommon.ErrBadRuntime
+	}
+	qf := NewQueryFactory(cid, runtime)
+
+	t := RuntimeTransaction{}
+	err := c.db.QueryRow(
```
pro-wh, line 1331: so this reads the first matching one, right? is that by default the least recently inserted one?

pro-wh (approved):

pro-wh:
oh you'll have to rebase

this helps with these branch chains
https://git-scm.com/docs/git-rebase#Documentation/git-rebase.txt---update-refs

csillag:
> oh you'll have to rebase

I was wondering why does it take 30 commits... so that's why.

csillag:
@mitjat could you please provide me with an (example) JSON return data, so that I can mock this endpoint for the frontend until this is merged and deployed? Thanks.

api/spec/v1.yaml
```diff
@@ -759,6 +759,26 @@ paths:
                 $ref: '#/components/schemas/RuntimeTransactionList'
         <<: *common_error_responses
 
+  /emerald/transactions/{tx_hash}:
+    get:
+      summary: Returns a list of Emerald transactions.
```
aefhm, line 764: No longer a list?

mitjat:
> this helps with these branch chains
> [git-scm.com/docs/git-rebase#Documentation/git-rebase.txt---update-refs](https://git-scm.com/docs/git-rebase#Documentation/git-rebase.txt---update-refs)

Ow sorry for scaring everyone with the huge commit list. I guess I forgot to set the base branch in github :/

Thanks, I I'm actually using update-refs. What I'd need instead is something like `update-refees` :) -- when I rebase the base branch, I'd want all the branches that are branched off of it (this time, I had multiple branches branching off of the big refactor PR) to rebase automatically. But that's problematic to do automatically because it can introduce conflicts.

api/spec/v1.yaml
mitjat: thanks, fixed

storage/client/client.go
mitjat: Answer before your comment:
`QueryRow()` throws an error if there's 0 or >1 result rows. So for ambiguous hashes, we'd return HTTP 500 right now. Same in consensus.

Now:
Still the same in consensus (I didn't touch it), but here I refactored so that we silently do a `LIMIT 1`, and yes, it's the most recent one that we return.

PR #287, opened 2023-01-19
bugfix: Autogenerate code before packaging a release

mitjat:
#280  introduces the use of autogenerated files that are not checked into the repo. It also fixed the CI targets to build those autogeenrated files before doign linting, building etc, but it forgot to update the CI rule for making a release.

I _think_ this PR fixes hings, but I also think the only way to verify is to merge the PR and try creating another release :/


csillag (approved):
Sure, it makes sense. Let's give it a spin!

Andrew7234 (approved):

Issue #256, opened 2022-12-20
Export readily available data through API for MVP


mitjat:
#247 extends our openapi spec with a number of fields, all required for the Explorer MVP.
None of those fields are actually populated by the server today; however, they should all be readily available from the existing DB.
Go through the diffs in #247 and implement all the new fields.

It would be somewhat nice to have #255 before this to ensure we've implemented everything, but it is not at all a blocker.

PR #288, opened 2023-01-20
Track native balances in runtimes. Per-denomination events from `accounts` module. 

mitjat:
Pre-work for #256 . It turns out not all data was "readily available"  :|

The PR brings two new features:
- A table with dead-reckoned balances for "oasis-sdk native" runtime balances. This is separate from Warren's work on runtime balances, which tracks ERC-20 tokens and is limited to runtimes using the `evm` module. This PR is for runtimes using the `accounts` module, which so far is all of them.
- Expands logged runtime events  (transfers, burns, mints) to include the denomination, which we completely ignored so far. Right now all runtimes only have a single denomination (ROSE or TEST), but it would be painful to expand the APIs later.

### DIFF
--- a/analyzer/modules/consensusaccounts.go
+++ b/analyzer/modules/consensusaccounts.go
@@ -7,6 +7,7 @@ import (
 	"github.com/oasisprotocol/oasis-indexer/analyzer"
 	"github.com/oasisprotocol/oasis-indexer/log"
 	"github.com/oasisprotocol/oasis-indexer/storage"
+	"github.com/oasisprotocol/oasis-sdk/client-sdk/go/modules/consensusaccounts"
 )
 
 const (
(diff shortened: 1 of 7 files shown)


analyzer/modules/consensusaccounts.go
```diff
@@ -97,28 +90,28 @@ func (h *ConsensusAccountsHandler) queueWithdraws(batch *storage.QueryBatch, dat
...
+			h.qf.RuntimeWithdrawInsertQuery(),
+			data.Round,
+			withdraw.From.String(),
+			withdraw.To.String(),
+			withdraw.Amount.Amount.String(),
+			withdraw.Nonce,
+			errorModule,
+			errorCode,
+		)
+		// Do not dencrease the recipient's runtime balance at this point;
```
Andrew7234, line 104:
```suggestion
		// Do not decrease the recipient's runtime balance at this point;
```

Andrew7234 (approved):
Looks good; hopefully Warren also has time to take a pass as well

PR #295, opened 2023-01-24
Api spec + db changes for indexing emerald events

Andrew7234:
Notes:
- We do not support events from the `contracts` module.
- We only index the first topic (hash of event signature). This means that unsupported event types (ie anything besides erc20 transfer + approve) will not be searchable by their indexed topics other than the event signature. For supported event types, we parse and store the event params in `evm_log_params` and index this column.

### DIFF
--- a/api/spec/v1.yaml
+++ b/api/spec/v1.yaml
@@ -86,6 +86,8 @@ x-examples:
     - &runtime_block_round_1 3283246
   runtime-block-hash:
     - &runtime_block_hash_1 '21c243cd34bedfc234f1b45615d10a868f0655f59578f063a7d2fc8c6e5b4009'
+  runtime-event-type:
+    - &runtime_event_type_1 'consensus_accounts.deposit'
   iso-timestamp:
     - &iso_timestamp_1 '2022-03-01T00:00:00Z'
     - &iso_timestamp_2 '2019-04-01T00:00:00Z'
@@ -766,6 +768,74 @@ paths:
             application/json:
               schema:
                 $ref: '#/components/schemas/RuntimeTransactionList'
+        <<: *common_error_responses
+  
+  /emerald/events:
+    get:
+      summary: Returns a list of Emerald events.
+      parameters:
+        - *limit
+        - *offset
+        - in: query
+          name: block
+          schema:
+            type: integer
+            format: int64
+          description: A filter on block round.
+          example: *runtime_block_round_1
+        - in: query
+          name: tx_index # TODO: Is it possible to get the tx_index from the oasis-sdk (currently only returns txhash)? Could make a second call to fetch the tx by its hash...
+          schema:
+            type: integer
+            format: int32
+          description: |
+            A filter on transaction index. The returned events all need to originate
+            from a transaction that appeared in `tx_index`-th position in the block.
+            It is invalid to specify this filter without also specifying a `block`.
+            Specifying `tx_index` and `round` is an alternative to specifying `tx_hash`;
+            either works to fetch events from a specific transaction.
+          example: 3
+        - in: query
+          name: tx_hash
+          schema:
+            type: string
+          description: |
+            A filter on the hash of the transaction that originated the events.
+            Specifying `tx_index` and `round` is an alternative to specifying `tx_hash`;
+            either works to fetch events from a specific transaction.
+          example: *tx_hash_1
+        - in: query
+          name: rel
+          schema:
+            type: string
+          description: |
+            A filter on related accounts. Every returned event will refer to
+            this account. For example, for a `accounts.Transfer` event, this will be
+            the sender or the recipient of tokens.
+          example: *staking_address_1
+        - in: query
+          name: type
+          schema:
+            $ref: '#/components/schemas/RuntimeEventType'
+          description: A filter on the event type.
+        - in: query
+          name: evm_log_signature
+          schema:
+            type: string
+          description: |
+            A filter on the first of `topics` in the EVM log structure, which typically contains the 
+            event _signature_, i.e. the keccak256 hash of the event name and parameter types.
+            Note: The filter will match on `topics[0]` even in the rare case of anonymous events 
+            when that field does not actually contain the signature.
+          example: '0x27f12abfe35860a9a927b465bb3d4a9c23c8428174b83f278fe45ed7b4da2662'
+      responses:
+        '200':
+          description: |
+            Runtime events matching the filters, sorted by most recent first.
+          content:
+            application/json:
+              schema:
+                $ref: '#/components/schemas/RuntimeEventList'
         <<: *common_error_responses
 
   /emerald/transactions/{tx_hash}:
@@ -1675,6 +1745,72 @@ components:
           example: 118597
       description: |
         A ParaTime block.
+
+    RuntimeEventList:
+      type: object
+      required: [events]
+      properties:
+        events:
+          type: array
+          items:
+            $ref: '#/components/schemas/RuntimeEvent'
+      description: |
+        A list of runtime events.
+
+    RuntimeEvent:
+      type: object
+      required: [round, type, body]
+      properties:
+        round:
+          type: integer
+          format: int64
+          description: The block height at which this event was generated.
+          example: *runtime_block_round_1
+        tx_hash:
+          type: string
+          nullable: true
+          description: |
+            Hash of this event's originating transaction
+            Absent if the event did not originate from a transaction.
+          example: *tx_hash_1
+        type: 
+          description: The type of the event.
+          $ref: '#/components/schemas/RuntimeEventType'
+        body:
+          type: object
+          description: |
+            The decoded event contents. This spec does not encode the many possible types;
+            instead, see [the Go API](https://pkg.go.dev/github.com/oasisprotocol/oasis-sdk/client-sdk/go/modules).
+            This object will conform to one of the `*Event` types two levels down
+            the hierarchy (e.g. `MintEvent` from `accounts > Event > MintEvent`),
+            OR `evm > Event`.
+        evm_log_name:
+          type: string
+          nullable: true
+          description: |
+            If the event type is `evm.log`, this field describes the human-readable type of evm event, e.g.
+            `Transfer`. We currently only support two types of evm events, ERC20 `Transfer` and `Approve`.
+            Absent if the event type is not `evm.log`.
+          example: 'Transfer'
+        evm_log_params:
+          type: object
+          description: |
+            The decoded `evm.log` event data. We currently support only two types of evm events, ERC20 `Transfer` 
+            and `Approve`.
+            Absent if the event type is not `evm.log`.
+      description: An event emitted by the runtime layer
+
+    RuntimeEventType:
+      type: string
+      enum:
+        - accounts.transfer
+        - accounts.burn
+        - accounts.mint
+        - consensus_accounts.deposit
+        - consensus_accounts.withdraw
+        - core.gas_used
+        - evm.log
+      example: *runtime_event_type_1
 
     RuntimeTransactionList:
       type: object
--- a/storage/migrations/02_oasis_3_emerald.up.sql
+++ b/storage/migrations/02_oasis_3_emerald.up.sql
@@ -232,6 +232,32 @@ CREATE TABLE oasis_3.runtime_sdk_balances (
 
 -- -- -- -- -- -- -- -- -- -- -- -- -- -- -- -- -- -- -- -- -- -- -- -- -- -- -- -- -- -- -- -- -- --
 
+-- Events emitted from the emerald paratime
+CREATE TABLE oasis_3.runtime_events
+(
+  runtime runtime NOT NULL,
+  round UINT63 NOT NULL,
+  tx_index UINT31,
+  FOREIGN KEY (runtime, round, tx_index) REFERENCES oasis_3.runtime_transactions(runtime, round, tx_index) DEFERRABLE INITIALLY DEFERRED,
+  
+  tx_hash HEX64,
+  -- TODO: add link to openapi spec section with runtime event types.
+  type TEXT NOT NULL,
+  -- The raw event, as returned by the oasis-sdk runtime client. 
+  -- `evm.log` events are further parsed into known event types, 
+  -- e.g. (ERC20) Transfer, to populate the `evm_log_name` and 
+  -- `evm_log_params` fields below.
+  body JSON NOT NULL,
+  evm_log_name TEXT,
+  -- The event signature, if it exists, will be the first topic.
+  evm_log_signature TEXT GENERATED ALWAYS AS (body->'topics'->>0) STORED,
+  evm_log_params JSONB,
+  related_accounts oasis_addr[]  
+);
+CREATE INDEX ix_runtime_events_related_accounts ON oasis_3.runtime_events USING gin(related_accounts);
+CREATE INDEX ix_runtime_events_evm_log_signature ON oasis_3.runtime_events(evm_log_signature);
+CREATE INDEX ix_runtime_events_evm_log_params ON oasis_3.runtime_events USING gin(evm_log_params);
+
 -- Grant others read-only use.
 -- (We granted already in 01_oasis_3_consensus.up.sql, but the grant does not apply to new tables.)
 GRANT SELECT ON ALL TABLES IN SCHEMA oasis_3 TO PUBLIC;
--- a/api/v1/strict_server.go
+++ b/api/v1/strict_server.go
@@ -267,3 +267,7 @@ func (srv *StrictServerImpl) GetEmeraldTransactionsTxHash(ctx context.Context, r
 
 	return apiTypes.GetEmeraldTransactionsTxHash200JSONResponse(apiTx), nil
 }
+
+func (srv *StrictServerImpl) GetEmeraldEvents(ctx context.Context, request apiTypes.GetEmeraldEventsRequestObject) (apiTypes.GetEmeraldEventsResponseObject, error) {
+	return apiTypes.GetEmeraldEvents200JSONResponse{}, nil // Not implemented
+}


api/spec/v1.yaml
```diff
@@ -1063,7 +1124,7 @@ components:
 
     ConsensusEvent:
       type: object
-      required: [height, type, body]
+      required: [block, type, body]
```
Andrew7234, line 1127: this is unrelated to the PR; will move it to a PR on mitja's branch for fixing api spec

api/spec/v1.yaml
```diff
@@ -1614,6 +1675,76 @@ components:
...
+              example: *runtime_address_1
+            to:
+              type: string
+              description: The evm address the value was transferred to.
+              example: *runtime_address_2
+            value:
+              type: uint64
+              description: The balance that was transferred. (in wei? nano-ROSE?)
+              example: 1000000000
+          description: An ERC20 Transfer Event.
```
Andrew7234, line 1746: I _don't_ think we can have object enums; but leaving this here as a discussion starting point

mitjat: You mean "sum types", aka [lots of other names](https://en.wikipedia.org/wiki/Tagged_union)?
openapi has `oneOf`, though I don't expect it maps elegantly to Go (which doesn't have native support for sum types, unlike e.g. Rust or Typescript)

Regardless, we don't want to enumerate the possible types here. Right now we only support erc20 Transfer events. But eventually, we'll have "contract verification", meaning that we'll be able to pull the ABIs of most (hopefully most, anyway) contracts, meaning we'll be able to "understand"/parse those contracts' events, and the result will be of an arbitrary type. It won't (necessarily) have any of the `from`, `to`, `value` fields. 

FWIW, right now, we're able to understand/parse erc20 transfer events because we [hardcoded](https://github.com/oasisprotocol/oasis-indexer/blob/1016ed82c9010370291719a13622a923f9d403a0/analyzer/evmabi/contracts/artifacts/ERC20.json#L230-L251) its ABI into our source code.

So the whole `EvmRuntimeEvent` type is too specific; we won't use it, or anything like it.

api/spec/v1.yaml
```diff
@@ -1614,6 +1675,76 @@ components:
...
+    RuntimeEventType:
+      type: string
+      enum:
+        - accounts.transfer
+        - accounts.burn
+        - accounts.mint
+        - consensus_accounts.deposit
+        - consensus_accounts.withdraw
+        - core.gas_used
+        - evm.erc20_transfer
```
mitjat, line 1727:
```suggestion
        - evm.log
```

At the paratime layer, all EVM events are of a single type: evm.Log.
The SDK [shows this](https://github.com/oasisprotocol/oasis-sdk/blob/b37e6da699df331f5a2ac62793f8be099c68469c/runtime-sdk/modules/evm/src/lib.rs#L302-L309), but then the Go library for interfacing with the SDK just hides away this only subtype, and instead makes `Error` directly contain the three fields that (in Rust) are in Error::Log.

I think it will be confusing if we mix the two kinds (paratime and evm) of event types here. So I'd stick with  just `evm.Log` here.

api/spec/v1.yaml
```diff
@@ -1614,6 +1675,76 @@ components:
...
+          description: The block height at which this event was generated.
+          example: *runtime_block_round_1
+        tx_hash:
+          type: string
+          nullable: true
+          description: |
+            Hash of this event's originating transaction
+            Absent if the event did not originate from a transaction.
+          example: *tx_hash_1
+        type:
```
mitjat, line 1705: See comment in DB and on RuntimeEventType. We probably want another filter here, `evm_log_name`. I think that one should be a free-form string; as mentioned elsewhere, we cannot enumerate all log names.
Also important, an example value here is "Transfer" (verbatim from the log), not "erc20.Transfer" or something more semantic. Yes the user might get some non-erc20 tokens' logs in response, but so it goes. If this is a problem in practice, we can add _yet another_ filter, "give me only events that come from an EVM contract that is of type X" (where X is the filter value, and is for example ERC20). But that will complicate the DB query, so let's not
do it until/if actually needed.

api/spec/v1.yaml
```diff
@@ -767,6 +769,65 @@ paths:
...
+            Specifying `tx_index` and `round` is an alternative to specifying `tx_hash`;
+            either works to fetch events from a specific transaction.
+          example: *tx_hash_1
+        - in: query
+          name: rel
+          schema:
+            type: string
+          description: |
+            A filter on related accounts. Every returned event will refer to
+            this account. For example, for a `Transfer` event, this will be the
```
mitjat, line 813:
```suggestion
            this account. For example, for a `Transfer` event, this will be
```

mitjat, line 813:
```suggestion
            this account. For example, for an `accounts.Transfer` event, this will be the
```

api/spec/v1.yaml
```diff
@@ -767,6 +769,65 @@ paths:
...
+            the sender or the recipient of tokens.
+          example: *staking_address_1
+        - in: query
+          name: type
+          schema:
+            $ref: '#/components/schemas/ConsensusEventType'
```
mitjat, line 819:
```suggestion
            $ref: '#/components/schemas/RuntimeEventType'
```

api/spec/v1.yaml
```diff
@@ -767,6 +769,65 @@ paths:
...
+          description: A filter on the event type.
+          example: *runtime_event_type_1
```
mitjat, line 821: Nit:
You can delete this line and bake the example into RuntimeEventType.
Also, no need for the YAML references (`&` and `*`). Just list the example directly, it makes the YAML more readable. Since the example is a part of the type, we won't need to reference it anywhere else.

mitjat:
This post does not cover _exactly_ what you need for this task, but there's good overlap so I think you'll find it useful; I did, anyway.
https://medium.com/mycrypto/understanding-event-logs-on-the-ethereum-blockchain-f4ae7ba50378

api/spec/v1.yaml
mitjat: Thanks! It was already fixed in #280 though, I thought.

storage/migrations/02_oasis_3_emerald.up.sql
```diff
@@ -204,6 +204,30 @@ CREATE TABLE oasis_3.emerald_withdraws
...
+  type TEXT NOT NULL, 
+  -- The raw event, as returned by the oasis-sdk runtime client. This field is optional 
+  -- and will only exist for events from the `contracts` or `evm` modules, see
+  -- [contracts](https://pkg.go.dev/github.com/oasisprotocol/oasis-sdk/client-sdk/go@v0.2.0/modules/contracts#Event)
+  -- [evm](https://pkg.go.dev/github.com/oasisprotocol/oasis-sdk/client-sdk/go@v0.2.0/modules/evm#Event)
+  raw JSON, -- Raw event byte blob
+  -- The parsed event. Events from the `accounts`, `consensusaccounts`, and `core` modules are unchanged.
+  -- Events from the `evm` and `contracts` modules are parsed into known event types, e.g. ERC20.transfer.
+  -- If the raw event cannot be decoded, this field is empty.
+  body JSON, -- Parsed event
```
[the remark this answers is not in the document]
Andrew7234: Nope we don't _need_ `raw` but it sounds like what you're suggesting is similar? Essentially the `raw` would now be stored in `body` and the parsed `evm` event in `body` would be stored in `evm_log_params` instead.

My original thinking was that `raw` would store the raw Event struct returned by the oasis-sdk `runtimeClient`, and that `body` would store the processed version of that event after being processed by the indexer. The indexer would define its own custom structs for possible `contract` or `evm` events and store those in `body`. When queried for event info, the indexer would only look at the `body` col and return info about events it knows how to parse. This should make queries simpler.

I'd prefer flattening the various module `Event` types with the `Evm` events because most users will want to see the `evm` event types they're familiar with. However, it doesn't necessarily mean we have to store it this way in the database, and I agree that what you're suggesting better mirrors the 2-layered hierarchy of runtime events.

api/spec/v1.yaml
Andrew7234: Neat, I didn't know about tagged unions before.

> we'll be able to "understand"/parse those contracts' events, and the result will be of an arbitrary type

Just curious - are we expecting to have some automation for the "contract verification," so parsing other runtimeEvent types won't require manual parsing / hardcoding?

> So the whole EvmRuntimeEvent type is too specific

Got it, will remove

api/spec/v1.yaml
Andrew7234: Hm would it be better to flatten/mix the paratime and evm event types? Assuming we're designing for ethereum developers, it would be nice to expose the events in recognizable form, eg `ERC20.transfer`. 

Alternatively, per your suggestion below we could intead return the `evm.Log` raw event, as well as the `evm_log_name` and `evm_log_params` and then the frontend could display it that way. Open to this as well.

storage/migrations/02_oasis_3_emerald.up.sql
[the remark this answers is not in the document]
mitjat: I think we've agreed on the format pretty well on the call and in slack, so I'll keep this shorter.

For posterity, here's the excellent summary you just posted and that touches on several of the points above:

> We won’t be flattening the event types bc 1) many potential event types once we have contract verification, 2) it’s confusing to have raw evm.Log events when we cannot parse the event but also parsed evm.Transfer event on the same level, 3) it doesn’t reflect the 2-level event hierarchy well.

Also, about this:

> better mirrors the 2-layered hierarchy of runtime events

I do think we should aim for this. Makes it easier to understand the output for power users that are familiar with (or trying to wrap their head around) the internals of the oasis blockchain. (Probably the great majority of our consumers are power users; non-power users will use the GUI, and the GUI engineers are power users themselves and will IMO like the rawer presentation better too.) Also, makes it easier for us to describe the output.

api/spec/v1.yaml
mitjat: yes, let's do the latter please, as discussed on the call and in the thread above. We've stuck pretty close to the raw on-chain formats and presentation in other parts of the API, I think it's a good high-level guiding principle.

api/spec/v1.yaml
Andrew7234: Got it, added!

api/spec/v1.yaml
```diff
@@ -1614,6 +1681,71 @@ components:
...
+          description: |
+            If the event type is `evm.Log`, this field describes the type of evm event, e.g. `Transfer`.
+            Absent if the event type is not `evm.Log`.
+          example: 'Transfer'
+        evm_log_params:
+          type: object
+          description: |
+            The decoded `evm.Log` event data. We currently support only two types of evm events, ERC20 `Transfer` 
+            and `Approve`.
+            Absent if the event type is not `evm.Log`.
```
mitjat, line 1734: nit here and elsewhere: We have `evm.log` (lowercase) in the enum. I'm fine with changing either. I'm low-key regretting one of the enums I made in the past that's all-lowercase, looks a little weird ... OTOH I was following the JSON names of fields in the top-level `Event` struct. Whatever, as long as we're consistent, I guess.

storage/migrations/02_oasis_3_emerald.up.sql
```diff
@@ -204,6 +204,29 @@ CREATE TABLE oasis_3.emerald_withdraws
...
+  -- TODO: add link to openapi spec section with runtime event types.
+  type TEXT NOT NULL,
+  -- The raw event, as returned by the oasis-sdk runtime client. 
+  -- `evm.Log` events are further parsed into known event types, 
+  -- e.g. (ERC20) Transfer, to populate the `evm_log_name` and 
+  -- `evm_log_params` fields below.
+  body JSON NOT NULL,
+  evm_log_name TEXT,
+  evm_log_params JSONB,
+  related_accounts TEXT[],
```
mitjat, line 223: nit:
```suggestion
  related_accounts oasis_account[],
```
unless something goes wrong on the pgx side then

api/spec/v1.yaml
```diff
@@ -1614,6 +1681,71 @@ components:
...
+          description: The type of the event.
+          $ref: '#/components/schemas/RuntimeEventType'
+        body:
+          type: object
+          description: |
+            The decoded event contents. This spec does not encode the many possible types;
+            instead, see [the Go API](https://pkg.go.dev/github.com/oasisprotocol/oasis-sdk/client-sdk/go/modules).
+            This object will conform to one of the `*Event` types two levels down
+            the hierarchy, e.g. `MintEvent` from `Event > accounts.Event > MintEvent`,
+            OR `evm.Event` directly.
```
mitjat, line 1721:
```suggestion
            the hierarchy (e.g. `MintEvent` from `accounts > Event > MintEvent`),
            OR `evm > Event`.
```

this better describes what they see / what they have to click if they follow the provided Go API link above.

api/spec/v1.yaml
```diff
@@ -767,6 +769,71 @@ paths:
...
+            this account. For example, for a `accounts.Transfer` event, this will be
+            the sender or the recipient of tokens.
+          example: *staking_address_1
+        - in: query
+          name: type
+          schema:
+            $ref: '#/components/schemas/RuntimeEventType'
+          description: A filter on the event type.
+        - in: query
+          name: evm_log_name
```
mitjat, line 822: I expect people will more so want to filter by `evm_event_topic`, i.e. the hash of the event type signature. That's as close as we can get to "event type" in EVM. Also, the topic is always available; the name is not, because we have to first know the ABI to parse the event to derive the name. So even for unparsed events, it will be nice to have a link like "show me other events of this type" in the GUI. It's also more selective than just the name.

To index it, you can stuff it in a separate field in the DB:
```sql
indexer=# create table foo(j json);
CREATE TABLE
indexer=# insert into foo values ('{"topic": "bar"}'::json);
INSERT 0 1
indexer=# insert into foo values ('{"not_a_topic": "bar"}'::json);
INSERT 0 1
indexer=# select j->'topic' from foo;
 ?column? 
----------
 "bar"
 
(2 rows)



indexer=# create table foo2 (j json, x text generated always as (j->'topic') stored);
CREATE TABLE
indexer=# insert into foo2(j) values ('{"not_a_topic": "bar"}'::json);
INSERT 0 1
indexer=# insert into foo2(j) values ('{"topic": "barzz"}'::json);
INSERT 0 1
indexer=# select * from foo2;
           j            |    x    
------------------------+---------
 {"not_a_topic": "bar"} | 
 {"topic": "barzz"}     | "barzz"
(2 rows)

indexer=# create index adfasdggg on foo2 (x);
CREATE INDEX
```

mitjat (commented):
Nice! I think this is the final round :+1:

api/spec/v1.yaml
Andrew7234: got it, updated!

api/spec/v1.yaml
Andrew7234: Got it, updated. 

Note: I chose not to extract+index the other possible event topics because they are a subset of the `evm_log_params` which is indexed

storage/migrations/02_oasis_3_emerald.up.sql
```diff
@@ -232,6 +232,32 @@ CREATE TABLE oasis_3.runtime_sdk_balances (
 
 -- -- -- -- -- -- -- -- -- -- -- -- -- -- -- -- -- -- -- -- -- -- -- -- -- -- -- -- -- -- -- -- -- --
 
+-- Events emitted from the emerald paratime
+CREATE TABLE oasis_3.emerald_events
```
mitjat, line 236: As of https://github.com/oasisprotocol/oasis-indexer/pull/296, this should be `evm_events` because it applies to every runtime with an `evm` module (notably, sapphire).

mitjat (approved):
Thank you for the edits! Looking good :+1: , just a few more tiny comments.

api/spec/v1.yaml
```diff
@@ -1676,6 +1744,78 @@ components:
...
+          $ref: '#/components/schemas/RuntimeEventType'
+        body:
+          type: object
+          description: |
+            The decoded event contents. This spec does not encode the many possible types;
+            instead, see [the Go API](https://pkg.go.dev/github.com/oasisprotocol/oasis-sdk/client-sdk/go/modules).
+            This object will conform to one of the `*Event` types two levels down
+            the hierarchy (e.g. `MintEvent` from `accounts > Event > MintEvent`),
+            OR `evm > Event`.
+        evm_log_signature:
```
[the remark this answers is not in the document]
Andrew7234: Good point; removed!

storage/migrations/02_oasis_3_emerald.up.sql
Andrew7234: Hrm I'm not a fan of `evm_events` since there could be some non-evm runtime events from other modules, ie `accounts.Burn`. Going with `runtime_events` instead.

api/spec/v1.yaml
```diff
@@ -767,6 +769,72 @@ paths:
...
+          schema:
+            $ref: '#/components/schemas/RuntimeEventType'
+          description: A filter on the event type.
+        - in: query
+          name: evm_log_signature
+          schema:
+            type: string
+          description: |
+            A filter on the evm log signature, used for searching `evm.log` events. The
+            event signature is the keccak256 hash of the event name and parameter types.
```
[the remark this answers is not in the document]
Andrew7234: Thanks! I like your version better; updated

api/spec/v1.yaml
```diff
@@ -1676,6 +1746,72 @@ components:
...
+            instead, see [the Go API](https://pkg.go.dev/github.com/oasisprotocol/oasis-sdk/client-sdk/go/modules).
+            This object will conform to one of the `*Event` types two levels down
+            the hierarchy (e.g. `MintEvent` from `accounts > Event > MintEvent`),
+            OR `evm > Event`.
+        evm_log_name:
+          type: string
+          nullable: true
+          description: |
+            If the event type is `evm.log`, this field describes the human-readable type of evm event, e.g.
+            `Transfer`. We currently only support two types of evm events, ERC20 `Transfer` and `Approve.
```
aefhm, line 1792:
```suggestion
            `Transfer`. We currently only support two types of evm events, ERC20 `Transfer` and `Approve`.
```

PR #296, opened 2023-01-26
Generalize emerald tables to multiple runtimes 

mitjat:
Generalize former emerald tables so they can store data from multiple runtimes:
- adds a `runtime` column to all of them
- renames them appropriately
- udpates SQL queries appropriately

This is the first step towards https://app.clickup.com/t/3u9y285
The main benefit of this PR is that we can support new runtimes without copy-pasting `CREATE TABLE` and `CREATE INDEX` from e.g. Emerald. 

The first two commits are only vaguely related to the rest of the PR:
- The first one improves the ergonomy of e2e regression tests, so I was able to test every change.
- The second one adds an `ORDER BY` to all `SELECT` statements that use `LIMIT`. Without `ORDER BY`, an arbitrary subset is returned, which is a serious bug for windowed queries. Thanks, regression tests (they found this bug because the API output was not stable).  

Testing: 
Before and after this PR, indexed 200 blocks of emerald. With `tests/e2e_regression/run.sh`, compared the API output for endpoints that exercise most of the API (see list in `run.sh`). No regressions/changes detected.

### DIFF
--- a/storage/migrations/02_oasis_3_emerald.up.sql
+++ b/storage/migrations/02_oasis_3_emerald.up.sql
@@ -40,13 +40,14 @@ CREATE TABLE oasis_3.runtime_transactions
   raw         BYTEA NOT NULL,
   -- result_raw is cbor(CallResult).
   result_raw  BYTEA NOT NULL,
-  PRIMARY KEY (round, tx_index)
+  PRIMARY KEY (runtime, round, tx_index)
 );
 CREATE INDEX ix_runtime_transactions_tx_hash ON oasis_3.runtime_transactions USING hash (tx_hash);
 CREATE INDEX ix_runtime_transactions_tx_eth_hash ON oasis_3.runtime_transactions USING hash (tx_eth_hash);
 
-CREATE TABLE oasis_3.emerald_transaction_signers
+CREATE TABLE oasis_3.runtime_transaction_signers
 (
+  runtime        runtime NOT NULL,
   round          UINT63 NOT NULL,
   tx_index       UINT31 NOT NULL,
   -- Emerald processes mainly Ethereum-format transactions with only one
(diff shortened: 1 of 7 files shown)


pro-wh:
looking good so far. any other testing you want to do?

pro-wh:
I'm doing the tokens part in #284 fyi

mitjat:
> looking good so far. any other testing you want to do?

I think the regression testing is good enough for now. It makes me pretty confident that all the SQL is valid. As for correctness ... the regression test also checks for that pretty ok, I think it' relatively hard to mess up beyond that. The exceptions are mistakes around missing a `WHERE` or a `JOIN` condition that involves the new `runtime` column. I think I did them all correctly, but I'll also test that once I have enough generalization to instantiate two emerald analyzers. The plan is to name them "emerald1" and "emerald2" or something, then run emerald1 to height X and emerald2 to height Y. Then regression test their outputs against a standalone analyzer that runs to height X (or Y). That will convince me further that there's no crosstalk across runtimes.

mitjat:
I have now confirmed that tests pass.

mitjat:
> I'm doing the tokens part in #284 fyi

Yes, I noticed that the new table is already runtime-agnostic, thank you. For the existing runtime_tokens table, I think the merge conflict will be pretty light :crossed_fingers:

tests/e2e_regression/run.sh
```diff
@@ -28,8 +28,8 @@ testCases=(
   'accounts                 /v1/consensus/accounts'
   'min_balance              /v1/consensus/accounts?minTotalBalance=1000000'
   'big_int_balance          /v1/consensus/accounts?minTotalBalance=999999999999999999999999999'
-  'bad_big_int              /v1/consensus/accounts?minTotalBalance=NA'
-  'extraneous_key           /v1/consensus/accounts?foo=bar'
+  'accounts_bad_big_int     /v1/consensus/accounts?minTotalBalance=NA'
+  'accoutns_extraneous_key  /v1/consensus/accounts?foo=bar'
```
pro-wh, line 32: accoutns

storage/migrations/02_oasis_3_emerald.up.sql
```diff
@@ -2,8 +2,11 @@
 
 BEGIN;
 
-CREATE TABLE oasis_3.emerald_rounds
+CREATE TYPE runtime AS ENUM ('emerald', 'sapphire', 'cipher');
+
+CREATE TABLE oasis_3.runtime_blocks
```
pro-wh, line 7: what's the rounds -> blocks change? seems fine btw

pro-wh (approved):
lg, didn't read through the scripts tho

storage/migrations/02_oasis_3_emerald.up.sql
mitjat: Not strictly related to the rest of this PR. It's just for consistency; most of our internal code and the API all refer to this as blocks.

mitjat:
> I'll also test that once I have enough generalization to instantiate two emerald analyzers

Hah, I did find an error that would only show up once we have multiple active runtimes: The FKs into runtime_blocks were all referencing the `round`, not the `(runtime, round)` pair. Fixed now.

PR #297, opened 2023-01-27
db client: Simplify, clean up

### LINKED ISSUE #279: Return HTTP 404 for non-existing objects
Requesting a non-existing object, e.g. `/v1/consensus/entities/aaaaaaaaaaaaaaa+5RKkkfOTUR+XheHIohlqMu+a9As=/nodes/1TDdhVK+R8N6KYNAvGFV4LeFQfp06+sjJI7SgTsN35c=`, currently returns `HTTP 500  internal server error` because the DB backend fails to fetch a row from the DB.

Recognize that type of error and propagate it in a way that it produces a 404.

mitjat:
This PR cleans up a bunch of small things around our DB client that have been bothering me for months, but even more so lately when I've been interacting with the DB more.
Changes/improvements:
- Removes copy-pasted error logs when a SQL query fails. We log these errors in our DB wrapper library, and `pgx` proper logs them also, so they were triple-logged previously. (For QueryRow(), `pgx` design is a little different and we cannot easily have a wrapper there, so we don't log the error ourselves. Those errors used to be double-logged, and are now single-logged, by pgx.)
- Obviates manual decoding of BigInts when pulled from the DB; I added a method to `BigInt` that `pgx` can use to do the decoding automatically. It doesn't work for `*BigInt` (= API fields that are nillable) due to a pgx bug/restriction. I spent way too much time trying alternatives, including pgx v5 (which has a different API for custom types), but it looks like we best stick with what we have.
- Propagates "no row found" from QueryRow() into a HTTP 404. This resolves #279 . This commit is also a good starting point for anybody wanting to create more custom errors.
- Simplifies the construction of QueryFactory objects. We had a bunch of copy-pasted code that always fetched the chainId and runtime from the `ctx`; now, this happens in a helper func.

### DIFF
--- a/api/errors.go
+++ b/api/errors.go
@@ -3,6 +3,7 @@ package api
 import (
 	"encoding/json"
 	"errors"
+	"fmt"
 	"net/http"
 	"reflect"
 
@@ -19,21 +20,32 @@ var (
 	// ErrBadRuntime is returned when a malformed or missing runtime name
 	// is provided.
 	ErrBadRuntime = errors.New("unable to resolve runtime")
-	// ErrStorageError is returned when the underlying storage suffers
-	// from an internal error.
-	ErrStorageError = errors.New("internal storage error")
+	// ErrNotFound is returned when handling a request for an item that
+	// does not exist in the DB.
+	ErrNotFound = errors.New("item not found")
 )
 
+type ErrStorageError struct{ Err error }
+
+func (e ErrStorageError) Error() string { return fmt.Sprintf("storage error: %s", e.Err.Error()) }
+
 func HttpCodeForError(err error) int {
-	errType := reflect.ValueOf(err).Elem().Type()
+	errVal := reflect.ValueOf(err)
+	// dereference an interface or pointer
+	if errVal.Kind() == reflect.Interface || errVal.Kind() == reflect.Ptr {
+		errVal = errVal.Elem()
+	}
+	errType := errVal.Type()
 
 	switch {
 	case err == ErrBadChainID:
 		return http.StatusNotFound
-	case err == ErrStorageError:
-		return http.StatusInternalServerError
 	case err == ErrBadRequest:
 		return http.StatusBadRequest
+	case err == ErrNotFound:
+		return http.StatusNotFound
+	case errType == reflect.TypeOf(ErrStorageError{}):
+		return http.StatusInternalServerError
 	case (errType == reflect.TypeOf(apiTypes.InvalidParamFormatError{}) ||
 		errType == reflect.TypeOf(apiTypes.RequiredHeaderError{}) ||
 		errType == reflect.TypeOf(apiTypes.RequiredParamError{}) ||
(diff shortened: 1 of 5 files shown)


api/middleware.go
```diff
@@ -235,9 +235,15 @@ func RuntimeFromURLMiddleware(baseURL string) func(next http.Handler) http.Handl
 			// The first part of the path (after the version) determines the runtime.
 			// Recognize only whitelisted runtimes.
 			runtime := ""
-			switch { //nolint:gocritic // allow single-case switch for future expansions
+			switch {
```
aefhm, line 238: Should we also error check or throw?

aefhm (approved):

common/types.go
```diff
@@ -33,6 +36,56 @@ func (b *BigInt) UnmarshalJSON(text []byte) error {
...
+	*b = bigInt
+	return err
+}
+
+func (b BigInt) EncodeBinary(ci *pgtype.ConnInfo, buf []byte) (newBuf []byte, err error) {
+	numeric := pgtype.Numeric{Int: &b.Int, Exp: 0, Status: pgtype.Present}
+	return numeric.EncodeBinary(ci, buf)
+}
+
+// TODO: DEDUPLICATE - DO NOT SUBMIT PR BEFORE FIXING THIS
```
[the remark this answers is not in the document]
mitjat: :+1: this is fixed already, you were probably looking at an intermediate commit

api/middleware.go
mitjat: You mean in case no runtime matches? No, there are legitimate URL that just don't have an associated runtime, not even a runtime in the loose sense of "proper runtime or consensus". Example: `/v1/spec/v1.yaml`.

api/errors.go
```diff
@@ -19,21 +20,32 @@ var (
...
 )
 
+type ErrStorageError struct{ Err error }
+
+func (e ErrStorageError) Error() string { return fmt.Sprintf("storage error: %s", e.Err.Error()) }
+
 func HttpCodeForError(err error) int {
-	errType := reflect.ValueOf(err).Elem().Type()
+	errVal := reflect.ValueOf(err)
+	// dereference an interface of pointer
```
Andrew7234, line 34:
```suggestion
	// dereference an interface or pointer
```

storage/client/client.go
```diff
@@ -256,29 +189,19 @@ func (c *StorageClient) Transactions(ctx context.Context, p apiTypes.GetConsensu
...
 		var code uint64
-		var feeNum pgtype.Numeric
 		if err := rows.Scan(
 			&t.Block,
 			&t.Index,
 			&t.Hash,
 			&t.Sender,
 			&t.Nonce,
-			&feeNum,
+			&t.Fee,
```
Andrew7234, line 198: Just checking - this works for the custom db types (`UINT63` + `UINT31`) now?

Andrew7234 (approved):

storage/client/client.go
mitjat: I believe it always worked for those. The new thing is that it now works for `UINT_NUMERIC` (and `NUMERIC` variants in general), thanks to commit [`e7de6a4`](https://github.com/oasisprotocol/oasis-indexer/pull/297/commits/e7de6a4172b2eedaca4cfd643c708ca23bf6c7c2)

PR #298, opened 2023-01-30
genesis-test: Create DB snapshot using full tx isolation

mitjat:
As discussed on Slack, `mainnet-indexer-statecheck` (the cron job that runs genesis_test on consensus) consistently fails. A possible cause for failure is that the test creates an inconsistent snapshot of the indexer state, due to not using a strong enough DB tx isolation level.

This PR uses `SERIALIZABLE`, the strongest tx isolation level, for snapshot creation. 

More about DB tx isolation levels:
https://www.postgresql.org/docs/current/transaction-iso.html

I created a dev docker image and I'm deploying this now to see what the effect is.
```
kubectl -n monitoring create job --from=cronjobs/mainnet-indexer-statecheck manual-indexer-statecheck-mitjat
```

### DIFF
--- a/storage/api.go
+++ b/storage/api.go
@@ -40,6 +40,9 @@ type QueryResults = pgx.Rows
 // QueryResult represents the result from a read query.
 type QueryResult = pgx.Row
 
+// TxOptions encodes the way DB transactions are executed.
+type TxOptions = pgx.TxOptions
+
 // Queue adds query to a batch.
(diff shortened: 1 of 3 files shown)


mitjat:
A test run completed. Unfortunately, it still fails in the same way as before: some nonces, allowances etc are off.
We should still merge this PR; it's a legitimate bug in the test harness.

aefhm (approved):

PR #299, opened 2023-01-30
analyzer: Refactor configs

mitjat:
This PR restructures configs, with two goals in mind:
- Allow different configs (wrt their type/attributes) for different analyzers. Now that we have runtime analyzers but also various metadata analyzers, the lowest-common-denominator config structure has been getting confusing, with each analyzer using only some of its fields.
- Factor out node connection config. Previously, every runtime (+ consensus) had to specify it separately, which invites typos and discrepancies, especially with the upcoming multiple runtimes.

This is working towards https://app.clickup.com/t/3u9y285

### DIFF
--- a/cmd/analyzer/analyzer.go
+++ b/cmd/analyzer/analyzer.go
@@ -3,7 +3,6 @@ package analyzer
 
 import (
 	"context"
-	"fmt"
 	"os"
 	"sync"
 
@@ -85,7 +84,7 @@ func Init(cfg *config.AnalysisConfig) (*Service, error) {
 	}
 
 	m, err := migrate.New(
-		cfg.Migrations,
+		cfg.Storage.Migrations,
 		cfg.Storage.Endpoint,
 	)
 	if err != nil {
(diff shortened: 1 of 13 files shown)


cmd/analyzer/analyzer.go
```diff
@@ -157,8 +157,8 @@ func NewService(cfg *config.AnalysisConfig) (*Service, error) {
 		switch analyzerCfg.Name {
 		case consensus.ConsensusDamaskAnalyzerName, "consensus_main_damask": // TODO: drop "main" variant; as of Oct 2022, it exists only to support legacy helmfiles
 			a, err = consensus.NewMain(analyzerCfg, client, logger)
-		case emerald.EmeraldDamaskAnalyzerName, "emerald_main_damask": // TODO: drop "main" variant; as of Oct 2022, it exists only to support legacy helmfiles
-			a, err = emerald.NewMain(analyzerCfg, client, logger)
+		case runtime.EmeraldDamaskAnalyzerName, "emerald_main_damask": // TODO: drop "main" variant; as of Oct 2022, it exists only to support legacy helmfiles
```
aefhm: 👍

tests/e2e_regression/reindex_and_run.sh
```diff
@@ -7,18 +7,20 @@ set -euo pipefail
 
 cat >/tmp/e2e_config.yaml <<EOF
```
aefhm, line 8: Maybe this can be a separate file now 🤔

storage/oasis/client.go
```diff
@@ -22,7 +22,7 @@ type ClientFactory struct {
 }
 
 // NewClientFactory creates a new oasis-node client factory.
-// Unless `skipChainContextCheck` is false, it also checks that
+// If `skipChainContextCheck` is true, it also checks that
```
aefhm, line 25: 😄

aefhm (approved):

tests/e2e_regression/reindex_and_run.sh
mitjat: I'm not strongly opposed to it. But I like having it all in one place, and the config is pretty short. The goal of this file is basically to set up a test fixture, and in test code, it's common to just inline the fixtures for readability, as the code tends to be read more times than it's changed. It's a judgement call tho.

PR #300, opened 2023-01-31
openapi: Generic runtime URLs

mitjat:
This PR makes runtimes' URLs generic, e.g. `/emerald/transaction` becomes `/{runtime}/transaction`.

`oapi-codegen` imports a symbol called `runtime` in the autogenerated code, which clashes with our URL param of the same name. The PR includes a bash script (called from `Makefile`) to work around this issue.

The PR also includes some general small cleanups -- I searched for "emerald" in the codebase and tweaked all mentions/comments that will soon apply to all runtimes, not just emerald.

Testing: No regressions on URLs from `https://github.com/oasisprotocol/oasis-indexer/blob/3c756580ea3fe1d7d5949b4062c30693074d8ad8/tests/e2e_regression/run.sh`.


go.mod
```diff
@@ -21,7 +21,7 @@ require (
 	github.com/stretchr/testify v1.8.1
 	golang.org/x/sync v0.0.0-20220722155255-886fb9371eb4
 	google.golang.org/grpc v1.49.0
-	gopkg.in/yaml.v2 v2.4.0
+	gopkg.in/yaml.v2 v2.4.0 // indirect
```
mitjat, line 24: All the changes in go.sum and in this file are just due to a routine `go mod tidy`.

storage/migrations/02_oasis_3_runtimes.up.sql
```diff
@@ -74,6 +74,32 @@ CREATE TABLE oasis_3.runtime_related_transactions
 );
 CREATE INDEX ix_runtime_related_transactions_address_height_index ON oasis_3.runtime_related_transactions (account_address, tx_round, tx_index);
 
+-- Events emitted from the runtimes. Includes deeply-parsed EVM events from EVM runtimes.
```
mitjat, line 77: This whole block is just moved from below, so that the file keeps its structure:
1) a section of tables that are (mostly) not tied to sdk-modules
2) one section per sdk module
3) permissions

analyzer/evmtokens/evm_tokens.go
```diff
@@ -26,7 +26,7 @@ import (
 // last_download_round columns. The block analyzer updates last_mutate_round,
 // this token analyzer updates last_download_round.
 
-const EmeraldDamaskTokensAnalyzerName = "emerald_damask_tokens"
+const EmeraldDamaskTokensAnalyzerName = "evm_tokens"
```
aefhm, line 29: Should we rename this variable too?

analyzer/runtime/runtime.go
```diff
@@ -204,7 +203,7 @@ func (m *Main) prework() error {
 	// Register special addresses.
 	batch.Queue(
 		m.qf.AddressPreimageInsertQuery(),
-		rewards.RewardPoolAddress.String(),          // oasis1qp7x0q9qahahhjas0xde8w0v04ctp4pqzu5mhjav for emerald on mainnet oasis-3
+		rewards.RewardPoolAddress.String(),          // oasis1qp7x0q9qahahhjas0xde8w0v04ctp4pqzu5mhjav on mainnet oasis-3
```
Andrew7234, line 206: this is guaranteed to be the same for all current+future runtimes?

analyzer/evmtokens/evm_tokens.go
```diff
@@ -191,8 +191,7 @@ func (m Main) Start() {
 	backoff, err := util.NewBackoff(
 		100*time.Millisecond,
 		6*time.Second,
-		// ^cap the timeout at the expected
-		// emerald round time
+		// ^cap the timeout at the expected emerald round time
```
Andrew7234, line 194: `emerald` specifically? Will we use this for sapphire?

scripts/namespace_codegen_templates.sh
```diff
@@ -0,0 +1,22 @@
...
+
+# Create a working copy of the templates.
+rm -rf /tmp/namespaced-templates
+cp -r "$codegenDir/templates" /tmp/namespaced-templates
+find /tmp/namespaced-templates/ -type d -print0 | xargs -0 chmod 755
+
+# Rename the "runtime" package to "codegen_runtime" in all templates.
+find /tmp/namespaced-templates -type f -print0 | xargs -0 sed -i -E 's/\bruntime\b/codegen_runtime/g'
+# Fix the import statement to use the new name.
+sed -i -E 's!"github.com/deepmap/oapi-codegen/pkg/codegen_runtime"!codegen_runtime "github.com/deepmap/oapi-codegen/pkg/runtime"!' /tmp/namespaced-templates/imports.tmpl
```
Andrew7234, line 22: 🤯 

(going to assume this works if `make oasis-indexer` works 😄 )

analyzer/evmtokens/evm_tokens.go
mitjat: Sure, lemme do that. I'll do it in my morning; can you give me a LGTM if things otherwise look OK?

analyzer/runtime/runtime.go
mitjat: looks like it. Comes from [here](https://github.com/oasisprotocol/oasis-sdk/blob/b37e6da699df331f5a2ac62793f8be099c68469c/client-sdk/go/modules/rewards/types.go#L10-L10). I guess the original comment here was misguided.

analyzer/evmtokens/evm_tokens.go
mitjat: Not emerald specific, thanks. Lemme fix before merging the PR.

scripts/namespace_codegen_templates.sh
mitjat: bash is a dangerous weapon :)
it's not as unreadable as it looks at first glance; the commands are actually simply, just the syntax is ugly.

it does work; I don't much worry about it breaking since we're using a fixed version of the codegen. However :exploding_head: is :no_good:, and I seriously considered not doing this and renaming `runtime` instead. But it's such a prominently displayed, outward-facing variable in URLs, and there's not a good alternative. `rtime`, `rt` ... too cryptic, inconsistent with other names. `paratime` ... not used in code (and the name does penetrate our Go code in many places).

aefhm (approved):

scripts/namespace_codegen_templates.sh
Andrew7234: Makes sense, thanks for explaining the thought process

api/spec/v1.yaml
```diff
@@ -42,6 +42,14 @@ x-query-params:
...
       The block height from which to query state. The Oasis Indexer does not
       make any guarantees about availability of historical state data.
+  - &runtime
+    in: path
+    name: runtime
+    required: true
+    schema:
+      $ref: '#/components/schemas/Runtime'
+    description: |
+      The layer for which to return the transaction volume timeline.
```
Andrew7234, line 52:
```suggestion
      The layer for which to return the requested data timeline.
```

Andrew7234 (approved):
Looks good!

api/spec/v1.yaml
mitjat: Uh bad copy paste. Rewrote from scratch. Thank you!

analyzer/runtime/runtime.go
```diff
@@ -136,9 +136,8 @@ func (m *Main) Start() {
 
 	backoff, err := util.NewBackoff(
 		100*time.Millisecond,
+		// Cap the timeout at the expected round time. All runtimes currently have the same round time.
```
pro-wh, line 139: oh no, my precious copied boilerplate code!

PR #301, opened 2023-02-01
Andrew7234/emerald events

Andrew7234:
Todo:

- [x] index non-tx events using new oasis client sdk
- [ ] cut new oasis-sdk release and bump version
- [ ] ~~clean up emerald analyzer?~~ -- tabled to later PR
- [x] ~~emerald analyzer seems slow; measure how much slower this PR makes things~~ -- not noticably slower

### DIFF
--- a/analyzer/runtime/incoming.go
+++ b/analyzer/runtime/incoming.go
@@ -6,6 +6,7 @@ import (
 	"fmt"
 	"math/big"
 
+	ethCommon "github.com/ethereum/go-ethereum/common"
 	"github.com/oasisprotocol/oasis-core/go/common/cbor"
 	"github.com/oasisprotocol/oasis-core/go/common/crypto/address"
 	"github.com/oasisprotocol/oasis-core/go/roothash/api/block"
@@ -19,6 +20,7 @@ import (
 	"github.com/oasisprotocol/oasis-indexer/analyzer"
 	"github.com/oasisprotocol/oasis-indexer/analyzer/modules"
 	common "github.com/oasisprotocol/oasis-indexer/analyzer/uncategorized"
+	apiTypes "github.com/oasisprotocol/oasis-indexer/api/v1/types"
 	"github.com/oasisprotocol/oasis-indexer/log"
 	"github.com/oasisprotocol/oasis-indexer/storage"
 )
@@ -27,7 +29,7 @@ import (
 
 type BlockTransactionSignerData struct {
 	Index   int
-	Address string
+	Address apiTypes.Address
 	Nonce   int
 }
 
@@ -38,7 +40,25 @@ type BlockTransactionData struct {
 	Raw                     []byte
 	RawResult               []byte
 	SignerData              []*BlockTransactionSignerData
-	RelatedAccountAddresses map[string]bool
+	RelatedAccountAddresses map[apiTypes.Address]bool
+}
+
+type EventBody interface{}
+
+type EventData struct {
+	TxIndex          int
+	TxHash           string
+	Type             apiTypes.RuntimeEventType
+	Body             EventBody
+	EvmLogName       string
+	EvmLogParams     []*apiTypes.EvmEventParam
+	RelatedAddresses map[apiTypes.Address]bool
+}
+
+type extractEventResult struct {
+	ExtractedEvents   []*EventData
+	FoundGasUsedEvent bool
+	TxGasUsed         uint64
 }
 
 type AddressPreimageData struct {
@@ -50,10 +70,10 @@ type AddressPreimageData struct {
 type TokenChangeKey struct {
 	// TokenAddress is the Oasis address of the smart contract of the
 	// compatible (e.g. ERC-20) token.
-	TokenAddress string
+	TokenAddress apiTypes.Address
 	// AccountAddress is the Oasis address of the owner of some amount of the
 	// compatible (e.g. ERC-20) token.
-	AccountAddress string
+	AccountAddress apiTypes.Address
 }
 
 type BlockData struct {
@@ -62,10 +82,11 @@ type BlockData struct {
 	GasUsed             uint64
 	Size                int
 	TransactionData     []*BlockTransactionData
-	AddressPreimages    map[string]*AddressPreimageData
+	EventData           []*EventData
+	AddressPreimages    map[apiTypes.Address]*AddressPreimageData
 	TokenBalanceChanges map[TokenChangeKey]*big.Int
 	// key is oasis bech32 address
-	PossibleTokens map[string]*modules.EVMPossibleToken
+	PossibleTokens map[apiTypes.Address]*modules.EVMPossibleToken
 }
 
 // Function naming conventions in this file:
@@ -116,7 +137,7 @@ func extractAddressPreimage(as *sdkTypes.AddressSpec) (*AddressPreimageData, err
 	}, nil
 }
 
-func registerAddressSpec(addressPreimages map[string]*AddressPreimageData, as *sdkTypes.AddressSpec) (string, error) {
+func registerAddressSpec(addressPreimages map[apiTypes.Address]*AddressPreimageData, as *sdkTypes.AddressSpec) (apiTypes.Address, error) {
 	addr, err := common.StringifyAddressSpec(as)
 	if err != nil {
 		return "", err
@@ -133,7 +154,7 @@ func registerAddressSpec(addressPreimages map[string]*AddressPreimageData, as *s
 	return addr, nil
 }
 
-func registerEthAddress(addressPreimages map[string]*AddressPreimageData, ethAddr []byte) (string, error) {
+func registerEthAddress(addressPreimages map[apiTypes.Address]*AddressPreimageData, ethAddr []byte) (apiTypes.Address, error) {
 	addr, err := common.StringifyEthAddress(ethAddr)
 	if err != nil {
 		return "", err
@@ -150,8 +171,7 @@ func registerEthAddress(addressPreimages map[string]*AddressPreimageData, ethAdd
 	return addr, nil
 }
 
-//nolint:unparam
-func registerRelatedSdkAddress(relatedAddresses map[string]bool, sdkAddr *sdkTypes.Address) (string, error) {
+func registerRelatedSdkAddress(relatedAddresses map[apiTypes.Address]bool, sdkAddr *sdkTypes.Address) (apiTypes.Address, error) {
 	addr, err := common.StringifySdkAddress(sdkAddr)
 	if err != nil {
 		return "", err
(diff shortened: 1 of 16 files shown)


analyzer/emerald/incoming.go
```diff
@@ -349,38 +413,76 @@ func extractRound(b *block.Block, txrs []*sdkClient.TransactionWithResults, logg
...
 							}
+							eventRelatedAddresses = append(eventRelatedAddresses, toAddr)
 							registerTokenIncrease(blockData.TokenBalanceChanges, eventAddr, toAddr, amount)
 						}
 						blockData.PossibleTokens[eventAddr] = &modules.EVMPossibleToken{
 							PossibleTypes: []modules.EVMTokenType{modules.EVMTokenTypeERC20},
 							EthAddr:       event.Address,
 						}
+						evmLogName := apiTypes.Erc20Transfer
+						evmLogParams := map[string]string{
```
mitjat, line 432: I've realized ... the order of args matters. We could introduce a separate structure that lists the args (and we could do it later too; for erc20 Transfer, everybody knows what's up), but I think it's a nicer interface to emit a single structure: instead of `map[string]string`, we can have `[]EvmEventArg`, with `type EvmEventArg struct { name: string, value: any, evm_type: string }`. That's my best guess on what we'll be able to extract and what will be relevant.

Differences:
- order is preserved because of the list
- value type is `any`. This should be the case anyway, even in the map. EVM supports [various types (int256, string, address, ...)](https://docs.soliditylang.org/en/v0.8.17/types.html#value-types) for event values. 
- a new attribute, `type`. We could even omit it for now, until we start supporting more kinds of events. It's a good idea to include it eventually probably because there's not a 1:1 mapping between JSON types (= what we're sending out) and EVM types.

I guess down the road we'll want to expose all event metadata that the ABI gives us; see [our ERC-20 example](https://github.com/oasisprotocol/oasis-indexer/blob/3c756580ea3fe1d7d5949b4062c30693074d8ad8/analyzer/evmabi/contracts/artifacts/ERC20.json#L230-L251) and [its documentation](https://docs.soliditylang.org/en/v0.8.17/abi-spec.html#events). But it's too early to worry about every fine detail, as long as we have an initial structure that's flexible enough that we can keep adding and exposing more properties. `EvmEventArg` is my attempt at striking that balance.

TLDR: Let's go with `[]EvmEventArg` here, and revisit in a few months if needed.

mitjat:
Nice , that was fast! Your consensus experience shows :)
About TODOs:
 - clean up emerald analyzer? -- I'll do that in a follow-up. Wanted to do it already but it would be too disruptive with you and Warren both working on it. We'll need to do it with the two other refactors in mind: 1) supporting multiple runtimes and 2) supporting multiple node versions. 
- emerald analyzer seems slow; measure how much slower this MR makes things -- It's slow mostly because it queries the EVM for token metadata on every round. #284 will soon fix that. Your stuff just wrangles local structs, shouldn't have an impact worth mentioning.

analyzer/emerald/incoming.go
```diff
@@ -41,6 +43,18 @@ type BlockTransactionData struct {
...
+type EventBody interface{}
+
+type EventData struct {
+	TxIndex          int
+	TxHash           string
+	Type             apiTypes.RuntimeEventType
+	Body             EventBody
+	EvmLogName       string
+	EvmLogParams     map[string]string
+	RelatedAddresses []string // todo: could use []sdkTypes.Address instead, or we can keep []string and be extra careful
```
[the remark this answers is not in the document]
Andrew7234: Got it, updated!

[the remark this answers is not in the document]
pro-wh: what is sdk.Address?

also please use map[string]bool

pro-wh:
> Some of the vars in this file are stringily typed because the whole code was imported in one go from an experimental draft of Warren's, and we didn't polish the code all the way at import time because it would have taken too long.

but don't use `Address` from the sdk. use a custom string type if you desire.

1. our uses for addresses is very different from what the runtime SDK and oasis-core do. we mostly want to shuffle human readable addresses around between TEXT-based database columns and JSON strings in the API. everywhere you look are places that want the bech32 string. get addresses from the source storage and convert them preferably early as they enter our indexer.
2. we might have to delve into the Go language spec to figure out if this is true or not, but strings are easier to work with for the kinds of things we do. see the current draft's ExtractUniqueAddresses for example. throw addresses in a map[string]... any day. assign, compare equality, etc.
3. using a separate type helps us check that we're ingesting data consistently, because we can associate certain tasks with the conversion routines. for example, this new code doesn't properly register address preimages. see `registerEthAddress` and `registerRelatedEthAddress` in incoming.go.

pro-wh:
this PR is meant to come after the new SDK that allows us to get a list of events with the tx hashes attached right? ugh I have a lot of ideas how to improve the block scanning performance when that happens

some ideas for now:

- extract the event processing code from extractRound into a function that takes a `blockData *BlockData` and a `relatedAccountAddresses map[string]bool`
- call that new function in extractRound and the new prepareEventData

storage/api.go
```diff
@@ -272,14 +275,21 @@ type AccountsData struct {
...
 
 	Deposits  []*consensusaccounts.DepositEvent
 	Withdraws []*consensusaccounts.WithdrawEvent
 }
 
+// RawEvents contains all undecoded events for a given height.
+type RawEvents struct {
+	Round uint64
+
+	Logs []*types.Event
```
pro-wh, line 290: at this layer they're still called events

analyzer/emerald/incoming.go
Andrew7234: Got it; updated

analyzer/emerald/emerald.go
```diff
@@ -320,3 +336,174 @@ func (m *Main) queueBlockAndTransactionInserts(ctx context.Context, batch *stora
...
+	nonTxEvents := []*EventData{}
+	for _, e := range events.Logs {
+		// TODO: uncomment this when oasis-sdk is bumped
+		// if e.TxHash != nil {
+		// 	continue
+		// }
+		if err = common.VisitSdkEvent(e, &common.SdkEventHandler{
+			Core: func(event *core.Event) error {
+				eventData := EventData{
+					Type: apiTypes.CoreGasUsed,
```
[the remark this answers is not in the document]
mitjat: Well that was ... surprising :grimacing: 

Documentation and code spelunking eventually led me to https://github.com/deepmap/oapi-codegen/blob/f4cf8f9a570380c24c6ba03ae04b9393cf120692/pkg/codegen/configuration.go#L66, and to the solution I pushed as a tiny commit to this PR. It introduces compile errors because the enum values are now renamed, I'll leave that one to you :). I checked on `main` that no previous enums are affected.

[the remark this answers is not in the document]
Andrew7234: O wow that's buried deep; thanks for digging into that. Updated!

Andrew7234:
> using a separate type helps us check that we're ingesting data consistently

Added a custom string type in apiTypes (`apiTypes.Address`). Updated the methods in incoming.go + emerald.go to use this type instead of `string`

> call that new function in extractRound and the new prepareEventData

Thanks for the comment! updated. Both the non-tx event and tx-event addresses should properly `Register` now.

> relatedAccountAddresses map[string]bool

Would it be ok to use `map[string]struct{}`?

storage/client/queries.go
```diff
@@ -404,6 +404,22 @@ func (qf QueryFactory) RuntimeTransactionsQuery() string {
...
+	return fmt.Sprintf(`
+		SELECT round, tx_index, tx_hash, type, body, evm_log_name, evm_log_params
+			FROM %[1]s.runtime_events
+			WHERE (runtime = '%[2]s') AND
+					($1::bigint IS NULL OR round = $1::bigint) AND
+					($2::integer IS NULL OR tx_index = $2::integer) AND
+					($3::text IS NULL OR tx_hash = $3::text) AND
+					($4::text IS NULL OR type = $4::text) AND
+					($5::text IS NULL OR evm_log_signature = $5::text) AND
+					($6::text IS NULL OR related_accounts @> ARRAY[$6::text])
```
mitjat, line 417: currently produces `operator does not exist: oasis_addr[] @> text[]`

Maybe it works with `$6::oasis_addr`; if not, let's just go with `text` here for the array also.

mitjat:
Some high level comments:
- I think this is largely good. I currently cannot test, but ...
- Let's add non-tx events in a separate PR; this one is already unwieldy.
- On a related note, `incoming.go` (heh) could use a refactor/cleanup, but let's not do anything more towards that in this PR.
- I rebased & pushed. I _think_ my rebase was ok, but it was a little hard to tell because there was a huge conflict around extractEvents() which Andy factored out, but also Warren and Andy both edited. Please be careful when reviewing those changes. FWIW, I did some local diffing of manually cut-n-pasted code blocks, plus regressions (for token metadata) pass.
- @pro-wh you're the main reviewer on this one.

analyzer/runtime/incoming.go
```diff
@@ -116,11 +143,12 @@ func extractAddressPreimage(as *sdkTypes.AddressSpec) (*AddressPreimageData, err
...
 }
 
-func registerAddressSpec(addressPreimages map[string]*AddressPreimageData, as *sdkTypes.AddressSpec) (string, error) {
-	addr, err := common.StringifyAddressSpec(as)
+func registerAddressSpec(addressPreimages map[apiTypes.Address]*AddressPreimageData, as *sdkTypes.AddressSpec) (apiTypes.Address, error) {
+	addrStr, err := common.StringifyAddressSpec(as)
 	if err != nil {
 		return "", err
 	}
+	addr := apiTypes.Address(addrStr)
```
pro-wh, line 151: is there much of this casting? we should modify all of uncategorized/addresses.go to use apiTypes.Address. this PR or elsewhere

pro-wh (commented):
I'm fine with map[sdkTypes.Address]struct{} :confused:

pro-wh (approved):
this looks good as it is. I see we're waiting for the sdk update.

I think in the future when we do the modules cleanup, there will be a lot we can do to simplify

analyzer/uncategorized/addresses.go
```diff
@@ -40,3 +41,19 @@ func StringifyEthAddress(ethAddr []byte) (string, error) {
 	ocAddr := address.NewAddress(ctx, ethAddr)
 	return StringifyOcAddress(ocAddr)
 }
+
+func ConvertEthAddressToSdkAddress(ethAddr []byte) sdkTypes.Address {
```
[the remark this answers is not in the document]
Andrew7234: Yup, removed

analyzer/runtime/incoming.go
Andrew7234: There was a bit more than I'd like yeah. updated the address fns in uncategorized/addresses.go

pro-wh: thanks!

storage/client/queries.go
Andrew7234: No luck with `ARRAY[$6::oasis_addr]`; the `ARRAY` converts the type to `text` which results in the error `operator does not exist: oasis_addr[] @> text[]`

Changed to `string`

Andrew7234:
> Let's add non-tx events in a separate PR; this one is already unwieldy.

Done. After refactoring out the event-parsing code, the non-tx event ingestion is quite short. However, I'd prefer not to block this PR on the oasis-sdk version bump (+ subsequent testing).

PR #304, opened 2023-02-02
/runtime/transactions: Populate the timestamp

mitjat:
After this PR:
- analyzer saves block timestamp into the tx DB table (this is denormailzed, but I opted for this rather than a JOIN every time we're serving txs)
- API reads it from the DB, presents it

Resolves https://app.clickup.com/t/861maqp0p

### DIFF
--- a/analyzer/queries.go
+++ b/analyzer/queries.go
@@ -375,8 +375,8 @@ func (qf QueryFactory) RuntimeRelatedTransactionInsertQuery() string {
 
 func (qf QueryFactory) RuntimeTransactionInsertQuery() string {
 	return fmt.Sprintf(`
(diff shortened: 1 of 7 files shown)


Andrew7234 (approved):
Looks good; might want to update the comment at https://github.com/oasisprotocol/oasis-indexer/blob/main/api/spec/v1.yaml#L1884 too

mitjat:
> Looks good; might want to update the comment at [`main`/api/spec/v1.yaml#L1884](https://github.com/oasisprotocol/oasis-indexer/blob/main/api/spec/v1.yaml?rgh-link-date=2023-02-03T09%3A07%3A55Z#L1884) too

I'm actually still not confident about what the answer is there :/

PR #305, opened 2023-02-02
(no merge) edits for running sapphire

pro-wh:
opening this for us to track what needs changing for us to run on sapphire. with ongoing work, this set will eventually reduce to nothing

### DIFF
--- a/analyzer/runtime/evm_test.go
+++ b/analyzer/runtime/evm_test.go
@@ -27,9 +27,11 @@ var (
 	// TODO: Would be nice to have an offline test.
 	PublicSourceConfig = &config.SourceConfig{
 		ChainName: ChainName,
-		Nodes: map[string]*config.NodeConfig{
+		Nodes: map[string]*config.ArchiveConfig{
 			CurrentArchiveName: {
-				RPC: sdkConfig.DefaultNetworks.All[string(ChainName)].RPC,
+				DefaultNode: &config.NodeConfig{
+					RPC: sdkConfig.DefaultNetworks.All[string(ChainName)].RPC,
+				},
 			},
 		},
 		FastStartup: false,
--- a/config/config.go
+++ b/config/config.go
@@ -68,6 +68,11 @@ func (cfg *AnalysisConfig) Validate() error {
 	} else if cfg.Source.ChainName != "" && cfg.Source.CustomChain != nil {
 		return fmt.Errorf("source.chain_name and source.custom_chain specified, can only use one")
 	}
+	for archiveName, archiveConfig := range cfg.Source.Nodes {
+		if archiveConfig.DefaultNode == nil && archiveConfig.ConsensusNode == nil && len(archiveConfig.RuntimeNodes) == 0 {
+			return fmt.Errorf("source.nodes[%v] has none of .default, .consensus, or .runtimes", archiveName)
+		}
+	}
 	if cfg.Analyzers.Consensus != nil {
 		if err := cfg.Analyzers.Consensus.Validate(); err != nil {
 			return err
@@ -131,7 +136,7 @@ type SourceConfig struct {
 	// Nodes describe the oasis-node(s) to connect to. Keys are "archive
 	// names," which are named after mainnet releases, in lowercase e.g.
 	// "cobalt" and "damask."
-	Nodes map[string]*NodeConfig `koanf:"nodes"`
+	Nodes map[string]*ArchiveConfig `koanf:"nodes"`
 
 	// If set, the analyzer will skip some initial checks, e.g. that
 	// `rpc` really serves the chain with the chain context we expect.
@@ -168,10 +173,12 @@ func CustomSingleNetworkSourceConfig(rpc string, chainContext string) *SourceCon
 	}
 }
 
-func SingleNetworkLookup(rpc string) map[string]*NodeConfig {
-	return map[string]*NodeConfig{
+func SingleNetworkLookup(rpc string) map[string]*ArchiveConfig {
+	return map[string]*ArchiveConfig{
 		"damask": {
-			RPC: rpc,
+			DefaultNode: &NodeConfig{
+				RPC: rpc,
+			},
 		},
 	}
 }
@@ -198,6 +205,34 @@ type CustomChainConfig struct {
 	// SDKNetwork is the oasis-sdk Network configuration of the latest
 	// network in the chain.
 	SDKNetwork *sdkConfig.Network `koanf:"sdk_network"`
+}
+
+// ArchiveConfig is information about the nodes for a network.
+type ArchiveConfig struct {
+	// DefaultNode is information about the node to get data from by default.
+	DefaultNode *NodeConfig `koanf:"default"`
+	// ConsensusNode is information about the node to get consensus data from,
+	// instead of the default node.
+	ConsensusNode *NodeConfig `koanf:"consensus"`
+	// RuntimeNodes is the information about the nodes to get runtime data
+	// from, instead of the default node.
+	RuntimeNodes map[common.Runtime]*NodeConfig `koanf:"runtimes"`
+}
+
+func (ac *ArchiveConfig) ResolvedConsensusNode() *NodeConfig {
+	nc := ac.ConsensusNode
+	if nc != nil {
+		return nc
+	}
+	return ac.DefaultNode
+}
+
+func (ac *ArchiveConfig) ResolvedRuntimeNode(runtime common.Runtime) *NodeConfig {
+	nc := ac.RuntimeNodes[runtime]
+	if nc != nil {
+		return nc
+	}
+	return ac.DefaultNode
 }
 
 // NodeConfig is information about one oasis-node to connect to.
--- a/config/docker-dev.yml
+++ b/config/docker-dev.yml
@@ -3,7 +3,8 @@ analysis:
     chain_name: mainnet
     nodes:
       damask:
-        rpc: unix:/tmp/node.sock #unix:/node/data/internal.sock
+        default:
+          rpc: unix:/tmp/node.sock #unix:/node/data/internal.sock
   analyzers:
     metadata_registry:
       interval: 1h
--- a/config/local-dev.yml
+++ b/config/local-dev.yml
@@ -3,7 +3,8 @@ analysis:
     chain_name: mainnet
     nodes:
       damask:
-        rpc: unix:/tmp/node.sock #unix:/node/data/internal.sock
+        default:
+          rpc: unix:/tmp/node.sock #unix:/node/data/internal.sock
   analyzers:
     metadata_registry:
       interval: 1h
--- a/storage/oasis/nodeapi/history/history.go
+++ b/storage/oasis/nodeapi/history/history.go
@@ -18,18 +18,18 @@ import (
 
 var _ nodeapi.ConsensusApiLite = (*HistoryConsensusApiLite)(nil)
 
-type APIConstructor func(ctx context.Context, chainContext string, nodeConfig *config.NodeConfig, fastStartup bool) (nodeapi.ConsensusApiLite, error)
+type APIConstructor func(ctx context.Context, chainContext string, archiveConfig *config.ArchiveConfig, fastStartup bool) (nodeapi.ConsensusApiLite, error)
 
 var APIConstructors = map[string]APIConstructor{
-	"damask": func(ctx context.Context, chainContext string, nodeConfig *config.NodeConfig, fastStartup bool) (nodeapi.ConsensusApiLite, error) {
-		sdkConn, err := connections.SDKConnect(ctx, chainContext, nodeConfig, fastStartup)
+	"damask": func(ctx context.Context, chainContext string, archiveConfig *config.ArchiveConfig, fastStartup bool) (nodeapi.ConsensusApiLite, error) {
+		sdkConn, err := connections.SDKConnect(ctx, chainContext, archiveConfig.ResolvedConsensusNode(), fastStartup)
 		if err != nil {
 			return nil, err
 		}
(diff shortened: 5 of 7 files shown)


analyzer/runtime/runtime.go
```diff
@@ -24,14 +24,14 @@ import (
 )
 
 const (
-	emerald = analyzer.RuntimeEmerald
+	sapphire = analyzer.RuntimeSapphire
 
-	EmeraldDamaskAnalyzerName = "emerald_damask"
+	SapphireDamaskAnalyzerName = "sapphire_damask"
```
pro-wh, line 29: this we ought to generalize, it's now just a runtime analyzer

analyzer/runtime/runtime.go
```diff
@@ -24,14 +24,14 @@ import (
...
 
 	ProcessRoundTimeout = 61 * time.Second
 )
 
-// Main is the main Analyzer for the Emerald Runtime.
+// Main is the main Analyzer for the Sapphire Runtime.
```
pro-wh, line 34: generalize comment

analyzer/runtime/runtime.go
```diff
@@ -44,7 +44,7 @@ type Main struct {
 
 var _ analyzer.Analyzer = (*Main)(nil)
 
-// NewRuntimeAnalyzer returns a new main analyzer for the Emerald Runtime.
+// NewRuntimeAnalyzer returns a new main analyzer for the Sapphire Runtime.
```
pro-wh, line 47: generalize comment

analyzer/runtime/runtime.go
```diff
@@ -66,11 +66,11 @@ func NewRuntimeAnalyzer(nodeCfg config.NodeConfig, cfg *config.BlockBasedAnalyze
 		return nil, err
 	}
 
-	id, err := emerald.ID(network)
+	id, err := sapphire.ID(network)
```
pro-wh, line 69: need to move this into a parameter or something

analyzer/runtime/runtime.go
```diff
@@ -66,11 +66,11 @@ func NewRuntimeAnalyzer(nodeCfg config.NodeConfig, cfg *config.BlockBasedAnalyze
...
 	if err != nil {
 		return nil, err
 	}
-	logger.Info("Emerald runtime ID determined", "runtime_id", id)
+	logger.Info("Sapphire runtime ID determined", "runtime_id", id)
```
pro-wh, line 73: move from log message into a key-value, take value from parameter

analyzer/runtime/runtime.go
```diff
@@ -88,14 +88,14 @@ func NewRuntimeAnalyzer(nodeCfg config.NodeConfig, cfg *config.BlockBasedAnalyze
 		Source: client,
 	}
 
-	qf := analyzer.NewQueryFactory(strcase.ToSnake(nodeCfg.ChainID), emerald.String())
+	qf := analyzer.NewQueryFactory(strcase.ToSnake(nodeCfg.ChainID), sapphire.String())
```
pro-wh, line 91: from parameter

analyzer/runtime/runtime.go
```diff
@@ -88,14 +88,14 @@ func NewRuntimeAnalyzer(nodeCfg config.NodeConfig, cfg *config.BlockBasedAnalyze
...
 
 	return &Main{
 		cfg:     ac,
 		qf:      qf,
 		target:  target,
-		logger:  logger.With("analyzer", EmeraldDamaskAnalyzerName),
-		metrics: metrics.NewDefaultDatabaseMetrics(EmeraldDamaskAnalyzerName),
+		logger:  logger.With("analyzer", SapphireDamaskAnalyzerName),
+		metrics: metrics.NewDefaultDatabaseMetrics(SapphireDamaskAnalyzerName),
```
pro-wh, line 98: probably split into two key-values. analyzer -> runtime, runtime -> that parameter

analyzer/runtime/runtime.go
```diff
@@ -177,7 +177,7 @@ func (m *Main) Start() {
 
 // Name returns the name of the Main.
 func (m *Main) Name() string {
-	return EmeraldDamaskAnalyzerName
+	return SapphireDamaskAnalyzerName
```
pro-wh, line 180: can maybe generate a runtime-bearing name programmatically

analyzer/runtime/runtime.go
```diff
@@ -188,7 +188,7 @@ func (m *Main) latestRound(ctx context.Context) (uint64, error) {
 		m.qf.LatestBlockQuery(),
 		// ^analyzers should only analyze for a single chain ID, and we anchor this
 		// at the starting round.
-		EmeraldDamaskAnalyzerName,
+		SapphireDamaskAnalyzerName,
```
pro-wh, line 191: bit of inconsistency with some recent hurried runtime db table generalization. a lot of our queries right now do string interpolation within the query factory with `qf.runtime`, whereas these leave the runtime parametric. let's choose one and make them uniform.

1. qf knows runtime, analyzer doesn't need to provide it. update these to remove the runtime sql placeholder and convert to a format string thing in query factory.
2. analyzer knows runtime, qf doesn't care. update runtime queries to use a sql placeholder and have analyzer pass it. qf no longer needs `runtime` field.

I'm voting 2

analyzer/runtime/runtime.go
```diff
@@ -257,7 +257,7 @@ func (m *Main) processRound(ctx context.Context, round uint64) error {
 		return err
 	}
 
-	opName := fmt.Sprintf("process_block_%s", emerald.String())
+	opName := fmt.Sprintf("process_block_%s", sapphire.String())
```
pro-wh, line 260: ok we'll need that parameter as a field in Main to use it here

analyzer/uncategorized/runtimes.go
```diff
@@ -15,7 +15,7 @@ func OpenUtxNoVerify(utx *sdkTypes.UnverifiedTransaction) (*sdkTypes.Transaction
 	if len(utx.AuthProofs) == 1 && utx.AuthProofs[0].Module != "" {
 		switch utx.AuthProofs[0].Module {
 		case "evm.ethereum.v0":
-			tx, err := decodeEthRawTx(utx.Body, 42262)
+			tx, err := decodeEthRawTx(utx.Body, 23294)
```
pro-wh, line 18: need to add a few layers of parameters. might make sense to wire this in all the way from the config

go.mod
```diff
@@ -14,8 +14,8 @@ require (
 	github.com/iancoleman/strcase v0.2.0
 	github.com/jackc/pgx/v4 v4.16.0
 	github.com/knadh/koanf v1.4.1
-	github.com/oasisprotocol/oasis-core/go v0.2202.3
-	github.com/oasisprotocol/oasis-sdk/client-sdk/go v0.2.1-0.20220823181727-bd5025a9857b
+	github.com/oasisprotocol/oasis-core/go v0.2202.5
+	github.com/oasisprotocol/oasis-sdk/client-sdk/go v0.2.1-0.20230131160755-5e0353622f04
```
pro-wh, line 18: go get github.com/oasisprotocol/oasis-sdk/client-sdk/go@main

go.mod
```diff
@@ -14,8 +14,8 @@ require (
 	github.com/iancoleman/strcase v0.2.0
 	github.com/jackc/pgx/v4 v4.16.0
 	github.com/knadh/koanf v1.4.1
-	github.com/oasisprotocol/oasis-core/go v0.2202.3
-	github.com/oasisprotocol/oasis-sdk/client-sdk/go v0.2.1-0.20220823181727-bd5025a9857b
+	github.com/oasisprotocol/oasis-core/go v0.2202.5
```
pro-wh, line 17: consequence of upgrading sdk

analyzer/uncategorized/runtimes.go
mitjat: This is completely separate from [RuntimeInfo.ID](https://github.com/oasisprotocol/oasis-sdk/blob/5e0353622f049fb53c52b3e7c8e54aa2d998bbdc/client-sdk/go/config/default.go#L90-L90)? I'm surprised we don't have this chainID anywhere in the Go codebase, only Rust.

Also, nit, prefer the hex version here because that's how it's used everywhere else in oasisprotocol repos.

The term chainID is already used in the indexer and has a fixed value of `oasis_3`, btw; perhaps we can get rid of it? I'm not sure when it would change.

analyzer/runtime/runtime.go
mitjat: This query is a tiny bit different from others because 
- it includes the `consensus_damask` analyzer: https://github.com/oasisprotocol/oasis-indexer/blob/699e4cccbde673c2e61bbc25822401512e756db4/analyzer/consensus/consensus.go#L202-L202
- the column technically refers to the analyzer, not the runtime. I can see a scenario (though I think it's unlikely) where we want to do some additional processing on already-processed blocks, and we'll whip up a temporary analyzer that will also track its progress in this table.

pro-wh: hmh

analyzer/uncategorized/runtimes.go
pro-wh: :+1: on hex

yes on this being separate from runtime ID. runtime ID is an oasis concept. chain ID in this usage is a concept from ethereum-compatible systems

analyzer/runtime/runtime.go
pro-wh: now we're doing analyzer -> that parameter

analyzer/runtime/runtime.go
pro-wh: ok let's leave `.runtime` in QueryFactory for now

analyzer/uncategorized/runtimes.go
pro-wh: getting rid of the check

analyzer/evmtokens/evm_tokens.go
```diff
@@ -62,11 +62,11 @@ func NewMain(nodeCfg *config.NodeConfig, target storage.TargetStorage, logger *l
 		return nil, err
 	}
 
-	id, err := analyzer.RuntimeEmerald.ID(network)
+	id, err := analyzer.RuntimeSapphire.ID(network)
```
pro-wh, line 65: parameter

analyzer/evmtokens/evm_tokens.go
```diff
@@ -62,11 +62,11 @@ func NewMain(nodeCfg *config.NodeConfig, target storage.TargetStorage, logger *l
...
 	if err != nil {
 		return nil, err
 	}
-	logger.Info("Emerald runtime ID determined", "runtime_id", id)
+	logger.Info("Sapphire runtime ID determined", "runtime_id", id)
```
pro-wh, line 69: generalize, show runtime name as kv pair

analyzer/evmtokens/evm_tokens.go
```diff
@@ -79,7 +79,7 @@ func NewMain(nodeCfg *config.NodeConfig, target storage.TargetStorage, logger *l
 	return &Main{
 		source: client,
 		target: target,
-		qf:     analyzer.NewQueryFactory("oasis_3", analyzer.RuntimeEmerald.String()),
+		qf:     analyzer.NewQueryFactory("oasis_3", analyzer.RuntimeSapphire.String()),
```
pro-wh, line 82: parameter

analyzer/evmtokens/evm_tokens.go
```diff
@@ -200,7 +200,7 @@ func (m Main) Start() {
 	backoff, err := util.NewBackoff(
 		100*time.Millisecond,
 		6*time.Second,
-		// ^cap the timeout at the expected emerald round time
+		// ^cap the timeout at the expected sapphire round time
```
pro-wh, line 203: this whole comment block is reworked. re-copy-paste from runtime analyzer

analyzer/modules/evm.go
```diff
@@ -85,7 +88,13 @@ func evmCallWithABICustom(
 	if err != nil {
 		return fmt.Errorf("packing evm simulate call data: %w", err)
 	}
-	outPacked, err := source.EVMSimulateCall(ctx, round, gasPrice, gasLimit, caller, contractEthAddr, value, inPacked)
+	inCBOR := cbor.Marshal(inPacked)
+	inCall := sdkTypes.Call{
+		Format: sdkTypes.CallFormatPlain,
+		Body:   inCBOR,
+	}
+	inEncoded := cbor.Marshal(inCall)
```
pro-wh, line 96: allegedly we won't need this when sapphire 0.4.0 comes out

go.mod
```diff
@@ -45,12 +47,15 @@ require (
...
 	github.com/go-logfmt/logfmt v0.5.1 // indirect
+	github.com/go-ole/go-ole v1.2.1 // indirect
+	github.com/go-stack/stack v1.8.0 // indirect
 	github.com/gogo/protobuf v1.3.2 // indirect
 	github.com/golang/glog v0.0.0-20160126235308-23def4e6c14b // indirect
 	github.com/golang/protobuf v1.5.2 // indirect
 	github.com/golang/snappy v0.0.4 // indirect
 	github.com/google/go-github/v39 v39.2.0 // indirect
 	github.com/google/go-querystring v1.1.0 // indirect
+	github.com/gorilla/websocket v1.5.0 // indirect
```
pro-wh, line 58: what are these changes?

analyzer/modules/evm.go
pro-wh: removed

go.mod
pro-wh: oh it's because we already have the go-ethereum monorepo added, but the ethclient etc pull in new transitive dependencies

pro-wh: moved to #318

PR #306, opened 2023-02-03
runtime: remove some Emerald references

pro-wh:
constructing a runtime analyzer now takes an analyzer.Runtime parameter

cross reference #305

last (?) step of https://app.clickup.com/t/3u9y285

### DIFF
--- a/analyzer/runtime/runtime.go
+++ b/analyzer/runtime/runtime.go
@@ -24,8 +24,6 @@ import (
 )
 
 const (
-	MajorCompat = "damask"
-
 	ProcessRoundTimeout = 61 * time.Second
 )
 
@@ -100,8 +98,8 @@ func NewRuntimeAnalyzer(
 		cfg:     ac,
 		qf:      qf,
 		target:  target,
-		logger:  logger.With("analyzer", runtime.String()+"_"+MajorCompat),
-		metrics: metrics.NewDefaultDatabaseMetrics(runtime.String() + "_" + MajorCompat),
+		logger:  logger.With("analyzer", runtime.String()),
+		metrics: metrics.NewDefaultDatabaseMetrics(runtime.String()),
 
 		// module handlers
 		moduleHandlers: []modules.ModuleHandler{
@@ -183,8 +181,7 @@ func (m *Main) Start() {
 
 // Name returns the name of the Main.
 func (m *Main) Name() string {
-	panic("we never actually call Name")
-	return m.runtime.String() + "_" + MajorCompat
+	return m.runtime.String()
 }
 
 // latestRound returns the latest round processed by the consensus analyzer.
@@ -258,7 +255,7 @@ func (m *Main) processRound(ctx context.Context, round uint64) error {
 		batch.Queue(
 			m.qf.IndexingProgressQuery(),
 			round,
-			m.runtime.String()+"_"+MajorCompat,
+			m.runtime.String(),
 		)
 		return nil
 	})
--- a/analyzer/runtime/runtime.go
+++ b/analyzer/runtime/runtime.go
@@ -24,15 +24,14 @@ import (
 )
 
 const (
-	emerald = analyzer.RuntimeEmerald
-
-	EmeraldDamaskAnalyzerName = "emerald_damask"
+	MajorCompat = "damask"
 
 	ProcessRoundTimeout = 61 * time.Second
 )
 
-// Main is the main Analyzer for the Emerald Runtime.
+// Main is the main Analyzer for runtimes.
 type Main struct {
+	runtime analyzer.Runtime
 	cfg     analyzer.RuntimeConfig
 	qf      analyzer.QueryFactory
 	target  storage.TargetStorage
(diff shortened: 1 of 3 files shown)


analyzer/runtime/runtime.go
```diff
@@ -88,14 +93,15 @@ func NewRuntimeAnalyzer(nodeCfg config.NodeConfig, cfg *config.BlockBasedAnalyze
...
 
 	return &Main{
+		runtime: runtime,
 		cfg:     ac,
 		qf:      qf,
 		target:  target,
-		logger:  logger.With("analyzer", EmeraldDamaskAnalyzerName),
-		metrics: metrics.NewDefaultDatabaseMetrics(EmeraldDamaskAnalyzerName),
+		logger:  logger.With("analyzer", runtime.String()+"_"+MajorCompat),
+		metrics: metrics.NewDefaultDatabaseMetrics(runtime.String() + "_" + MajorCompat),
```
mitjat, line 104: Let's get rid of the `MajorCompat` const altogether, here and elsewhere.
The analyzer is and will be independent of the underlying storage.
We don't cosume these prometheus metrics yet so it's easy to change the name now, but we'll soon base some alerts on them.

mitjat (approved):
Hah I have an almost identical commit on my machine :)
Thank you!

analyzer/runtime/runtime.go
```diff
@@ -177,7 +183,8 @@ func (m *Main) Start() {
 
 // Name returns the name of the Main.
 func (m *Main) Name() string {
-	return EmeraldDamaskAnalyzerName
+	panic("we never actually call Name")
```
mitjat, line 186: :)
Let's just get rid of it at the interface level then

mitjat: Edit: We call it [here](https://github.com/oasisprotocol/oasis-indexer/blob/699e4cccbde673c2e61bbc25822401512e756db4/cmd/analyzer/analyzer.go#L156-L156)
But it's true that from I can tell, we never truly use that return value. We do use it as a map key, but then only ever iterate over the values of the map and ignore the key. So if we change the map to a list, we're good to go.

analyzer/runtime/runtime.go
```diff
@@ -66,11 +71,11 @@ func NewRuntimeAnalyzer(nodeCfg config.NodeConfig, cfg *config.BlockBasedAnalyze
 		return nil, err
 	}
 
-	id, err := emerald.ID(network)
+	id, err := runtime.ID(network)
```
mitjat, line 74: I tested running multiple runtime analyzers (emerald and sapphire) in parallel using this PR. It was a slightly fake setup: I changed this line so that the "sapphire" analyzer was really pulling emerald data. There were no changes in regression tests. :tada: So it looks like we're fully ready for the sapphire impl.

I'll post the tests as top-level comments for visibility.

mitjat:
As per the comment https://github.com/oasisprotocol/oasis-indexer/pull/306/files#r1095643489, here are the two tests that I ran to verify that this PR basically resolves https://app.clickup.com/t/3u9y285 and makes us ready to use multiple analyzers.

In both tests, I ran `reindex_and_run.sh` regression test with the config posted below, once with sapphire commented out and once present.

**Test 1 - config**:
Both analyzers scan the same blocks
```yaml
  analyzers:
    emerald:
      from: 1_003_298  # round at Damask genesis
      to: 1_003_308  # 10 blocks
    sapphire:
      from: 1_003_298  # round at Damask genesis
      to: 1_003_308  # 10 blocks
```
No differences in emerald queries on the resulting DB. This strongly suggests no erroneous double dead-reckoning, or similar overlap between analyzers' output.

**Test 2 - config**:
Analyzers overlap in half of the blocks
```yaml
  analyzers:
    emerald:
      from: 1_003_298  # round at Damask genesis
      to: 1_003_318  # 20 blocks
    sapphire:
      from: 1_003_308  # emerald + 10
      to: 1_003_328  # emerald + 10
```
No difference in any queries, except `/consensus/accounts` which presents a union of all true consensus accounts _and_ accounts encountered in runtimes ([via the address_preimages table](https://github.com/oasisprotocol/oasis-indexer/blob/6d3c08e4eedb387cde9a353c228e31fd7f666775/storage/client/queries.go#L197-L197)). This makes sense, as the "sapphire" analyzer scanned 10 extra blocks, which involved not-seen-before eth addresses.

analyzer/runtime/runtime.go
pro-wh: :raised_hands: k

analyzer/runtime/runtime.go
pro-wh: :O

pro-wh: keeping then. gonna remove the panic

analyzer/runtime/runtime.go
pro-wh: cool, and I tested with the public node on sapphire mainnet

pro-wh:
thanks for the testing!

analyzer/runtime/runtime.go
pro-wh: removed. y'all metrics-lookers-at are hereby warned

analyzer/runtime/runtime.go
pro-wh: removed

PR #308, opened 2023-02-03
codegen helper: Use POSIX sed (OS X compatibility)

mitjat:


mitjat:
Works on My Machine (TM) (linux).  @Andrew7234 please test and approve if it works for you.

scripts/namespace_codegen_templates.sh
```diff
@@ -17,6 +17,7 @@ cp -r "$codegenDir/templates" /tmp/namespaced-templates
 find /tmp/namespaced-templates/ -type d -print0 | xargs -0 chmod 755
 
 # Rename the "runtime" package to "codegen_runtime" in all templates.
-find /tmp/namespaced-templates -type f -print0 | xargs -0 sed -i -E 's/\bruntime\b/codegen_runtime/g'
+# It happens to be always used as " runtime.", so find that. "\b" doesn't work in POSIX sed (which OS X uses).
+find /tmp/namespaced-templates -type f -print0 | xargs -0 sed -i -E --posix 's/ runtime./ codegen_runtime./g'
```
Andrew7234: blegh turns out mac `sed` doesn't support the `--posix` flag

```
sed: illegal option -- -
usage: sed script [-Ealnru] [-i extension] [file ...]
	sed [-Ealnu] [-i extension] [-e script] ... [-f script_file] ... [file ...]
```

Andrew7234 (approved):
Thanks for putting this together! I added a check to use mac-compatible `sed` commands; and confirmed that it works on my machine. Please let me know if it doesn't work on your machine.

Alternatively I'm open to replacing my local `sed` with  `gnu-sed` but it would be a somewhat cumbersome to require all osx devs to do so.

scripts/namespace_codegen_templates.sh
mitjat: Blerg. It would be nice to not have two versions of the command.
What if you just rip out `--posix`, does that work for you?  I only added it to guard me (on linux) from using non-posix extensions. But it works for me without `--posix` too.

> it would be a somewhat cumbersome to require all osx devs to do so (install gnu-sed).
+1

Andrew7234: Ah that worked! Removed and updated

PR #309, opened 2023-02-03
Andrew7234/emerald non tx events

Andrew7234:
Updates the analyzer to index non-tx emerald events.

### DIFF
--- a/analyzer/runtime/runtime.go
+++ b/analyzer/runtime/runtime.go
@@ -15,7 +15,9 @@ import (
 
 	"github.com/oasisprotocol/oasis-indexer/analyzer"
 	"github.com/oasisprotocol/oasis-indexer/analyzer/modules"
+	common "github.com/oasisprotocol/oasis-indexer/analyzer/uncategorized"
 	"github.com/oasisprotocol/oasis-indexer/analyzer/util"
+	apiTypes "github.com/oasisprotocol/oasis-indexer/api/v1/types"
 	"github.com/oasisprotocol/oasis-indexer/config"
 	"github.com/oasisprotocol/oasis-indexer/log"
 	"github.com/oasisprotocol/oasis-indexer/metrics"
@@ -234,6 +236,7 @@ func (m *Main) processRound(ctx context.Context, round uint64) error {
 	type prepareFunc = func(context.Context, uint64, *storage.QueryBatch) error
 	for _, f := range []prepareFunc{
 		m.prepareBlockData,
+		m.prepareEventData,
 	} {
 		func(f prepareFunc) {
 			group.Go(func() error {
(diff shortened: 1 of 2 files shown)


go.mod
```diff
@@ -14,8 +14,8 @@ require (
 	github.com/iancoleman/strcase v0.2.0
 	github.com/jackc/pgx/v4 v4.16.0
 	github.com/knadh/koanf v1.4.1
-	github.com/oasisprotocol/oasis-core/go v0.2202.3
-	github.com/oasisprotocol/oasis-sdk/client-sdk/go v0.2.1-0.20220823181727-bd5025a9857b
+	github.com/oasisprotocol/oasis-core/go v0.2202.5
+	github.com/oasisprotocol/oasis-sdk/client-sdk/go v0.2.1-0.20230131160755-5e0353622f04
```
mitjat, line 18: this probably doesn't belong in this PR?

mitjat (approved):
Thank you! LGTM assuming you tested this e2e manually.

go.mod
Andrew7234: yeah i should've based it off of warren's oasis-sdk bump pr. Will update

> tested this e2e manually.

I ran the indexer for ~1hr and haven't picked up any non-tx events... Not sure if they're rare or if there's an issue, will keep it running. I did confirm the raw events are being fetched + checked correctly.

pro-wh: https://github.com/oasisprotocol/sapphire-paratime/blob/v0.3.1/runtime/src/lib.rs#LL185C22-L185C28 there should be transfers made by the reward module

Andrew7234: Hmm we pick up the `evm.Call` transaction associated with the transfer, but not any events. Do rewards transfers emit events like normal transfers do?

Perhaps of note: the rewards module itself [has no events](https://pkg.go.dev/github.com/oasisprotocol/oasis-sdk/client-sdk/go/modules/rewards)

pro-wh: https://github.com/oasisprotocol/oasis-sdk/blob/runtime-sdk/v0.3.0/runtime-sdk/src/modules/rewards/mod.rs#L197

- that shouldn't make any transaction
- that shouldn't do anything involving the evm module
- correct that the rewards module doesn't have its own event types
- it makes a transfer using the accounts module, and the accounts module should emit an event

Andrew7234: Update: The issue was that I was using [hash.IsEmpty()](https://github.com/oasisprotocol/oasis-core/blob/master/go/common/crypto/hash/hash.go#L118) instead of directly checking for the zero-valued hash (`0000000000000000000000000000000000000000000000000000000000000000`).

After the fix, the indexer now indexes non-tx events properly.

aefhm: > Update: The issue was that I was using [hash.IsEmpty()](https://github.com/oasisprotocol/oasis-core/blob/master/go/common/crypto/hash/hash.go#L118) instead of directly checking for the zero-valued hash (0000000000000000000000000000000000000000000000000000000000000000).
After the fix, the indexer now indexes non-tx events properly.

And that is not a bug?

mitjat: nope. consensus uses the hash of the empty string, runtimes use the all-zero hash :shrug: :nauseated_face: 
offline discussion: https://oasis-labs.slack.com/archives/C03T4953DFX/p1676279544349129

Andrew7234: Yep; auto-merging now

PR #310, opened 2023-02-06
api: Add runtime transaction sender and receiver Eth address

aefhm:
#### Description
Address https://app.clickup.com/t/3u9y3yp and add Eth address to API endpoint

#### TODO
- [x] perf check
- [x] solve perf problem
- [x] investigate missing preimage

### DIFF
--- a/analyzer/uncategorized/addresses.go
+++ b/analyzer/uncategorized/addresses.go
@@ -3,6 +3,8 @@ package common
 import (
 	"fmt"
 
+	ethCommon "github.com/ethereum/go-ethereum/common"
+
 	"github.com/oasisprotocol/oasis-core/go/common/crypto/address"
 	apiTypes "github.com/oasisprotocol/oasis-indexer/api/v1/types"
 	sdkTypes "github.com/oasisprotocol/oasis-sdk/client-sdk/go/types"
@@ -40,6 +42,15 @@ func StringifyEthAddress(ethAddr []byte) (apiTypes.Address, error) {
 	ctx := sdkTypes.AddressV0Secp256k1EthContext
 	ocAddr := address.NewAddress(ctx, ethAddr)
 	return StringifyOcAddress(ocAddr)
+}
+
+func EthAddrReference(eth_addr_hex string) *string {
+	if len(eth_addr_hex) == 0 {
+		return nil
+	}
+
+	ref := ethCommon.HexToAddress(eth_addr_hex).String()
+	return &ref
 }
 
 func ExtractAddresses(accounts map[apiTypes.Address]bool) []string {
--- a/api/spec/v1.yaml
+++ b/api/spec/v1.yaml
@@ -1909,6 +1909,11 @@ components:
             Retrieving the other signers is currently not supported by this API.
           # NOTE: To expose the other signers, use the transaction_signers table.
           example: oasis1qz670t637yyxshnlxhjj5074wgwl94d0x5x69zqd
+        sender_eth_0:
+          type: string
+          description: |
+            The Ethereum address of this transaction's 0th signer.
+          example: 0xd8A2Ae03f6Edd58999a0F1005db7a6532F2AA79e
         nonce_0:
           type: integer
           format: uint64
--- a/api/spec/v1.yaml
+++ b/api/spec/v1.yaml
@@ -86,6 +86,8 @@ x-examples:
   staking-address:
     - &staking_address_1 'oasis1qpg2xuz46g53737343r20yxeddhlvc2ldqsjh70p'
     - &staking_address_2 'oasis1qprtzrg97jk0wxnqkhxwyzy5qys47r7alvfl3fcg'
+  eth-address:
+    - &eth_address_1 0xd8A2Ae03f6Edd58999a0F1005db7a6532F2AA79e
   proposal-id:
     - &proposal_id_1 1
   proposal-state:
@@ -1909,11 +1911,11 @@ components:
             Retrieving the other signers is currently not supported by this API.
           # NOTE: To expose the other signers, use the transaction_signers table.
           example: oasis1qz670t637yyxshnlxhjj5074wgwl94d0x5x69zqd
-        sender_eth_0:
+        sender_0_eth:
           type: string
           description: |
             The Ethereum address of this transaction's 0th signer.
-          example: 0xd8A2Ae03f6Edd58999a0F1005db7a6532F2AA79e
+          example: *eth_address_1
         nonce_0:
           type: integer
           format: uint64
@@ -1951,6 +1953,11 @@ components:
               - For `method = "evm.Create"`, this is the address of the newly created smart contract.
               - For `method = "evm.Call"`, this is the address of the called smart contract
           example: "oasis1qq6ulxmcagnp5nr56ylva7nhmwnxtf0krumg9dkq"
+        to_eth:
+          type: string
+          description: |
+            A reasonable "to" Ethereum address associated with this transaction,
+          example: *eth_address_1
         amount:
           type: string
           description: |
--- a/api/v1/logic.go
+++ b/api/v1/logic.go
@@ -23,6 +23,12 @@ func renderRuntimeTransaction(storageTransaction client.RuntimeTransaction) (api
 	if err != nil {
 		return apiTypes.RuntimeTransaction{}, fmt.Errorf("utx open no verify: %w", err)
 	}
+	sender0Addr, err := uncategorized.StringifyAddressSpec(&tx.AuthInfo.SignerInfo[0].AddressSpec)
+	if err != nil {
+		return apiTypes.RuntimeTransaction{}, fmt.Errorf("signer 0: %w", err)
+	}
+	sender0 := string(sender0Addr)
+	sender0Eth := uncategorized.EthAddrReference(storageTransaction.AddressPreimage[sender0])
 	var cr types.CallResult
 	if err = cbor.Unmarshal(storageTransaction.ResultRaw, &cr); err != nil {
 		return apiTypes.RuntimeTransaction{}, fmt.Errorf("result unmarshal: %w", err)
@@ -37,8 +43,8 @@ func renderRuntimeTransaction(storageTransaction client.RuntimeTransaction) (api
 		Hash:       storageTransaction.Hash,
 		EthHash:    storageTransaction.EthHash,
 		Timestamp:  storageTransaction.Timestamp,
-		Sender0:    *storageTransaction.Sender0,
-		SenderEth0: storageTransaction.SenderEth0,
+		Sender0:    sender0,
+		Sender0Eth: sender0Eth,
 		Nonce0:     tx.AuthInfo.SignerInfo[0].Nonce,
 		Fee:        tx.AuthInfo.Fee.Amount.Amount.String(),
 		GasLimit:   tx.AuthInfo.Fee.Gas,
@@ -68,9 +74,12 @@ func renderRuntimeTransaction(storageTransaction client.RuntimeTransaction) (api
 					return fmt.Errorf("to: %w", err2)
 				}
 				to := string(toAddr)
+				toEth := uncategorized.EthAddrReference(storageTransaction.AddressPreimage[to])
 				apiTransaction.To = &to
+				apiTransaction.ToEth = toEth
 			} else {
-				apiTransaction.To = storageTransaction.Sender0
+				apiTransaction.To = &sender0
+				apiTransaction.ToEth = sender0Eth
 			}
 			amount, err2 := uncategorized.StringifyNativeDenomination(&body.Amount)
 			if err2 != nil {
@@ -86,12 +95,15 @@ func renderRuntimeTransaction(storageTransaction client.RuntimeTransaction) (api
 					return fmt.Errorf("to: %w", err2)
 				}
 				to := string(toAddr)
+				toEth := uncategorized.EthAddrReference(storageTransaction.AddressPreimage[to])
 				// Beware, this is the address of an account in the consensus
 				// layer, not an account in the runtime indicated in this API
 				// request.
 				apiTransaction.To = &to
+				apiTransaction.ToEth = toEth
 			} else {
-				apiTransaction.To = storageTransaction.Sender0
+				apiTransaction.To = &sender0
+				apiTransaction.ToEth = sender0Eth
 			}
 			// todo: ensure native denomination?
 			amount := body.Amount.Amount.String()
@@ -106,7 +118,9 @@ func renderRuntimeTransaction(storageTransaction client.RuntimeTransaction) (api
 					return fmt.Errorf("created contract: %w", err2)
 				}
 				to := string(toAddr)
+				toEth := uncategorized.EthAddrReference(storageTransaction.AddressPreimage[to])
 				apiTransaction.To = &to
+				apiTransaction.ToEth = toEth
 			}
 			amount := uncategorized.StringifyBytes(body.Value)
 			apiTransaction.Amount = &amount
@@ -118,7 +132,9 @@ func renderRuntimeTransaction(storageTransaction client.RuntimeTransaction) (api
 				return fmt.Errorf("to: %w", err2)
 			}
 			to := string(toAddr)
+			toEth := uncategorized.EthAddrReference(storageTransaction.AddressPreimage[to])
 			apiTransaction.To = &to
+			apiTransaction.ToEth = toEth
 			amount := uncategorized.StringifyBytes(body.Value)
 			apiTransaction.Amount = &amount
 			return nil
--- a/api/v1/logic.go
+++ b/api/v1/logic.go
@@ -23,11 +23,6 @@ func renderRuntimeTransaction(storageTransaction client.RuntimeTransaction) (api
 	if err != nil {
 		return apiTypes.RuntimeTransaction{}, fmt.Errorf("utx open no verify: %w", err)
 	}
-	sender0Addr, err := uncategorized.StringifyAddressSpec(&tx.AuthInfo.SignerInfo[0].AddressSpec)
-	if err != nil {
-		return apiTypes.RuntimeTransaction{}, fmt.Errorf("signer 0: %w", err)
-	}
-	sender0 := string(sender0Addr)
 	var cr types.CallResult
 	if err = cbor.Unmarshal(storageTransaction.ResultRaw, &cr); err != nil {
 		return apiTypes.RuntimeTransaction{}, fmt.Errorf("result unmarshal: %w", err)
@@ -37,18 +32,19 @@ func renderRuntimeTransaction(storageTransaction client.RuntimeTransaction) (api
 		return apiTypes.RuntimeTransaction{}, fmt.Errorf("body unmarshal: %w", err)
 	}
 	apiTransaction := apiTypes.RuntimeTransaction{
-		Round:     storageTransaction.Round,
-		Index:     storageTransaction.Index,
-		Hash:      storageTransaction.Hash,
-		EthHash:   storageTransaction.EthHash,
-		Timestamp: storageTransaction.Timestamp,
-		Sender0:   sender0,
-		Nonce0:    tx.AuthInfo.SignerInfo[0].Nonce,
-		Fee:       tx.AuthInfo.Fee.Amount.Amount.String(),
-		GasLimit:  tx.AuthInfo.Fee.Gas,
-		Method:    tx.Call.Method,
-		Body:      body,
-		Success:   cr.IsSuccess(),
+		Round:      storageTransaction.Round,
+		Index:      storageTransaction.Index,
+		Hash:       storageTransaction.Hash,
+		EthHash:    storageTransaction.EthHash,
+		Timestamp:  storageTransaction.Timestamp,
+		Sender0:    *storageTransaction.Sender0,
+		SenderEth0: storageTransaction.SenderEth0,
+		Nonce0:     tx.AuthInfo.SignerInfo[0].Nonce,
+		Fee:        tx.AuthInfo.Fee.Amount.Amount.String(),
+		GasLimit:   tx.AuthInfo.Fee.Gas,
+		Method:     tx.Call.Method,
+		Body:       body,
+		Success:    cr.IsSuccess(),
 	}
 	if err = uncategorized.VisitCall(&tx.Call, &cr, &uncategorized.CallHandler{
 		AccountsTransfer: func(body *accounts.Transfer) error {
@@ -74,7 +70,7 @@ func renderRuntimeTransaction(storageTransaction client.RuntimeTransaction) (api
 				to := string(toAddr)
 				apiTransaction.To = &to
 			} else {
-				apiTransaction.To = &sender0
+				apiTransaction.To = storageTransaction.Sender0
 			}
 			amount, err2 := uncategorized.StringifyNativeDenomination(&body.Amount)
 			if err2 != nil {
@@ -95,7 +91,7 @@ func renderRuntimeTransaction(storageTransaction client.RuntimeTransaction) (api
 				// request.
 				apiTransaction.To = &to
 			} else {
-				apiTransaction.To = &sender0
+				apiTransaction.To = storageTransaction.Sender0
 			}
 			// todo: ensure native denomination?
 			amount := body.Amount.Amount.String()
--- a/storage/client/queries.go
+++ b/storage/client/queries.go
@@ -394,35 +394,38 @@ func (qf QueryFactory) RuntimeBlocksQuery() string {
 
 func (qf QueryFactory) RuntimeTransactionsQuery() string {
 	return fmt.Sprintf(`
-		SELECT 
-				txs.round,
-				txs.tx_index,
-				txs.tx_hash,
-				txs.tx_eth_hash,
-				txs.timestamp,
-				txs.raw,
-				txs.result_raw,
-				%[1]s.address_preimages.address,
-				%[1]s.address_preimages.address_data
-			FROM %[1]s.runtime_transactions AS txs
-			LEFT JOIN %[1]s.runtime_related_transactions AS rel_accounts ON txs.round = rel_accounts.tx_round 
-				AND txs.tx_index = rel_accounts.tx_index
-				AND txs.runtime = rel_accounts.runtime
-				-- When related_address ($3) is NULL and hence we do no filtering on it, avoid the join altogether.
-				-- Otherwise, every tx will be returned as many times as there are related addresses for it. 
-				AND $3::text IS NOT NULL
-			JOIN %[1]s.runtime_transaction_signers ON txs.round = %[1]s.runtime_transaction_signers.round 
-				AND txs.tx_index = %[1]s.runtime_transaction_signers.tx_index
-				-- Currently, transactions only have a single signer in practice.
-				AND %[1]s.runtime_transaction_signers.signer_index = 0
-			JOIN %[1]s.address_preimages ON %[1]s.runtime_transaction_signers.signer_address = %[1]s.address_preimages.address
-			WHERE (txs.runtime = '%[2]s') AND
-						($1::bigint IS NULL OR txs.round = $1::bigint) AND
-						($2::text IS NULL OR txs.tx_hash = $2::text OR txs.tx_eth_hash = $2::text) AND
-						($3::text IS NULL OR rel_accounts.account_address = $3::text)
+		SELECT
+			txs.round,
+			txs.tx_index,
+			txs.tx_hash,
+			txs.tx_eth_hash,
+			txs.timestamp,
+			txs.raw,
+			txs.result_raw,
+			(
+				SELECT
+					json_object_agg(pre.address, encode(pre.address_data, 'hex'))
+				FROM %[1]s.runtime_related_transactions AS rel
+				JOIN %[1]s.address_preimages AS pre ON rel.account_address = pre.address
+				WHERE txs.runtime = rel.runtime
+					AND txs.round = rel.tx_round
+					AND txs.tx_index = rel.tx_index
+			) AS eth_addr_lookup
+		FROM %[1]s.runtime_transactions AS txs
+		LEFT JOIN %[1]s.runtime_related_transactions AS rel ON txs.round = rel.tx_round
+			AND txs.tx_index = rel.tx_index
+			AND txs.runtime = rel.runtime
+			-- When related_address ($3) is NULL and hence we do no filtering on it, avoid the join altogether.
+			-- Otherwise, every tx will be returned as many times as there are related addresses for it.
+			AND $3::text IS NOT NULL
+		WHERE (txs.runtime = '%[2]s') AND
+					($1::bigint IS NULL OR txs.round = $1::bigint) AND
+					($2::text IS NULL OR txs.tx_hash = $2::text OR txs.tx_eth_hash = $2::text) AND
+					($3::text IS NULL OR rel.account_address = $3::text)
 		ORDER BY txs.round DESC, txs.tx_index DESC
 		LIMIT $4::bigint
-		OFFSET $5::bigint`, qf.chainID, qf.runtime)
+		OFFSET $5::bigint
+		`, qf.chainID, qf.runtime)
 }
 
 func (qf QueryFactory) RuntimeEventsQuery() string {
--- a/storage/client/queries.go
+++ b/storage/client/queries.go
@@ -401,7 +401,9 @@ func (qf QueryFactory) RuntimeTransactionsQuery() string {
 				txs.tx_eth_hash,
 				txs.timestamp,
 				txs.raw,
-				txs.result_raw
+				txs.result_raw,
+				%[1]s.address_preimages.address,
+				%[1]s.address_preimages.address_data
 			FROM %[1]s.runtime_transactions AS txs
 			LEFT JOIN %[1]s.runtime_related_transactions AS rel_accounts ON txs.round = rel_accounts.tx_round 
 				AND txs.tx_index = rel_accounts.tx_index
@@ -409,6 +411,11 @@ func (qf QueryFactory) RuntimeTransactionsQuery() string {
 				-- When related_address ($3) is NULL and hence we do no filtering on it, avoid the join altogether.
 				-- Otherwise, every tx will be returned as many times as there are related addresses for it. 
 				AND $3::text IS NOT NULL
+			JOIN %[1]s.runtime_transaction_signers ON txs.round = %[1]s.runtime_transaction_signers.round 
+				AND txs.tx_index = %[1]s.runtime_transaction_signers.tx_index
+				-- Currently, transactions only have a single signer in practice.
+				AND %[1]s.runtime_transaction_signers.signer_index = 0
+			JOIN %[1]s.address_preimages ON %[1]s.runtime_transaction_signers.signer_address = %[1]s.address_preimages.address
 			WHERE (txs.runtime = '%[2]s') AND
 						($1::bigint IS NULL OR txs.round = $1::bigint) AND
 						($2::text IS NULL OR txs.tx_hash = $2::text OR txs.tx_eth_hash = $2::text) AND
--- a/docker/node/etc/config.yml
+++ b/docker/node/etc/config.yml
@@ -29,7 +29,3 @@ consensus:
           #   p2p:
           #     # External P2P configuration.
           #     enabled: true
-runtime:
-  mode: client
-  paths:
-    - /node/runtimes/emerald-paratime.orc
--- a/storage/client/client.go
+++ b/storage/client/client.go
@@ -8,6 +8,9 @@ import (
 	"github.com/dgraph-io/ristretto"
 	"github.com/jackc/pgtype"
 	"github.com/jackc/pgx/v4"
+
+	ethCommon "github.com/ethereum/go-ethereum/common"
+
 	oasisErrors "github.com/oasisprotocol/oasis-core/go/common/errors"
 
 	beacon "github.com/oasisprotocol/oasis-core/go/beacon/api"
@@ -998,6 +1001,7 @@ func (c *StorageClient) RuntimeTransactions(ctx context.Context, p apiTypes.GetR
 	}
 	for rows.Next() {
 		var t RuntimeTransaction
+		var address []byte
 		if err := rows.Scan(
 			&t.Round,
 			&t.Index,
@@ -1006,9 +1010,14 @@ func (c *StorageClient) RuntimeTransactions(ctx context.Context, p apiTypes.GetR
 			&t.Timestamp,
 			&t.Raw,
 			&t.ResultRaw,
+			&t.Sender0,
+			&address,
 		); err != nil {
 			return nil, wrapError(err)
 		}
+
+		senderEth0 := ethCommon.BytesToAddress(address).String()
+		t.SenderEth0 = &senderEth0
 
 		ts.Transactions = append(ts.Transactions, t)
 	}
--- a/storage/client/client.go
+++ b/storage/client/client.go
@@ -9,14 +9,10 @@ import (
 	"github.com/jackc/pgtype"
 	"github.com/jackc/pgx/v4"
 
-	ethCommon "github.com/ethereum/go-ethereum/common"
-
-	oasisErrors "github.com/oasisprotocol/oasis-core/go/common/errors"
-
 	beacon "github.com/oasisprotocol/oasis-core/go/beacon/api"
-	staking "github.com/oasisprotocol/oasis-core/go/staking/api"
-
 	"github.com/oasisprotocol/oasis-core/go/common/crypto/signature"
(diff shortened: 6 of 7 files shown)


api/spec/v1.yaml
```diff
@@ -1903,6 +1903,12 @@ components:
             Retrieving the other signers is currently not supported by this API.
           # NOTE: To expose the other signers, use the transaction_signers table.
           example: oasis1qz670t637yyxshnlxhjj5074wgwl94d0x5x69zqd
+        sender_eth_0:
+          type: string
+          description: |
+            The Ethereum address of this transaction's 0th signer.
+          # TODO: use a better example
```
aefhm, line 1910:
```
address            | oasis1qz670t637yyxshnlxhjj5074wgwl94d0x5x69zqd
context_identifier | oasis-runtime-sdk/address: secp256k1eth
context_version    | 0
address_data       | \xd8a2ae03f6edd58999a0f1005db7a6532f2aa79e
```

api/v1/strict_server.go
```diff
@@ -248,6 +250,12 @@ func (srv *StrictServerImpl) GetRuntimeTransactions(ctx context.Context, request
 		if err2 != nil {
 			return nil, fmt.Errorf("round %d tx %d: %w", storageTransaction.Round, storageTransaction.Index, err2)
 		}
+		// address preimage may not be available
```
aefhm, line 253: Alternatively move this logic into `api/v1/logic.go`?

Unfortunately, we don't get the Oasis address until post raw body parsing of the transaction, so we have to do a second loop for look up 🤢 🤔

aefhm: Thanks! Rabbit hole took me to `runtime_transaction_signers` eventually.

pro-wh: signers is only a subset though

api/spec/v1.yaml
```diff
@@ -1903,6 +1903,11 @@ components:
             Retrieving the other signers is currently not supported by this API.
           # NOTE: To expose the other signers, use the transaction_signers table.
           example: oasis1qz670t637yyxshnlxhjj5074wgwl94d0x5x69zqd
+        sender_eth_0:
+          type: string
+          description: |
+            The Ethereum address of this transaction's 0th signer.
+          example: 0xd8A2Ae03f6Edd58999a0F1005db7a6532F2AA79e
```
mitjat, line 1910: I think our output is all-lowercase. Also, do we prefix the `0x`?

api/spec/v1.yaml
```diff
@@ -1903,6 +1903,11 @@ components:
             Retrieving the other signers is currently not supported by this API.
           # NOTE: To expose the other signers, use the transaction_signers table.
           example: oasis1qz670t637yyxshnlxhjj5074wgwl94d0x5x69zqd
+        sender_eth_0:
```
mitjat, line 1906: nit: `eth` is the modifier of `sender_0`, so let's either prefix or postfix it, not stick it in the middle of the name. Realistically, we should postfix it for consistency with other fields.

```suggestion
        sender_0_eth:
```

mitjat (commented):
Thank you Xi!
Largely LGTM, with a performance concern where I'd like to ask you for some more reassurances/testing.

api/v1/logic.go
```diff
@@ -90,12 +97,15 @@ func renderRuntimeTransaction(storageTransaction client.RuntimeTransaction) (api
 					return fmt.Errorf("to: %w", err2)
 				}
 				to := string(toAddr)
+				toEth := ethCommon.BytesToAddress([]byte(storageTransaction.AddressLookup[body.To.String()])).String()
```
Andrew7234, line 100: here and elsewhere: Can you use [ethCommon.HexToAddress](https://pkg.go.dev/github.com/ethereum/go-ethereum/common#HexToAddress) instead of the byte-wrangling?

Andrew7234 (approved):

api/spec/v1.yaml
```diff
@@ -1940,6 +1945,11 @@ components:
               - For `method = "evm.Create"`, this is the address of the newly created smart contract.
               - For `method = "evm.Call"`, this is the address of the called smart contract
           example: "oasis1qq6ulxmcagnp5nr56ylva7nhmwnxtf0krumg9dkq"
+        to_eth:
+          type: string
+          description: |
+            A reasonable "to" Ethereum address associated with this transaction,
+          example: 0xd8A2Ae03f6Edd58999a0F1005db7a6532F2AA79e
```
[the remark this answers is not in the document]
aefhm: Legit.

api/v1/logic.go
```diff
@@ -72,9 +76,12 @@ func renderRuntimeTransaction(storageTransaction client.RuntimeTransaction) (api
...
```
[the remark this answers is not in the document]
aefhm: You are right. That was gross. Refactored to use just `to`.

api/spec/v1.yaml
aefhm: > I think our output is all-lowercase.

Yes, it is!

> Also, do we prefix the 0x?

Not that I know of.

I thought that mixed case would be more consistent with [explorer](https://testnet.explorer.sapphire.oasis.dev/tx/0x837d852282ba29546f04f57d535a38879d9eca530b22b37e52ed9c95349c82c1) and [Etherscan](https://etherscan.io/address/0x1f9090aae28b8a3dceadf281b0f12828e676c326)? I don't feel strongly about this and can go either way.

api/spec/v1.yaml
aefhm: Ah fair. I like the index at the end, but you are right, `sender_0_eth` is oddly more consistent.

api/v1/logic.go
aefhm: Yeah, this is better.

storage/client/queries.go
```diff
@@ -394,12 +394,32 @@ func (qf QueryFactory) RuntimeBlocksQuery() string {
 
 func (qf QueryFactory) RuntimeTransactionsQuery() string {
 	return fmt.Sprintf(`
-		SELECT round, tx_index, tx_hash, tx_eth_hash, timestamp, raw, result_raw
+		WITH address_lookups AS (
```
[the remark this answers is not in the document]
aefhm: Oof, `LIMIT 1000` is high, but regardless, this query is disgusting, even just on `5280995` rows in production.

<details><summary>Query</summary>

```
oasisindexer=> EXPLAIN WITH address_lookups AS (
  SELECT
    txs.round,
    txs.tx_index,
    json_object_agg(pre.address, pre.address_data) AS lookup
  FROM oasis_3.runtime_transactions AS txs
  JOIN oasis_3.runtime_related_transactions AS rel ON txs.round = rel.tx_round
    AND txs.tx_index = rel.tx_index
  JOIN oasis_3.address_preimages AS pre ON rel.account_address = pre.address
  GROUP BY txs.round, txs.tx_index
)
SELECT
    txs.round,
    txs.tx_index,
    txs.tx_hash,
    txs.tx_eth_hash,
    txs.timestamp,
    txs.raw,
    txs.result_raw,
    address_lookups.lookup
  FROM oasis_3.runtime_transactions AS txs
  JOIN address_lookups ON txs.round = address_lookups.round AND txs.tx_index = address_lookups.tx_index
  WHERE (txs.runtime = 'emerald')
ORDER BY txs.runtime DESC, txs.round DESC, txs.tx_index DESC
LIMIT 1000;
```

</details>

<details><summary>Plan</summary>

```                                                                                    QUERY PLAN                                                                                  
--------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
 Limit  (cost=9659943.41..9667236.21 rows=1000 width=750)
   ->  Merge Join  (cost=9659943.41..11022077.80 rows=186778 width=750)
         Merge Cond: (((txs.round)::bigint = (address_lookups.round)::bigint) AND ((txs.tx_index)::integer = (address_lookups.tx_index)::integer))
         ->  Index Scan Backward using runtime_transactions_pkey on runtime_transactions txs  (cost=0.43..1330276.36 rows=5215771 width=718)
               Index Cond: (runtime = 'emerald'::runtime)
         ->  Sort  (cost=9659942.98..9661246.92 rows=521577 width=44)
               Sort Key: address_lookups.round DESC, address_lookups.tx_index DESC
               ->  Subquery Scan on address_lookups  (cost=2544678.32..9594365.17 rows=521577 width=44)
                     ->  GroupAggregate  (cost=2544678.32..9589149.40 rows=521577 width=44)
                           Group Key: txs_1.round, txs_1.tx_index
                           ->  Merge Join  (cost=2544678.32..8296204.27 rows=171523388 width=82)
                                 Merge Cond: (((txs_1.round)::bigint = (rel.tx_round)::bigint) AND ((txs_1.tx_index)::integer = (rel.tx_index)::integer))
                                 ->  Gather Merge  (cost=626367.93..1233830.41 rows=5215771 width=12)
                                       Workers Planned: 2
                                       ->  Sort  (cost=625367.90..630801.00 rows=2173238 width=12)
                                             Sort Key: txs_1.round, txs_1.tx_index
                                             ->  Parallel Index Only Scan using runtime_transactions_pkey on runtime_transactions txs_1  (cost=0.43..322335.23 rows=2173238 width=12)
                                 ->  Materialize  (cost=1918012.90..4086200.42 rows=18225197 width=82)
                                       ->  Gather Merge  (cost=1918012.90..4040637.43 rows=18225197 width=82)
                                             Workers Planned: 2
                                             ->  Sort  (cost=1917012.88..1935997.46 rows=7593832 width=82)
                                                   Sort Key: rel.tx_round, rel.tx_index
                                                   ->  Hash Join  (cost=1388.08..322420.70 rows=7593832 width=82)
                                                         Hash Cond: ((rel.account_address)::text = (pre.address)::text)
                                                         ->  Parallel Seq Scan on runtime_related_transactions rel  (cost=0.00..301096.32 rows=7593832 width=59)
                                                         ->  Hash  (cost=963.59..963.59 rows=33959 width=70)
                                                               ->  Seq Scan on address_preimages pre  (cost=0.00..963.59 rows=33959 width=70)
(27 rows)
```
</details>

Solutions:

1. The 1:1 or fast query route is possible if we choose not to show the `toEth` address. This wasn't explicitly called out in the ticket, but I think we wanted this, with @pro-wh suggestion.

2. Let me try to subquery my way out of this first?

3. The two queries multiple pass messes up the clean logic of `renderRuntimeTransaction` function, but would be preferable to me than probably more denormalizing.

api/spec/v1.yaml
mitjat: The mixed case is an eth kinda-cool kinda-hacky feature where the lower/uppercase distinction forms a series of bits, and those bits represent the checksum of the address. It's optional; all eth tools should accept all-lowercase addresses. We don't implement the mixed-case anywhere (we'd have to recompute it) right now.

storage/client/queries.go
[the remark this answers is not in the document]
mitjat: Thank you! Bummer for the results; oh well.

BTW, if you run `EXPLAIN ANALYZE` instead of just `EXPLAIN`, you get actual timing info, which is usually quite helpful. Though here ... there's a step that materializes an estimated 1.8M rows, that's going to be slow no matter what :)

Re 1: I agree `toEth` is very useful, let's try hard to keep it. 
Re 2: Sure. Maybe you can do sth like `WITH txs AS (SELECT FROM oasis_3.runtime_transactions LIMIT 1000) SELECT ... FROM txs JOIN blah JOIN blah`? That will encourage postgres to do index lookups because there's only 1000 rows (x4 JOINs). But don't forget to rebase first; Andy just introduced additional complexity into this query that we'll have to accommodate; I don't remember if it's problematic in that sense. Also, this assumes that the JOINs will not change the number of rows in either direction.
Re 3: We can break `renderRuntimeTransaction` into two stages no problem. What would be even better would be to run most of `renderRuntimeTransaction` at analysis time and modify the table so that it contains a much more structured version of the tx, with each field in a separate column. At render time, we wouldn't need `renderRuntimeTransaction` any more, we'd just pull from the DB directly. Warren initially set up thsi parsing/rendering step outside analysis so we'd have more flexibility in the parsing (eth is tricky, and you don't want to restructure the table every time you figure out how to parse the tx more/better), but we're pretty stable now and it would be cleaner to do the tx parsing during analysis.

Denormalization: I'm not opposed to it philosophically. The indexer uses mostly INSERTs, and only few UPDATEs, so it's unusually easy to keep even fairly denormalized data in sync, I feel.
And FWIW, the last branch `ANALYZE` output builds a denormalized table on the fly: It joins `runtime_related_transactions` and `address_preimages` and materializes the join :). Though actually, doing that denorm ourselves makes it a little trickier later on to implement parallel processing of blocks, because in that scenario, by the time we observe a tx-related address we don't necessarily know its preimage yet. But we could work around it with a final UPDATE after the parallel work is done. So denorm is on the table, even this particular denorm is on the table, it's just not my first choice.

Denormalization 2: The first thing I'd probably recommend going for is getting rid of `runtime_related_transactions` (terrible name btw) and sticking that info into `runtime_transactions` directly as a column, much like [we do for events](https://github.com/oasisprotocol/oasis-indexer/blob/96a140711e77d460da56a0e5f5bf4e6bbaf8283b/storage/migrations/02_oasis_3_runtimes.up.sql#L98-L98). We're using these two separate approaches (`TEXT[]` vs `JOIN`) for random reasons only; it would be nice to settle on one anyway.

If you do any denorm work, or move around the parsing from `renderRuntimeTransaction` (see "re 3" above), that's best done in a separate prework PR.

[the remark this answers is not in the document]
mitjat: Up to you what you wanted to try first. I mostly wrote a wall of text to nudge us towards changes that will likely be beneficial for this issue but at the same time were on my mental tech-debt wishlist even before the PR.

Also BTW, setting `LIMIT 10` doesn't change the execution plan (or time, which is ~50s). `VACUUM ANALYZE` and `REINDEX SCHEMA` don't help either.

api/spec/v1.yaml
aefhm: `ethCommon.HexToAddress` which we use does return the mixed-case.

storage/client/queries.go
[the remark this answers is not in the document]
aefhm: > Denormalization 2

This would be my preferred denormalization strategy. I don't see why we couldn't have done that in one pass with `runtime_transactions`.

I may disappoint you though. I think this is looking a lil better. What is our perf expectation 😅?

<details><summary>Query</summary>

```
SELECT 
    txs.round,
    txs.tx_index,
    txs.tx_hash,
    json_object_agg(pre.address, pre.address_data) AS lookup
FROM oasis_3.runtime_transactions AS txs
JOIN oasis_3.runtime_related_transactions AS rel ON txs.round = rel.tx_round
    AND txs.tx_index = rel.tx_index
    AND txs.round = rel.tx_round
JOIN oasis_3.address_preimages AS pre ON rel.account_address = pre.address
WHERE (txs.round, txs.tx_index, txs.tx_hash) IN

(
    SELECT 
    txs.round,
    txs.tx_index,
    txs.tx_hash
    FROM oasis_3.runtime_transactions AS txs
    WHERE (txs.runtime = 'emerald')
    ORDER BY txs.round DESC, txs.tx_index DESC
    LIMIT 1000
)

GROUP BY txs.round, txs.tx_index, txs.tx_hash;
```
</details>

<details><summary>Planner</summary>

```

-------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
 GroupAggregate  (cost=552829.82..552841.47 rows=466 width=109) (actual time=14065.302..14069.309 rows=1000 loops=1)
   Group Key: txs.round, txs.tx_index, txs.tx_hash
   ->  Sort  (cost=552829.82..552830.99 rows=466 width=147) (actual time=14065.181..14065.411 rows=2707 loops=1)
         Sort Key: txs.round, txs.tx_index, txs.tx_hash
         Sort Method: quicksort  Memory: 816kB
         ->  Nested Loop  (cost=8236.11..552809.17 rows=466 width=147) (actual time=3515.101..14061.673 rows=2707 loops=1)
               ->  Hash Join  (cost=8235.69..552339.24 rows=466 width=124) (actual time=3512.093..13829.490 rows=2708 loops=1)
                     Hash Cond: (((rel.tx_round)::bigint = (txs.round)::bigint) AND ((rel.tx_index)::integer = (txs.tx_index)::integer))
                     ->  Seq Scan on runtime_related_transactions rel  (cost=0.00..407409.97 rows=18225197 width=59) (actual time=1.452..8104.875 rows=18220073 loops=1)
                     ->  Hash  (cost=8235.48..8235.48 rows=14 width=89) (actual time=724.715..724.721 rows=1000 loops=1)
                           Buckets: 1024  Batches: 1  Memory Usage: 130kB
                           ->  Nested Loop  (cost=272.98..8235.48 rows=14 width=89) (actual time=5.370..722.444 rows=1000 loops=1)
                                 ->  HashAggregate  (cost=272.98..282.98 rows=1000 width=77) (actual time=2.885..5.002 rows=1000 loops=1)
                                       Group Key: ("ANY_subquery".round)::bigint, ("ANY_subquery".tx_index)::integer, ("ANY_subquery".tx_hash)::text
                                       Batches: 1  Memory Usage: 577kB
                                       ->  Subquery Scan on "ANY_subquery"  (cost=0.43..265.48 rows=1000 width=77) (actual time=1.245..2.157 rows=1000 loops=1)
                                             ->  Limit  (cost=0.43..255.48 rows=1000 width=77) (actual time=1.244..1.880 rows=1000 loops=1)
                                                   ->  Index Scan Backward using runtime_transactions_pkey on runtime_transactions txs_1  (cost=0.43..1330276.36 rows=5215771 width=77) (actual time=1.243..1.765 rows=1000 loops=1)
                                                         Index Cond: (runtime = 'emerald'::runtime)
                                 ->  Index Scan using ix_runtime_transactions_tx_hash on runtime_transactions txs  (cost=0.00..7.94 rows=1 width=77) (actual time=0.713..0.716 rows=1 loops=1000)
                                       Index Cond: ((tx_hash)::text = ("ANY_subquery".tx_hash)::text)
                                       Rows Removed by Index Recheck: 0
                                       Filter: ((("ANY_subquery".round)::bigint = (round)::bigint) AND (("ANY_subquery".tx_index)::integer = (tx_index)::integer))
               ->  Index Scan using address_preimages_pkey on address_preimages pre  (cost=0.41..1.01 rows=1 width=70) (actual time=0.085..0.085 rows=1 loops=2708)
                     Index Cond: ((address)::text = (rel.account_address)::text)
 Planning Time: 13.361 ms
 Execution Time: 14069.796 ms
```
</details>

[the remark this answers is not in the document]
aefhm: If this isn't fast enough, probably have to denormalize the `runtime_transactions` table to swallow `runtime_related_transactions`.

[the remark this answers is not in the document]
wh0: woa, I gotta look into this tomorrow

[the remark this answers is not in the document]
mitjat: With this new approach, we cannot filter results by related address ([like this](https://github.com/oasisprotocol/oasis-indexer/blob/288d291fca9fc0e8ccb9f5d0ac68fc74932f285b/storage/client/queries.go#L415-L415), I did warn to rebase first somewhere in the wall of text above :)).
It's also still way too slow, unfortunately. I expect this will be one of the most commonly loaded web pages, and so one of the most common queries we run. **Twice**, once we [start returning  the total number of pagination pages](https://app.clickup.com/t/861mcbbhg). So we should be shooting for about 100ms I reckon.

I'm not too worried yet. This task just happened to expose our tech debt and is forcing us to pay it down:
1. Parse txs during analysis time so we can fetch the txs and then join `address_preimages` just to `sender` and `to` fields. (Alternative: Fetch txs into go, parse txs, go to the DB again for preimages. That's uglier though, and I don't think it's easier to implement either.)
2. Put related addresses directly into `runtime_transactions` in a `TEXT[]` column.

Once we have those two PRs done, let's revisit this and see how hard (or not) the problem is.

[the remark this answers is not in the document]
mitjat: Update: I'm hopeful. Joining 15k addresses with address_preimages takes 40ms. Filtering by a `TEXT[]` field can be mega expensive, but when using a gin index and limiting results to at most 1000 (as we will), the cost is under 20ms. And that should be all we need.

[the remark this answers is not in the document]
aefhm: > we cannot filter results by related address

I rebased this PR, but didn't include that `WHERE` because I thought that use case was narrow and more performant than the general case. I'd imagine filtering to return the primary key of `(round, tx_index, tx_hash)` works as expected.

> It's also still way too slow, unfortunately. I expect this will be one of the most commonly loaded web pages, and so one of the most common queries we run

Fair. I do think going 1K to 2K for max limit shouldn't change much.

> 1. Parse txs during analysis time so we can fetch the txs and then join address_preimages just to sender and to fields. (Alternative: Fetch txs into go, parse txs, go to the DB again for preimages. That's uglier though, and I don't think it's easier to implement either.)

Slight groan on this haha, but yeah, I do agree we should just pay the one time cost up front.

> 2. Put related addresses directly into runtime_transactions in a TEXT[] column.

Very aligned on this.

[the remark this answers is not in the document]
pro-wh: > I'm quite worried about the performance of this quadruple JOIN.

what was the original quadruple join?

> Staging has 2M runtime txs right now; it will have >10M

maybe this used to be different, but we never have to scan all transactions to complete this query right?

> * stoer related_addresses inside runtime_txs

we do this for events, maybe we can do it here. is it efficient enough to find txs that are related to a given address this way?

> * preimage inside related_addresses

might end up being annoying to update the related addresses table when we find out a new preimage. I hope we don't have to do this :skull:

> * parse the txs to get oasis addresses

don't need to do this in a separate round trip. the accounts related transactions table knows the addresses you'll need

> Materialize  (cost=1918012.90..4086200.42 rows=18225197 width=82)

omg why does it materialize that? and why does it go up from 7593832 to 18225197?

> 1. The 1:1 or fast query route is possible if we choose not to show the `toEth` address. This wasn't explicitly called out in the ticket, but I think we wanted this, with @pro-wh suggestion.

what's the 1:1? what's the fast query route? what's `toEth`? what was @pro-wh suggestion? why does clickup render `inline code` in a proportional font?

> 3\. two queries multiple pass messes up the clean logic

won't need multiple passes because that accounts related transactions table already knows. still two queries though, which should be doable

> there's a step that materializes an estimated 1.8M rows

no 18 million

> by the time we observe a tx-related address we don't necessarily know its preimage yet

that should already be the case. for example, deposits into Emerald are made out to an oasis-style address, and that account may never have been used in an ethereum style transaction before

> random reasons only

being economical in how many postgresql-specific bells and whistles we use :stuck_out_tongue: 

> WHERE (txs.round, txs.tx_index, txs.tx_hash) IN

what's the `WHERE ... IN ...` for? also, take out the `txs.tx_hash`. its making the planner do some extra hashing



> AND txs.round = rel.tx_round

duplicated condition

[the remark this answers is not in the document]
pro-wh: ->  Seq Scan on runtime_related_transactions rel  (cost=0.00..407409.97 rows=18225197 width=59) (actual time=1.452..8104.875 rows=18220073 loops=1)

this is where things go wrong btw. we should definitely not be doing a full scan on related transactions. fixing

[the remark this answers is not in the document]
pro-wh: ok please try with #320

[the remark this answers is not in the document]
aefhm: Post #320 

<details><summary>New</summary>

```
                                                                                                             QUERY PLAN                                                           
-------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
 GroupAggregate  (cost=552740.13..552751.78 rows=466 width=109) (actual time=8004.759..8008.638 rows=1000 loops=1)
   Group Key: txs.round, txs.tx_index, txs.tx_hash
   ->  Sort  (cost=552740.13..552741.30 rows=466 width=147) (actual time=8004.693..8004.912 rows=2707 loops=1)
         Sort Key: txs.round, txs.tx_index, txs.tx_hash
         Sort Method: quicksort  Memory: 816kB
         ->  Nested Loop  (cost=8236.11..552719.48 rows=466 width=147) (actual time=2774.860..8000.603 rows=2707 loops=1)
               ->  Hash Join  (cost=8235.69..552249.55 rows=466 width=124) (actual time=2771.679..7761.576 rows=2708 loops=1)
                     Hash Cond: (((rel.tx_round)::bigint = (txs.round)::bigint) AND ((rel.tx_index)::integer = (txs.tx_index)::integer))
                     ->  Seq Scan on runtime_related_transactions rel  (cost=0.00..407358.72 rows=18220072 width=59) (actual time=0.104..3621.805 rows=18220073 loops=1)
                     ->  Hash  (cost=8235.48..8235.48 rows=14 width=89) (actual time=721.130..721.134 rows=1000 loops=1)
                           Buckets: 1024  Batches: 1  Memory Usage: 130kB
                           ->  Nested Loop  (cost=272.98..8235.48 rows=14 width=89) (actual time=11.787..718.852 rows=1000 loops=1)
                                 ->  HashAggregate  (cost=272.98..282.98 rows=1000 width=77) (actual time=4.259..5.872 rows=1000 loops=1)
                                       Group Key: ("ANY_subquery".round)::bigint, ("ANY_subquery".tx_index)::integer, ("ANY_subquery".tx_hash)::text
                                       Batches: 1  Memory Usage: 577kB
                                       ->  Subquery Scan on "ANY_subquery"  (cost=0.43..265.48 rows=1000 width=77) (actual time=1.447..2.856 rows=1000 loops=1)
                                             ->  Limit  (cost=0.43..255.48 rows=1000 width=77) (actual time=1.445..2.427 rows=1000 loops=1)
                                                   ->  Index Scan Backward using runtime_transactions_pkey on runtime_transactions txs_1  (cost=0.43..1330276.36 rows=5215771 width=77) (actual time=1.444..2.258 rows=1000 loops=1)
                                                         Index Cond: (runtime = 'emerald'::runtime)
                                 ->  Index Scan using ix_runtime_transactions_tx_hash on runtime_transactions txs  (cost=0.00..7.94 rows=1 width=77) (actual time=0.708..0.711 rows=1 loops=1000)
                                       Index Cond: ((tx_hash)::text = ("ANY_subquery".tx_hash)::text)
                                       Rows Removed by Index Recheck: 0
                                       Filter: ((("ANY_subquery".round)::bigint = (round)::bigint) AND (("ANY_subquery".tx_index)::integer = (tx_index)::integer))
               ->  Index Scan using address_preimages_pkey on address_preimages pre  (cost=0.41..1.01 rows=1 width=70) (actual time=0.087..0.087 rows=1 loops=2708)
                     Index Cond: ((address)::text = (rel.account_address)::text)
 Planning Time: 16.883 ms
 Execution Time: 8009.534 ms
(27 rows)
```
</details>

Much faster, but still meh.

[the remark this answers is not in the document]
aefhm: Blarf! Sorry, sorry, I had a dup in my query.

<details><summary>Revised query</summary>

```
SELECT 
    txs.round,
    txs.tx_index,
    txs.tx_hash,
    json_object_agg(pre.address, pre.address_data) AS lookup
FROM oasis_3.runtime_transactions AS txs
JOIN oasis_3.runtime_related_transactions AS rel
    ON txs.runtime = rel.runtime
    AND txs.tx_index = rel.tx_index
    AND txs.round = rel.tx_round
JOIN oasis_3.address_preimages AS pre ON rel.account_address = pre.address
WHERE (txs.round, txs.tx_index, txs.tx_hash) IN

(
    SELECT 
    txs.round,
    txs.tx_index,
    txs.tx_hash
    FROM oasis_3.runtime_transactions AS txs
    WHERE (txs.runtime = 'emerald')
    ORDER BY txs.round DESC, txs.tx_index DESC
    LIMIT 1000
)

GROUP BY txs.round, txs.tx_index, txs.tx_hash;
```
</details>


<details><summary>Res</summary>

```
                                                                                                          QUERY PLAN                                                              
-------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
 GroupAggregate  (cost=8314.13..8315.36 rows=49 width=109) (actual time=45.768..49.623 rows=1000 loops=1)
   Group Key: txs.round, txs.tx_index, txs.tx_hash
   ->  Sort  (cost=8314.13..8314.26 rows=49 width=147) (actual time=45.745..45.980 rows=2707 loops=1)
         Sort Key: txs.round, txs.tx_index, txs.tx_hash
         Sort Method: quicksort  Memory: 816kB
         ->  Nested Loop  (cost=273.96..8312.76 rows=49 width=147) (actual time=2.133..42.887 rows=2707 loops=1)
               ->  Nested Loop  (cost=273.54..8263.34 rows=49 width=124) (actual time=2.077..14.216 rows=2708 loops=1)
                     ->  Nested Loop  (cost=272.98..8235.48 rows=14 width=93) (actual time=1.629..7.359 rows=1000 loops=1)
                           ->  HashAggregate  (cost=272.98..282.98 rows=1000 width=77) (actual time=1.613..2.151 rows=1000 loops=1)
                                 Group Key: ("ANY_subquery".round)::bigint, ("ANY_subquery".tx_index)::integer, ("ANY_subquery".tx_hash)::text
                                 Batches: 1  Memory Usage: 577kB
                                 ->  Subquery Scan on "ANY_subquery"  (cost=0.43..265.48 rows=1000 width=77) (actual time=0.024..0.997 rows=1000 loops=1)
                                       ->  Limit  (cost=0.43..255.48 rows=1000 width=77) (actual time=0.023..0.658 rows=1000 loops=1)
                                             ->  Index Scan Backward using runtime_transactions_pkey on runtime_transactions txs_1  (cost=0.43..1330276.36 rows=5215771 width=77) (actual time=0.023..0.544 rows=1000 loops=1)
                                                   Index Cond: (runtime = 'emerald'::runtime)
                           ->  Index Scan using ix_runtime_transactions_tx_hash on runtime_transactions txs  (cost=0.00..7.94 rows=1 width=81) (actual time=0.004..0.005 rows=1 loops=1000)
                                 Index Cond: ((tx_hash)::text = ("ANY_subquery".tx_hash)::text)
                                 Rows Removed by Index Recheck: 0
                                 Filter: ((("ANY_subquery".tx_index)::integer = (tx_index)::integer) AND (("ANY_subquery".round)::bigint = (round)::bigint))
                     ->  Index Scan using ix_runtime_related_transactions_round_index on runtime_related_transactions rel  (cost=0.56..1.97 rows=2 width=63) (actual time=0.005..0.006 rows=3 loops=1000)
                           Index Cond: ((runtime = txs.runtime) AND ((tx_round)::bigint = (txs.round)::bigint) AND ((tx_index)::integer = (txs.tx_index)::integer))
               ->  Index Scan using address_preimages_pkey on address_preimages pre  (cost=0.41..1.01 rows=1 width=70) (actual time=0.010..0.010 rows=1 loops=2708)
                     Index Cond: ((address)::text = (rel.account_address)::text)
 Planning Time: 1.591 ms
 Execution Time: 49.836 ms
(25 rows)
```
</details>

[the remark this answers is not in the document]
aefhm: > > I'm quite worried about the performance of this quadruple JOIN.
> 
> what was the original quadruple join?

I think he meant the `JOIN` across `runtime_transactions`, `runtime_related_transactions`, `address_preimages`, and on the subquery again.
> 
> > Staging has 2M runtime txs right now; it will have >10M
> 
> maybe this used to be different, but we never have to scan all transactions to complete this query right?

I believe this.
> 
> > * stoer related_addresses inside runtime_txs
> 
> we do this for events, maybe we can do it here. is it efficient enough to find txs that are related to a given address this way?

addresses could be one to many right? I think reaching across a different table is okay.

> 
> > * preimage inside related_addresses
> 
> might end up being annoying to update the related addresses table when we find out a new preimage. I hope we don't have to do this 💀

Agree.
> 
> > * parse the txs to get oasis addresses
> 
> don't need to do this in a separate round trip. the accounts related transactions table knows the addresses you'll need

Aye.
> 
> > Materialize  (cost=1918012.90..4086200.42 rows=18225197 width=82)
> 
> omg why does it materialize that? and why does it go up from 7593832 to 18225197?

Bad querying from moi.
> 
> > 1. The 1:1 or fast query route is possible if we choose not to show the `toEth` address. This wasn't explicitly called out in the ticket, but I think we wanted this, with @pro-wh suggestion.
> 
> what's the 1:1? what's the fast query route? what's `toEth`? what was @pro-wh suggestion? why does clickup render `inline code` in a proportional font?

1:1 was referring to the `JOIN` to the `runtime_transaction_signers` to get the address in the first version of this PR, but @pro-wh helpfully called out that the `toEth` address is meaningful. "why does clickup _______?"
> 
> > 3. two queries multiple pass messes up the clean logic
> 
> won't need multiple passes because that accounts related transactions table already knows. still two queries though, which should be doable

K.
> 
> > there's a step that materializes an estimated 1.8M rows
> 
> no 18 million

Self-cringe. It is better now.
> 
> > by the time we observe a tx-related address we don't necessarily know its preimage yet
> 
> that should already be the case. for example, deposits into Emerald are made out to an oasis-style address, and that account may never have been used in an ethereum style transaction before
> 
> > random reasons only
> 
> being economical in how many postgresql-specific bells and whistles we use 😛
> 
> > WHERE (txs.round, txs.tx_index, txs.tx_hash) IN
> 
> what's the `WHERE ... IN ...` for? also, take out the `txs.tx_hash`. its making the planner do some extra hashing

K. I should just use the correct primary key.
> 
> > AND txs.round = rel.tx_round
> 
> duplicated condition

Yes.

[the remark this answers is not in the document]
pro-wh: :raised_hands: thanks!

storage/client/queries.go
```diff
@@ -394,28 +394,50 @@ func (qf QueryFactory) RuntimeBlocksQuery() string {
...
+			)
+		GROUP BY
+			txs.round,
+			txs.tx_index,
+			txs.tx_hash,
+			txs.tx_eth_hash,
+			txs.timestamp,
+			txs.raw,
+			txs.result_raw
+		`, qf.chainID, qf.runtime)
```
mitjat, line 440: I find this more readable when the inner and outer query are inverted:
```sql
SELECT 
  txs.*,  -- all the fields from GROUP BY
  (              
    SELECT json_object_agg(pre.address, pre.address_data)
    FROM oasis_3.runtime_related_transactions rel
    JOIN oasis_3.address_preimages pre ON rel.account_address = pre.address
    WHERE txs.runtime = rel.runtime AND txs.round = rel.tx_round AND txs.tx_index = rel.tx_index
  ) AS eth_addr_lookup
  FROM oasis_3.runtime_transactions txs
  -- LEFT JOIN rel_accounts
  -- WHERE ... 
  LIMIT 1000;
```

It avoids the expansion-and-then-contraction of the number of rows where we first JOIN and then GROUP many of the resulting rows.

I also tested the speed, and it's a little faster (for limit=10 or 100; a little slower for limit=1000, which will probably see less use).

WDYT?

analyzer/uncategorized/addresses.go
```diff
@@ -42,6 +45,15 @@ func StringifyEthAddress(ethAddr []byte) (apiTypes.Address, error) {
 	return StringifyOcAddress(ocAddr)
 }
 
+func EthAddrReference(preimage string) *string {
+	if len(preimage) == 0 {
+		return nil
+	}
+
+	ref := ethCommon.HexToAddress(strings.Trim(preimage, "\\x")).String()
```
mitjat, line 53: Here we're hardcoding a peculiarity of postgres (`bytea` is returned serialized as hex, with "\\x" prepended), in a module completely separate from sql.

Let's fix the encoding on the sql side instead:
`select encode(address_data, 'hex') from ...`
https://stackoverflow.com/questions/44813386/mysqls-hex-and-unhex-equivalent-in-postgres

This Trim() can go away then.

analyzer/uncategorized/addresses.go
```diff
@@ -42,6 +45,15 @@ func StringifyEthAddress(ethAddr []byte) (apiTypes.Address, error) {
 	return StringifyOcAddress(ocAddr)
 }
 
+func EthAddrReference(preimage string) *string {
+	if len(preimage) == 0 {
```
mitjat, line 49: Does this ever happen? Feels like it shouldn't.

analyzer/uncategorized/addresses.go
```diff
@@ -42,6 +45,15 @@ func StringifyEthAddress(ethAddr []byte) (apiTypes.Address, error) {
 	return StringifyOcAddress(ocAddr)
 }
 
+func EthAddrReference(preimage string) *string {
```
mitjat, line 48:
```suggestion
func EthAddrReference(eth_addr_hex string) *string {
```
to make it clearer what the fn does.

api/spec/v1.yaml
mitjat: Oh my bad, TIL. Thank you.

storage/client/queries.go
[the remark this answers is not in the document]
mitjat: Thank you Warren for realizing the bad query plan was not because postgres was too stupid to use index, or because the indexes were too expensive, but because they were missing :facepalm: 

I'm a little sad for losing the opportunity/excuse to do the cleanup we discussed :), but it's good news that we can move ahead faster.

storage/client/queries.go
```diff
@@ -394,28 +394,50 @@ func (qf QueryFactory) RuntimeBlocksQuery() string {
...
+			AND txs.tx_index = rel.tx_index
+		JOIN %[1]s.address_preimages AS pre ON rel.account_address = pre.address
+		WHERE (txs.runtime, txs.round, txs.tx_index) IN
+			(
+				SELECT
+					txs.runtime,
+					txs.round,
+					txs.tx_index
+				FROM %[1]s.runtime_transactions AS txs
+				LEFT JOIN %[1]s.runtime_related_transactions AS rel_accounts ON txs.round = rel_accounts.tx_round
```
mitjat, line 418: nit: we refer to this table once as `rel` and once as `rel_accounts`; let's pick one.
(I think I suggested at least one of the names; my bad)

mitjat (approved):
Thank you Xi and Warren!

analyzer/uncategorized/addresses.go
pro-wh: aaah the json object making us not able to have byte arrays

storage/client/queries.go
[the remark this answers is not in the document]
aefhm: Not anti-cleanup, but mounting guilt from not finishing tickets, maybe in a few sprints 🙏

analyzer/uncategorized/addresses.go
aefhm: Does :(

```
{
  "amount": "715000000000000000000",
  "body": {
    "amount": [
      "JsKeR8SETAAA",
      ""
    ],
    "to": "ABSac93FbIpG2PVK4dp6tXP4tkWw"
  },
  "fee": "2000000000000000",
  "gas_limit": 15000,
  "hash": "be42e50e8b65ce95ad6677c429659d7a6133c4e9e786e5cf8cd38be161ad7b37",
  "index": 1,
  "method": "consensus.Withdraw",
  "nonce_0": 140,
  "round": 1009434,
  "sender_0": "oasis1qqa20swrvp8hl4fz6cc2c5g7dq5eeuz28g00hf3j",
  "sender_0_eth": "0xAc09Ca840c02e9883fEB40A95eC7f7E506799601",
  "success": true,
  "timestamp": "2022-04-11T19:00:02-05:00",
  "to": "oasis1qq2f5u7ac4kg53kc749wrkn6k4el3dj9kqeq85ek"
}
```

We don't have a preimage for `oasis1qq2f5u7ac4kg53kc749wrkn6k4el3dj9kqeq85ek`.

Uh, alright, I'll dig a little more and see if we missed anything in prework.

storage/client/queries.go
aefhm: Cool! Going with `rel`.

storage/client/queries.go
aefhm: Oh yeah. This is much nice than the `GROUP BY`.

analyzer/uncategorized/addresses.go
mitjat: It's OK to not have preimages for some addrs. We only learn the preimage when the account actively submits a tx. So if somebody sends tokens _to_ an address that's never submitted anything before ... we have the above situation.

I assumed that for such addresses, we'd never reach `EthAddrReference()` in Go, but I reread now and I see I was wrong.

pro-wh: https://github.com/oasisprotocol/oasis-indexer/blob/xz/3u9y3yp-add-transaction-sender-eth-address/storage/migrations/02_oasis_3_runtimes.up.sql#L252

although wouldn't this reference constraint have prevented us from scanning that block in the first place?

mitjat: > wouldn't this reference constraint have prevented us from scanning that block in the first place?

The constraint is only on the tx sender. Xi's example has a `to` with an unknown eth address, not a `sender_0`.

pro-wh: ah

aefhm: Coolio. Thanks for saving me a rabbit hole trip.

PR #311, opened 2023-02-07
EVM: remove chain ID check

pro-wh:
for #305 

remove the chain ID check. it's arguably part of verifying an transaction, which we don't do in the indexer for runtimes

this makes the `decodeEthRawTx` function a little more footgun if anyone else discovers it though

### DIFF
--- a/analyzer/uncategorized/evm_raw_tx.go
+++ b/analyzer/uncategorized/evm_raw_tx.go
@@ -11,7 +11,7 @@ import (
 	sdkTypes "github.com/oasisprotocol/oasis-sdk/client-sdk/go/types"
 )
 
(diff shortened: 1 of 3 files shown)


Andrew7234 (approved):
If the analyzer `chainId` is misconfigured, will the indexer error when connecting to the node?

Just want to check that this doesn't permit any silent errors to happen.

pro-wh:
> If the analyzer `chainId` is misconfigured, will the indexer error when connecting to the node?

which, the oasis-3 one?

Andrew7234:
Nvm I see that it was hardcoded before `tx, err := decodeEthRawTx(utx.Body, 42262)`

PR #312, opened 2023-02-09
metrics: Normalize endpoints, use histogram for latencies

abukosek:
This is the first part of the metrics fixes -- the endpoint URLs are normalized based on a dumb but working heuristic and the latencies are now reported as a histogram.

### DIFF
--- a/api/middleware.go
+++ b/api/middleware.go
@@ -23,6 +23,31 @@ var (
 	maxLimit                 = uint64(1000)
 )
 
+// normalizeEndpoint removes all unique identifiers from the URL in order to
+// make it possible to group the Prometheus metrics nicely.
+func normalizeEndpoint(url string) string {
+	var nels []string
+
+	els := strings.Split(url, "/")
+	for _, e := range els {
+		// All unique IDs that we use are some hashes or integers, so we can
+		// just cut everything that's too long or looks like an int here.
+		//
+		// In the future, a better solution would be to look at the OpenAPI
+		// declaration and pass only non-parametrized parts of the query
+		// through, but that might be over-engineering.
+		isTooLong := len(e) >= 32
+		isInt := len(e) > 0 && strings.IndexFunc(e, func(c rune) bool { return c < '0' || c > '9' }) == -1
+		if isTooLong || isInt {
+			nels = append(nels, "*")
+		} else {
+			nels = append(nels, e)
+		}
+	}
+
+	return strings.Join(nels, "/")
+}
+
 // MetricsMiddleware is a middleware that measures the start and end of each request,
 // as well as other useful request information.
 // It should be used as the outermost middleware, so it can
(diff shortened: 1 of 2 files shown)


aefhm (approved):
Makes sense to me and seems like a good heuristic.

abukosek:
Thanks!

api/middleware.go
```diff
@@ -23,6 +23,31 @@ var (
...
+	els := strings.Split(url, "/")
+	for _, e := range els {
+		// All unique IDs that we use are some hashes or integers, so we can
+		// just cut everything that's too long or looks like an int here.
+		//
+		// In the future, a better solution would be to look at the OpenAPI
+		// declaration and pass only non-parametrized parts of the query
+		// through, but that might be over-engineering.
+		isTooLong := len(e) >= 32
+		isInt := len(e) > 0 && strings.IndexFunc(e, func(c rune) bool { return c < '0' || c > '9' }) == -1
```
mitjat, line 40: We also care about the following values that can be a part of the URL:
- hashes (contain letters a-f, case-insensitive)
- bech32 addresses

If there's a new endpoint in the future (e.g. hypothetically one that allows us to access e.g. ERC-20 token data by ticker symbol, meaning it takes an almost-arbitrary string as part of the URL, like `/v1/evm_tokens/DAI`), it would be better to break in a way where we over-coarsen the URL (consequence: we'll get slightly less informative grafana charts) than in a way where we under-coarse the URL (consequence: the number of timer names explodes and the entire set of prometheus timers becomes unresponsive).

abukosek: A whitelist needs to be constantly updated as new endpoints are added or existing ones are renamed, but this heuristic does not and it already cuts out all addresses and other hashes.

The `evm_tokens` example can be very simply handled here by just adding something like:
```
	if (e == "evm_tokens") {
		break;
	}
```
inside the loop and it will cut off anything after this path element.

mitjat: :facepalm: 
I somehow misinterpreted `isTooLong || isInt` as `isTooLong && isInt`. Yeah this is good, thank you. I'd probably still opt for a whitelist :), but as you say, there are tradeoffs.
 
I did find this explosion of labels, which whitelisting would solve: URLs in failed requests:
image
(there's a lot more labels hidden behind the scrollbar)

But that's all just due to bots trying bogus URLs. It's probably best not to worry about those too much, and just completely abstract away the URL (to some constant, e.g. `"ignored"`) for 4xx requests. Would you be OK to add that? Otherwise LMK. Thanks!

abukosek: :D
Sure, no problem, I'll make a PR tomorrow :)

abukosek: Implemented in https://github.com/oasisprotocol/oasis-indexer/pull/317 :)

PR #314, opened 2023-02-14
[api] add rel filter on emerald/transactions

Andrew7234:


storage/client/queries.go
```diff
@@ -394,14 +394,28 @@ func (qf QueryFactory) RuntimeBlocksQuery() string {
...
+			LEFT JOIN %[1]s.runtime_related_transactions ON %[1]s.runtime_transactions.round = %[1]s.runtime_related_transactions.tx_round 
+				AND %[1]s.runtime_transactions.tx_index = %[1]s.runtime_related_transactions.tx_index
+				AND %[1]s.runtime_transactions.runtime = %[1]s.runtime_related_transactions.runtime
+				-- When related_address ($3) is NULL and hence we do no filtering on it, avoid the join altogether.
+				-- Otherwise, every tx will be returned as many times as there are related addresses for it. 
+				AND $3::text IS NOT NULL
+			WHERE (%[1]s.runtime_transactions.runtime = '%[2]s') AND
+						($1::bigint IS NULL OR %[1]s.runtime_transactions.round = $1::bigint) AND
+						($2::text IS NULL OR %[1]s.runtime_transactions.tx_hash = $2::text) AND
+						($3::text IS NULL OR %[1]s.runtime_related_transactions.account_address = $3::text)
```
mitjat, line 415: Nit: Doesn't make much sense to cast NULLs to anything. I'm fine to leave it though if you prefer.
```suggestion
						($1 IS NULL OR %[1]s.runtime_transactions.round = $1::bigint) AND
						($2 IS NULL OR %[1]s.runtime_transactions.tx_hash = $2::text) AND
						($3 IS NULL OR %[1]s.runtime_related_transactions.account_address = $3::text)
```

mitjat (approved):
Thank you! LGTM provided you tested e2e locally.

storage/client/queries.go
Andrew7234: I'll leave it for now and change this and the rest of the queries in a follow-up MR

storage/client/queries.go
```diff
@@ -394,14 +394,28 @@ func (qf QueryFactory) RuntimeBlocksQuery() string {
 
 func (qf QueryFactory) RuntimeTransactionsQuery() string {
 	return fmt.Sprintf(`
-		SELECT round, tx_index, tx_hash, tx_eth_hash, timestamp, raw, result_raw
+		SELECT 
+				%[1]s.runtime_transactions.round,
```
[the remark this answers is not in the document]
Andrew7234: Got it, fixed. Will update the rest in a follow-up MR as well

Andrew7234:
> tested e2e locally.

Yep tested locally and confirmed that the returned transactions are associated with the `rel` address. It should be noted that for `evm.Call` transactions, the `rel` address often only appears in the events associated with that transaction and is not displayed within the transaction itself (eg as `sender` or `to`). This can be checked by querying `emerald/events?tx_hash={some_tx_hash}` and inspecting the returned events for the `rel` address

storage/client/queries.go
Andrew7234: Ergh I didn't test this thoroughly; removing the typecasts causes errors of the flavor `"msg": "storage error: ERROR: could not determine data type of parameter $1 (SQLSTATE 42P08)"`

Reverted in #315

PR #316, opened 2023-02-14
EVM tokens: fix missing token type

pro-wh:
this had resulted in the database having zeros in the type column, making the mutate workflow fail when it sees an unknown token type


pro-wh:
image

sample db built after the correction, token_type now correctly says 20

mitjat (approved):
:tada:

PR #317, opened 2023-02-15
metrics: Ignore 4xx results

abukosek:
Group all 4xx results under the same "ignored" label, as they're mostly just bots and we're not interested in those requests for metrics anyway.

### DIFF
--- a/api/middleware.go
+++ b/api/middleware.go
@@ -63,8 +63,8 @@ func MetricsMiddleware(m metrics.RequestMetrics, logger log.Logger) func(next ht
 				"request_id", requestID,
 			)
 			t := time.Now()
-			timerName := normalizeEndpoint(r.URL.Path)
-			timer := m.RequestTimer(timerName)
+			metricName := normalizeEndpoint(r.URL.Path)
+			timer := m.RequestTimer(metricName)
 
 			// Serve the request.
 			next.ServeHTTP(w, r.WithContext(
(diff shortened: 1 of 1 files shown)


api/middleware.go
```diff
@@ -73,17 +73,30 @@ func MetricsMiddleware(m metrics.RequestMetrics, logger log.Logger) func(next ht
...
-			timer.ObserveDuration()
 
 			statusTxt := "failure"
 			if httpStatus >= 200 && httpStatus < 400 {
 				statusTxt = "success"
+			} else if httpStatus >= 400 && httpStatus < 500 {
+				// Group all 4xx errors into the "ignored" label, as they're
+				// almost always just bots and are not relevant to the metrics
+				// we're interested in.
+				statusTxt = "ignored"
```
mitjat, line 98: Let's do `failure_4xx` or something here. We do know the status, we're only ignoring the URL.

api/middleware.go
```diff
@@ -73,17 +73,30 @@ func MetricsMiddleware(m metrics.RequestMetrics, logger log.Logger) func(next ht
...
+				timerName = "ignored"
```
mitjat, line 99: nit: Can you rename this to `metricName` or (perhaps better) `normalizedURL` or similar? It was already weird that we were using `timerName` to name a counter, and it's doubly weird now that we're using this variable more actively.

mitjat (approved):
Thank you!

abukosek:
Thanks, fixed :)

PR #318, opened 2023-02-16
storage: add eth runtime source storage impl 

pro-wh:
if things get too weird with deploying an SGX-enabled node with exposed gRPC, here's this

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -24,6 +24,7 @@ require (
 )
 
 require (
+	github.com/StackExchange/wmi v0.0.0-20180116203802-5d049714c4a6 // indirect
 	github.com/apapsch/go-jsonmerge/v2 v2.0.0 // indirect
 	github.com/beorn7/perks v1.0.1 // indirect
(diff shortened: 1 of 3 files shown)


go.mod
```diff
@@ -25,6 +25,7 @@ require (
 )
 
 require (
+	github.com/StackExchange/wmi v0.0.0-20180116203802-5d049714c4a6 // indirect
```
mitjat, line 28: Do you know why all the new deps?

mitjat (commented):
Oh nice, I didn't realize you got that far along with cleaning this up. We should have started pushing for gRPC access sooner, might have avoided some code churn :|. It was a bad combo of Gregor's and my OOO one after another.

I think no need to merge until we know we need this? But it's nice that it's ready.

go.mod
pro-wh: should be from ethclient

pro-wh:
#305 is moved over to a proper sapphire node

PR #320, opened 2023-02-17
 runtimes: add related transactions index by tx 

pro-wh:
in #310, we want to look up related addresses by transaction. add an index so that we can quickly get those and pull the address preimages

also some indexes didn't get the `runtime` column added when we unified the runtime tables in #296. adding it


Andrew7234 (approved):

storage/migrations/02_oasis_3_runtimes.up.sql
```diff
@@ -73,7 +73,7 @@ CREATE TABLE oasis_3.runtime_related_transactions
   tx_index        UINT31 NOT NULL,
   FOREIGN KEY (runtime, tx_round, tx_index) REFERENCES oasis_3.runtime_transactions(runtime, round, tx_index) DEFERRABLE INITIALLY DEFERRED
 );
-CREATE INDEX ix_runtime_related_transactions_address_height_index ON oasis_3.runtime_related_transactions (account_address, tx_round, tx_index);
```
aefhm, line 76: Ohh, I didn't know that you could get transitive properties in indices.

PR #322, opened 2023-02-21
Add `/runtime/accounts/{addr}`; add stats to runtime account endpoint

### LINKED ISSUE #325: Add runtime account stats (totalSent, totalReceived, numTxns)
https://app.clickup.com/t/861mccnhp

(We are going with the alternative option)

Andrew7234:
Resolves #325 

Open questions:
- We'd like to support querying runtime accounts by the runtime address (eth address) as well as the consensus address. Whether or not to scope that into the MVP/this ticket is up for debate.
- Do we want to support multiple oasis-sdk balances per runtime? Doing so slightly complicates the schema (need a list of `runtimeSdkBalance`). AFAIK we only currently use 'ROSE' as the native denomination and have no plans to add more. I'd prefer to keep it simple if possible.
- Do we want to include `AccountStats` for consensus accounts if we can only show the `numTxs` for them? As Mitja pointed out, we cannot show total sent/received bc we have no consensus transfers table

Edit** 
Impl open questions:
- ~~If there is no matching entry in the `address_preimages` table, we have not seen this address in runtimes before and return 404; is this alright? Before, in the old `/consensus/account/{addr}` endpoint, we'd ignore this and leave default values because we did a full outer join and an account could only have consensus data. On the frontend, we def want to display a zero-valued account; but I'm not as sure what the backend api should display.~~
- Edit*** After thinking more I'm in favor of not doing 404s and instead returning zero-valued account objects. Better to be consistent with the frontend. (and consensus endpoint)
- Do we want a generic `/{runtime}/accounts` endpoint similar to the consensus one?

Note: For both runtime_balances and evm_balances, we now return an empty list instead of making the entire field nullable (and removing the key). I believe this was was originally nullable for the same reasons as the above point (full outer join; the addr might be valid but have no runtime data). I prefer to make this change since it more accurately describes the account state; it's probably more ergonomic for frontend folks, and it mirrors what we do in `consensus/account/{addr}` for `allowances`. But lmk if anyone feels strongly otherwise.

OTOH, if we want to implement `/emerald/accounts` we may want to make `evm_balances` optional for performance reasons.

### DIFF
--- a/api/spec/v1.yaml
+++ b/api/spec/v1.yaml
@@ -895,6 +895,26 @@ paths:
                 $ref: '#/components/schemas/EvmTokenList'
         <<: *common_error_responses
 
+  /{runtime}/accounts/{address}:
+    get:
+      summary: Returns a runtime account.
+      parameters:
+        - *runtime
+        - in: path
+          name: address
+          required: true
+          schema:
+            <<: *StakingAddressType
+          description: The staking address of the account to return.
+      responses:
+        '200':
+          description: A JSON object containing a runtime layer account.
+          content:
+            application/json:
+              schema:
+                $ref: '#/components/schemas/RuntimeAccount'
+        <<: *common_error_responses
+
   /{layer}/stats/tx_volume:
     get:
       summary: |
@@ -1490,25 +1510,14 @@ components:
             is the Ethereum address (in base64, not hex!).
           example: 'INLp2Ih3YIdcA+zFNhM+SIGyFgKsYYc9SKQeKRKe2uI='
 
-    RuntimeName:
-      description: |
-        The name of a runtime. This is a human-readable identifier, and should
-        stay stable across runtime upgrades/versions.
-      type: string
-      enum:
-        - emerald
-      example: emerald
-
     RuntimeSdkBalance:
       description: Balance of an account for a specific runtime and oasis-sdk token (e.g. ROSE).
       type: object
-      required: [balance, runtime, token_symbol, token_decimals]
+      required: [balance, token_symbol, token_decimals]
       properties:
         balance:
           <<: *BigIntType
           description: Number of tokens held, in base units.
-        runtime:
-          $ref: "#/components/schemas/RuntimeName"
         token_symbol:
           type: string
           description: The token ticker symbol. Unique across all oasis-sdk tokens in the same runtime.
@@ -1521,13 +1530,11 @@ components:
     RuntimeEvmBalance:
       description: Balance of an account for a specific runtime and EVM token.
       type: object
-      required: [balance, runtime, token_contract_addr, token_decimals]
+      required: [balance, token_contract_addr, token_decimals]
       properties:
         balance:
           <<: *BigIntType
           description: Number of tokens held, in base units.
-        runtime:
-          $ref: "#/components/schemas/RuntimeName"
         token_contract_addr:
           type: string
           description: The EVM address of this token's contract. Encoded as a lowercase hex string.
@@ -1553,29 +1560,11 @@ components:
           type: string
           description: The staking address for this account.
           example: *staking_address_1
-        address_preimage:
-          $ref: '#/components/schemas/AddressPreimage'
         nonce:
           type: integer
           format: int64
           description: A nonce used to prevent replay.
           example: 0
-        runtime_sdk_balances:
-          description: |
-            The balances of this account in each runtime, as managed by oasis-sdk.
-            NOTE 1: This field is omitted for efficiency when listing multiple accounts.
-            NOTE 2: This field is limited to 1000 entries. If you need more, please let us know in a GitHub issue.
-          type: array
-          items:
-            $ref: '#/components/schemas/RuntimeSdkBalance'
-        runtime_evm_balances:
-          description: |
-            The balances of this account in each runtime, as managed by EVM smart contracts (notably, ERC-20).
-            NOTE 1: This field is omitted for efficiency when listing multiple accounts.
-            NOTE 2: This field is limited to 1000 entries. If you need more, please let us know in a GitHub issue.
-          type: array
-          items:
-            $ref: '#/components/schemas/RuntimeEvmBalance'
         available:
           <<: *BigIntType
           description: The available balance, in base units.
@@ -2008,7 +1997,7 @@ components:
           type: string
           description: |
             A reasonable "amount" associated with this transaction, if
-            applicable. The meaning varies based on the transaction mehtod.
+            applicable. The meaning varies based on the transaction method.
             Usually in native denomination, ParaTime units. As a string.
           example: "100000001666393459"
         success:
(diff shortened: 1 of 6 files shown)


mitjat:
Thank you! Wrapping up for today so I won't look at the code right now, just replying:

> * We'd like to support querying runtime accounts by the runtime address (eth address) as well as the consensus address. Whether or not to scope that into the MVP/this ticket is up for debate.

I'd say not in he MVP. It's more important that it can be done from the Explorer GUI, and that's already enabled; frontend converts teh eth address to oasis format and queries us with that. Def not this PR. I think we do have a ticket for this somewhere already, but not sure.

> * Do we want to support multiple oasis-sdk balances per runtime? Doing so slightly complicates the schema (need a list of `runtimeSdkBalance`). AFAIK we only currently use 'ROSE' as the native denomination and have no plans to add more. I'd prefer to keep it simple if possible.

Yeah it's a little silly right now, and might even stay silly forever, but for this one I think it's good to use the same level of generality as the accounts module. If at some point a runtime appears that does use multiple denominations, we'd have to either break the API (balance type changes from scalar to dict; we don't want that) or present the non-native balances in some really awkward way. So to me the risk (of a simplified struct here) outweighs the benefits. 

> * Do we want to include `AccountStats` for consensus accounts if we can only show the `numTxs` for them? As Mitja pointed out, we cannot show total sent/received bc we have no consensus transfers table

I just meant it's not completely trivial to add that stat, but it's still not hard for us to add a consensus transfers table and then use it. I think let's do it in a separate PR. It might be nice to have the transfers in a table anyway for debugging.

api/spec/v1.yaml
```diff
@@ -2024,6 +2038,22 @@ components:
...
+      required: [num_txns]
+      properties:
+        total_sent:
+          description: The total number of tokens sent, in base units.
+          <<: *BigIntType
+        total_received:
+          description: The total number of tokens received, in base units.
+          <<: *BigIntType
+        num_txns:
+          description: The total number of transactions this account was involved with.
```
mitjat, line 2053:
```suggestion
          description: The total number of transactions this account submitted.
```

I'd say that's more indicative of account activity?

api/spec/v1.yaml
```diff
@@ -1959,6 +1949,30 @@ components:
...
+      type: object
+      required: [address, address_preimage, runtime, sdk_balance, stats]
+      properties:
+        address:
+          type: string
+          description: The staking address for this account.
+          example: *staking_address_1
+        address_preimage:
+          $ref: '#/components/schemas/AddressPreimage'
+        sdk_balance:
```
mitjat, line 1962: I know I named this, but now I'm wondering if it's too technical. Runtime users don't know about the SDK. Is `runtime_balance` better, or too vague? Does anybody have other ideas?

api/spec/v1.yaml
```diff
@@ -2024,6 +2038,22 @@ components:
             The number of addresses that have a nonzero balance of this token,
             as calculated from Transfer events.
           example: 123
+    
+    AccountStats:
+      type: object
+      required: [num_txns]
```
mitjat, line 2044: Why not also the total_* fields?

mitjat (commented):
Thanks, the API changes LGTM also! Small comments only.

api/spec/v1.yaml
Andrew7234: Hm I'm ambivalent on this one. I agree that submitted txs are perhaps more interesting, but I looked at a couple other block explorers and while most don't have this field, the ones that do track transactions that an account was involved with, [example](https://www.blockchain.com/explorer/addresses/btc/bc1q6ufjmqnzz0ck2mvdx5t6yq5enguz86ur29wqhx7wn7dlg5j5cg0swy9wfl)

api/spec/v1.yaml
Andrew7234: I like `runtime_balance` better; changed to that for now. I'd be ok using `balance` as well

api/spec/v1.yaml
Andrew7234: Oh this was in case we added the stats to consensus accounts but couldn't derive the sent/received amounts easily. Since we'll add those in a different PR, I'll make all 3 fields required for now.

api/spec/v1.yaml
aefhm: +1 to ambivalence. I'd think block explorers generally want a higher activity stat.

aefhm:
> Edit*** After thinking more I'm in favor of not doing 404s and instead returning zero-valued account objects. Better to be consistent with the frontend. (and consensus endpoint)

I can go either way. The consistency is nice (a better ux), but I think the distinction is clear either way.

> Do we want a generic /{runtime}/accounts endpoint similar to the consensus one?

I don't see a need for this query.

mitjat:
> > Edit*** After thinking more I'm in favor of not doing 404s and instead returning zero-valued account objects. Better to be consistent with the frontend. (and consensus endpoint)
> 
> I can go either way. The consistency is nice (a better ux), but I think the distinction is clear either way.

+1 for keeping it consistent. Also philosophically/technically, all accounts with valid addresses do "exist" from day 0. If not, does account creation happen when sb generates its private key (and derives the address)? If so, what if I send tokens to a random address for which nobody has the key? I'm struggling to phrase/explain nicely. But it's murky waters, I feel :)

> > Do we want a generic /{runtime}/accounts endpoint similar to the consensus one?
> 
> I don't see a need for this query.

Hm what will the Explorer use to show the list of accounts? The `/consensus/accounts` endpoint won't be much use, especially now that we've removed all runtime data from the `[Runtime]Account` struct. Edit: I see there _is_ no "list of accounts" screen in the MVP. I'm surprised, so I asked for more info in Slack.

For the sake of this PR: looks like we don't need it now, and I think it's likely we'll need it at some point. I'd lean towards just doing; now that you're in the middle of this code, it will be fast. And it feels like a basic indexer feature, in as much as we want to think of the indexer as a product separate from the explorer.
But I'm also OK with leaving it out, your call.

mitjat (commented):
This pretty much LGTM, thank you!
It does look like it needs a PR or commit to implement the runtime_transfers table.

storage/client/queries.go
```diff
@@ -434,6 +426,27 @@ func (qf QueryFactory) RuntimeEventsQuery() string {
...
+		SELECT context_identifier, context_version, address_data
+			FROM %[1]s.address_preimages
+			WHERE address = $1::text`, qf.chainID)
+}
+
+func (qf QueryFactory) RuntimeAccountStatsQuery() string {
+	return fmt.Sprintf(`
+		SELECT
+			COALESCE (
+				(SELECT sum(amount) from %[1]s.runtime_transfers where sender=$1::text)
```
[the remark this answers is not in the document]
Andrew7234: Ah actually Warren added it already; it's populated by looking at `accounts` module transfers/burns/mints. I tested it manually as well.

api/spec/v1.yaml
```diff
@@ -1959,6 +1948,32 @@ components:
...
+      properties:
+        address:
+          type: string
+          description: The staking address for this account.
+          example: *staking_address_1
+        address_preimage:
+          $ref: '#/components/schemas/AddressPreimage'
+        runtime_balances:
+          description: |
+            The balances of this account, as managed by the oasis-sdk.
```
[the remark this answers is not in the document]
Andrew7234: Thanks for the suggestion, I like it much better. Updated!

PR #324, opened 2023-02-23
Andrew7234/pagination total

### LINKED ISSUE #323: Paginated endpoints: Return total (up to a limit) number of results
[Clickup issue](https://app.clickup.com/t/861mcbbhg)

Andrew7234:
Resolves #323


storage/client/client.go
```diff
@@ -25,6 +26,8 @@ import (
 const (
 	blockCost = 1
 	txCost    = 1
+
+	maxTotalCount = 2000
```
aefhm, line 30: I'd guess we also expose this in documentation?

api/spec/v1.yaml
```diff
@@ -929,6 +929,20 @@ components:
       type: string
       enum: [emerald, sapphire, cipher]
 
+    List:
+      type: object
+      required: [total_count, is_total_count_clipped]
+      properties:
+        total_count:
```
aefhm, line 936: I also wonder if the block height at query time is important enough to return.

mitjat: Interesting question, it applies to almost every single response we generate. I guess "not for now" -- it would be clumsy to surface this in the UI, or to force the UI to refetch all data if it performs multiple fetches from the indexer and they end up being generated at different heights.

storage/client/client.go
```diff
@@ -85,6 +94,46 @@ func wrapError(err error) error {
 	return apiCommon.ErrStorageError{Err: err}
 }
 
+// Returns the rows for a given query as well as the total count of matching records
+// for queries that return multiple rows.
```
mitjat, line 98:
```suggestion
// For queries that return multiple rows, returns the rows for a given query, as well as
// the total count of matching records, i.e. the number of rows the query would return
// with limit=infinity.
```

(Putting "For queries" first to make it clear which part of the sentence it refers to; explaining how "total count" is different from len(rows))

I think it would be useful to be similarly explicit/wordy in openapi because it's a subtle distinction.

mitjat (approved):
Thank you, looks nice! I'm open to renaming `is_total_count_clipped`, I agree it's long. OTOH I feel the "X -> is_X_clipped" pairing is good and not too verbose. So a shorter X != total_count would probably be best, but I have a hard time coming up with something that is not (even more) confusable with "the number of rows".

storage/client/client.go
Andrew7234: Nice, updated!

storage/client/client.go
Andrew7234: Going to leave the actual 1000 out of the documentation for now; if we get lots of requests from users asking about it we can update it.

PR #326, opened 2023-02-23
refactor: Initial support for Cobalt nodes

mitjat:
This PR brings architectural support for Cobalt nodes, but implements it for a single endpoint (`consensus.GetGenesisDocument()`) only, to showcase the changes. More implementation is TODO in further PRs once we settle on this approach.

Main changes:
- Adds a helper script for cloning the types from an arbitrary old version of oasis-core into the indexer codebase. See the long comment at the start of `vendor_oasis_core.sh`. 
- Copies and checks in those types for Cobalt. **The entire `coreapi/` folder is autogenerated**, don't spend much time on it.
- Implements conversions from Cobalt types to indexer-internal(*) types (for the genesis `Document` type only, but that's probably as complex as structs get).
- Introduces a new layer of abstraction in node-accessing: We used to use the [`consensus.ClientBackend`](https://github.com/oasisprotocol/oasis-core/blob/32c31ea781301f782e0e3588f08fc28d1aa1fcd5/go/consensus/api/api.go#L93-L93) interface from oasis-core to send RPCs to the node. Now there's a new `ConsensusApiLite` interface that provides the same functionality, but supports multiple node versions: For Cobalt, it uses the grpc library directly, and for Damask, it reuses the convenience of  `consensus.ClientBackend`.

(*) indexer-internal types are currently just the types from the current oasis-core. We should probably change that, otherwise going from Damask to oasis-4 will be pretty painful. Later though; this will require changes throughout the indexer.

Testing:
- If the .yml config specifies a damask node, everything works as before.
- If it specifies a cobalt node, we successfully fetch and decode the genesis doc; things fail from there on because we dont' have oasis-2 tables. To be fixed in later PRs.

Note on naming:
The core folks in particular have been adamant that it's a bad idea to have "cobalt" and "damask" float around the codebase, and that we should use numeric versions like v21.1.1. I did follow that naming suggestion in `coreapi/v21.1.1` where the types are literally cloned from oasis-core v21.1.1. But for internal variables, interfaces etc, I strongly prefer `cobalt` to `21_1_1` or similar; it's much more readable and memorable. We *might* need to get a little creative at some point if testnet has a version that is CBOR-incompatbile with any of the official named releases on mainnet (testnet releases are not named), but that's fine, we can create a name and a comment, it will still be more readable. 

I will outline the remaining TODOs for supporting multiple versions in a small doc, to get a high level pre-review from everyone and to give you context for future PRs.


aefhm:
> we can create a name and a comment, it will still be more readable.

"name" being say Damask-test?

aefhm:
Would you say that the type conversion process to internal type was okay/okay-ish?

scripts/vendor-oasis-core/remove_func.py
```diff
@@ -0,0 +1,58 @@
...
+
+lines = sys.stdin.readlines()
+out = []
+
+def find_end(lines, start):
+  for i, line in enumerate(lines[start:]):
+    if line.startswith("}") or line.startswith(")"):
+      # if line != "}\n":
+      #   sys.stderr.write("Warning: } not on its own line\n")
+      return i + start
```
Andrew7234, line 24: unused?

scripts/vendor-oasis-core/remove_func.py
```diff
@@ -0,0 +1,58 @@
...
+in_func = False  # are we inside a function or other block we want to skip
+for i, line in enumerate(lines):
+  if in_func and i <= end:
+    continue
+  else:
+    in_func = False
+
+  if line.startswith("func"):
+    # We'll remove functions in general, but we'll keep the (un)marshalers
+    if "Marshal" in line or "Unmarshal" in line or re.match(r"^func \([^)]+\) String\(\).*", line) or "func (s SlashReason) checkedString()" in line:
```
Andrew7234, line 35: What's `func (s SlashReason) checkedString()`?

scripts/vendor-oasis-core/vendor_oasis_core.sh
```diff
@@ -0,0 +1,118 @@
+#!/bin/bash
+
+# This script vendors (=clones) type definitions from a given version of
+# oasis-core into a local directory, coreapi/$VERSION.
+#
+# The intention is to use them to communicate with archive nodes that use
+# old oasis-core. (The indexer needs to be able to talk to archive nodes
+# and the curernt node; ee cannot directly include multiple versions of
```
Andrew7234, line 8:
```suggestion
# and the current node; ie cannot directly include multiple versions of
```

storage/oasis/consensus.go
```diff
@@ -4,55 +4,35 @@ import (
...
-
-// StakingGenesis returns the staking genesis document at the provided height.
-func (cc *ConsensusClient) StakingGenesis(ctx context.Context, height int64) (*stakingAPI.Genesis, error) {
-	return cc.client.Staking().StateToGenesis(ctx, height)
-}
-
-// SchedulerGenesis returns the scheduler genesis document at the provided height.
-func (cc *ConsensusClient) SchedulerGenesis(ctx context.Context, height int64) (*schedulerAPI.Genesis, error) {
-	return cc.client.Scheduler().StateToGenesis(ctx, height)
-}
```
Andrew7234, line 51: These were unused?

mitjat:
> > we can create a name and a comment, it will still be more readable.
> "name" being say Damask-test?

IDK, we'll see when the time comes, if it ever does. Possibly no "test" in the name (?); it's a version in its own right, and could be later deployed on mainnet as well. Maybe damask++ or damaskoid or even something unrelated like popeye. As long as it's documented.

> Would you say that the type conversion process to internal type was okay/okay-ish?

I guess so. Tedious but with no surprises. vscode copilot was very helpful.

Andrew7234 (approved):
Nice! Would be nice if another set of eyes could go over the `convert.go` section though

pro-wh (commented):
convert.go looks pretty un-crazy :+1: if you'd like to count this as having been reviewed additionally

mitjat:
Thanks both! Being back in the US will take a little adjusting. I was completely taken aback -- "they reviewed it WHILE I'M  WORKING? And TWO reviews? They should be asleep!". Also I inadvertently blew off a 1:1 yesterday because it happened at an entirely wrong time of day :)

PR #327, opened 2023-02-23
include codegen files in docker image

Andrew7234:
In support of https://github.com/oasislabs/private-charts/pull/552

We now copy the repo _after_ `make codegen-go` has been run in the build process.

Without the codegen files; the statecheck tests fail. We already copy the repo into the dockerfile, so the added size of the generated files is marginal.


mitjat (approved):
Thank you!

PR #328, opened 2023-02-24
log: automatically include caller and timestamp

ptrus:
Changes:
```
{"analyzer":"consensus","height":8052362,"level":"info","module":"analysis_service","msg":"processed block"}
```

To:
```
{"analyzer":"consensus","caller":"consensus.go:173","height":8052362,"level":"info","module":"analysis_service","msg":"processed block","ts":"2023-02-24T19:27:30.720709236Z"}
```

Seems like a useful thing to have.

### DIFF
--- a/log/log.go
+++ b/log/log.go
@@ -19,31 +19,40 @@ type Logger struct {
 
 // NewDefaultLogger initializes a new logger instance with default settings.
 func NewDefaultLogger(module string) *Logger {
-	return &Logger{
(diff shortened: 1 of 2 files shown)


mitjat (approved):
Thanks! We were OK without those because the caller is usually clear from the message, and the timestamp is added by GCP (though I've wished for it on occasion when running locally). But I agree, they're both good to have.

Andrew7234 (approved):

PR #329, opened 2023-02-24
stats: Implement Daily active accounts statistics

ptrus:
This PR adds  `/{layer}/stats/active_accounts` endpoint, which returns a timeline with daily number of active accounts per layer.

The API only supports daily buckets (the buckets in response are always 86400 seconds big), but it offers overlapping buckets (e.g. 5-minute bucket granularity as in the example below). The window size is fixed to 1 day, while the window step is configurable via config - 5 or 10 minute is likely ok to use in production deployment.

Example responses (on a minimally synced mainnet indexer):
```
curl http://localhost:8008/v1/consensus/stats/active_accounts?limit=5&window_step_seconds=300
{
    "window_size_seconds": 86400,
    "windows": [
        {
            "active_accounts": 1251,
            "window_end": "2022-04-16T02:35:00Z"
        },
        {
            "active_accounts": 1246,
            "window_end": "2022-04-16T02:30:00Z"
        },
        {
            "active_accounts": 1243,
            "window_end": "2022-04-16T02:25:00Z"
        },
        {
            "active_accounts": 1242,
            "window_end": "2022-04-16T02:20:00Z"
        },
        {
            "active_accounts": 1243,
            "window_end": "2022-04-16T02:15:00Z"
        }
    ]
}

curl http://localhost:8008/v1/consensus/stats/active_accounts?window_step_seconds=30
{"msg":"Invalid format for parameter window_step_seconds: invalid value: 30"}

curl http://localhost:8008/v1/consensus/stats/active_accounts
{
    "window_size_seconds": 86400,
    "windows": [
        {
            "active_accounts": 1230,
            "window_end": "2022-04-16T00:00:00Z"
        },
        {
            "active_accounts": 1326,
            "window_end": "2022-04-15T00:00:00Z"
        },
        {
            "active_accounts": 1367,
            "window_end": "2022-04-14T00:00:00Z"
        },
        {
            "active_accounts": 1119,
            "window_end": "2022-04-13T00:00:00Z"
        },
        {
            "active_accounts": 666,
            "window_end": "2022-04-12T00:00:00Z"
        }
    ]
}

curl http://localhost:8008/v1/emerald/stats/active_accounts?limit=5
{
    "window_size_seconds": 86400,
    "windows": [
        {
            "active_accounts": 3569,
            "window_end": "2022-04-15T00:00:00Z"
        },
        {
            "active_accounts": 3852,
            "window_end": "2022-04-14T00:00:00Z"
        },
        {
            "active_accounts": 3776,
            "window_end": "2022-04-13T00:00:00Z"
        },
        {
            "active_accounts": 2418,
            "window_end": "2022-04-12T00:00:00Z"
        }
    ]
}
```

TODO:
- [x] Run the updated analyzer on mainnet and sanity check the results
- [x] Likely some code cleanup


storage/client/queries.go
```diff
@@ -534,3 +534,15 @@ func (qf QueryFactory) TxVolumesQuery() string {
 		OFFSET $3::bigint
 	`
 }
+
+func (qf QueryFactory) DailyActiveAccountsQuery() string {
+	return `
+		SELECT window_end - INTERVAL '1 DAY' AS bucket_start, active_accounts
```
ptrus, line 540: This `window_end - INTERVAL '1 DAY'` part could be avoided if the analyzer backend and the API used unified representation for buckets  

BUT:
- it made more sense to define buckets with `window_end` (instead of `window_start`) within the analyzer backend code (personally I think it would make the analyzer code a bit less readable if window_start would be used)
- on the other hand, the existing stats API (`stats/tx_volume`) defines buckets with start time already, and it is nice to use the unified API for `stats/active_accounts` as well
- since API and the analyzer are nicely decoupled already, I think that using a separate representation and handling the conversion here is ok/better than alternative of updating the analyzer code.

config/local-dev.yml
```diff
@@ -7,7 +7,13 @@ analysis:
     metadata_registry:
       interval: 1h
     aggregate_stats:
-      interval: 10m
+      tx_volume_interval: 10m
+      daily_active_accounts_window_step: 5m
+      daily_active_accounts_layers:
```
ptrus, line 12: Not sure if it even makes sense for these to be configurable, could always index all layers.

mitjat (commented):
Looks great, thank you!
I much appreciate the comments sprinkled everywhere, and the clean refactoring of existing code where needed.

I have some comments/suggestions, but nothing major.

analyzer/aggregate_stats.go
```diff
@@ -79,3 +126,143 @@ func (a *AggregateStatsAnalyzer) writeToDB(ctx context.Context, batch *storage.Q
...
+				latestStatsTs = nextWindow
+			}
+			if err := a.writeToDB(ctx, batch, fmt.Sprintf("update_daily_active_account_%s_stats", layer)); err != nil {
+				a.logger.Error("failed to insert daily active account stats update", "err", err, "layer", layer)
+			}
+			a.logger.Info("updated daily active account stats", "layer", layer, "num_inserts", batch.Len())
+		}
+
+		select {
+		case <-time.After(windowStep):
```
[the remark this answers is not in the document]
ptrus: Makes sense, although I'm not totally sure about a good tolerance value to use.

In order to compute stats for a window, there needs to be at least one block indexed in the window after (since the analyzer tracks time by looking at timestamps of indexed blocks).
`1s` is surely too small, since consensus blocks are generated every ~6 seconds, therefore we could miss the first block in the new window and not start analyzing the previous window. There is also some lag, so I would use something like 10 seconds.

There are also two cases where the next iteration should be done sooner:
- batch limit reached: the analyzer is way behind latest indexed block, start next iteration sooner
- no stats computed: we likely hit the issue described above (no blocks indexed) retry sooner

---

All in all, I think the best solution is to just use a short constant here (e.g. 1 min), not related to the `windowStep` since the iteration itself is quite lightweight if there's no new data (one latest stats + one latest block query for each layer).

api/spec/v1.yaml
```diff
@@ -921,6 +921,31 @@ paths:
                 $ref: '#/components/schemas/TxVolumeList'
         <<: *common_error_responses
 
+  /{layer}/stats/active_accounts:
+    get:
+      summary: |
+        Returns a (sliding) timeline of the recorded daily unique active accounts for
+        either consensus or one of the paratimes.
+      parameters:
```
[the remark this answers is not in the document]
ptrus: Will add. Although I think what you mean is actually `window_step_seconds`? Because the window size will currently always be 24 hours, even if the step is smaller - we only compute the unique accounts over the 24 hours intervals.

> Hum, makes me realize we don't probably limit the size of the output in any way right now. Oh well, we can fix that later.

Yeah, likely good idea to handle this in a unified way for all paginated endpoints.

config/local-dev.yml
```diff
@@ -7,7 +7,13 @@ analysis:
     metadata_registry:
       interval: 1h
     aggregate_stats:
-      interval: 10m
+      tx_volume_interval: 10m
+      daily_active_accounts_window_step: 5m
```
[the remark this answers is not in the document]
ptrus: I thought about this, and I didn't find it that weird. The indexer would start at `last_analyzed_slot + floor(current_step)` unless I missed something.   

So for example, changing from `5m` to `10m` there would be a point in DB where the time between recorded stats would start being `10m`, previously `5m`.  Since all the API response buckets are timestamped, and the bucket size remains the same, I imagine the consumers would likely be minimally affected by such a change.


However, I'm also not against just hardcoding this.

config/local-dev.yml
ptrus: Yeah, the code should handle this just fine (only one `info` log message per missing layer per iteration would be emitted).

Will probably remove this from config.

storage/client/queries.go
ptrus: Agreed on all points, will update.

analyzer/queries.go
```diff
@@ -519,3 +524,93 @@ func (qf QueryFactory) RefreshMin5TxVolumeQuery() string {
 		REFRESH MATERIALIZED VIEW stats.min5_tx_volume
 	`
 }
+
+// LatestDailyAccountStatsQuery returns the query to get the timestamp of the latest daily active accounts stat.
+func (qf QueryFactory) LatestDailyAccountStatsQuery() string {
+	return `
+		SELECT
```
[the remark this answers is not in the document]
ptrus: Updated the query formatting (also for some of the ones below).

config/local-dev.yml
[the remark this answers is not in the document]
ptrus: Actually, by adding support for `window_step_seconds` to the API and this also effecting what can be queried, maybe moving this into a constant does indeed make more sense.

ptrus:
Thanks for the thoughtful review, I think I addressed all the comments, please take another look @mitjat

storage/migrations/03_agg_stats.up.sql
```diff
@@ -45,6 +45,19 @@ CREATE MATERIALIZED VIEW stats.daily_tx_volume AS
...
+(
+  layer           TEXT NOT NULL,
+  window_end      TIMESTAMP WITH TIME ZONE NOT NULL,
+  active_accounts uint63 NOT NULL,
+
+  PRIMARY KEY (layer, window_end)
+);
+-- Index for efficient query of the daily samples:
+-- WHERE (layer == 'layer' AND window_end AT TIME ZONE 'UTC')::time = '00:00:00').
+CREATE INDEX ix_stats_daily_active_accounts_window_end_time ON stats.daily_active_accounts (layer, ((window_end AT TIME ZONE 'UTC')::time));
```
mitjat, line 60: Does this end up getting used? I think it's better to go with

```suggestion
CREATE INDEX ix_stats_daily_active_accounts_window_end ON stats.daily_active_accounts (layer, window_end);
CREATE INDEX ix_stats_daily_active_accounts_daily_windows ON stats.daily_active_accounts (layer, window_end) WHERE ((window_end AT TIME ZONE 'UTC')::time = '00:00:00'));
```

The first index matches `WHERE` and the `ORDER BY` and will help the db walk the rows in order. Otherwise, it has to do a full table scan and then sort the rows.
The second index is the same, but for the daily query. Compared to the current query: Using `WHERE` instead of indexing the `::time` keeps the index smaller. Also, indexing the `window_end` makes it so the db an just walk the index and obtain sorted rows, instead of using the index to filter on 00:00:00 and then having to sort separately.

tests/e2e_regression/run.sh
```diff
@@ -52,6 +52,9 @@ testCases=(
   'tx_volume                /v1/consensus/stats/tx_volume'
   'bucket_size              /v1/consensus/stats/tx_volume?bucket_size_seconds=300'
   'nonstandard_bucket_size  /v1/consensus/stats/tx_volume?bucket_size_seconds=301'
+  'active_accounts          /v1/consensus/stats/active_accounts'
+  'active_accounts_window   /v1/consensus/stats/active_accounts?window_step_seconds=300'
+  'active_accounts          /v1/consensus/stats/active_accounts'
```
mitjat, line 57: Nit: duplicate.

FYI, we're not using this test in any automated way yet, but we expect to.

storage/client/client.go
```diff
@@ -1257,3 +1257,51 @@ func (c *StorageClient) TxVolumes(ctx context.Context, layer apiTypes.Layer, p a
...
+		return nil, wrapError(err)
+	}
+	defer rows.Close()
+
+	ts := DailyActiveAccountsList{
+		WindowSizeSeconds: 86400, // Day.
+		Windows:           []apiTypes.ActiveAccounts{},
+	}
+	for rows.Next() {
+		var d struct {
```
mitjat, line 1288: Optional: `d` could be directly `apiTypes.ActiveAccounts`. You'd just need to modify the query to do the UTC conversion in postgres:
`SELECT window_end AT TIME ZONE 'UTC' AT TIME ZONE 'UTC'`

That said, the syntax is pretty stupid. The first `AT TIME ZONE` converts to a timestamp without timezone, the second one re-adds the UTC timezone. See [docs](https://www.postgresql.org/docs/current/functions-datetime.html#FUNCTIONS-DATETIME-ZONECONVERT). Your call whether we should go for it.

analyzer/aggregate_stats.go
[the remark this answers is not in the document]
mitjat: :+1: Good thoughts on the tolerance, and the other two cases! I agree with your thinking; a constant wait is probably good enough for out scenario, so keeping it simple is the way to go. We can always tweak/complicate later if needed, not that I expect we'll need to.

mitjat (approved):
:tada: 

Only very minor comments left.

storage/migrations/03_agg_stats.up.sql
ptrus: The first index already exists, due to `(layer, window_end)` being the primary key, that's why I didn't define it explicitly.

The second index did also end up being used for the daily query, but your suggestion is much better indeed. I completely forgot about the option of using a `where` filter in the index.

tests/e2e_regression/run.sh
ptrus: Right, this was supposed to be a query for `emerald` :-) Will update, thanks!

api/spec/v1.yaml
```diff
@@ -2149,6 +2187,36 @@ components:
           description: The transaction volume on this day.
           example: 420
 
+    ActiveAccountsList:
+      type: object
+      required: [window_size_seconds, windows]
+      properties:
+        window_size_seconds:
```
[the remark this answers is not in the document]
ptrus: I like it because it explicitly states the window size of the response windows. This cannot be inferred from the window data itself, because the endpoint can return overlapping windows.

storage/migrations/03_agg_stats.up.sql
mitjat: Ah I missed the primary key. Sounds good!

storage/client/client.go
ptrus: Thanks for the suggestion. I have mixed feeling about this, on one hand It would be nice to use  `apiTypes.ActiveAccounts` directly, but I feel that the `"AT TIME ZONE 'UTC' AT TIME ZONE 'UTC'"` merits a doc comment explaining it. But then there are 4 queries where this should be used (2 for active accounts and 2 for tx volume), so maybe there can be a helper function in the `queries.go` to append this.

But there is also a different way to avoid the `d` helper struct, I'm thinking the nicest solution for all this is to just do:

```go
var t apiTypes.ActiveAccounts
if err := rows.Scan(
	&t.WindowEnd,
	&t.ActiveAccounts,
); err != nil {
	return nil, wrapError(err)
}
t.WindowEnd = t.WindowEnd.UTC() // Ensure timestamp is in UTC.

ts.Windows = append(ts.Windows, t)
```

PR #337, opened 2023-03-01
api: add gas_used and size to runtime transactions

### LINKED ISSUE #336: Add size, gas_used to runtime transactions
https://app.clickup.com/t/861m45ehr

Andrew7234:
Resolves #336 

Testing:
Manually tested locally against a production node and checked that the `gas_used` matched the value reported by the current emerald explorer. `size` is a bit more tricky to verify since the emerald explorer doesn't show it. Warren and I also briefly discussed this and it seems like the way we re-serialize transactions could possibly differ from the way the original transaction was cbor-encoded if the original transaction was serialized in a non-standard manner (which the runtime accepts). Our calculated `size` may be inaccurate in these cases, but to our knowledge this is rare.

### DIFF
--- a/analyzer/runtime/incoming.go
+++ b/analyzer/runtime/incoming.go
@@ -38,6 +38,8 @@ type BlockTransactionData struct {
 	Index                   int
 	Hash                    string
 	EthHash                 *string
+	GasUsed                 uint64
+	Size                    int
 	Raw                     []byte
 	RawResult               []byte
 	SignerData              []*BlockTransactionSignerData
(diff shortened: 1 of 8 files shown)


analyzer/runtime/incoming.go
```diff
@@ -246,6 +248,7 @@ func extractRound(b *block.Block, txrs []*sdkClient.TransactionWithResults, logg
 		}
 		blockTransactionData.Raw = cbor.Marshal(txr.Tx)
 		blockTransactionData.RawResult = cbor.Marshal(txr.Result)
+		blockTransactionData.Size = len(blockTransactionData.Raw)
```
pro-wh, line 251: can you replace `txSize` below (~line 343) with this and move the comment up here

Andrew7234: Sure, updated

pro-wh (approved):

mitjat (approved):
Noice

Can you extend the PR description with how you tested, please?

PR #338, opened 2023-03-01
db: consensus events: add index

mitjat:
Similar to https://github.com/oasisprotocol/oasis-indexer/pull/333, but in consensus.
I added the filters based on what the [currently supported query](https://github.com/oasisprotocol/oasis-indexer/blob/30b5669bec4220b0e0ce05cb01f3b0bfa7f91b71/storage/client/queries.go#L123-L135) uses in its WHERE.

Testing/rollout:
Already applied in staging (where this was observed during group debugging today). Slowness went away. Applying in prod now.


storage/migrations/01_oasis_3_consensus.up.sql
```diff
@@ -74,6 +74,9 @@ CREATE TABLE oasis_3.events
   FOREIGN KEY (tx_block, tx_index) REFERENCES oasis_3.transactions(block, tx_index) DEFERRABLE INITIALLY DEFERRED
 );
 CREATE INDEX ix_events_related_accounts ON oasis_3.events USING gin(related_accounts);
+CREATE INDEX ix_events_tx_block ON oasis_3.events (tx_block);  -- for fetching events without filters
+CREATE INDEX ix_events_tx_hash ON oasis_3.events (tx_hash);
+CREATE INDEX ix_events_type ON oasis_3.events (type);
```
pro-wh, line 79: add tx_block to the end so that we can get them in order

storage/migrations/01_oasis_3_consensus.up.sql
```diff
@@ -74,6 +74,9 @@ CREATE TABLE oasis_3.events
   FOREIGN KEY (tx_block, tx_index) REFERENCES oasis_3.transactions(block, tx_index) DEFERRABLE INITIALLY DEFERRED
 );
 CREATE INDEX ix_events_related_accounts ON oasis_3.events USING gin(related_accounts);
+CREATE INDEX ix_events_tx_block ON oasis_3.events (tx_block);  -- for fetching events without filters
+CREATE INDEX ix_events_tx_hash ON oasis_3.events (tx_hash);
```
pro-wh, line 78: should be joining through the transactions table and looking up events by tx_block, tx_index. do we ever not?

pro-wh (approved):
approving but I have a lot of opinions lol just do whatever though

storage/migrations/01_oasis_3_consensus.up.sql
```diff
@@ -74,6 +74,9 @@ CREATE TABLE oasis_3.events
   FOREIGN KEY (tx_block, tx_index) REFERENCES oasis_3.transactions(block, tx_index) DEFERRABLE INITIALLY DEFERRED
 );
 CREATE INDEX ix_events_related_accounts ON oasis_3.events USING gin(related_accounts);
+CREATE INDEX ix_events_tx_block ON oasis_3.events (tx_block);  -- for fetching events without filters
```
Andrew7234, line 77: should the index be appended too? eg, `(tx_block, tx_index)`

Andrew7234 (approved):

storage/migrations/01_oasis_3_consensus.up.sql
pro-wh: theory was that they can seek to the block and then walk through a sane number of events. should see if any query plans go haywire

storage/migrations/01_oasis_3_consensus.up.sql
mitjat: cc @Andrew7234 you'll find this interesting too I think

pro-wh: hm good point about the two different query plans

PR #340, opened 2023-03-02
analyzer: remove concurrency in runtime analyzer

### LINKED ISSUE #339: analyzer/runtime: Mutating shared `storage.QueryBatch` in multiple go-runtines
In [`processRound`](https://github.com/oasisprotocol/oasis-indexer/blob/7d8ec81cf36af20dbb2772e6a1a021ee67569739/analyzer/runtime/runtime.go#L228-L283) the `batch` variable is shared and mutated from multiple go-routines concurrently (`group.Go` is running functions in new go-routines). This is a race condition and leads to undefined behavior.

The possible solutions are to make the `storage.QueryBatch{}` thread-safe to use, or (likely the better alternative) fixing the `processRound` to not share the batch variable.

Note: didn't check if there are any other such cases, noticed this one by chance.

Andrew7234:
One way to resolve #339 

This PR removes concurrency from the runtime analyzer. It also refactors the runtime analyzer to fetch all the data initially and then process it sequentially, similar to the consensus analyzer. 

Open Question:
- Should we keep the added mutex in `storage.QueryBatch`? (see the second commit) During local testing, the additional lock/unlock did not degrade the analyzer speed since the bottleneck is the network requests to the node. Although this mutex is now unnecessary, having a thread-safe `QueryBatch` would prevent this type of bug from sneaking in (again) though.

### DIFF
--- a/analyzer/modules/accounts.go
+++ b/analyzer/modules/accounts.go
@@ -1,8 +1,6 @@
 package modules
 
 import (
-	"context"
-	"fmt"
 	"math/big"
 
 	"github.com/oasisprotocol/oasis-indexer/analyzer"
@@ -30,18 +28,13 @@ func NewAccountsHandler(source storage.RuntimeSourceStorage, runtime analyzer.Ru
 
 // PrepareAccountsData prepares raw data from the `accounts` module for insertion.
 // into target storage.
-func (h *AccountsHandler) PrepareData(ctx context.Context, round uint64, batch *storage.QueryBatch) error {
-	data, err := h.source.AccountsData(ctx, round)
-	if err != nil {
-		return fmt.Errorf("error retrieving accounts data: %w", err)
-	}
-
+func (h *AccountsHandler) PrepareData(batch *storage.QueryBatch, data *storage.RuntimeAllData) error {
 	for _, f := range []func(*storage.QueryBatch, *storage.AccountsData) error{
 		h.queueMints,
 		h.queueBurns,
 		h.queueTransfers,
 	} {
-		if err := f(batch, data); err != nil {
+		if err := f(batch, data.AccountsData); err != nil {
 			return err
 		}
 	}
--- a/analyzer/runtime/runtime.go
+++ b/analyzer/runtime/runtime.go
@@ -10,7 +10,6 @@ import (
 	oasisConfig "github.com/oasisprotocol/oasis-sdk/client-sdk/go/config"
 	"github.com/oasisprotocol/oasis-sdk/client-sdk/go/modules/rewards"
 	"github.com/oasisprotocol/oasis-sdk/client-sdk/go/types"
-	"golang.org/x/sync/errgroup"
 
 	"github.com/oasisprotocol/oasis-indexer/analyzer"
 	"github.com/oasisprotocol/oasis-indexer/analyzer/modules"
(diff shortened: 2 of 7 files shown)


storage/api.go
```diff
@@ -53,16 +59,22 @@ func (b *QueryBatch) Queue(cmd string, args ...interface{}) {
 
 // Extend merges another batch into the current batch.
 func (b *QueryBatch) Extend(qb *QueryBatch) {
+	b.mu.Lock()
+	defer b.mu.Unlock()
```
ptrus, line 63: This should probably? also lock `qb` as it accesses `qb.items`.

Andrew7234: good eye; updated!

storage/api.go
```diff
@@ -32,7 +33,10 @@ type batchItem struct {
 
 // QueryBatch represents a batch of queries to be executed atomically.
 // We use a custom type that mirrors pgx.Batch but also allows introspection for debugging.
-type QueryBatch struct{ items []*batchItem }
+type QueryBatch struct {
+	items []*batchItem
+	mu    sync.Mutex
```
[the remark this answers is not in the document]
mitjat: I'm in favor of the mutex as it makes so that if we make the same mistake again (and it's happened 3 times now: consensus, runtimes, evm token analyzer), we won't pay the terrible price of a heisenbug.

> not providing the thread-safe version by default also encourages the above.

... Much like how removing seatbelts would encourage more careful driving. It's technically true! Maybe :)
I admit that I haven't been driving the Go-kart (heh) for too long and maybe my perspective would be different with more experience. But I like seatbelts, even when I don't intend to crash.

Sorry for the tone. Go somehow keeps surprising me with its philosophy, not in a good way.

analyzer/modules/accounts.go
```diff
@@ -29,18 +27,13 @@ func NewAccountsHandler(source storage.RuntimeSourceStorage, qf *analyzer.QueryF
 
 // PrepareAccountsData prepares raw data from the `accounts` module for insertion.
 // into target storage.
-func (h *AccountsHandler) PrepareData(ctx context.Context, round uint64, batch *storage.QueryBatch) error {
```
mitjat, line 32: Big :heart: for the loss of `ctx`. I very clearly proves the separation of data fetching from data processing that we implemented a while ago in consensus but kept delaying in runtimes.

analyzer/runtime/runtime.go
```diff
@@ -283,16 +276,11 @@ func (m *Main) processRound(ctx context.Context, round uint64) error {
...
 // prepareBlockData adds block data queries to the batch.
-func (m *Main) prepareBlockData(ctx context.Context, round uint64, batch *storage.QueryBatch) error {
-	data, err := m.cfg.Source.BlockData(ctx, round)
-	if err != nil {
-		return err
-	}
-
-	for _, f := range []func(context.Context, *storage.QueryBatch, *storage.RuntimeBlockData) error{
+func (m *Main) prepareBlockData(batch *storage.QueryBatch, data *storage.RuntimeAllData) error {
+	for _, f := range []func(*storage.QueryBatch, *storage.RuntimeBlockData) error{
```
mitjat, line 280: Nit: WDYT about inlining here? I think the whole func is just `return queueBlockAndTransactionInserts(batch, data.blockData)`.

analyzer/runtime/runtime.go
```diff
@@ -228,47 +227,41 @@ func (m *Main) prework() error {
...
+		return err
+	}
 
 	// Prepare and perform updates.
 	batch := &storage.QueryBatch{}
-
-	type prepareFunc = func(context.Context, uint64, *storage.QueryBatch) error
+	type prepareFunc = func(*storage.QueryBatch, *storage.RuntimeAllData) error
+	// Process block and event data.
 	for _, f := range []prepareFunc{
```
mitjat, line 244: Optional: Now that we don't have goroutines any more, maybe just explicitly call the two funcs and skip the for loop? You call on what's more readable.

mitjat (approved):
Nice!

storage/api.go
```diff
@@ -53,16 +59,24 @@ func (b *QueryBatch) Queue(cmd string, args ...interface{}) {
 
 // Extend merges another batch into the current batch.
 func (b *QueryBatch) Extend(qb *QueryBatch) {
+	b.mu.Lock()
+	qb.mu.Lock()
+	defer b.mu.Unlock()
+	defer qb.mu.Unlock()
```
ptrus, line 65: Let's do:
```golang
	if qb != b {
		qb.mu.Lock()
		defer qb.mu.Unlock()
	}
```

That way, this won't silently deadlock in case  `Extend` is (irrationally) called on itself (e.g. `qb.Extend(&qb)`).

storage/api.go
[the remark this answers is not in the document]
ptrus: Fine with me :+1:

analyzer/runtime/runtime.go
Andrew7234: Hmm I think I slightly prefer the loop to avoid repeating the error handling boilerplate. Also I'm hoping that we'll eventually re-add goroutines (gated by some flag) for faster indexing performance once the indexer is more stable.

analyzer/runtime/runtime.go
Andrew7234: Good point, updated!

storage/api.go
Andrew7234: Thanks! updated

PR #341, opened 2023-03-03
db: schema `oasis_3` -> `chain`

mitjat:
Resolves https://app.clickup.com/t/8669n79am

**What this PR does**
Reorganizes the DB as described in the ticket:
     - renames the `oasis_3` schema to `chain`
     - moves the `processed_geneses` table from `multichain` schema (which only contained this one table) to just `chain`; `multichain` doesn't make sense any more now that we're treating the sequence of oasis-3, oasis-4 etc as a single chain
  
**Testing**
Analysis:
I ran emerald from block 1_003_298 to 1_029_386 and consensus from block 8_048_956 to 8_058_462. EVM token analyzer and aggregate_stats analyzer were running in parallel at a 10min interval. No `error` or `warning` level logs were emitted. 

API:
I ran e2e_regressions; consensus and emerald ran for 100 blocks. None of the tested URLs found any differences.

Statecheck:
Ran it with `OASIS_GENESIS_DUMP=$PWD/cache/genesis@8049056.json HEALTHCHECK_TEST_CONN_STRING='postgresql://rwuser:password@localhost:5432/indexer?sslmode=disable' HEALTHCHECK_TEST
_CHAIN_CONTEXT=b11b369e0da5bb230b220127f5e7b242d385ef8c6f54906243f30af63c815535 HEALTHCHECK_TEST_NODE_RPC=unix:/tmp/node.sock OASIS_INDEXER_HEALTHCHECK=1 go test -v -run TestGenesisFull github.com/oasisprotocol/oasis-indexer/tests/
statecheck`. Test partially fails functionally (detects some node registry discrepancies, as is common) but the technical SQL-accessing part shows no errors.


storage/migrations/01_consensus.up.sql
```diff
@@ -280,8 +280,14 @@ CREATE TABLE oasis_3.processed_blocks
   PRIMARY KEY (height, analyzer)
 );
 
+-- Keeps track of chains for which we've already processed the genesis data.
+CREATE TABLE chain.processed_geneses (
+    chain_context TEXT NOT NULL PRIMARY KEY,  -- identifies the genesis data; derived from its hash
+    processed_time TIMESTAMP WITH TIME ZONE NOT NULL
```
Andrew7234, line 286: We removed the `analyzer` field because only consensus will ever have a genesis document that we'll process?

analyzer/consensus/genesis.go
```diff
@@ -46,9 +45,8 @@ func (mg *GenesisProcessor) Process(document *genesis.Document) ([]string, error
 	}
 
 	// Rudimentary templating.
-	chainID := strcase.ToSnake(document.ChainID)
 	for i, query := range queries {
-		queries[i] = strings.ReplaceAll(query, "{{ChainId}}", chainID)
+		queries[i] = strings.ReplaceAll(query, "{{ChainId}}", "chain")
```
Andrew7234, line 49: Should we just replace all the `{{ChainId}}` with `chain` in
```           mg.addRegistryBackendMigrations,
		mg.addStakingBackendMigrations,
		mg.addGovernanceBackendMigrations,```

Andrew7234 (approved):

analyzer/consensus/genesis.go
mitjat: Yes, we should, I was just thinking we'd do it in a later PR. I left this in the description of #343 and forgot to move/copy it here after I split that PR into three:

> A similar-to-QF lightweight templating approach remains in genesis.go. I didn't touch it because we'll majorly clean up its SQL generation (from manually stringifying args to using SQL templates) in an upcoming PR.

storage/migrations/01_consensus.up.sql
mitjat: Yes! Runtimes don't really have a genesis doc.
There is a genesis struct that runtimes can specify with every new version, but it's hardcoded into the binary (IIRC) and opaque, so I think the best we can do there is manually encode it in indexer, should a significant genesis ever come out. The ones so far have been no-ops.

PR #342, opened 2023-03-03
Minor fixes: improve db ergonomics, e2e testing

mitjat:
Various minor fixes in preparation for https://github.com/oasisprotocol/oasis-indexer/pull/341:
- Expands e2e_regression script so it works smoothly when running&testing multiple block analyzers (e.g. consensus and emerald)
- Tweaks the mechanism via which we submit query batches to the DB; the new approach is possibly (unconfirmed) a little slower but produces more helpful error messages
- Changes the analyzer-internal Runtime enum so its underlying type is `string` rather than `int`; this way, it can be passed directly into SQL queries and pgx knows how to interpret it.
- Makes the internal `QueryBatch.Queries()` method return a more structured type (cmd and args separately), for cleaner usage and easier-to-read logs (in which queries appear when there's an error).

### DIFF
--- a/storage/postgres/client.go
+++ b/storage/postgres/client.go
@@ -5,6 +5,7 @@ package postgres
 import (
 	"context"
 	"fmt"
+	"os"
 
 	"github.com/jackc/pgx/v4"
 	"github.com/jackc/pgx/v4/pgxpool"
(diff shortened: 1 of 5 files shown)


storage/postgres/client.go
```diff
@@ -93,13 +94,26 @@ func (c *Client) SendBatch(ctx context.Context, batch *storage.QueryBatch) error
...
 
 func (c *Client) SendBatchWithOptions(ctx context.Context, batch *storage.QueryBatch, opts pgx.TxOptions) error {
-	pgxBatch := batch.AsPgxBatch()
 	if err := c.pool.BeginTxFunc(ctx, opts, func(tx pgx.Tx) error {
-		batchResults := tx.SendBatch(ctx, &pgxBatch)
-		defer common.CloseOrLog(batchResults, c.logger)
-		for i := 0; i < pgxBatch.Len(); i++ {
-			if _, err := batchResults.Exec(); err != nil {
-				return fmt.Errorf("query %d %v: %w", i, batch.Queries()[i], err)
+		// NOTE: Sending txs with tx.SendBatch(batch.AsPgxBatch()) is possibly more
```
ptrus, line 98: I would assume the performance gain would be noticeable for bigger batches, as I believe the batch version should do considerably fewer round-trips to the DB, while for `tx.Exec` each transaction is sent separately. Although it's likely more nuanced than this.

---

Also, there seems to be some recent improvements in the upstream library regarding these errors: https://github.com/jackc/pgx/pull/1441

Looks like indexer is using a somewhat old release: 
https://github.com/oasisprotocol/oasis-indexer/blob/a05eb962b1588e11ad495d2e907e94bdf3db075c/go.mod#LL15

Have you maybe tested the reported errors with an updated version of the library?

mitjat: Nice find!
I tried to switch to pgx v5 in the past but ended up deciding against it. I forget the exact reason, but I think it was support for custom types that felt a lot more half-baked and less battle-tested based on github issues.
This library makes very liberal use of breaking changes, so it's not completely trivial to switch to v5 (though probably still not more than a day or two of work). In particular, the (de)serialization of custom types takes some work to get right, though IIRC there are other differences too.
So I'm prioritizing this lower. But it's good to know that switching to v5 might buy performance, especially as we'll probably be focusing on performance more in the future.

ptrus: I think v4 is also actively maintained and has some recent releases. Sorry am on mobile so cannot easily check if this fix is included in any recent v4 releases.

Andrew7234 (approved):
Nice!

storage/postgres/client.go
mitjat: I checked now, it's not. We might be switching to v5 anyway; pgx has been deadlocking for us in the last week :/

PR #343, opened 2023-03-03
refactor: Remove QueryFactory

mitjat:
Follow-up to #341 now that it is no longer required to know the ChainID to query the DB.

This PR removes the QueryFactory (QF) abstraction. It was hardly needed any more after its main parameter (chainID) was rendered obsolete. The remaining uses of QueryFactory's `runtime` field were converted so that the runtime is passed into each query as an explicit parameter.

Benefits of killing QF:
- reduces global-ish state (in QF)
- reduces number of ways in which we pass parameters into queries
- removes a layer of templated strings so SQL is easier to reason about and easier to copy-paste (for debugging etc)

A similar-to-QF lightweight templating approach remains in genesis.go. I didn't touch it because we'll majorly clean up its SQL generation (from manually stringifying args to using SQL templates) in an upcoming PR.

For reviewers: Code at most commits does not compile; I still left them there because they each represent a logical step, so it's easier to follow along (IMO).

**Testing**
I ran e2e_regressions; consensus and emerald ran for 100 blocks. None of the tested URLs found any differences.

### DIFF
--- a/analyzer/modules/consensusaccounts.go
+++ b/analyzer/modules/consensusaccounts.go
@@ -17,14 +17,15 @@ const (
 // ConsensusAccountsHandler implements support for transforming and inserting data from the
 // `consensus_accounts` module for a runtime into target storage.
 type ConsensusAccountsHandler struct {
-	source storage.RuntimeSourceStorage
-	qf     *analyzer.QueryFactory
-	logger *log.Logger
+	source      storage.RuntimeSourceStorage
+	runtimeName string
+	qf          *analyzer.QueryFactory
+	logger      *log.Logger
 }
 
 // NewConsensusAccountsHandler creates a new handler for `consensus_accounts` module data.
-func NewConsensusAccountsHandler(source storage.RuntimeSourceStorage, qf *analyzer.QueryFactory, logger *log.Logger) *ConsensusAccountsHandler {
-	return &ConsensusAccountsHandler{source, qf, logger}
+func NewConsensusAccountsHandler(source storage.RuntimeSourceStorage, runtimeName string, qf *analyzer.QueryFactory, logger *log.Logger) *ConsensusAccountsHandler {
+	return &ConsensusAccountsHandler{source, runtimeName, qf, logger}
 }
 
 // PrepareConsensusAccountsData prepares raw data from the `consensus_accounts` module for insertion.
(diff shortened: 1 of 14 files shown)


Andrew7234 (approved):

storage/client/client.go
```diff
@@ -55,6 +56,17 @@ func toString(b *BigInt) *string {
 	return &s
 }
 
+func runtimeFromCtx(ctx context.Context) string {
+	// Extract the runtime name. It's populated by a middleware based on the URL.
+	runtime, ok := ctx.Value(common.RuntimeContextKey).(string)
+	if !ok {
+		// We're being called from a non-runtime-specific endpoint.
+		// This shouldn't happen. Return a dummy value, let the caller deal with it.
+		return "__NO_RUNTIME__"
```
ptrus, line 65: Should this just panic as this is likely a developers mistake? e.g. calling this method from a non-runtime specific endpoint?

analyzer/modules/consensusaccounts.go
```diff
@@ -17,14 +17,14 @@ const (
 // ConsensusAccountsHandler implements support for transforming and inserting data from the
 // `consensus_accounts` module for a runtime into target storage.
 type ConsensusAccountsHandler struct {
-	source storage.RuntimeSourceStorage
-	qf     *analyzer.QueryFactory
-	logger *log.Logger
+	source      storage.RuntimeSourceStorage
+	runtimeName string
```
ptrus, line 21: I think it would make sense to use the `analyzer.api.Runtime` type for runtimes everywhere.

storage/client/queries/queries.go
```diff
@@ -0,0 +1,462 @@
...
+		SELECT window_end, active_accounts
+		FROM stats.daily_active_accounts
+		WHERE layer = $1::text
+		ORDER BY
+			window_end DESC
+		LIMIT $2::bigint
+		OFFSET $3::bigint
+	`
+
+	// DailyActiveAccountsQuery returns the query for daily sampled daily active account windows.
```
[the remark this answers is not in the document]
mitjat: Thanks, now replaced throughout both queries.go.

storage/client/client.go
mitjat: It is a developers mistake for sure yeah. However, this is only used on the API side, i.e. when we're serving data from the DB. If we have a bug in a single endpoint handler, I'd rather produce an error log (which this __NO_RUNTIME__ value will probably eventually result in) than crash the entire API server.

analyzer/modules/consensusaccounts.go
mitjat: Yes, that was sloppy; I used the minimally-invasive, minimal-thought-required option (using a string directly) when I was in the middle of the big refactor. Should have cleaned up later. Thanks for calling it out, fixed now.

PR #346, opened 2023-03-06
EVM: handle deterministic errors for mutated token

pro-wh:
if, for example, a contract is stateful and pretends to support ERC-20 when we first scan it then refuses totalSupply later, we'd crash. don't.

### DIFF
--- a/analyzer/evmtokens/evm_tokens.go
+++ b/analyzer/evmtokens/evm_tokens.go
@@ -153,6 +153,7 @@ func (m Main) processBatch(ctx context.Context) (int, error) {
 		group.Go(func() error {
 			m.logger.Info("downloading", "stale_token", staleToken)
 			// todo: assert that addr context is secp256k1
+			//nolint:nestif
 			if staleToken.LastDownloadRound == nil {
 				tokenData, err := modules.EVMDownloadNewToken(
 					groupCtx,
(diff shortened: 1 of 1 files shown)


analyzer/evmtokens/evm_tokens.go
```diff
@@ -190,10 +190,12 @@ func (m Main) processBatch(ctx context.Context) (int, error) {
 				if err != nil {
 					return fmt.Errorf("downloading mutated token %s: %w", staleToken.Addr, err)
 				}
-				batch.Queue(m.qf.RuntimeEVMTokenUpdateQuery(),
-					staleToken.Addr,
-					mutable.TotalSupply.String(),
-				)
+				if mutable != nil {
```
Andrew7234, line 193: does it make sense to add an error log here if `mutable` is nil?

Andrew7234 (approved):

analyzer/evmtokens/evm_tokens.go
pro-wh: https://github.com/oasisprotocol/oasis-indexer/blob/78c85ac6f16f7fcfd8662a684e39c90d0414e450/analyzer/modules/evm.go#L140 it logs deep within :grimacing:

PR #348, opened 2023-03-07
bugfix: reorder paginated queries

Andrew7234:
Tested by running multiple batches of 50 queries in parallel to `localhost:8008/v1/emerald/transactions?limit=3`

### DIFF
--- a/storage/client/client.go
+++ b/storage/client/client.go
@@ -115,15 +115,13 @@ func (c *StorageClient) withTotalCount(ctx context.Context, sql string, args ...
 	if len(args) < 2 {
 		return nil, fmt.Errorf("list queries must have at least two params (limit and offset)")
 	}
-	rows, err := c.db.Query(
-		ctx,
-		sql,
-		args...,
-	)
-	if err != nil {
-		return nil, err
-	}
 
+	// A note on ordering: We query the totalCount before querying for the rows in order to
+	// avoid deadlocks. The row returned by the totalCount query is Scan-ed immediately
+	// and thus the underlying db connection is also released. However, the `rows` are
+	// `Scan`-ed in the calling function, which means that the underlying db connection is
+	// held (and unavailable to other goroutines) in the meantime.
+	limit := args[len(args)-2]
 	args[len(args)-2] = maxTotalCount + 1 // limit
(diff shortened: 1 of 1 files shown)


pro-wh (approved):
good, this also prevents us from leaking rows if the count query errors

storage/client/client.go
```diff
@@ -115,15 +115,13 @@ func (c *StorageClient) withTotalCount(ctx context.Context, sql string, args ...
...
-		return nil, err
-	}
 
+	// A note on ordering: We query the totalCount before querying for the rows in order to
+	// avoid deadlocks. The row returned by the totalCount query is Scan-ed immediately
+	// and thus the underlying db connection is also released. However, the `rows` are
+	// `Scan`-ed in the calling function, which means that the underlying db connection is
+	// held (and unavailable to other goroutines) in the meantime.
+	limit := args[len(args)-2]
 	args[len(args)-2] = maxTotalCount + 1 // limit
```
mitjat, line 125: Readability nit:

```suggestion
	origLimit := args[len(args)-2]
	// Temporarily set limit to just high enough to learn
	// if there are >maxTotalCount matching items in the DB.
	args[len(args)-2] = maxTotalCount + 1
```

(as it stands now, the two mentions of "limit" are confusable/confusing)

oh, or possibly even better so we don't need to muck up and then unmuck `args`:
```go
// totalCountArgs is like args, but with the limit set just high enough to
// learn if there are >maxTotalCount matching items in the DB.
totalCountArgs = args[:len(args)-2] 
totalCountArgs = append(totalCountArgs, maxTotalCount + 1 /*limit*/, args[len(args)-1] /*offset*/)
```

mitjat (approved):
Good find! I'll be curious what the results are. Either way it's probably good hygiene to merge in this change :+1: 

Edit: I tested in staging, and after this fix, server reliably takes 150 parallel queries.

storage/client/client.go
Andrew7234: Thanks! Updated with the first suggestion/comment. I held off on the latter option since `totalCountArgs` uses the same backing array. We could also explicitly copy `args` but imo the rename/comment you suggested is cleaner.

PR #349, opened 2023-03-08
migrate to pgx v5

mitjat:
I started this migration for the wrong reasons (thinking v4 might be the culprit for a db deadlock; it's not), but now that it's done, let's use v5. The future is now.


storage/client/client.go
```diff
@@ -1202,26 +1202,18 @@ func (c *StorageClient) RuntimeTokens(ctx context.Context, p apiTypes.GetRuntime
...
 			&t.Symbol,
 			&t.Decimals,
-			&totalSupplyNum,
+			&t.TotalSupply,
 			&t.Type,
 			&t.NumHolders,
 		); err2 != nil {
 			return nil, wrapError(err)
 		}
-		if totalSupplyNum.Status == pgtype.Present {
```
mitjat, line 1218: v4 had a bug where it didn't support scanning into _pointers to_ custom types. v5 does that just fine, so we simplify here.

storage/client/queries/queries.go
```diff
@@ -236,7 +236,7 @@ const (
 				chain.entities.address AS entity_address,
 				chain.nodes.id AS node_address,
 				chain.accounts.escrow_balance_active AS escrow,
-				chain.commissions.schedule AS commissions_schedule,
+				COALESCE(chain.commissions.schedule, '{}'::json) AS commissions_schedule,
```
mitjat, line 239: this json field gets scanned into a struct. v4 apparently scanned NULLs into a default-valued struct; v5 errors out in that case.

analyzer/consensus/consensus.go
```diff
@@ -413,6 +413,11 @@ func (m *Main) queueTransactionInserts(batch *storage.QueryBatch, data *storage.
 			signedTx.Signature.PublicKey,
 		).String()
 
+		bodyBytes, err := tx.Body.MarshalCBOR()
```
Andrew7234, line 416: Is this because v5 doesn't use `MarshalBytes` anymore, and instead marshals to a pgx.Value?

Also, on the api side does this need to be unmarshalled as cbor?

storage/postgres/client.go
```diff
@@ -94,38 +97,85 @@ func (c *Client) SendBatch(ctx context.Context, batch *storage.QueryBatch) error
...
+	if os.Getenv("PGX_FAST_BATCH") == "1" { //nolint:nestif
+		pgxBatch := batch.AsPgxBatch()
+		var batchResults pgx.BatchResults
+		var emptyTxOptions pgx.TxOptions
+		var tx pgx.Tx
+		var err error
+
+		// Begin a transaction.
+		useExplicitTx := opts != emptyTxOptions
+		if useExplicitTx {
```
Andrew7234, line 116:
```suggestion
		if !useExplicitTx {
```

Andrew7234 (approved):
Looks good pending the comment!

PR #352, opened 2023-03-08
config: archive node configuration

pro-wh:
new config format to allow setting multiple nodes, for archive access. see the sample configs for how to migrate your config files.

indexer devs: no more client factory system. just create a consensus client or runtime client directly. see existing analyzers for how to do this.

indexer devs: the analyzer no longer runs with a set chain context. in general, don't verify signatures. our node(s) will do that.

### DIFF
--- a/analyzer/consensus/consensus.go
+++ b/analyzer/consensus/consensus.go
@@ -14,7 +14,6 @@ import (
 	"github.com/oasisprotocol/oasis-core/go/common/cbor"
 	"github.com/oasisprotocol/oasis-core/go/consensus/api/transaction"
 	staking "github.com/oasisprotocol/oasis-core/go/staking/api"
-	oasisConfig "github.com/oasisprotocol/oasis-sdk/client-sdk/go/config"
 
 	"github.com/oasisprotocol/oasis-indexer/analyzer"
 	"github.com/oasisprotocol/oasis-indexer/analyzer/queries"
@@ -42,6 +41,13 @@ type parsedEvent struct {
 	relatedAddresses []staking.Address
 }
 
+// OpenSignedTxNoVerify decodes the Transaction inside a Signed transaction
+// without verifying the signature. Callers should be sure to check if the
+// transaction actually succeeded. The indexer trusts its oasis-node to
+// provide the correct transaction result, which will indicate if there was an
+// authentication problem. Skipping the verification saves CPU on the indexer.
+// Due to the chain context being global, we cannot verify transactions for
+// multiple networks anyway.
 func OpenSignedTxNoVerify(signedTx *transaction.SignedTransaction) (*transaction.Transaction, error) {
 	var tx transaction.Transaction
 	if err := cbor.Unmarshal(signedTx.Blob, &tx); err != nil {
@@ -61,22 +67,11 @@ type Main struct {
 var _ analyzer.Analyzer = (*Main)(nil)
 
 // NewMain returns a new main analyzer for the consensus layer.
-func NewMain(nodeCfg config.NodeConfig, cfg *config.BlockBasedAnalyzerConfig, target storage.TargetStorage, logger *log.Logger) (*Main, error) {
+func NewMain(sourceConfig *config.SourceConfig, cfg *config.BlockBasedAnalyzerConfig, target storage.TargetStorage, logger *log.Logger) (*Main, error) {
 	ctx := context.Background()
 
 	// Initialize source storage.
-	networkCfg := oasisConfig.Network{
-		ChainContext: nodeCfg.ChainContext,
-		RPC:          nodeCfg.RPC,
-	}
-	factory, err := source.NewClientFactory(ctx, &networkCfg, nodeCfg.FastStartup)
-	if err != nil {
-		logger.Error("error creating client factory",
-			"err", err.Error(),
-		)
-		return nil, err
-	}
-	client, err := factory.Consensus()
+	client, err := source.NewConsensusClient(ctx, sourceConfig)
 	if err != nil {
 		logger.Error("error creating consensus client",
 			"err", err.Error(),
@@ -90,10 +85,9 @@ func NewMain(nodeCfg config.NodeConfig, cfg *config.BlockBasedAnalyzerConfig, ta
 		To:   cfg.To,
 	}
 	ac := analyzer.ConsensusConfig{
-		ChainID:      nodeCfg.ChainID,
-		ChainContext: nodeCfg.ChainContext,
-		Range:        blockRange,
-		Source:       client,
+		GenesisChainContext: sourceConfig.History().CurrentRecord().ChainContext,
+		Range:               blockRange,
+		Source:              client,
 	}
 
 	logger.Info("Starting consensus analyzer", "config", ac)
@@ -112,7 +106,7 @@ func (m *Main) Start() {
 	// Get block to be indexed.
 	var height int64
 
-	isGenesisProcessed, err := m.isGenesisProcessed(ctx)
+	isGenesisProcessed, err := m.isGenesisProcessed(ctx, m.cfg.GenesisChainContext)
 	if err != nil {
 		m.logger.Error("failed to check if genesis is processed",
 			"err", err.Error(),
@@ -120,7 +114,7 @@ func (m *Main) Start() {
 		return
 	}
 	if !isGenesisProcessed {
-		if err = m.processGenesis(ctx); err != nil {
+		if err = m.processGenesis(ctx, m.cfg.GenesisChainContext); err != nil {
 			m.logger.Error("failed to process genesis",
 				"err", err.Error(),
 			)
@@ -212,19 +206,19 @@ func (m *Main) latestBlock(ctx context.Context) (int64, error) {
 	return latest, nil
 }
 
-func (m *Main) isGenesisProcessed(ctx context.Context) (bool, error) {
+func (m *Main) isGenesisProcessed(ctx context.Context, chainContext string) (bool, error) {
 	var processed bool
 	if err := m.target.QueryRow(
 		ctx,
 		queries.IsGenesisProcessed,
-		m.cfg.ChainContext,
+		chainContext,
 	).Scan(&processed); err != nil {
 		return false, err
 	}
 	return processed, nil
 }
 
-func (m *Main) processGenesis(ctx context.Context) error {
+func (m *Main) processGenesis(ctx context.Context, chainContext string) error {
 	m.logger.Info("fetching genesis document")
 	genesisDoc, err := m.cfg.Source.GenesisDocument(ctx)
 	if err != nil {
@@ -255,7 +249,7 @@ func (m *Main) processGenesis(ctx context.Context) error {
 	}
 	batch.Queue(
 		queries.GenesisIndexingProgress,
-		m.cfg.ChainContext,
+		chainContext,
 	)
 	if err := m.target.SendBatch(ctx, batch); err != nil {
 		return err
--- a/analyzer/consensus/consensus.go
+++ b/analyzer/consensus/consensus.go
@@ -42,6 +42,14 @@ type parsedEvent struct {
 	relatedAddresses []staking.Address
 }
 
+func OpenSignedTxNoVerify(signedTx *transaction.SignedTransaction) (*transaction.Transaction, error) {
+	var tx transaction.Transaction
+	if err := cbor.Unmarshal(signedTx.Blob, &tx); err != nil {
+		return nil, fmt.Errorf("signed tx unmarshal: %w", err)
+	}
+	return &tx, nil
+}
+
 // Main is the main Analyzer for the consensus layer.
 type Main struct {
 	cfg     analyzer.ConsensusConfig
@@ -400,8 +408,13 @@ func (m *Main) queueTransactionInserts(batch *storage.QueryBatch, data *storage.
 		signedTx := txr.Transaction
 		result := txr.Result
 
-		var tx transaction.Transaction
-		if err := signedTx.Open(&tx); err != nil {
+		tx, err := OpenSignedTxNoVerify(&signedTx)
+		if err != nil {
+			m.logger.Info("couldn't parse transaction",
+				"err", err,
+				"height", data.Height,
+				"tx_index", i,
+			)
 			continue
 		}
 
@@ -443,7 +456,7 @@ func (m *Main) queueTransactionInserts(batch *storage.QueryBatch, data *storage.
 
 		// TODO: Use event when available
 		// https://github.com/oasisprotocol/oasis-core/issues/4818
-		if tx.Method == "staking.AmendCommissionSchedule" {
+		if tx.Method == "staking.AmendCommissionSchedule" && result.IsSuccess() {
 			var rawSchedule staking.AmendCommissionSchedule
 			if err := cbor.Unmarshal(tx.Body, &rawSchedule); err != nil {
 				return err
--- a/analyzer/evmtokens/evm_tokens.go
+++ b/analyzer/evmtokens/evm_tokens.go
@@ -5,7 +5,6 @@ import (
 	"fmt"
 	"time"
 
-	oasisConfig "github.com/oasisprotocol/oasis-sdk/client-sdk/go/config"
 	"golang.org/x/sync/errgroup"
 
 	"github.com/oasisprotocol/oasis-indexer/analyzer"
@@ -45,37 +44,14 @@ var _ analyzer.Analyzer = (*Main)(nil)
 
 func NewMain(
 	runtime analyzer.Runtime,
-	nodeCfg *config.NodeConfig,
+	sourceConfig *config.SourceConfig,
 	target storage.TargetStorage,
 	logger *log.Logger,
 ) (*Main, error) {
 	ctx := context.Background()
 
 	// Initialize source storage.
-	networkCfg := oasisConfig.Network{
-		ChainContext: nodeCfg.ChainContext,
-		RPC:          nodeCfg.RPC,
-	}
-	factory, err := oasis.NewClientFactory(ctx, &networkCfg, nodeCfg.FastStartup)
-	if err != nil {
-		logger.Error("error creating client factory",
-			"err", err,
-		)
-		return nil, err
-	}
-
-	network, err := analyzer.FromChainContext(nodeCfg.ChainContext)
-	if err != nil {
-		return nil, err
-	}
-
-	id, err := runtime.ID(network)
-	if err != nil {
-		return nil, err
-	}
-	logger.Info("Runtime ID determined", "runtime", runtime.String(), "runtime_id", id)
-
-	client, err := factory.Runtime(id)
+	client, err := oasis.NewRuntimeClient(ctx, sourceConfig, runtime)
 	if err != nil {
 		logger.Error("error creating runtime client",
 			"err", err,
--- a/api/spec/v1.yaml
+++ b/api/spec/v1.yaml
@@ -75,8 +75,6 @@ x-query-params:
       86400 (1 day). Requests with other values may be rejected.
 
 x-examples:
-  chain-id:
-    - &chain_id_1 'oasis-3'
   block-height:
     - &block_height_1 8048956
     - &block_height_2 8049555
@@ -1023,11 +1021,6 @@ components:
       type: object
       required: [latest_chain_id, latest_block, latest_update]
       properties:
-        latest_chain_id:
-          x-go-name: LatestChainID  # oapi-codegen generates LatestChainId (with a lowercase d) by default.
-          type: string
-          description: The most recently indexed chain ID.
-          example: *chain_id_1
         latest_block:
           type: integer
           format: int64
--- a/cmd/api/api.go
+++ b/cmd/api/api.go
@@ -86,7 +86,6 @@ func Init(cfg *config.ServerConfig) (*Service, error) {
 // Service is the Oasis Indexer's API service.
 type Service struct {
 	address string
-	chainID string
 	target  *storage.StorageClient
 	logger  *log.Logger
 }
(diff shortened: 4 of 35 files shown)


storage/oasis/nodeapi/history/history.go
```diff
@@ -0,0 +1,177 @@
...
+	Records []*Record
+}
+
+var Mainnet = History{
+	Records: []*Record{
+		{
+			// https://github.com/oasisprotocol/mainnet-artifacts/releases/tag/2022-04-11
+			ArchiveName:   "damask",
+			ConstructAPI:  damask.NewDamaskConsensusApiLite,
+			GenesisHeight: 8048956,
```
mitjat, line 43: I think we'll also need the lowest round of each supported runtime. That way, we'll know which node (URL) to hit for each round.

mitjat: Agreed. I should have been clearer; I intended my comment in the context of this single-node config.
For each runtime, we want the lowest round for which the data is available _on this node version_ (cobalt, damask).

In case you meant something even more fine-grained by "node dependent": It's technically true that each individual node deployment could have differently pruned history, but that's also true in consensus. We should assume unpruned history, and then make sure at the ops level that that's how our nodes are deployed.

pro-wh: what's the limit in that case?

mitjat: For consensus, it's the height at which the dump-restore happened, i.e. GenesisHeight.
For runtimes, the first round that was finalized at or after that height. ... oh ... do you reckon a Damask node can potentially have all of Emerald history including Cobalt, because the runtime's statetree format hasn't changed? I haven't tested but I think that's not the case.

storage/oasis/runtime.go
```diff
@@ -4,25 +4,55 @@ import (
...
 // RuntimeClient is a client to a ParaTime.
 type RuntimeClient struct {
 	client  connection.RuntimeClient
-	network *config.Network
+	network *sdkConfig.Network
+	info    *sdkTypes.RuntimeInfo
+	rtCtx   sdkSignature.Context
+}
+
+func NewRuntimeClient(ctx context.Context, sourceConfig *config.SourceConfig, runtime analyzer.Runtime) (*RuntimeClient, error) {
```
pro-wh, line 30: todo: move this back to client.go for smaller diff

storage/oasis/consensus.go
```diff
@@ -13,16 +13,27 @@ import (
...
 )
 
 // ConsensusClient is a client to the consensus backends.
 type ConsensusClient struct {
 	nodeApi nodeapi.ConsensusApiLite
-	network *config.Network
+	network *sdkConfig.Network
+}
+
+func NewConsensusClient(ctx context.Context, sourceConfig *config.SourceConfig) (*ConsensusClient, error) {
```
pro-wh, line 30: todo: move this back to client.go for smaller diff

storage/oasis/client.go
```diff
@@ -4,107 +4,79 @@ package oasis
...
 	if err != nil {
 		return nil, err
 	}
 
-	rtCtx := runtimeSignature.DeriveChainContext(info.ID, cf.network.ChainContext)
+	rtCtx := sdkSignature.DeriveChainContext(info.ID, sdkNet.ChainContext)
 	return &RuntimeClient{
 		client:  client,
-		network: cf.network,
+		network: sdkNet,
```
pro-wh, line 79: how did this used to work? wasn't cf.network the fake one we used with empty .ParaTimes?

storage/oasis/client.go
```diff
@@ -4,107 +4,57 @@ package oasis
...
 	}
-
-	// Configure chain context for all signatures using chain domain separation.
-	signature.SetChainContext(cf.network.ChainContext)
-
-	c := &ConsensusClient{
-		nodeApi: nodeApi,
-		network: cf.network,
+	fakePT := &sdkConfig.ParaTime{
+		ID: runtime.String(),
```
pro-wh, line 41: oop this is wrong

storage/oasis/client_test.go
```diff
@@ -5,50 +5,82 @@ import (
...
 	pastHeight   = 8048956      // oasis-3 genesis height
 	futureHeight = 100000000000 // an unreasonably high height
 )
 
-func newClientFactory() (*ClientFactory, error) {
-	network := &config.Network{
-		ChainContext: os.Getenv("CI_TEST_CHAIN_CONTEXT"),
-		RPC:          os.Getenv("CI_TEST_NODE_RPC"),
+func init() {
+	history.Chains["ci-test"] = &history.History{
```
pro-wh, line 23: okay this stuff sucks. so much code relies on internally looking at `history.Chains[aChainNameYouGaveMe]`. but that makes it less nice to use custom chains, e.g. a local testnet. I wonder if we could make it more viable to pass an actual *history.History around and chain names less

pro-wh:
ugh and the docker e2e test doesn't have the right configs and stuff

pro-wh:
alright let's allow there to be a single-oasis-node-with-given-chain-context setting as well.

storage/oasis/client_test.go
mitjat: That sounds reasonable from afar. My guess is it's just the difference between having the chain name + a bunch of utility functions that look at a hardcoded history, or representing those functions as methods of the history object.

config/config.go
```diff
@@ -109,26 +112,27 @@ type AnalyzersList struct {
...
-	ChainContext string `koanf:"chaincontext"`
+	// Nodes describe the oasis-node(s) to connect to. Keys are "archive
+	// names," e.g. "damask."
+	Nodes map[string]*NodeConfig `koanf:"nodes"`
 
 	// If set, the analyzer will skip some initial checks, e.g. that
-	// `rpc` really serves the chain with `chain_context`.
+	// `rpc` really serves the chain with the chain context we expect.
 	// NOT RECOMMENDED in production; intended for faster testing.
 	FastStartup bool `koanf:"fast_startup"`
```
mitjat, line 127: I'm OK to get rid of this BTW, and just default to true (= never check).
I don't think this makes it harder to misconfigure a mainnet analyzer to query a testnet node, or vice versa. If you accidentally use a testnet config (via partial node config copy-paste, or via using the wrong config file) on a mainnet analyzer, you'll use it wholesale, with the chain name _and_ context at the same time.
I expect the overwhelming scenario in which the name and context will disagree will be when you change the name intentionally but forgot to hunt down the corresponding chain context.

pro-wh: :+1: but let's consider that outside this PR

storage/oasis/client_test.go
pro-wh: added an alternative SourceConfig to specify one rpc url and a custom chain context

config/config.go
```diff
@@ -186,9 +216,6 @@ func (cfg *AggregateStatsConfig) Validate() error {
 
 // ServerConfig contains the API server configuration.
 type ServerConfig struct {
-	// ChainID is the chain ID (normally latest) for the server API
-	ChainID string `koanf:"chain_id"`
```
pro-wh, line 190: turns out this is still used in one of our API responses. keeping it for now. we can remove it later if we want

pro-wh: actually let me try removing it in a separate commit

pro-wh:
I've just been made aware of analyzer/consensus.Main.source(height)

pro-wh:
shout out to #39 and #63

config/config.go
pro-wh: removing that field. good luck to everyone downstream

storage/oasis/client_test.go
pro-wh: added, use SingleNetwork

storage/oasis/nodeapi/history/history.go
pro-wh: I think we can get the round number from the genesis roothash entries

mitjat:
> shout out to #39 and #63

I also realized half way through my archive refactor(s) that `ConsensusSourceStorage` and `RuntimeSourceStorage` are basically the spritual precursors to `ConsensusApiLite` and `RuntimeApiLite`, and maybe I should have built multi-version support in there. Their sin was that they did data fetching from the node, but then also some data munging, so the abstraction boundary was not quite right. Still, in retrospect, it would have been probably easier to take out the data munging and repurpose them for just RPC access. Alas, I saw this only later.

storage/oasis/client.go
```diff
@@ -4,107 +4,95 @@ package oasis
...
-		// Cobalt mainnet.
-		creds := credentials.NewTLS(&tls.Config{MinVersion: tls.VersionTLS12})
-		dialOpts := []grpc.DialOption{grpc.WithTransportCredentials(creds)}
-		grpcConn, err := cmnGrpc.Dial(cf.network.RPC, dialOpts...)
+func NewRuntimeClient(ctx context.Context, sourceConfig *config.SourceConfig, runtime analyzer.Runtime) (*RuntimeClient, error) {
+	var sdkClient connection.RuntimeClient
+	var sdkNet *sdkConfig.Network
+	switch {
+	case sourceConfig.History != nil:
+		// Use only the current node for runtime interaction for now.
```
mitjat, line 72: Can you open a clickup ticket for the remaining work please?
- encode lowest runtime round for each runtime and chain (Damask etc) in `history.DefaultChains`
- support per-runtime node configs (e.g. using a separate node for Sapphire)

I think a single ticket/PR can handle it, WDYT?

config/config.go
```diff
@@ -61,6 +61,25 @@ type AnalysisConfig struct {
 
 // Validate validates the analysis configuration.
 func (cfg *AnalysisConfig) Validate() error {
+	sourceConfigured := false
+	if cfg.Source.History != nil {
+		sourceConfigured = true
+		if cfg.Source.History.ChainName == "" {
+			return fmt.Errorf("source.history.chian_name not provided")
```
mitjat, line 68: nit: typo

storage/oasis/nodeapi/history/history.go
```diff
@@ -0,0 +1,293 @@
...
+				ConstructAPI:  NotImplemented,
+				GenesisHeight: 1,
+			},
+		},
+	},
+}
+
+type HistoryConsensusApiLite struct {
+	History *History
+	// APIs. Keys are archive names.
```
mitjat, line 164:
```suggestion
	// APIs. Keys are lowercase archive names (cobalt, damask, ...).
```

I'm worried about naming. Not that I have any great solutions. But cobalt and damask are chains, aren't they. They each have their _chain_ ID (oasis-2, oasis-3), their _chain_ context. So maybe instead, mainnet vs testnet are ... megachains? metachains? I'm open to suggestions.

Quite possibly, we should just leave names as they are. Maybe there's no hope in hell of getting the naming straight; "chain" seems to be everything from "mainnet" to "damask" to "emerald", not just here but in oasis-core land too. It's partly for that reason that I like peppering chain-related docs with little parenthesized examples to clarify, like in this suggested diff.

mitjat (commented):
Looks nice! My only meatier comment is around trying to special-case e2e tests less, and not having the SingleNetwork construct.

It might be a little gnarly to resolve merge conflicts in https://github.com/oasisprotocol/oasis-indexer/pull/362 regardless of which PR goes in first, but it looks like they'll be the kind of conflicts where once the compiler is happy, things just work :crossed_fingers: .

cmd/api/api.go
```diff
@@ -86,7 +86,6 @@ func Init(cfg *config.ServerConfig) (*Service, error) {
 // Service is the Oasis Indexer's API service.
 type Service struct {
 	address string
-	chainID string
```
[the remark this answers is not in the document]
pro-wh: still compiles lol :muscle:

storage/oasis/nodeapi/history/history.go
```diff
@@ -0,0 +1,293 @@
...
+				// https://github.com/oasisprotocol/testnet-artifacts/releases/tag/2020-09-15
+				ArchiveName:   "2020-09-15",
+				ConstructAPI:  NotImplemented,
+				GenesisHeight: 1,
+			},
+		},
+	},
+}
+
+type HistoryConsensusApiLite struct {
```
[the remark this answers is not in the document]
pro-wh: different style. dependencies above dependents

storage/oasis/nodeapi/cobalt/node.go
```diff
@@ -45,6 +45,10 @@ func NewCobaltConsensusApiLite(grpcConn *grpc.ClientConn) *CobaltConsensusApiLit
 	}
 }
 
+func NewCobaltConsensusApiLiteErased(grpcConn *grpc.ClientConn) nodeapi.ConsensusApiLite {
```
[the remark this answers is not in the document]
pro-wh: k I'll try making the default type-erased

analyzer/consensus/consensus.go
```diff
@@ -42,6 +42,14 @@ type parsedEvent struct {
 	relatedAddresses []staking.Address
 }
 
+func OpenSignedTxNoVerify(signedTx *transaction.SignedTransaction) (*transaction.Transaction, error) {
```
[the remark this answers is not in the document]
pro-wh: gonna add comment, but the explanation is a little different. the node did not filter out txs with invalid signatures, but the whole committee checks signatures and determines which txs had invalid authentication

storage/oasis/nodeapi/history/history.go
```diff
@@ -0,0 +1,293 @@
...
+			{
+				// https://github.com/oasisprotocol/mainnet-artifacts/releases/tag/2021-04-28
+				ArchiveName:   "cobalt",
+				ConstructAPI:  BindConstructorGRPC(cobalt.NewCobaltConsensusApiLiteErased),
+				GenesisHeight: 3027601,
+				ChainContext:  "53852332637bacb61b91b6411ab4095168ba02a50be4c3f82448438826f23898",
+			},
+			{
+				// https://github.com/oasisprotocol/mainnet-artifacts/releases/tag/2020-11-18
+				ArchiveName:   "mainnet",
```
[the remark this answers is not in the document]
pro-wh: k I'll add something about this

storage/oasis/nodeapi/history/history.go
```diff
@@ -0,0 +1,293 @@
...
+			APIs[record.ArchiveName] = api
+		}
+	}
+	return &HistoryConsensusApiLite{
+		History: history,
+		APIs:    APIs,
+	}, nil
+}
+
+func (c *HistoryConsensusApiLite) APIForHeight(height int64) (nodeapi.ConsensusApiLite, error) {
```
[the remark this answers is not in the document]
pro-wh: would need to check, but no motivation. this is something we can provide publicly correctly

storage/oasis/nodeapi/history/history.go
```diff
@@ -0,0 +1,293 @@
...
+func (h *History) RecordForHeight(height int64) (*Record, error) {
+	if height == consensus.HeightLatest {
+		return h.CurrentRecord(), nil
+	}
+	for _, r := range h.Records {
+		if height >= r.GenesisHeight {
+			return r, nil
+		}
+	}
+	return nil, fmt.Errorf("height earliest than earliest history record")
```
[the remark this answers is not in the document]
pro-wh: hm ok

storage/oasis/client.go
pro-wh: oh nvm we rummage around in sdkConfig.DefaultNetworks.All[?].ParaTimes.All[?]

config/config.go
```diff
@@ -109,26 +128,37 @@ type AnalyzersList struct {
...
-	// RPC is the node endpoint.
-	RPC string `koanf:"rpc"`
-
-	// ChainContext is the domain separation context.
-	// It uniquely identifies the chain; it is derived as a hash of the genesis data.
-	// Used as safety check to prevent accidental use of the wrong RPC endpoint.
-	ChainContext string `koanf:"chaincontext"`
+// SourceConfig has some controls about what chain we're analyzing and how to connect.
+type SourceConfig struct {
+	SingleNetwork *SingleNetworkConfig `koanf:"single_network"`
```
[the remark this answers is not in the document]
pro-wh: > Hardcode the chain context into history.DefaultChains, with a comment and understanding that this will change as test fixtures change.

are the test chain contexts actually stable?

the other thing about history config is that the chain name is also tied to how we access sdkConfig.DefaultNetworks.

I'm going to try changing it to where SourceConfig has:

- Nodes map[string]*NodeConfig (always)
- ChainName string (when using a known chain)
- CustomChain struct (when not using a known chain)
  - History History (you just put your single network with chain context in your yaml config file)
  - SDKNetwork sdkConfig.Network (for you to provide paratime and denomination info)

analyzer/evmtokens/evm_tokens.go
```diff
@@ -45,37 +44,14 @@ var _ analyzer.Analyzer = (*Main)(nil)
...
-		return nil, err
-	}
-
-	id, err := runtime.ID(network)
-	if err != nil {
-		return nil, err
-	}
-	logger.Info("Runtime ID determined", "runtime", runtime.String(), "runtime_id", id)
-
-	client, err := factory.Runtime(id)
```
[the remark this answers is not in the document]
pro-wh: :heart:

storage/oasis/client.go
pro-wh: k

pro-wh: https://app.clickup.com/t/866a0fpr3

config/config.go
pro-wh: oo good find

storage/oasis/nodeapi/history/history.go
pro-wh: > I'm open to suggestions.

continuity. or continuum



> cobalt and damask are chains

:face_with_spiral_eyes: also been using the term 'network' due to that one SDK struct

internally I've been calling them 'beta,' 'mainnet-a,' 'mainnet-b,' 'mainnet-c'

storage/oasis/nodeapi/history/history.go
pro-wh: continuing in clickup https://app.clickup.com/t/866a0fpr3

config/config.go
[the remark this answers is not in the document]
mitjat: :+1: Sounds like a reasonable plan too.

To answer your questions and advocate for my suggestion a little more (but disregard if you've already gone ahead with yours):
- "are the test chain contexts actually stable?" I don't know for sure, but I assume so. The context is a hash of the genesis IIRC, and for our e2e tests, the genesis is a fixture. I guess it _might_ include something ugly like current time that changes the context every time it runs ... but I doubt it, and it IMO shouldn't.
- "chain name is also tied to how we access sdkConfig.DefaultNetworks" - I'm in favor of not using `sdkConfig.DefaultNetworks`. It only covers the current version, we already replicate a large part of it, and we'll need to replicate even more of it once we add proper support for runtimes. Just copy paste the whole thing into indexer, munge it into whatever format suits us best, and be done with it. Its contents are stable are valid forever.

storage/oasis/nodeapi/history/history.go
mitjat: I kinda like continuum :)
But it's also a rare enough concept that I'd probably still want to mention it as "continuum (e.g. mainnet/testnet/localnet)" in most docstrings.

storage/oasis/nodeapi/cobalt/node.go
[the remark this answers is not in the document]
pro-wh: ok with the separation of History and APIConstructors, we won't have to repeat code so much. so I'll try instead un-extracting these mind-bending BindConstructor* functions, and we should be able to retain the un-type-erased constructors

config/config.go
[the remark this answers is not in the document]
pro-wh: k, going with `custom_chain` for now though. let's copy over the paratime info and move away from sdkConfig.DefaultNetworks some time in the future

config/config.go
pro-wh: rewrote the validation function when I changed the structure

storage/oasis/nodeapi/history/history.go
pro-wh: introduce the term separately (:

storage/oasis/nodeapi/history/sdk.go
```diff
@@ -0,0 +1,35 @@
...
+
+	sdkConfig "github.com/oasisprotocol/oasis-sdk/client-sdk/go/config"
+	"github.com/oasisprotocol/oasis-sdk/client-sdk/go/connection"
+
+	"github.com/oasisprotocol/oasis-indexer/config"
+)
+
+// SDKConnect creates an Oasis SDK Connection to a node running the current
+// version of oasis-node.
+func SDKConnect(ctx context.Context, chainContext string, nodeConfig *config.NodeConfig, fastStartup bool) (connection.Connection, error) {
```
mitjat, line 15: Nit, nonblocking: Not sure this has to do with history much ... Can you think of a better home? `storage/oasis`? Same for raw_grpc.go.

mitjat (approved):

storage/oasis/nodeapi/cobalt/node.go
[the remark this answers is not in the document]
mitjat: great! The `BindConstructor` stuff was too clever for my liking too.

pro-wh:
`custom_chain` is added, action-oriented comments resolved. we can feel free to add test-oriented chains to the go code as well. but I like having it possible to specify one in the config

tests/e2e/config/e2e-dev.yml
```diff
@@ -13,8 +20,7 @@ analysis:
     migrations: file:///storage/migrations
 
 server:
-  chain_id: oasis-3  # a fake mainnet, based on oasis-net-runner
-  chain_name: localnet
+  chain_name: localnet  # a fake mainnet, based on oasis-net-runner
```
mitjat, line 23:
```suggestion
  chain_name: localnet  # unused
```

mitjat (approved):

storage/oasis/nodeapi/history/sdk.go
pro-wh: oh ok I can move these into a separate directory :+1:

api/spec/v1.yaml
```diff
@@ -1023,11 +1021,6 @@ components:
       type: object
       required: [latest_chain_id, latest_block, latest_update]
       properties:
-        latest_chain_id:
-          x-go-name: LatestChainID  # oapi-codegen generates LatestChainId (with a lowercase d) by default.
```
lukaw3d, line 1027: Leftover latest_chain_id in `required`

pro-wh: o thanks

pro-wh: #393

PR #356, opened 2023-03-15
analyzer/consensus: Full Cobalt support. Internal types for node responses.

mitjat:
Resolves https://app.clickup.com/t/8669n7f27, partially resolves https://app.clickup.com/t/8669n7gt3 (for most of consensus types)

This PR:
- Implements the rest of `ConsensusApiLite` (introduced in #326), an interface that unifies communication with Cobalt and Damask nodes.
- Migrates that interface from using the latest oasis-core types to using internal types.

The latter has considerable fallout, and introduces considerable boilerplate :/. That said, it's the best solution I see. Re-posting my Slack reasoning for visibility:
>  The initial plan was to grab the Cobalt response, convert it to a Damask response, and then reuse our existing analyzer code from there. There are two issues with this:
> * It's more work than it should be. For example, in `TransactionsWithResults`, which includes all the events, some types are more complex than I expected, and can differ quite a lot between versions. For example, when a runtime registers, the registration includes a ton of parameters (executors, TEE hardware, admission policies for nodes, staking rules, etc), and many of them changed their types between versions. I plowed into converting those types, but 150 lines in, I gave up. They were lines that come slowly, because you have to manually hunt down every field and try to cast it, then failing that, convert manually, carefully comparing the fields. With 150 lines of boilerplate-looking, easy-to-make-mistakes code, I was nowhere near done with the conversion. The kicker is that indexer doesn't really use almost any of those fields. But I cannot just replace them with `nil`, read on.
> * It loses information. We _do_ use all events at a surface level in that we index them, i.e. we store their raw JSON into the DB (and will probably expose it via the web UI). When converting cobalt events to the damask-event format, some fields are lost, some are made up ... and only then does the analyzer get to see the value and store it into the DB.
> * It is not future-proof because indexer will eventually depend on an even newer oasis-core. We've known this, and we planned for it tentatively in https://app.clickup.com/t/8669n7gt3; that task now got largely folded into this PR.

**Testing:**
- I ran 10k rounds of consensus pre- and post-PR, will compare DB dumps. Results:
    - There was a bug in storing the body of `staking.allowance_change` events. The old analyzer stored the base64-encoded JSON representation, but all other events are just plain JSON. This PR fixes that.
    - For nodes that had no TLS addresses (= they were not listening for TLS connections; I'm not clear on what protocol under TLS), we used to store their addresses as `{''}`. Now, it's correctly `{}`. (Curlies are how postgres represents arrays.) That said, I see those addresses are now [deprecated](https://github.com/oasisprotocol/oasis-core/blob/7e226c391c477cbcbb492a8dbe4cfce88bc8cb46/go/common/node/node.go#L455-L455) anyway :). Same for P2P addresses, which are _not_ deprecated.
    - For nodes, various addresses were incorrectly stored doubly-quoted (= as a string with literal `'` at beginning and end); fixed now.
    - No other differences.
- I ran the consensus analyzer against a Cobalt node (except non-tx registry events, because they're not whitelisted by our current node instance) for a good number of rounds. No errors or crashes. I did not go to the trouble of retrofitting statecheck for Cobalt.

### DIFF
--- a/storage/oasis/client.go
+++ b/storage/oasis/client.go
@@ -66,7 +66,7 @@ func (cf *ClientFactory) Consensus() (*ConsensusClient, error) {
 		if err != nil {
 			return nil, err
 		}
-		nodeApi = cobalt.NewCobaltConsensusApiLite(grpcConn, (*cf.connection).Consensus())
+		nodeApi = cobalt.NewCobaltConsensusApiLite(grpcConn)
 	} else {
 		// Assume Damask.
 		client := (*cf.connection).Consensus()
--- a/storage/oasis/client.go
+++ b/storage/oasis/client.go
@@ -66,7 +66,7 @@ func (cf *ClientFactory) Consensus() (*ConsensusClient, error) {
 		if err != nil {
 			return nil, err
 		}
-		nodeApi = cobalt.NewCobaltConsensusApiLite(grpcConn)
+		nodeApi = cobalt.NewCobaltConsensusApiLite(grpcConn, (*cf.connection).Consensus())
 	} else {
 		// Assume Damask.
 		client := (*cf.connection).Consensus()
--- a/storage/oasis/nodeapi/cobalt/node.go
+++ b/storage/oasis/nodeapi/cobalt/node.go
@@ -150,15 +150,19 @@ func (c *CobaltConsensusApiLite) RoothashEvents(ctx context.Context, height int6
 	return events, nil
 }
 
-func (c *CobaltConsensusApiLite) GetValidators(ctx context.Context, height int64) ([]*scheduler.Validator, error) {
+func (c *CobaltConsensusApiLite) GetValidators(ctx context.Context, height int64) ([]nodeapi.Validator, error) {
 	rsp, err := c.damaskClient.Scheduler().GetValidators(ctx, height)
 	if err != nil {
 		return nil, fmt.Errorf("calling GetValidators() on Cobalt node using Damask ABI: %w", err)
 	}
-	return rsp, nil
+	validators := make([]nodeapi.Validator, len(rsp))
+	for i, v := range rsp {
+		validators[i] = nodeapi.Validator(*v)
+	}
+	return validators, nil
 }
 
-func (c *CobaltConsensusApiLite) GetCommittees(ctx context.Context, height int64, runtimeID common.Namespace) ([]*scheduler.Committee, error) {
+func (c *CobaltConsensusApiLite) GetCommittees(ctx context.Context, height int64, runtimeID common.Namespace) ([]nodeapi.Committee, error) {
 	rsp, err := c.damaskClient.Scheduler().GetCommittees(ctx, &scheduler.GetCommitteesRequest{
 		Height:    height,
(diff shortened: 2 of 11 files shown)


storage/oasis/nodeapi/cobalt/node.go
```diff
@@ -2,37 +2,47 @@ package cobalt
...
 type CobaltConsensusApiLite struct {
 	grpcConn *grpc.ClientConn
+	// Used as a convenience for calling methods that are ABI-compatible between Cobalt and Damask.
+	damaskClient consensus.ClientBackend
 }
 
 var _ nodeapi.ConsensusApiLite = (*CobaltConsensusApiLite)(nil)
 
-func NewCobaltConsensusApiLite(grpcConn *grpc.ClientConn) *CobaltConsensusApiLite {
+func NewCobaltConsensusApiLite(grpcConn *grpc.ClientConn, damaskClient consensus.ClientBackend) *CobaltConsensusApiLite {
```
pro-wh, line 42: don't do this yet, #352 relies on a specific constructor signature.

storage/oasis/client.go
```diff
@@ -66,7 +66,7 @@ func (cf *ClientFactory) Consensus() (*ConsensusClient, error) {
 		if err != nil {
 			return nil, err
 		}
-		nodeApi = cobalt.NewCobaltConsensusApiLite(grpcConn)
+		nodeApi = cobalt.NewCobaltConsensusApiLite(grpcConn, (*cf.connection).Consensus())
```
pro-wh, line 69: can't proceed with this. making an sdk connection to an old node might break with some combinations. e.g. early nodes didn't have a dedicated GetChainContext, and you'd instead get the whole genesis document and hash it if you really needed it

mitjat: It's not SDK, it's oasis-core. Just makes the conversions a little simpler; for example, if the struct remains the same across Cobalt and Damask but contains an enum, we can either a) interpret the result's CBOR as a Damask object, or b) interpret the result's CBOR as a Cobalt object, then manually convert every single field just so we can explicitly cast the enum.

It did work in practice, and it was only used for a select few methods. ("	// Used as a convenience for calling methods that are ABI-compatible between Cobalt and Damask.", said the docstring for `damaskClient` in the cobalt client).

I don't think it was risky, because Cobalt is frozen in time, so if it worked once, it would work always. That said, it was hacky. Removed in favor of explicit conversions.

storage/oasis/nodeapi/cobalt/node.go
mitjat: Removed, see below.

storage/oasis/client.go
pro-wh: I meant cf.connection was connection.Connection from `connection "github.com/oasisprotocol/oasis-sdk/client-sdk/go/connection"`, from the SDK, i.e. tracks latest rather than being pinned to Damask. Thanks for the changes.

pro-wh (approved):
let's try it out with this. haven't read through, but the regression test passing is a good sign

Andrew7234 (approved):

PR #358, opened 2023-03-16
analyzer/runtime: Refactor: Separate download, extract, sqlize steps

mitjat:
Prework for https://app.clickup.com/t/8669n7f41

The runtime analyzer is currently an amalgamation of the early Emerald code and the imported code from oasis-nexus.

This PR refactors the analyzer to unify their approaches more, though there's still a lot of cleanup opportunities let.
In particular, the PR refactors the analyzer so that each block is processed in three stages:
- fetch all data from node (this was already separate as of recently; thanks Andy)
- preprocess structures from the node into a rich `BlockData` struct (this was the approach from nexus/EVM/Warren, but the data was only used in some places, while others used the raw structures directly. That makes it a lot harder to rework the structures coming from the node, as we need to do to support multiple versions of oasis-core/oasis-sdk)
- convert the `BlockData` structure into sql statements

**Testing:** 
Analyzed 5k blocks before and after the refactor, compared DB dumps. No diffs except that the unused `chain.runtime_gas_used` table is gone.


pro-wh (commented):
looks good so far

Andrew7234 (approved):
Looks good, thanks!

PR #359, opened 2023-03-17
bugfix: analyzer: evm events: Represent big int values as string, not float

mitjat:
Before this PR, large-enough integer values in EVM event params were encoded as floats in the API response:
```
{"evm_type":"uint256","name":"value","value":2.70272e+22}
```

The value was correct in the DB's internal JSON representation (270272000...00) but got incorrectly serialized as a float when inserting it into the API response.

This PR changes the return type of the web API such that all integer values are represented as strings.
(first approach taken by this PR shown with ~strikeout~)

It does so by
~1. changing the internal API type of the `value` field to be `json.RawMessage`, essentially meaning "take whatever encoding the DB gives you, trust that it's valid JSON, and when you're asked to serialize yourself, just regurgitate the bytes". This makes the API correctly return `270272000...00`.~
~2. encoding integer values as strings on the fly as they are fetched from the DB. While `270272000...00` is technically correct, `jq` and presumably most JS clients mis-parse such large int as floats (again). Our web API already represents all BigInts as strings, so this is consistent with the rest of the API.~
1. Serializing the uint256 EVM type into a JSON string, rather than a JSON number.

**Testing:**
- Tested the API manually by querying `localhost:8008/v1/emerald/events`. Numbers appear as strings, other kinds of values are unaffected.
- Tested the modified analyzer by running for 1000 blocks and diffing the DB dump. No diffs.

### DIFF
--- a/analyzer/runtime/incoming.go
+++ b/analyzer/runtime/incoming.go
@@ -490,7 +490,9 @@ func extractEvents(blockData *BlockData, relatedAccountAddresses map[apiTypes.Ad
 						{
 							Name:    "value",
 							EvmType: "uint256",
-							Value:   amount,
+							// JSON supports encoding big integers, but many clients (javascript, jq, etc.)
+							// will incorrectly parse them as floats. So we encode uint256 as a string instead.
+							Value: amount.String(),
 						},
 					}
 					eventData := EventData{
@@ -537,7 +539,9 @@ func extractEvents(blockData *BlockData, relatedAccountAddresses map[apiTypes.Ad
 						{
 							Name:    "value",
 							EvmType: "uint256",
-							Value:   amount,
+							// JSON supports encoding big integers, but many clients (javascript, jq, etc.)
+							// will incorrectly parse them as floats. So we encode uint256 as a string instead.
+							Value: amount.String(),
 						},
 					}
 					eventData := EventData{
--- a/api/spec/v1.yaml
+++ b/api/spec/v1.yaml
@@ -1961,8 +1961,12 @@ components:
           description: The solidity type of the event parameter.
         value:
           description: The parameter value.
-      description: A decoded parameter of an event emitted from an evm runtime.
-
+      description: |
+        A decoded parameter of an event emitted from an EVM runtime.
+        Values of EVM type `int128`, `uint128`, `int256`, `uint256`, `fixed`, and `ufixed` are represented as strings.
+        Values of EVM type `address` and `address payable` are represented as lowercase hex strings with a "0x" prefix.
+        Values of EVM type `bytes` and `bytes<N>` are represented as base64 strings.
+        Values of other EVM types (integer types, strings, arrays, etc.) are represented as their JSON conterpart.
     RuntimeTransactionList:
       allOf:
         - $ref: '#/components/schemas/List'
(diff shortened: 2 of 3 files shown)


storage/client/client.go
```diff
@@ -1076,6 +1077,8 @@ func (c *StorageClient) RuntimeTransactions(ctx context.Context, p apiTypes.GetR
 	return &ts, nil
 }
 
+var digitTest = regexp.MustCompile(`^-?[0-9]+$`)
```
Andrew7234, line 1080: what's the purpose of the `-?`

storage/client/client.go
```diff
@@ -1114,6 +1117,23 @@ func (c *StorageClient) RuntimeEvents(ctx context.Context, p apiTypes.GetRuntime
...
+		//    somewhere in the stack.
+		//  - For JS clients, big ints are best represented as strings. This is also
+		//    consistent with how the rest of the API represents the BigInt type.
+		if e.EvmLogParams != nil {
+			for i := range *e.EvmLogParams {
+				param := &(*e.EvmLogParams)[i]
+				// param.Value is a json.RawMessage, so "marshaling" it just returns
+				// the same raw JSON-encoded bytes that came from the DB.
+				JSONValue, _ := param.Value.MarshalJSON()
+				if digitTest.MatchString(string(JSONValue)) {
```
Andrew7234, line 1132: Since `json.RawMessage` is []byte, it looks like we can maybe just use `param.Value` directly instead of calling [MarshalJSON()](https://cs.opensource.google/go/go/+/refs/tags/go1.20.2:src/encoding/json/stream.go;l=265)?

ie, something like
```
MatchString(string(param.Value))
```

analyzer/runtime/incoming.go
```diff
@@ -523,21 +527,24 @@ func extractEvents(blockData *BlockData, relatedAccountAddresses map[apiTypes.Ad
 					}
 					amount := &big.Int{}
 					amount.SetBytes(amountU256)
+					ownerJSON, _ := json.Marshal(ethCommon.BytesToAddress(ownerEthAddr))
+					spenderJSON, _ := json.Marshal(ethCommon.BytesToAddress(spenderEthAddr))
+					amountJSON, _ := json.Marshal(amount)
```
Andrew7234, line 532: The PR description says the db stores the number correctly, so this is functionally a no-op? It's just because the `EvmEventParam` type changed, right?

Andrew7234 (approved):

pro-wh (commented):
I'm in favor of having them be strings in the db too

storage/client/client.go
mitjat: We could, and it works fine, I just tried. However this relies on what is technically an internal implementation detail of the `json` library. I'd rather preserve the abstraction boundaries and use the intended API for converting to []byte (and eventually string).

mitjat (commented):
Thanks Andy! Answers inline.

mitjat:
> I'm in favor of having them be strings in the db too

That is simpler, but I felt that it was ugly because it forced a JS limitation deep into our Go/postgres stack.

But then thinking about it more, I realized that JSON is just an intermediary format anyway; the real format/type is the EVM one, and we already cannot natively represent some of those types, notably `address`. It's entirely our choice how to best represent them in JSON.
So I went ahead and encoded our `uint256` as strings. I also slightly prematurely documented (in openapi) the suggested/expected mappings from EVM types to JSON types for us to use in the future, once we support arbitrary EVM events.

Also, it's not just simpler, it's _much_ simpler. So thanks for the suggestion!

pro-wh (approved):
image
12 of those being comments and docs :heart:

PR #360, opened 2023-03-20
Andrew7234/api tests

Andrew7234:
https://app.clickup.com/t/861md6gue

The initial idea of this PR is to 
* Update the indexer to be able to read node data from a file to facilitate e2e testing offline or in CI without access to a node
* To support ^, add a tool to dump node state into a file
* Add api tests that utilize the existing `run.sh` tool to verify both analyzer and indexer api functionality

Open questions:
- Runtime support is more tricky because of the `runtimeClient.SimulateEVMCall`. Do we include runtime api tests in this PR, and if so, what's the best way of doing that


storage/oasis/nodeapi/file/node.go
```diff
@@ -0,0 +1,149 @@
...
+
+// FileConsensusApiLite provides access to the consensus API of an Oasis node.
+// Since FileConsensusApiLite is backed by a file containing the cached responses
+// to `ConsensusApiLite` calls, this data is inherently compatible with the
+// current indexer and can thus handle heights from both Cobalt/Damask.
+type FileConsensusApiLite struct {
+	currHeight      int64
+	currData        *nodeapi.ConsensusApiLiteBlockData // cache the current height's data to speed sequential reads.
+	genesisDocument *genesis.Document
+	filename        string
```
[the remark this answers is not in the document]
Andrew7234: got it, switched to pogreb!

tests/api/api-dev.yml
mitjat, line 1: is this file accidentally checked in? if not: you have an `andrewlow` reference in it.

storage/oasis/nodeapi/file/node.go
```diff
@@ -0,0 +1,293 @@
...
+		damaskApi: damaskApi,
+	}, nil
+}
+
+func generateCacheKey(methodName string, params ...interface{}) ([]byte, error) {
+	var buf bytes.Buffer
+	enc := gob.NewEncoder(&buf)
+	err := enc.Encode(methodName)
+	if err != nil {
+		return nil, err
```
mitjat, line 53: WDYT about just panicking here and below, and simplify the return value to not include `error`? It would simplify callsites, and I don't see a reason `.Encode()` to suddenly return an error anyway. Only in some corner cases like null pointers or weird types, but those properties should not change for us from height to height.

storage/oasis/nodeapi/file/node.go
```diff
@@ -0,0 +1,293 @@
...
+	"github.com/oasisprotocol/oasis-indexer/storage/oasis/nodeapi/damask"
+)
+
+// FileConsensusApiLite provides access to the consensus API of an Oasis node.
+// Since FileConsensusApiLite is backed by a file containing the cached responses
+// to `ConsensusApiLite` calls, this data is inherently compatible with the
+// current indexer and can thus handle heights from both Cobalt/Damask.
+type FileConsensusApiLite struct {
+	db        pogreb.DB
+	damaskApi *damask.DamaskConsensusApiLite
```
mitjat, line 26: Prefer the `ConsensusApiLite` interface, not the specific `DamaskConsensusApiLite` impl. This way, we can back this cache with the upcoming "universal" ApiLite that works for all heights (by switching between Cobalt and Damask)

storage/oasis/nodeapi/file/node.go
```diff
@@ -0,0 +1,293 @@
+package file
+
+import (
+	"bytes"
+	"context"
+	"encoding/gob"
+	"encoding/json"
```
mitjat, line 7: Why are we doing both?

In particular, why json? I was surprised by the commit that switches to json. (Thank you for relatively clean commits, btw!)  It adds verbose annotations all over prod types, and it bloats the output by a lot. It's nice to have a human-readable cache, but it's hidden in pogreb's DB anyway. And if wanted to make enable fast reindexing by always writing to cache and essentially reproducing the relevant node state on indexer's disk (that was one of my hopes for this work), we'll end up with hundreds of GB, and then json's verbosity might matter. Though to be honest, it also might not :).

storage/oasis/consensus.go
```diff
@@ -23,13 +23,13 @@ import (
 //     boilerplate. Consider removing most types from this package, and
 //     using nodeapi types directly.
 type ConsensusClient struct {
-	nodeApi nodeapi.ConsensusApiLite
+	NodeApi nodeapi.ConsensusApiLite
```
mitjat, line 26: Looks like this was made public just so you can do `cc.NodeApi = ...` in a test. Not very clean; can we create an alternative constructor instead, where you pass in the NodeApi?

Same comment for the newly-public `Main.Cfg` in `analyzer/consensus/consensus.go`.

tests/api/util_test.go
```diff
@@ -0,0 +1,87 @@
...
+	"github.com/oasisprotocol/oasis-indexer/config"
+	"github.com/oasisprotocol/oasis-indexer/log"
+	source "github.com/oasisprotocol/oasis-indexer/storage/oasis"
+	fileSource "github.com/oasisprotocol/oasis-indexer/storage/oasis/nodeapi/file"
+)
+
+const ConfigFile = "/Users/andrewlow/oasis/oasis-block-indexer/tests/api/api-dev.yml"
+const DestFile = "/Users/andrewlow/oasis/oasis-block-indexer/tests/api/data/consensus"
+
+func TestDumpConsensusNodeData(t *testing.T) {
```
mitjat, line 27: This runs the analyzer so as to populate the cache, right?

I would much prefer for this to be expressible through the config, so we can run with pogreb caching enabled even in prod. (And we'd manually run it once in a blue moon to generate the node responses for, say, ~1000 blocks so that we have inputs for e2e tests.)
I have hopes for this caching beyond just tests :)

mitjat (commented):
Nice! Really curious about the speed of running the analyzer against the cached data.

Also, looks like you already intended to split it this way, but let's worry about integrating this into CI in a separate PR.

storage/oasis/nodeapi/file/node.go
```diff
@@ -0,0 +1,293 @@
...
+	return json.Unmarshal(res, result)
+}
+
+func (c *FileConsensusApiLite) put(key []byte, val interface{}) error {
+	valBytes, err := json.Marshal(val)
+	if err != nil {
+		fmt.Printf("file backend: error encoding value for put " + err.Error())
+		return err
+	}
+	return c.db.Put(key, valBytes)
```
mitjat, line 78: Can you check if `pogreb` is thread-safe? If not, let's hide `Put()` and `Get()` and `Has()` any other relevant methods behind a mutex.

Andrew7234: Yep it's goroutine-safe

storage/oasis/consensus.go
Andrew7234: Agreed, these kludges will be removed in favor of config file changes + custom constructors. I'm planning to add the config changes once 352 goes in

tests/api/data/consensus/00000-1.psg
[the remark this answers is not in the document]
Andrew7234: Yep they're all generated by pogreb. I didn't mean to check these files in; will remove.

If/when we want to use the file-based cache to speed up reindexing, we can check in those files or host them somewhere. For this PR, however, I'm expecting this to only be used for tests.

tests/api/util_test.go
Andrew7234: Yup, I'm using this as a placeholder for now until I make changes to the config. I just didn't want to create unnecessary conflcts with #352 . I think I'll eventually be able to remove this file entirely and use makefile commands + two diff config files instead.

storage/oasis/nodeapi/file/node.go
Andrew7234: Good catch, I'll update. Thanks!

storage/oasis/nodeapi/file/node.go
Andrew7234: Hm I want to test that this assumption is still true after adding RuntimeApiLite support. In particular, the `EVMSimulateCall(ctx context.Context, round uint64, gasPrice []byte, gasLimit uint64, caller []byte, address []byte, value []byte, data []byte) ([]byte, error)` seems like it could be potentially problematic.

That being said; there's no real point in error handling/retrying if this fails so I'll update it to panic. Thanks!

tests/api/data/consensus/00000-1.psg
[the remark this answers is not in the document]
mitjat: SG!
If/when we use the file cache to speed up reindexing, we'll host them locally; they're too big for git, and they don't conceptually belong with the code either.
OTOH as soon as we integrate this test with CI, we'll check in these files with some small number (1000?) of blocks. That way, we can run e2e tests a) without network IO flakiness and b) even in CI, which doesn't have access to any nodes.

storage/oasis/nodeapi/file/node.go
mitjat: :+1: 
Yeah I wasn't trying to say these won't ever fail (especially as we add/change types), just that it is very unlikely to start failing at some random block. If it fails, it will do so immediately, and we'll catch it in local testing or e2e, where panicking is fine.

storage/oasis/nodeapi/file/node.go
mitjat: Interesting. Thank you for the link. If the first commenter is right, our problem is specifically `Quantity` values that appear in Go maps. Funky.

Regardless, my suggestion here would be to go with cbor everywhere. We know that the types that come from oasis-node can be round-tripped (encode-decode) via cbor, because we already use that functionality. All the custom (de)serializers are implemented and continuously tested in the real world. With JSON and/or gob, I'm slightly concerned that some custom types might change in subtle ways for edge case values, for example, because we never tested all the node types with those encodings.

storage/oasis/consensus.go
mitjat: Oh ok great.
325 is pretty much ready now; if you want to work on this PR, you can safely rebase on top of 325 even before it goes in.

storage/oasis/nodeapi/file/consensus.go
```diff
@@ -0,0 +1,226 @@
...
+	}, nil
+}
+
+func (c *FileConsensusApiLite) get(key []byte, result interface{}) error {
+	res, err := c.db.Get(key)
+	if err != nil {
+		return err
+	}
+	return cbor.Unmarshal(res, result)
+}
```
Andrew7234, line 44: hrm `cbor` complains with `cbor: invalid map key type: slice` due to unmarshalling into the `result` interface here.

storage/oasis/nodeapi/file/runtime.go
```diff
@@ -0,0 +1,124 @@
...
+	var blockHeader nodeapi.RuntimeBlockHeader
+	err := r.get(key, &blockHeader)
+	if err != nil {
+		return nil, err
+	}
+	return &blockHeader, nil
+}
+
+func (r *FileRuntimeApiLite) GetTransactionsWithResults(ctx context.Context, round uint64) ([]nodeapi.RuntimeTransactionWithResults, error) {
+	key := generateCacheKey("GetRuntimeTransactionsWithResults", round) // todo: maybe remove "runtime" from key; if we don't need to worry about collisions with the consensus api method of the same name. are we always guaranteed the diff db between consensus/runtime?
```
mitjat, line 80: All keys for the runtime API need to contain the runtime name to prevent clashes between e.g. emerald and sapphire. That will also take care of differentiating from consensus.

storage/oasis/nodeapi/file/consensus.go
pro-wh: this was that event `Body interface{}` problem right?

Andrew7234: yep

storage/oasis/nodeapi/api.go
```diff
@@ -79,36 +80,41 @@ type TxError consensusResults.Error
...
-	RegistryNode              *NodeEvent
-	RegistryNodeUnfrozen      *NodeUnfrozenEvent
-
-	RoothashExecutorCommitted *ExecutorCommittedEvent
-
-	GovernanceProposalSubmitted *ProposalSubmittedEvent
-	GovernanceProposalExecuted  *ProposalExecutedEvent
-	GovernanceProposalFinalized *ProposalFinalizedEvent
-	GovernanceVote              *VoteEvent
+	Type apiTypes.ConsensusEventType `json:"type"`
```
mitjat, line 96: If we no longer need the json annotations, I'let's remove them to keep things simple.
(I think they were not strictly needed even when we were serializing into JSON.)

tests/e2e_regression/run.sh
```diff
@@ -103,7 +104,7 @@ for (( i=0; i<nCases; i++ )); do
     >/tmp/pretty 2>/dev/null \
   && cp /tmp/pretty "$outDir/$name.body" || true
   # Sanitize the current timestamp out of the response header so that diffs are stable
-  sed -E -i 's/^(Date|Content-Length|Last-Modified): .*/\1: UNINTERESTING/g' "$outDir/$name.headers"
+  sed -E -i '' 's/^(Date|Content-Length|Last-Modified): .*/\1: UNINTERESTING/g' "$outDir/$name.headers"
```
mitjat, line 107:
```suggestion
  sed -E -i='' 's/^(Date|Content-Length|Last-Modified): .*/\1: UNINTERESTING/g' "$outDir/$name.headers"
```

Otherwise in linux/bash (really, On My Machine (TM)), `''` is taken to be the command and `s/.../` is taken to be another file path.

tests/e2e_regression/run.sh
```diff
@@ -69,18 +69,19 @@ testCases=(
 nCases=${#testCases[@]}
 
 # Kill background processes on exit. (In our case the indexer API server.)
+# todo: this doesn't work on MacOS.
```
mitjat, line 72: Do you know what the culprit is? Longer term, it might be worth coding this up in Go or python I guess ...

tests/e2e_regression/run.sh
```diff
@@ -69,18 +69,19 @@ testCases=(
...
-./oasis-indexer --config="${SCRIPT_DIR}/e2e_config.yaml" serve &
+./oasis-indexer --config="${SCRIPT_DIR}/../api/api-dev.yml" serve &
 while ! curl --silent localhost:8008/v1/ >/dev/null; do
   echo "Waiting for API server to start..."
   sleep 1
 done
 
 # Run the test cases.
-seen=()
+seen=("placeholder") # avoids 'seen[*]: unbound variable' error on zsh
```
mitjat, line 84: Just curious: because of the `#!/bin/bash` shebang, don't you have to go to extra trouble to run this in zsh? Alternatively, is there an invocation where you can explicitly invoke the script with bash and avoid zsh compat issues?

config/config.go
```diff
@@ -117,6 +117,9 @@ type AnalyzersList struct {
 
 // SourceConfig has some controls about what chain we're analyzing and how to connect.
 type SourceConfig struct {
+	// Cache holds the configuration for a file-based caching backend.
+	Cache *CacheList `koanf:"cache"`
```
mitjat, line 121: Do you have a use case in mind that benefits from having separate configs for consensus and runtime? I can't, and would suggest a single `Cache *CacheConfig` here.

Even if they ever needed to be separate, I think it's more likely we'd want to have per-runtime control, e.g. "cache sapphire but not emerald or consensus". But let's not go there.

Makefile
```diff
@@ -58,6 +58,18 @@ test-e2e: export OASIS_INDEXER_E2E = true
 test-e2e:
 	@$(GO) test -race -coverpkg=./... -coverprofile=coverage.txt -covermode=atomic -v ./tests/e2e
 
+dump-state: oasis-indexer
+	sed -E -i '' 's/query_on_cache_miss: false/query_on_cache_miss: true/g' "tests/api/api-dev.yml"
```
mitjat, line 62: `i=''`, please. Here and a few lines below. See comment in `run.sh`.

Makefile
```diff
@@ -58,6 +58,18 @@ test-e2e: export OASIS_INDEXER_E2E = true
...
+	./oasis-indexer --config tests/api/api-dev.yml
+
+# Run the api tests locally, assuming the environment is set up with an oasis-node that is 
+# accessible as specified in the config file.
+test-api: oasis-indexer
+	sed -E -i '' 's/query_on_cache_miss: true/query_on_cache_miss: false/g' "tests/api/api-dev.yml"
```
mitjat, line 68: Here and in dump-state: The analyzer we'll be running on top of this cache is configured by `tests/e2e_regression/e2e_config.yaml`, so the cache prework should happen based on that too. Edit: Oh, I see you changed the paths. I don't agree with the new name: This is an e2e test (analyzer+api), not just an api test. We do already have a couple e2e tests based on synthetic data in `tests/e2e`, so I put the `run.sh` test into `tests/e2e_regression` for disambiguation: The term "regression testing" is often used somewhat loosely to mean testing against (a snapshot of) real-world data and usage patterns.
So please do move `tests/api/*` back under `tests/e2e_regression`. I'm also open to other names if e2e_regression is confusing/misleading, but not this one :)

Thanks for trying to make names make sense though. It's easy to leave behind a naming mess that made sense when you (the generic you; here, Mitja) were writing it, but does not objectively work very well.

config/config.go
```diff
@@ -168,6 +171,30 @@ func SingleNetworkLookup(rpc string) map[string]*NodeConfig {
...
+	Runtime *CacheConfig `koanf:"runtime"`
+}
+
+type CacheConfig struct {
+	// File is the directory where the cache data is stored
+	File string `koanf:"file"`
+
+	// If set, the indexer will query the node upon any cache
+	// misses.
+	QueryOnCacheMiss bool `koanf:"query_on_cache_miss"`
```
mitjat, line 188: Is it possible to get rid of this flag? 
I see two scenarios:
- We expect cache misses. Then we want this to be true, obviously.
- We do not expect cache misses. Expected use case are offline e2e tests. If there really are no cache misses, great, and this flag doesn't matter. If there _is_ a cache miss though ... this flag will only change the behavior from getting a network error to getting some slightly different error message. That seems not worth it.

(It does become a more meaningful flag if we want to verify that a run has no cache misses even when we do have access to a node. But I don't see a use case for that, other than e2e tests, and CI will alert us to any problems there, albeit slightly less ergonomically. Or, to test before CI, one could just set the node URL to a bad string to ensure the analyzer is "offline".)

Makefile
```diff
@@ -58,6 +58,18 @@ test-e2e: export OASIS_INDEXER_E2E = true
...
+	./oasis-indexer --config tests/api/api-dev.yml
```
mitjat, line 69:
```suggestion
	./oasis-indexer --config tests/api/api-dev.yml analyze
```
or else the server part will keep this process running forever

storage/oasis/nodeapi/file/runtime.go
Andrew7234: Good call; updated!

storage/oasis/nodeapi/file/consensus.go
```diff
@@ -0,0 +1,226 @@
...
+	var genesisDocument genesis.Document
+	err := c.get(key, &genesisDocument)
+	if err != nil {
+		return nil, err
+	}
+	return &genesisDocument, nil
+}
+
+func (c *FileConsensusApiLite) StateToGenesis(ctx context.Context, height int64) (*genesis.Document, error) {
+	key := generateCacheKey("StateToGenesis", height)
```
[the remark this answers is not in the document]
Andrew7234: Good catch; updated! The file-based analyzer will now try to route the call to the nodeApi if it exists; otherwise it will error.

Makefile
Andrew7234: thanks, updated!

config/config.go
```diff
@@ -168,6 +171,30 @@ func SingleNetworkLookup(rpc string) map[string]*NodeConfig {
...
+type CacheList struct {
+	// Consensus holds the configuration for the consensus cache
+	Consensus *CacheConfig `koanf:"consensus"`
+
+	// Runtime holds the configuration for the runtime cache
+	Runtime *CacheConfig `koanf:"runtime"`
+}
+
+type CacheConfig struct {
+	// File is the directory where the cache data is stored
```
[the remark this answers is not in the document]
Andrew7234: Oops; updated!

config/config.go
Andrew7234: Ah I was actually going to move further in the direction of per-runtime configurations. I think it'll be useful to store the consensus data and runtime(s) data separately. That way, changes to the consensus or runtime analyzers would only require updating the corresponding dataset. And if we do end up using the file-based indexer to speed up full reindexes, it'll save us from having to re-pull a lot of data from node.

An alternative would be to just list out all the cache directories in a single `CacheConfig`, and have a single `QueryOnCacheMiss` flag.

config/config.go
Andrew7234: Ah good eye; I don't think it's possible to remove the flag entirely but first some context:

This flag was originally intended to tell the analyzer whether to try to connect to a node on startup. This is necessary for two reasons. 1) During startup, the analyzer tries to connect to the node and if the connection cannot be made (either a bad url config or unreachable node), the analyzer errors out. 2) In CI(outside of docker-compose), a node connection will not be available. Therefore, the indexer needs to know whether to _attempt_ to connect to the node, and that was the original purpose of the `QueryOnCacheMiss` flag. 

However, during a later commit the logic was removed and the indexer currently always tries to connect to the node on startup. I think it's still important to allow file-based indexers to be able to run without a node connection (for CI / full reindexes) though, so I'm planning to re-add this logic in this PR, at least for the consensus analyzer. The runtime analyzer is a bit trickier to handle because it fetches the `runtimeInfo` from node during startup, so we'd need to store that offline in the config as well.

storage/oasis/nodeapi/file/node.go
Andrew7234: got it; updated to serialize all the data using oasis-core's `cbor`. I kept gob for serializing the cache keys since I'm not sure that oasis-core's `cbor` is meant for this use case.

tests/e2e_regression/run.sh
Andrew7234: Er not off the top of my head; but I'll try running it with `bash` from now on instead of `zsh` which should hopefully resolve it.

Makefile
Andrew7234: Thanks for the thorough explanation! I have moved everything from `tests/api/*` to `tests/e2e_regression/*`. I was hesitant to modify the `e2e_regression` dir in case there was a broader plan for it, but I see now that the file-based indexer tests do indeed belong there.

Makefile
```diff
@@ -58,6 +58,18 @@ test-e2e: export OASIS_INDEXER_E2E = true
 test-e2e:
 	@$(GO) test -race -coverpkg=./... -coverprofile=coverage.txt -covermode=atomic -v ./tests/e2e
 
+dump-state: oasis-indexer
+	sed -E -i='' 's/query_on_cache_miss: false/query_on_cache_miss: true/g' "tests/e2e_regression/e2e_config.yml"
+	./oasis-indexer --config tests/e2e_regression/e2e_config.yml analyze
```
mitjat, line 63:
```suggestion
fill-cache-for-e2e-regression: oasis-indexer
	cp tests/e2e_regression/e2e_config.yml /tmp/indexer_fill_e2e_regression_cache.yml
	sed -E -i='' 's/query_on_cache_miss: false/query_on_cache_miss: true/g' /tmp/indexer_fill_e2e_regression_cache.yml
	./oasis-indexer --config /tmp/indexer_fill_e2e_regression_cache.yml analyze
```

Two nits:
1) Renamed the target to relate to the name of the corresponding test target
2) I'd prefer to keep the use case for `tests/e2e_regression/e2e_config.yml` clear: Running the test. That's what its name+location imply, and also we'll be running the test a lot more often than we'll be filling the cache. So the suggested change above makes a temporary copy of the .yml and modifies it just for the purpose of filling the cache.

Makefile
```diff
@@ -58,6 +58,18 @@ test-e2e: export OASIS_INDEXER_E2E = true
...
+
+# Run the api tests locally, assuming the environment is set up with an oasis-node that is 
+# accessible as specified in the config file.
+test-api: oasis-indexer
```
mitjat, line 67:
```suggestion
test-e2e-regression: oasis-indexer
```

Makefile
```diff
@@ -58,6 +58,18 @@ test-e2e: export OASIS_INDEXER_E2E = true
...
+	sed -E -i='' 's/query_on_cache_miss: true/query_on_cache_miss: false/g' "tests/e2e_regression/e2e_config.yml"
```
mitjat, line 68: This line can go away after the suggested change in `dump-state` target.

config/config.go
mitjat: > That way, changes to the consensus or runtime analyzers would only require updating the corresponding dataset.

Hopefully, changes to analyzers will not require any changes to the cached dataset, because we're caching the _input_ to the analyzers. The cache might need to be expanded with new results of new RPCs, but there's no benefit to per-runtime configs there. Or, in the rare case, the NodeApiLite might need to start producing a differently-normalized version of the node response. In that case, we might want to throw away some cache keys, but again at a different granularity than per-runtime. If that happens, we can still modify the FileApiLite layer so it updates the cached entry if it's not using the newest format.

The suggestion in your last paragraph is very close to my suggestion actually. And there is potential value to having separate DBs per runtime. Let's go with your last paragraph; I would just further suggest to simplify a tiny bit more and take a single directory path in the config, then auto-derive subfolders in the FileApiLite impls. So config would say something like `/data/oasis-indexer/rpc-cache`, and DBs would be created in `/data/oasis-indexer/rpc-cache/consensus`, `/data/oasis-indexer/rpc-cache/emerald`, itd.

config/config.go
mitjat: Yes, it's crucial we allow fully-offline execution, CI being the strongest / most immediate reason.

Bummer about the underlying "real" ApiLite impls trying to connect immediately; I missed that. In an ideal world I would prefer to change those ApiLites so they connect lazily (= when the first RPC requires a connection) and then we wouldn't need this flag and its complications. But let's not go there; this flag is a fine alternative. Thanks for explaining!

storage/oasis/client.go
```diff
@@ -19,10 +20,28 @@ const (
...
+		nodeApi, err := file.NewFileConsensusApiLite(sourceConfig.Cache.Consensus.CacheDir, nil)
+		if err != nil {
+			return nil, fmt.Errorf("error instantiating cache-based consensusApi: %w", err)
+		}
+		return &ConsensusClient{
+			nodeApi: nodeApi,
+			network: sourceConfig.SDKNetwork(),
+		}, nil
+	}
+	var nodeApi nodeapi.ConsensusApiLite
```
mitjat, line 34:
```suggestion

	// Create an API that connects to the real node, then wrap it in a caching layer.
```

I think we don't need the `var`? But mostly I just wanted to visually separate the two `QueryOnCacheMiss` alternatives, with a newline in between and a comment for each.

storage/oasis/client.go
```diff
@@ -49,8 +68,24 @@ func NewRuntimeClient(ctx context.Context, sourceConfig *config.SourceConfig, ru
 		return nil, err
 	}
 
+	// todo: short circuit if using purely a file-based backend and avoid connecting
```
mitjat, line 71: Do you plan this for this PR still? Looks like it should be very similar to the solution in `NewConsensusClient`.

Fine to leave it for a later PR (likely the one that integrates into CI) too.

mitjat (approved):
Thank you! I left only one more not-entirely-trivial comment, but I think it's not contentious either.

mitjat:
Reviewing the code made me wonder if it can be written with less repetition. My proof of concept ended up working out; I pushed it to this PR as three additional commits. Benefits of the slight refactor:
- less code + less repetition
- for calls where the value is not yet in the cache, avoids deserialization
Downside: Potentially hared-to-understand code?

I only made the changes on the consensus side; I feel the runtime side has too few methods to warrant generalizing them. Although I guess the whole `getFromCacheOrCall()` could be moved into a new `KVStore` struct that is an abstraction over `pogreb`; a good idea regardless in case we want to mock or replace pogreb later.

Disclaimer: This is untested beyond satisfying the compiler. And we don't have the habit of slipping into each other's PRs. So I'm happy to submit this as a follow-up PR too, or we can leave it as a part of this one, whichever you prefer.

mitjat:
Uh a pretty big problem I noticed: the docs for `pogreb.Open()` state: "opens or creates a new DB. The DB must be closed after use, by calling Close method."
We'll have to give a Close() or Shutdown() method to ApiLite structs and analyzers and pipe that through to pogreb :/. Otherwise, it sounds like we're risking data corruption.

storage/oasis/client.go
Andrew7234: I'd prefer to leave it for a future PR; storing the `runtimeInfo` offline requires additional config changes, and I'd like to do more digging into why we fetch the info from the node and how it's used.

storage/oasis/client.go
Andrew7234: Hm so we need the `var` or else golang's assumed type for `nodeApi` is `HistoryConsensusApiLite` which is too strict if we wrap/reassign it as a `FileConsensusApiLite` right after. I added the newline and comment; thanks!

Andrew7234:
> submit this as a follow-up PR too, or we can leave it as a part of this one, whichever you prefer.

Thanks for the commits! I'm good with keeping them in this PR

>  pipe that through to pogreb
Hrm we have quite a few layers to get through :\ I can get a commit in later today for this as well if you'd prefer, just lmk!

mitjat:
>> We'll have to give a Close() or Shutdown() method to ApiLite structs and analyzers and pipe that through to pogreb

> Hrm we have quite a few layers to get through :\ I can get a commit in later today for this as well if you'd prefer

I've actually already implemented it, but haven't pushed it to github yet. Let's not block this PR further, and do that in a separate PR; it won't be tiny either. In my early experiments, pogreb often (always?) recovered a not-cleanly-closed DB on the next start.

storage/oasis/client.go
```diff
@@ -31,13 +36,19 @@ func NewConsensusClient(ctx context.Context, sourceConfig *config.SourceConfig)
...
+	// Create an API that connects to the real node, then wrap it in a caching layer.
 	var nodeApi nodeapi.ConsensusApiLite
 	nodeApi, err := history.NewHistoryConsensusApiLite(ctx, sourceConfig.History(), sourceConfig.Nodes, sourceConfig.FastStartup)
 	if err != nil {
 		return nil, fmt.Errorf("instantiating history consensus API lite: %w", err)
 	}
-	if sourceConfig.Cache != nil && sourceConfig.Cache.Consensus != nil {
-		nodeApi, err = file.NewFileConsensusApiLite(sourceConfig.Cache.Consensus.CacheDir, nodeApi)
+	if sourceConfig.Cache != nil {
+		cachePath := sourceConfig.Cache.CacheDir + "/consensus"
```
mitjat, line 47: Nit: Prefer `filepath.Join()` to avoid the risk of missing or duplicated `/`.
In fact, the current code sort-of assumes that `CacheDir` does not have a trailing `/` here, and then assumes that `CacheDir` does have one in `NewRuntimeClient()` :)

storage/oasis/client.go
```diff
@@ -31,13 +36,19 @@ func NewConsensusClient(ctx context.Context, sourceConfig *config.SourceConfig)
...
+		if err := os.MkdirAll(cachePath, os.ModeDir); err != nil {
+			return nil, fmt.Errorf("error creating cache dir for consensusApi: %w", err)
+		}
```
mitjat, line 50: Not needed; `pogreb` will do this.

storage/oasis/client.go
mitjat: Ah thank you, I overlooked the `info`. And you even mentioned it in a previous comment IIRC, but I didn't make the connection. Yeah SG to leave for a future PR.

mitjat (approved):
Thank you for the final tweaks! Finally time to rebase and :sheep: it.

I created subtasks for https://app.clickup.com/t/861md6gue to capture the remaining work we decided to keep out of this PR.

storage/oasis/client.go
Andrew7234: Oops my bad, thanks for the catch! Updated

analyzer/runtime/extract.go
```diff
@@ -248,7 +248,7 @@ func registerTokenDecrease(tokenChanges map[TokenChangeKey]*big.Int, contractAdd
 	change.Sub(change, amount)
 }
 
-func ExtractRound(blockHeader nodeapi.RuntimeBlockHeader, txrs []*nodeapi.RuntimeTransactionWithResults, rawEvents []*nodeapi.RuntimeEvent, logger *log.Logger) (*BlockData, error) { //nolint:gocyclo
+func ExtractRound(blockHeader nodeapi.RuntimeBlockHeader, txrs []nodeapi.RuntimeTransactionWithResults, rawEvents []nodeapi.RuntimeEvent, logger *log.Logger) (*BlockData, error) { //nolint:gocyclo
```
pro-wh, line 251: what did these * removals do? originally they were pointers so that we don't have to copy structs around. not that that's particularly expensive.

common/types.go
```diff
@@ -90,6 +91,16 @@ func Ptr[T any](v T) *T {
 	return &v
 }
 
+// Returns `v` as a JSON string. If `v` cannot be marshaled,
+// returns the string "null" instead.
+func TryAsJSON(v interface{}) json.RawMessage {
```
pro-wh, line 96: didn't this appear in an earlier PR? what happened to that? now I can't find it

storage/oasis/client.go
pro-wh: this is because we didn't go for that type-erased-constructors change last sprint :smiling_face_with_tear:

common/types.go
pro-wh: oh that was #382, turns out it never got merged

analyzer/runtime/extract.go
pro-wh: oh it's so that we don't need a zillion different GetFromCacheOrCall variants, right?

pro-wh:
nice, thanks!

PR #361, opened 2023-03-23
statecheck: Do not use indexer-internal APIs for node access

mitjat:
Before this PR, statecheck used indexer's internal libraries to access the node. This
- Makes for less than ideal separation: Indexer proper now has methods that exist only for tests to run.
- Makes it slightly harder/messier/more boilerplate-y to refactor those libraries for multiple-version support.

This PR makes the statecheck depend on the indexer less (actually not at all, except for DB utility wrappers). It does mean that we concede we won't run statecheck on anything but the version of oasis-core and oasis-sdk that it was compiled against. Not a loss IMO; we don't need huge generality there.

Testing: None :grimacing: because it's a bit of a pain to do, and the changes are such that the compiler should have caught any problems. Also, the statecheck is not hooked up to any alerts yet.

### DIFF
--- a/tests/statecheck/consensus_test.go
+++ b/tests/statecheck/consensus_test.go
@@ -10,6 +10,7 @@ import (
 
 	"github.com/oasisprotocol/oasis-core/go/common/entity"
 	"github.com/oasisprotocol/oasis-core/go/common/node"
+	consensusAPI "github.com/oasisprotocol/oasis-core/go/consensus/api"
 	genesisAPI "github.com/oasisprotocol/oasis-core/go/genesis/api"
 	governanceAPI "github.com/oasisprotocol/oasis-core/go/governance/api"
 	registryAPI "github.com/oasisprotocol/oasis-core/go/registry/api"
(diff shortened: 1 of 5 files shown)


aefhm (approved):

pro-wh:
o good, I can make the diff in #352 smaller

tests/statecheck/consensus_test.go
```diff
@@ -137,11 +137,9 @@ func TestGenesisFull(t *testing.T) {
 
 	ctx := context.Background()
 
-	factory, err := newSourceClientFactory()
-	require.Nil(t, err)
-
-	oasisClient, err := factory.Consensus()
+	conn, err := newSdkConnection(ctx)
 	require.Nil(t, err)
+	oasisClient := conn.Consensus()
```
pro-wh, line 142: :information_source:  the factory would do signature.SetChainContext, where the SDK connection doesn't. we have a few `.Open()`, but luckily they're on registry things that don't use the chain context

Andrew7234: thanks for the tip; i was making the same mistake in the api-tests

PR #366, opened 2023-03-28
bugfix: Ignore entities with no address

mitjat:
Fixes a bug where the the indexer API potentially returned a HTTP 5xx because it tried to scan a NULL address for an entity into a string. See inline comments for explanation.

The bug was introduced in https://github.com/oasisprotocol/oasis-indexer/pull/351 which changed `UPDATE chain.entities SET meta=...` into an upsert, potentially creating a new entity row with no known address. The regression wasn't caught by that PR's e2e tests because it depends on a specific sequencing of two analyzers that (can) work in parallel.

We could introduce the old behavior (UPDATE instead of upsert), but the current behavior has the advantage that we'll be able to show an entity with all its metadata as soon as we scan its registration on the chain, instead of waiting for the metadata analyzer to do its periodic re-scan of git. I'm open to both solutions though; the advantage of UPDATE is that it's conceptually a little simpler.

### DIFF
--- a/storage/client/queries/queries.go
+++ b/storage/client/queries/queries.go
@@ -98,15 +98,26 @@ const (
 
 	Entities = `
 		SELECT id, address
-			FROM chain.entities
+		FROM chain.entities
+		-- The metadata_registry analyzer does not receive entity addresses from git, only metadata.
+		-- Addresses come from registry events in the consensus layer. Ignore entities that only the
+		-- metadata_registry analyzer knows about; they are irrelevant to the current state of the
+		-- chain, and our API guarantees a non-null address.
+		WHERE (address IS NOT NULL)
 		ORDER BY id
 		LIMIT $1::bigint
 		OFFSET $2::bigint`
 
 	Entity = `
 		SELECT id, address
-			FROM chain.entities
-			WHERE id = $1::text`
+		FROM chain.entities
+		WHERE
+			(id = $1::text) AND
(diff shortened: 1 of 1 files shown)


pro-wh:
what are the circumstances for this happening? what kind of data does the metadata registry analyzer produce? how is it keyed? what do we need this address for? the registry uses public keys as keys

mitjat:
> what are the circumstances for this happening? what kind of data does the metadata registry analyzer produce? how is it keyed? what do we need this address for? the registry uses public keys as keys

The circumstances: Block analyzer falls behind (or is inherently behind because it's doing a reindex), metadata analyzer always fetches newest data from git. So it can easily see entities that haven't appeared on the part of the chain that had been indexed so far.

Metadata is usually a human-readable name (e.g. "BitCat"). I think we also support things like homepage URL, avatar, etc. It's a freeform JSON from indexer's perspective.

Entities are keyed by ID, which is `base64_ed25519_pubkey`. The address ... fair question, it's just derived from the ID. IDK how it's used, if at all. I assumed it was derived somewhere deep within oasis-core, and I didn't want to go hunt down the generation code and then reproduce it (potentially in a brittle way) in the indexer. But it turns out that we [always derive it in the indexer](https://github.com/oasisprotocol/oasis-indexer/blob/90968f1819dcf486c74df323d52b48f081a7ebfb/analyzer/consensus/consensus.go#L535-L535)!

Following up in https://github.com/oasisprotocol/oasis-indexer/pull/371

PR #367, opened 2023-03-28
 storage client: get entire address preimage 

pro-wh:
~~fixes #365~~

1. adjust some specs to use our custom string Address type
2. ~~get all related address preimage info when querying transaction~~
3. ~~don't set incorrect xxx_eth address when preimage is not secp256k1eth~~

EthAddrReference is moved out to client package. changed to EthChecksumAddrFromPreimage

PR #368, opened 2023-03-29
refactor: analyzer/runtime: Store a more-parsed tx in the db

mitjat:
This PR changes the stage at which we process runtime txs: The processing used to be spread between the analyzer and a "renderer" at http api call time; it now happens entirely in the analyzer, and the parsed tx is stored in the db.

This is a no-op externally, but allows us
1) one-off queries/analyses in the db
2) implementation of filters on inner fields, such as `method = accounts.Transfer`

The first two commits are just pre-work and could well be separate small PRs. I recommend reviewing separately.

**Testing:** `reindex_and_run.sh` for 5000 rounds.
Diffs: The body of `consensus.Withdraw` txs is now serialized in a more verbose/interpretable way: Before:
```json
      "body": {
        "amount": [
          "HLs6P/CNCAAA",
          ""
        ],
        "to": "AJifyI0wEVN16b6H61PxYQ/Hkvaw"
      },
```
After:
```json
      "body": {
        "amount": {
          "Amount": "530000000000000000000",
          "Denomination": ""
        },
        "to": "oasis1qzvfljydxqg4xa0fh6r7k5l3vy8u0yhkkqak5uaq"
      },
```


storage/client/queries/queries.go
```diff
@@ -313,33 +313,46 @@ const (
...
 			-- Otherwise, every tx will be returned as many times as there are related addresses for it.
-			AND $4::text IS NOT NULL
-		WHERE (txs.runtime = $1) AND
-					($2::bigint IS NULL OR txs.round = $2::bigint) AND
-					($3::text IS NULL OR txs.tx_hash = $3::text OR txs.tx_eth_hash = $3::text) AND
-					($4::text IS NULL OR rel.account_address = $4::text)
+			($4::text IS NOT NULL)
+		WHERE
+			(txs.runtime = $1) AND
+			(signer0.signer_index = 0) AND
```
Andrew7234, line 352: might be a dumb q, but is it worth moving this line to the join above? eg, `LEFT JOIN chain.runtime_transaction_signers AS signer0 USING (runtime, round, tx_index) ON (signer0.signer_index = 0)`

storage/migrations/02_runtimes.up.sql
```diff
@@ -34,19 +34,30 @@ CREATE TABLE chain.runtime_transactions
...
+  timestamp TIMESTAMP WITH TIME ZONE NOT NULL,
+
   tx_hash     HEX64 NOT NULL,
   tx_eth_hash HEX64,
-  gas_used UINT63 NOT NULL,
+  -- NOTE: The signer(s) and their nonce(s) are stored separately in runtime_transaction_signers.
+
+  fee         UINT_NUMERIC NOT NULL,
+  gas_limit   UINT63 NOT NULL,
+  gas_used    UINT63 NOT NULL,
```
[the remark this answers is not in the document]
mitjat: We do assume a number. Having `NULL` here is an interesting option ... one not discussed in https://github.com/oasisprotocol/oasis-indexer/pull/362#discussion_r1153702055.
I'd like to keep it NOT NULL here so we're not introducing another feature/change in this PR. This PR just puts into the DB what we already had in Go variables, with unchanged semantics.

storage/client/types.go
```diff
@@ -98,30 +96,11 @@ type RuntimeBlockList = api.RuntimeBlockList
...
-	Timestamp       time.Time
-	Raw             []byte
-	ResultRaw       []byte
-	AddressPreimage map[string]string
-}
+type (
+	RuntimeTransactionList = api.RuntimeTransactionList
+	RuntimeTransaction     = api.RuntimeTransaction
+	TxError                = api.TxError
+)
```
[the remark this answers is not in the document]
mitjat: gofumpt/linter wants it this way

mitjat:
> would appreciate another pair of eyes on `extract.go` to double-check that it's not missing anything from `logic.go`.

That is my biggest concern too; e2e regressions at least cover the most common events though. (They might well miss rarer events if they don't happen to appear in that single API query that run.sh does.)

storage/client/queries/queries.go
mitjat: Not dumb at all, might be a little more readable. I debated it myself. A gotcha is that you cannot mix `ON` and `USING`, so we'd have to expand the `USING` into a much longer `ON` block, which I think hurts readability overall. Happy to change it if you have stronger feelings here; I don't.

storage/client/queries/queries.go
```diff
@@ -313,33 +313,46 @@ const (
...
+		LEFT JOIN chain.address_preimages AS signer_eth ON
+			(signer0.signer_address = signer_eth.address) AND
+			(signer_eth.context_identifier = 'oasis-runtime-sdk/address: secp256k1eth') AND (signer_eth.context_version = 0)
+		LEFT JOIN chain.address_preimages AS to_eth ON
+			(txs.to = to_eth.address) AND
+			(to_eth.context_identifier = 'oasis-runtime-sdk/address: secp256k1eth') AND (to_eth.context_version = 0)
+		LEFT JOIN chain.runtime_related_transactions AS rel ON 
+			(txs.round = rel.tx_round) AND
+			(txs.tx_index = rel.tx_index) AND
+			(txs.runtime = rel.runtime) AND
```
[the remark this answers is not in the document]
mitjat: I actually expect it to be faster. (Famous last words ...)
They're very lightweight JOINs, they each look up a single row per tx, and using an index. The nested SELECT we had before is where the engine is more likely to have a hard time optimizing it into a reasonable plan. This just replaces the nested SELECT (with a JOIN of its own, no less!) with two new JOINs.

analyzer/runtime/runtime.go
```diff
@@ -289,18 +289,34 @@ func (m *Main) queueDbUpdates(batch *storage.QueryBatch, data *BlockData) {
...
+			transactionData.GasLimit,
 			transactionData.GasUsed,
 			transactionData.Size,
 			data.Header.Timestamp,
-			transactionData.Raw,
-			transactionData.RawResult,
+			transactionData.Method,
+			transactionData.Body,
+			transactionData.To,
+			transactionData.Amount,
```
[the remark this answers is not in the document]
mitjat: This conversion finally works "automatically" (we do provide a special serialization method on BigInt) as it should since we've migrated to pgx v5. (Edit: Not really, read on)

If you're alluding to `transactionData.Fee.String()` above, that did require the `String()` for some reason ... The difference is that `Fee` is `BigInt` whereas `Amount` is `*BigInt`. I change `Fee.String()` to `&Fee` and now everything (still) works, with no explicit `String()`. :grimacing:

storage/client/queries/queries.go
```diff
@@ -355,7 +355,7 @@ const (
 					($5::text IS NULL OR type = $5::text) AND
 					($6::text IS NULL OR evm_log_signature = $6::text) AND
 					($7::text IS NULL OR related_accounts @> ARRAY[$7::text])
-			ORDER BY round DESC, tx_index
+			ORDER BY round DESC, tx_index, type, body::text
```
pro-wh, line 371: bleh, we ought to file a low priority issue about preserving event order from the node

storage/migrations/02_runtimes.up.sql
```diff
@@ -34,18 +34,34 @@ CREATE TABLE chain.runtime_transactions
...
+
   size UINT31 NOT NULL,
-  timestamp TIMESTAMP WITH TIME ZONE NOT NULL,
+  
+  -- Transaction contents.
+  method      TEXT NOT NULL,  -- accounts.Transter, consensus.Deposit, consensus.Withdraw, evm.Create, evm.Call
+  body        JSON NOT NULL,  -- For EVM txs, the EVM method and args are encoded in here.
```
pro-wh, line 52: is this EVM only? or is the comment about this having a different meaning in some cases?

storage/client/queries/queries.go
```diff
@@ -313,33 +313,42 @@ const (
...
-					json_object_agg(pre.address, encode(pre.address_data, 'hex'))
-				FROM chain.runtime_related_transactions AS rel
-				JOIN chain.address_preimages AS pre ON rel.account_address = pre.address
-				WHERE txs.runtime = rel.runtime
-					AND txs.round = rel.tx_round
-					AND txs.tx_index = rel.tx_index
-			) AS eth_addr_lookup
+			txs.method,
+			txs.body,
+			-- .to, .to_eth, and .amout have to be inferred from the body post-hoc
```
pro-wh: amout -> amount

storage/client/client.go
```diff
@@ -1054,21 +1055,38 @@ func (c *StorageClient) RuntimeTransactions(ctx context.Context, p apiTypes.GetR
...
+			&t.Error.Message,
 		); err != nil {
 			return nil, wrapError(err)
 		}
+		if t.Success != nil && *t.Success {
+			t.Error = nil
+		}
+		// Fancy-format eth addresses: Apply checksum capitalization, prepend 0x.
+		if t.Sender0Eth != nil {
+			*t.Sender0Eth = ethCommon.HexToAddress(*t.Sender0Eth).Hex()
```
pro-wh, line 1091: view logic in storage layer :upside_down_face:

storage/client/queries/queries.go
pro-wh: question about this: does factoring out this WHERE signer0.signer_index = 0 change the meaning of the join outer-sidedness? if we tried doing the same for a hypothetical signer1 field, would the dbms conceptually first come up with a record:

method=accounts.Transfer, signer0=oasis1xxxx, signer1=NULL

and then determine NULL != 0 and discard it?

we're lucky here that everything actually has a signer0

storage/client/queries/queries.go
```diff
@@ -313,33 +313,42 @@ const (
...
+			txs.success,
+			txs.error_module,
+			txs.error_code,
+			txs.error_message
 		FROM chain.runtime_transactions AS txs
-		LEFT JOIN chain.runtime_related_transactions AS rel ON txs.round = rel.tx_round
-			AND txs.tx_index = rel.tx_index
-			AND txs.runtime = rel.runtime
+		LEFT JOIN chain.runtime_transaction_signers AS signer0 USING (runtime, round, tx_index)
+		LEFT JOIN chain.address_preimages AS signer_eth ON
```
pro-wh, line 337: AS signer0_eth

storage/client/queries/queries.go
pro-wh: what's our strategy for to_eth now? second query?

storage/client/queries/queries.go
```diff
@@ -325,7 +325,9 @@ const (
 			txs.size,
 			txs.method,
 			txs.body,
-			-- .to, .to_eth, and .amout have to be inferred from the body post-hoc
+			txs.to,
```
pro-wh, line 328: oh ok

analyzer/uncategorized/quantities.go
```diff
@@ -21,6 +21,6 @@ func StringifyBytes(value []byte) string {
 
 func QuantityFromBytes(value []byte) quantity.Quantity {
 	q := *quantity.NewQuantity()
-	q.FromBigInt(new(big.Int).SetBytes(value))
+	_ = q.FromBigInt(new(big.Int).SetBytes(value))
```
pro-wh, line 24: here's this to put the errors into :hole:

analyzer/runtime/runtime.go
```diff
@@ -304,7 +304,7 @@ func (m *Main) queueDbUpdates(batch *storage.QueryBatch, data *BlockData) {
 			transactionData.Index,
 			transactionData.Hash,
 			transactionData.EthHash,
-			transactionData.Fee.String(),
+			&transactionData.Fee, // pgx bug? Needs a *BigInt (not BigInt) to know how to serialize.
```
pro-wh, line 307: maybe we defined a method only on *BigInt?

pro-wh (commented):
ah that was less hectic than I was afraid of

analyzer/runtime/runtime.go
Andrew7234: v tangential but `encoding/gob` complains if you use pointer receivers like `func (t *MyType) MarshalBytes`... go pointers are funky i guess

storage/client/queries/queries.go
mitjat: I thought they were equivalent. I now see there's a subtle difference:
> a restriction placed in the ON clause is processed before the join, while a restriction placed in the WHERE clause is processed after the join. That does not matter with inner joins, but it matters a lot with outer joins.

See "Notice that placing the restriction in the WHERE clause produces a different result:" [here](https://www.postgresql.org/docs/current/queries-table-expressions.html).

Thanks. For our case ... Perhaps we should keep the WHERE even though it's less "pure"? It makes the join less "outer": In the hypothetical case of a tx without a signer, WHERE would discard that tx, whereas ON would return the tx but with a null signer. But our data types don't support null signers. Hrm it's actually good that we'd then crash and return a 500 for a result involving such a tx.

OK I can change the query. It's all pretty philosophical because we always have a signer, but still.

pro-wh: changing in #367

PR #373, opened 2023-04-03
bugfix: analyzer: Normalize error messages

mitjat:
The analyzer failed at round `370744` of Emerald with `ERROR: invalid byte sequence for encoding "UTF8": 0x00` when trying to insert an error message rendered in logs as 
image
 (can't even be copy-pasted from the web interface). See diff for explanation.

This PR makes it so we slightly normalize error messages into valid UTF-8 without zero bytes. Since error messages are meant for humans, this is a reasonable thing to do even if it technically discards a tiny bit of data.

**Testing:** Indexed the problematic block. Success.


Andrew7234 (approved):

analyzer/consensus/consensus.go
```diff
@@ -420,7 +420,10 @@ func (m *Main) queueTransactionInserts(batch *storage.QueryBatch, data *storage.
 		}
 		var message *string
 		if len(result.Error.Message) > 0 {
-			message = &result.Error.Message
+			// The message shuld be well-formed since it comes from oasis-core.
```
aefhm, line 423:
```suggestion
			// The message should be well-formed since it comes from oasis-core.
```

aefhm (approved):

PR #374, opened 2023-04-04
bugfix: db: Allow NULL tx body (for malformed txs)

mitjat:
Fixes a crash of sapphire indexer on round 144, which seems to be the first one to contain an encrypted tx.
Explanation in comments.

Testing: None, haven't figured out a tunnel to the sapphire node yet.


pro-wh (approved):
also null for encrypted txs and txs with unrecognized methods

pro-wh:
not gonna test on my side

aefhm (approved):

PR #375, opened 2023-04-05
track native token balance changes from evm txs

Andrew7234:
https://app.clickup.com/t/866a11n9q

Todo:
- Testing this against cobalt

### DIFF
--- a/analyzer/runtime/runtime.go
+++ b/analyzer/runtime/runtime.go
@@ -15,7 +15,6 @@ import (
 	"github.com/oasisprotocol/oasis-indexer/analyzer"
 	"github.com/oasisprotocol/oasis-indexer/analyzer/queries"
 	common "github.com/oasisprotocol/oasis-indexer/analyzer/uncategorized"
-	uncategorized "github.com/oasisprotocol/oasis-indexer/analyzer/uncategorized"
 	"github.com/oasisprotocol/oasis-indexer/analyzer/util"
 	"github.com/oasisprotocol/oasis-indexer/config"
 	"github.com/oasisprotocol/oasis-indexer/log"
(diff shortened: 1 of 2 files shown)


aefhm (approved):
LG caveat my context may be limited

analyzer/runtime/runtime.go
```diff
@@ -318,6 +319,23 @@ func (m *Main) queueDbUpdates(batch *storage.QueryBatch, data *BlockData) {
...
 		)
+
+		// Handle native-token transfers from transactions.
+		// These transfers do not emit events and are not covered
+		// by the accounts module, so we process them here.
+		if transactionData.Method == "evm.Call" {
+			body := transactionData.Body.(*evm.Call)
+			if len(body.Data) == 0 {
+				amount := common.QuantityFromBytes(body.Value)
+				batch.Queue(
```
pro-wh, line 330: can we do that here https://github.com/oasisprotocol/oasis-indexer/blob/316cd3b3bf5eaa217e026d33c9a5086bb98307d5/analyzer/runtime/extract.go#L382

mitjat: +1, the runtime analyzer is more structured than the consensus analyzer, and tends to first store all the analysis results into a big struct, then convert that struct into sql statements in a separate step.

Andrew7234: updated!

analyzer/runtime/runtime.go
```diff
@@ -318,6 +318,16 @@ func (m *Main) queueDbUpdates(batch *storage.QueryBatch, data *BlockData) {
 			error_code,
 			error_message,
 		)
+
+		if transactionData.IsNativeTransfer {
+			batch.Queue(
+				queries.RuntimeNativeBalanceUpdate,
```
pro-wh, line 324: do we ever subtract from the sender account?

Andrew7234:
Closing this since we decided to track native balances differently.

PR #378, opened 2023-04-06
runtime: store unrecognized evm log events

pro-wh:
sample run on block 5204809 https://explorer.emerald.oasis.dev/tx/0x0c72020e799e069c6ff2e82a33f5eaaf2919bd6a6eb0cd845bdb62eef723a574/logs

image


mitjat (approved):
Nice. Less complicated than you feared, I take it :)

PR #379, opened 2023-04-07
db: Prefer JSONB column type over JSON

mitjat:
This PR switches `JSON` to the newer `JSONB` ("JSON but Better") type. From [postgres docs](https://www.postgresql.org/docs/current/datatype-json.html):
> PostgreSQL offers two types for storing JSON data: json and jsonb. 
> The major practical difference is one of efficiency. The json data type stores an exact copy of the input text, which processing functions must reparse on each execution; while jsonb data is stored in a decomposed binary format that makes it slightly slower to input due to added conversion overhead, but significantly faster to process, since no reparsing is needed. jsonb also supports indexing, which can be a significant advantage.
> Because the json type stores an exact copy of the input text, it will preserve semantically-insignificant white space between tokens, as well as the order of keys within JSON objects. By contrast, jsonb does not preserve white space, does not preserve the order of object keys, and does not keep duplicate object keys. 
> In general, **most applications should prefer to store JSON data as jsonb**, unless there are quite specialized needs, such as legacy assumptions about ordering of object keys.

Prompted by @pro-wh 's offhand comment in slack where he noticed we use both JSON and JSONB.

*Testing*: reindex-and-run.sh, 100 blocks.

### DIFF
--- a/storage/client/queries/queries.go
+++ b/storage/client/queries/queries.go
@@ -227,7 +227,7 @@ const (
 				chain.entities.address AS entity_address,
 				chain.nodes.id AS node_address,
 				chain.accounts.escrow_balance_active AS escrow,
-				COALESCE(chain.commissions.schedule, '{}'::json) AS commissions_schedule,
+				COALESCE(chain.commissions.schedule, '{}'::JSONB) AS commissions_schedule,
 				CASE WHEN EXISTS(SELECT null FROM chain.nodes WHERE chain.entities.id = chain.nodes.entity_id AND voting_power > 0) THEN true ELSE false END AS active,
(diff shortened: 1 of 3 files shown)


storage/client/queries/queries.go
```diff
@@ -239,7 +239,7 @@ const (
 				chain.entities.address AS entity_address,
 				chain.nodes.id AS node_address,
 				chain.accounts.escrow_balance_active AS escrow,
-				COALESCE(chain.commissions.schedule, '{}'::json) AS commissions_schedule,
+				COALESCE(chain.commissions.schedule, '{}'::JSONB) AS commissions_schedule,
```
mitjat, line 242: caught by regression tests :) -- `cannot coalesce JSONB and JSON`

pro-wh (approved):
ooh thanks for looking into what the actual difference is

Andrew7234 (approved):

PR #380, opened 2023-04-12
EVM: download tokens at last processed round

pro-wh:
a preview of what it will be like to do EVM queries for downloading token data + token balances at the ~latest round instead of the last mutated round

this is for sapphire where blocks far in the past aren't accessible

tested on emerald, where I ran the block scanner over a range of blocks with token and token balance analyzers disabled then ran with only token analyzer and token balance analyzer. checked that tokens and token balances are still downloaded successfully and recorded to have been downloaded at the latest scanned block round

### DIFF
--- a/analyzer/queries/queries.go
+++ b/analyzer/queries/queries.go
@@ -313,12 +313,12 @@ const (
 	RuntimeEVMTokenAnalysisStale = `
     SELECT
       evm_token_analysis.token_address,
-      evm_token_analysis.last_mutate_round,
       evm_token_analysis.last_download_round,
       evm_tokens.token_type,
       address_preimages.context_identifier,
       address_preimages.context_version,
-      address_preimages.address_data
+      address_preimages.address_data,
+      (SELECT MAX(height) FROM chain.processed_blocks WHERE analyzer = evm_token_analysis.runtime::text) AS download_round
     FROM chain.evm_token_analysis
     LEFT JOIN chain.evm_tokens USING (runtime, token_address)
     LEFT JOIN chain.address_preimages ON
@@ -367,7 +367,6 @@ const (
     SELECT
       evm_token_balance_analysis.token_address,
       evm_token_balance_analysis.account_address,
-      evm_token_balance_analysis.last_mutate_round,
       evm_tokens.token_type,
       evm_token_balances.balance,
       token_address_preimage.context_identifier,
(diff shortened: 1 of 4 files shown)


analyzer/queries/queries.go
```diff
@@ -384,6 +386,8 @@ const (
 		token_address_preimage.address = evm_token_balance_analysis.token_address
 	LEFT JOIN chain.address_preimages AS account_address_preimage ON
 		account_address_preimage.address = evm_token_balance_analysis.account_address
+    LEFT JOIN chain.processed_blocks ON
```
pro-wh, line 389: what the heck whitespace

analyzer/queries/queries.go
```diff
@@ -313,16 +313,18 @@ const (
...
       address_preimages.context_identifier,
       address_preimages.context_version,
-      address_preimages.address_data
+      address_preimages.address_data,
+      processed_blocks.height AS download_round
     FROM chain.evm_token_analysis
     LEFT JOIN chain.evm_tokens USING (runtime, token_address)
     LEFT JOIN chain.address_preimages ON
       address_preimages.address = evm_token_analysis.token_address
+    LEFT JOIN chain.processed_blocks ON
```
mitjat, line 326: Won't this join against all already-processed heights? Causing each token to be returned 1e7 times when we're fully synced :)
You want `SELECT ..., (SELECT max(height) FROM chain.processed_blocks WHERE chain.processed_blocksanalyzer=evm_token_analysis.runtime) AS download_round, ...`, probably.

BTW we're lacking a useful index for effective `SELECT max(height)`, you should add it. Right now, we only run that query when an analyzer intiializes, so it was fine if it was slow, but I noticed even so.

pro-wh: oh shoot it's a list of all processed blocks? then yeah we want the max haha

pro-wh: I think postgresql had some complaint about doing it roughly this way. although I was using `$1` instead of the `evm_token_analysis.runtime`, let me try with the latter

pro-wh: ah `evm_token_analysis.runtime::text` works, to cast it to be comparable to the `processed_blocks.analyzer text` column

mitjat (approved):
Thank you! LGTM provided you tested this. Can you please describe the testing in the PR description? (Once we have e2e tests in CI, that will be hopefully less needed/meaningful.)

pro-wh:
testing description added :+1:

PR #385, opened 2023-04-20
analyzer: remove the 100ms timeout before processing every block

ptrus:
Removes the artificial 100ms timeout before processing every block. I noticed this while working on parallelizing the block processing, and it looks to me that this timeout is unintended.

Should help with https://github.com/oasisprotocol/oasis-indexer/issues/179 - likely also affects the analysis done in that issue, since the indexer was artificially slowing itself down.

## Testing

Only quickly tested consensus analyzer. On my node, this increases the consensus analyzer block process rate from:~6 per second, to: ~20 per second

### DIFF
--- a/analyzer/util/util.go
+++ b/analyzer/util/util.go
@@ -10,8 +10,7 @@ import (
 )
 
 const (
-	InitialTimeoutLowerBoundSeconds = 0
-	MaximumTimeoutUpperBoundSeconds = 60
+	timeoutUpperBound = 60 * time.Second
 	// SHA256 of empty string.
 	// oasis-core tags event that are not associated with any real transactions.
 	// For example, the DebondingStart escrow event.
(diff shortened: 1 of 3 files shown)


Andrew7234 (approved):
😅

analyzer/util/util.go
```diff
@@ -64,6 +64,9 @@ func (b *Backoff) Success() {
 
 // Failure increases the backoff interval.
 func (b *Backoff) Failure() {
+	if b.currentTimeout == 0 {
+		b.currentTimeout = 10 * time.Millisecond
```
pro-wh, line 68: :information_source: 10ms here instead of 100ms used before (albeit incorrectly)

analyzer/util/util.go
```diff
@@ -31,7 +31,7 @@ type Backoff struct {
 
 // NewBackoff returns a new backoff.
 func NewBackoff(initialTimeout time.Duration, maximumTimeout time.Duration) (*Backoff, error) {
```
pro-wh, line 33: could use some description, as we're finding this incorrect usage and not changing the API

analyzer/consensus/consensus.go
```diff
@@ -138,7 +138,7 @@ func (m *Main) Start() {
 	}
 
 	backoff, err := util.NewBackoff(
-		100*time.Millisecond,
+		0,
```
pro-wh, line 141: would we ever want nonzero initial "timeout"? seems like it was used wrong later on, we should only `backoff.Wait()` on errors, rather than in every iteration

pro-wh: oh although if we did that we wouldn't have this gradual speedup when an error stops happening

analyzer/util/util.go
ptrus: Added description after refactoring things a bit.

analyzer/consensus/consensus.go
ptrus: Refactored the struct a bit. This `100*time.Millisecond` now defines the "minimum timeout step".

pro-wh (approved):
sounds good, thanks

PR #390, opened 2023-04-25
config-related cleanup and opinions

### LINKED ISSUE #387: config-related cleanup
extracting ideas from #383

---

              unused, let's remove

_Originally posted by @pro-wh in https://github.com/oasisprotocol/oasis-indexer/pull/383#discussion_r1167339159_
            
---
              currently unused. let's replace with a new `type Network string` like how `Runtime` is and start using it in places where we use `"mainnet"` etc. literals

_Originally posted by @pro-wh in https://github.com/oasisprotocol/oasis-indexer/pull/383#discussion_r1167339527_
            
---
              only used in statecheck, should move there

_Originally posted by @pro-wh in https://github.com/oasisprotocol/oasis-indexer/pull/383#discussion_r1167339790_
            
---
              only used in statecheck, should move there

_Originally posted by @pro-wh in https://github.com/oasisprotocol/oasis-indexer/pull/383#discussion_r1167340349_

pro-wh:
fixes #387

- ChainID: unused, removing
- Network: we no longer talk about networks, as they're too short-lived. we're letting the SDK have that term. the indexer now talks about "chains." thus we're providing a new typed string, "ChainName" for that purpose
- "consensus" as the 'runtime' in an API request context: seems unused, and I don't remember if we had anything planned for it. removing. but let me know if we shouldn't
- `.String()`: had been carried over from when these were special uint8s, but lately we're more into types that actually are strings. removing. use `string(...)` where needed
- StringifyDenomination: we mixed that into RuntimeClient because it had conveniently had the sdkPT there. but it's configuration rather than storage, and it's the only remaining thing needing the sdkPT field, and it's the only thing structurally differentiating RuntimeClient from RuntimeLiteApi. moving it to runtime.Main. runtime.Main.cfg now has the equivalent .ParaTime field, if anyone needs it
- runtime-related analyzers (incl. token and token balance) `runtime common.Runtime` field moved to `RuntimeConfig.RuntimeName` (i.e. `.runtime` becomes `.cfg.RuntimeName`)

### DIFF
--- a/analyzer/api.go
+++ b/analyzer/api.go
@@ -5,6 +5,7 @@ import (
 
 	sdkConfig "github.com/oasisprotocol/oasis-sdk/client-sdk/go/config"
 
+	"github.com/oasisprotocol/oasis-indexer/common"
 	"github.com/oasisprotocol/oasis-indexer/storage"
 )
 
@@ -55,6 +56,9 @@ type BlockRange struct {
 // RuntimeConfig specifies configuration parameters for
 // processing the runtime layer.
 type RuntimeConfig struct {
+	// RuntimeName is which runtime to analyze.
+	RuntimeName common.Runtime
+
 	// ParaTime is the SDK ParaTime structure describing the runtime.
 	ParaTime *sdkConfig.ParaTime
 
--- a/analyzer/api.go
+++ b/analyzer/api.go
@@ -3,6 +3,8 @@ package analyzer
 import (
 	"errors"
 
+	sdkConfig "github.com/oasisprotocol/oasis-sdk/client-sdk/go/config"
+
 	"github.com/oasisprotocol/oasis-indexer/storage"
 )
 
@@ -53,6 +55,9 @@ type BlockRange struct {
 // RuntimeConfig specifies configuration parameters for
 // processing the runtime layer.
 type RuntimeConfig struct {
+	// ParaTime is the SDK ParaTime structure describing the runtime.
+	ParaTime *sdkConfig.ParaTime
+
 	// Range is the range of rounds to process.
 	// If this is set, the analyzer analyzes rounds in the provided range.
 	Range RoundRange
--- a/config/config.go
+++ b/config/config.go
@@ -263,7 +263,8 @@ func (cfg *AggregateStatsConfig) Validate() error {
 
 // ServerConfig contains the API server configuration.
 type ServerConfig struct {
-	// ChainName is the name of the chain (e.g. mainnet/testnet/local).
+	// ChainName is the name of the chain (i.e. mainnet/testnet). Custom/local nets are not supported.
+	// This is only used for the runtime status endpoint.
 	ChainName common.ChainName `koanf:"chain_name"`
 
 	// Endpoint is the service endpoint from which to serve the API.
--- a/config/config.go
+++ b/config/config.go
@@ -148,7 +148,7 @@ func (sc *SourceConfig) History() *History {
 
 func (sc *SourceConfig) SDKNetwork() *sdkConfig.Network {
 	if sc.ChainName != "" {
-		return sdkConfig.DefaultNetworks.All[sc.ChainName.String()]
+		return sdkConfig.DefaultNetworks.All[string(sc.ChainName)]
 	}
 	return sc.CustomChain.SDKNetwork
 }
--- a/config/config.go
+++ b/config/config.go
@@ -12,6 +12,7 @@ import (
 	"github.com/knadh/koanf/providers/file"
 	sdkConfig "github.com/oasisprotocol/oasis-sdk/client-sdk/go/config"
 
+	"github.com/oasisprotocol/oasis-indexer/common"
 	"github.com/oasisprotocol/oasis-indexer/log"
 )
 
@@ -122,7 +123,7 @@ type SourceConfig struct {
 
 	// ChainName is the name of the chain (e.g. mainnet/testnet). Set
 	// this to use one of the default chains.
-	ChainName string `koanf:"chain_name"`
+	ChainName common.ChainName `koanf:"chain_name"`
 	// CustomChain is information about a custom chain. Set this to use a
 	// chain other than the default chains, e.g. a local testnet.
 	CustomChain *CustomChainConfig `koanf:"custom_chain"`
@@ -147,7 +148,7 @@ func (sc *SourceConfig) History() *History {
 
 func (sc *SourceConfig) SDKNetwork() *sdkConfig.Network {
 	if sc.ChainName != "" {
-		return sdkConfig.DefaultNetworks.All[sc.ChainName]
+		return sdkConfig.DefaultNetworks.All[sc.ChainName.String()]
 	}
 	return sc.CustomChain.SDKNetwork
 }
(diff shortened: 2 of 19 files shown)


pro-wh:
inventory of typed strings/ints, annotated:

```console
$ ag 'type [A-Z]\w+ [a-z]\w+$'
api/v1/types/openapi.gen.go
176:type AddressDerivationContext string
274:type ConsensusEventType string
277:type ConsensusTxMethod string
438:type EvmTokenType string
441:type Layer string <- interesting, this that 4-state consensus-cipher-emerald-sapphire enum. I had removed "consensus" from common.Runtime, if you need "consensus" then use this. possibly move this type to common, other uses are in storage system
575:type Runtime string <- deduplicate with common.Runtime
689:type RuntimeEventType string
1194:type GetLayerStatsActiveAccountsParamsWindowStepSeconds uint32

api/v1/types/util.go
12:type Address string <- analyzer uses this too. imo move to common

common/types.go <- that's us!
104:type ContextKey string
127:type ChainName string
139:type Runtime string

config/config.go
291:type StorageBackend uint <- how far are we on giving up cockroachdb? could rip out the whole backend selection config field

scripts/vendor-oasis-core/patches/v21.1.1/reuse_epoch_time.patch <- here and below imported from oasis-core
20:-type EpochTime uint64

coreapi/v21.1.1/governance/api/proposal.go
12:type ProposalState uint8
117:type Vote uint8

coreapi/v21.1.1/registry/api/runtime.go
22:type RuntimeKind uint32
264:type RuntimeGovernanceModel uint8

coreapi/v21.1.1/roothash/api/block/header.go
13:type HeaderType uint8

coreapi/v21.1.1/roothash/api/commitment/executor.go
39:type ExecutorCommitmentFailure uint8

coreapi/v21.1.1/roothash/api/api.go
61:type EvidenceKind uint8

coreapi/v21.1.1/consensus/api/transaction/gas.go
10:type Gas uint64
36:type Op string

coreapi/v21.1.1/beacon/api/pvss.go
46:type RoundState uint8

coreapi/v21.1.1/consensus/api/transaction/transaction.go
80:type MethodPriority uint8
99:type MethodName string

coreapi/v21.1.1/consensus/api/api.go
31:type FeatureMask uint8

coreapi/v21.1.1/staking/api/api.go
256:type ThresholdKind int
329:type StakeClaim string

coreapi/v21.1.1/scheduler/api/api.go
18:type Role uint8
84:type CommitteeKind uint8 <- here and above imported from oasis-core

log/format.go
9:type Format uint

log/level.go
9:type Level uint

coreapi/v21.1.1/staking/api/slashing.go <- oh this too. weird sorting
11:type SlashReason uint8

analyzer/runtime/evm.go
18:type EVMTokenType int

analyzer/consensus/consensus.go
638:type TransferType string
```

analyzer/api.go
```diff
@@ -53,6 +55,9 @@ type BlockRange struct {
 // RuntimeConfig specifies configuration parameters for
 // processing the runtime layer.
 type RuntimeConfig struct {
```
mitjat, line 57: Since `RuntimeConfig` now aims to contain all metadata about the runtime, can you please also add `runtimeName common.Runtime` and remove it from analyzers' `Main` structs? Then for consistent naming, `ParaTime` can maybe become `runtimeInfo` or `runtimeSdkInfo`. Your call on the exact names, but let's definitely not have `runtime` for the name and `ParaTime` for other metadata :)

Background: The existence of this type (and even more so `ConsensusConfig`, seeing as how there's only one consensus) has annoyed me for a long time because it seemed to contain such a random subset of an analyzer's config. It felt to me that its fields should just be absorbed into the containing struct, which is always an analyzer `Main`. But well ... after all these cleanups that we've done, and after introducing all these new analyzers that deal with the runtime(s), the separation seems to make good sense. All this struct is missing is the runtime name.

config/config.go
```diff
@@ -259,7 +264,7 @@ func (cfg *AggregateStatsConfig) Validate() error {
 // ServerConfig contains the API server configuration.
 type ServerConfig struct {
 	// ChainName is the name of the chain (e.g. mainnet/testnet/local).
```
mitjat, line 266:
```suggestion
	// ChainName is the name of the chain (i.e. mainnet/testnet). Custom/local nets are not supported.
	// This is only used for the runtime status endpoint.
```

The second line is optional, but it mellows out the second sentence. I guess the second sentence is optional too ...

FYI, the only place this param is ever used is here: https://github.com/oasisprotocol/oasis-indexer/blob/dc3834f3132e79a9e9b07ce61d41b0dea6b81898/storage/client/client.go#L1333-L1333, which is then only used to list the compute nodes that support a particular runtime: https://github.com/oasisprotocol/oasis-indexer/blob/dc3834f3132e79a9e9b07ce61d41b0dea6b81898/storage/client/client.go#L1363-L1363

This should ideally hook into your `custom_chain` approach but since it lives on the API-server side (as opposed to the analysis side), we don't have access to that. But looks like we don't need that kind of generality here, at least for now.

mitjat (approved):
Thank you. This will be very helpful for anybody working on the code, now and in the future. Will be cool to potentially get rid of `RuntimeClient` and `ConsensusClient` at some point ...

Comments on the type inventory:
- `openapi.gen.go: Layer`: Cannot move to common, type is autogenerated from openapi spec. Could alias it in common though. We don't have great hygiene around using openapi types in places that are not API-related; it would be ideal to always have a separate type in non-API areas, even if it's just an alias of the API type.
- cockroach: The ship has sailed long ago, it has more incompatibilities with Postgres than advertised, and we're using some relatively advanced Postgres now; we're not switching. I'd be very OK with simplifying the setup/code more.
- All other suggestions I'm also on board with :+1:

analyzer/api.go
pro-wh: ok let's try that. might be a little awkward fit for the token and token balance analyzers, which don't use this yet. they don't use the block range part, but everything else I think is in common

config/config.go
pro-wh: oh good point. if we later go about adding custom chain support for the api side, we could also consider moving the chain info up a level so that it's not analyzer-specific. then access that piece of configuration from the api too

pro-wh:
> `openapi.gen.go: Layer`: Cannot move to common, type is autogenerated from openapi spec. Could alias it in common though. We don't have great hygiene around using openapi types in places that are not API-related; it would be ideal to always have a separate type in non-API areas, even if it's just an alias of the API type.

we can get it with x-go-type in the spec

mitjat:
> > `openapi.gen.go: Layer`: Cannot move to common, type is autogenerated from openapi spec. Could alias it in common though. We don't have great hygiene around using openapi types in places that are not API-related; it would be ideal to always have a separate type in non-API areas, even if it's just an alias of the API type.
> 
> we can get it with x-go-type in the spec

I think it takes more than that? To define the path to the package (not just its name), we'd have to do a separate invocation of oapi-codegen too, with a different output file. And then we'd have to adjust all the CI/Makefile parts that assume that `common` can be built as-is and does not need to be code-generated first.

mitjat:
> no I'm picturing having a `type Layer string` outside of the openapi spec, then the spec refers to it. 

ahh right :facepalm: 
yes, that works :+1:

Andrew7234 (approved):
thanks! all the config stuff is easier to understand now

analyzer/api.go
pro-wh: thusly moved

pro-wh:
thanks for the feedback. going to merge with this current set of changes.

PR #392, opened 2023-04-27
analyzers, cache: Allow to shut down gracefully

mitjat:
This PR was motivated by #360 which introduces an on-disk cache (KV store), but does not close it cleanly when exiting. This PR adds a cleanup mechanism to the cache, and (partly by necessity, partly for code hygiene) extends the cleanup mechanism to all the analyzers.

At a high level, all analyzers are now provided with an external `context.Context` in their `Start()` method, and they are on the lookout for the `ctx.Done()` channel closing. The top-level code closes the channel on regular program termination and on SIGINT (Ctrl+C).

There is an additional change piggybacking on this PR:
- The recently-added KVStore (cache) now uses `cbor` to serialize its keys. It previously used `gob`, which in my testing complained for certain keys. (I think null pointers? I didn't investigate further, as I think it's preferable to not introduce new serialization formats (here, gob) into the project anyway.)

**Testing:** Ran in three different scenarios, verified that KV store (which is the furthest down the stack of cleanup methods) closes correctly:
- analyzers with a limited range in config; allow them to finish "on their own"
- kill analyzers with a single ctrl+C
- kill analyzers with a double ctrl+C (by introducing an artificial delay into cleanup); verify that the program is killable at the cost of incomplete cleanup

### DIFF
--- a/analyzer/evmtokenbalances/evm_token_balances.go
+++ b/analyzer/evmtokenbalances/evm_token_balances.go
@@ -258,6 +258,8 @@ func (m Main) processBatch(ctx context.Context) (int, error) {
 }
 
 func (m Main) Start(ctx context.Context) {
+	defer m.cleanup()
+
 	backoff, err := util.NewBackoff(
 		100*time.Millisecond,
 		// Cap the timeout at the expected round time. All runtimes currently have the same round time.
--- a/analyzer/evmtokenbalances/evm_token_balances.go
+++ b/analyzer/evmtokenbalances/evm_token_balances.go
@@ -257,9 +257,7 @@ func (m Main) processBatch(ctx context.Context) (int, error) {
 	return len(staleTokenBalances), nil
 }
 
-func (m Main) Start() {
-	ctx := context.Background()
-
+func (m Main) Start(ctx context.Context) {
 	backoff, err := util.NewBackoff(
 		100*time.Millisecond,
 		// Cap the timeout at the expected round time. All runtimes currently have the same round time.
@@ -273,7 +271,14 @@ func (m Main) Start() {
 	}
 
 	for {
-		backoff.Wait()
+		select {
+		case <-time.After(backoff.Timeout()):
+			// Process another batch of token balances.
+		case <-ctx.Done():
+			m.logger.Warn("shutting down evm_token_balances analyzer", "reason", ctx.Err())
+			m.cleanup()
+			return
+		}
 
 		numProcessed, err := m.processBatch(ctx)
 		if err != nil {
@@ -291,6 +296,10 @@ func (m Main) Start() {
 
 		backoff.Success()
 	}
+}
+
+func (m *Main) cleanup() {
+	m.cfg.Source.Close()
 }
 
 func (m Main) Name() string {
--- a/analyzer/evmtokenbalances/evm_token_balances.go
+++ b/analyzer/evmtokenbalances/evm_token_balances.go
@@ -301,7 +301,9 @@ func (m Main) Start(ctx context.Context) {
 }
 
 func (m *Main) cleanup() {
-	m.cfg.Source.Close()
+	if err := m.cfg.Source.Close(); err != nil {
+		m.logger.Error("error closing data source", "err", err)
+	}
 }
 
 func (m Main) Name() string {
--- a/analyzer/runtime/runtime.go
+++ b/analyzer/runtime/runtime.go
@@ -74,9 +74,7 @@ func NewRuntimeAnalyzer(
 	}, nil
 }
 
-func (m *Main) Start() {
-	ctx := context.Background()
-
+func (m *Main) Start(ctx context.Context) {
 	if err := m.prework(); err != nil {
 		m.logger.Error("error doing prework",
 			"err", err,
@@ -115,7 +113,14 @@ func (m *Main) Start() {
 	}
 
 	for m.cfg.Range.To == 0 || round <= m.cfg.Range.To {
-		backoff.Wait()
+		select {
+		case <-time.After(backoff.Timeout()):
+			// Process next block.
+		case <-ctx.Done():
+			m.logger.Warn("shutting down runtime analyzer", "reason", ctx.Err())
+			m.cleanup()
+			return
+		}
 		m.logger.Info("attempting block", "round", round)
 
 		if err := m.processRound(ctx, round); err != nil {
@@ -141,6 +146,10 @@ func (m *Main) Start() {
 	m.logger.Info(
 		fmt.Sprintf("finished processing all blocks in the configured range [%d, %d]",
 			m.cfg.Range.From, m.cfg.Range.To))
+}
+
+func (m *Main) cleanup() {
+	m.cfg.Source.Close()
 }
 
 // Name returns the name of the Main.
--- a/analyzer/runtime/runtime.go
+++ b/analyzer/runtime/runtime.go
@@ -77,7 +77,7 @@ func NewRuntimeAnalyzer(
 func (m *Main) Start(ctx context.Context) {
 	defer m.cleanup()
 
-	if err := m.prework(); err != nil {
+	if err := m.prework(ctx); err != nil {
 		m.logger.Error("error doing prework",
 			"err", err,
 		)
@@ -193,9 +193,8 @@ func (m *Main) latestRound(ctx context.Context) (uint64, error) {
 }
 
 // prework performs tasks that need to be done before the main loop starts.
-func (m *Main) prework() error {
+func (m *Main) prework(ctx context.Context) error {
 	batch := &storage.QueryBatch{}
-	ctx := context.Background()
 
 	// Register special addresses.
 	batch.Queue(
--- a/analyzer/runtime/runtime.go
+++ b/analyzer/runtime/runtime.go
@@ -75,6 +75,8 @@ func NewRuntimeAnalyzer(
 }
 
 func (m *Main) Start(ctx context.Context) {
+	defer m.cleanup()
+
 	if err := m.prework(); err != nil {
 		m.logger.Error("error doing prework",
 			"err", err,
@@ -118,7 +120,6 @@ func (m *Main) Start(ctx context.Context) {
 			// Process next block.
 		case <-ctx.Done():
 			m.logger.Warn("shutting down runtime analyzer", "reason", ctx.Err())
-			m.cleanup()
 			return
 		}
 		m.logger.Info("attempting block", "round", round)
@@ -149,7 +150,11 @@ func (m *Main) Start(ctx context.Context) {
 }
 
 func (m *Main) cleanup() {
-	m.cfg.Source.Close()
+	if err := m.cfg.Source.Close(); err != nil {
+		m.logger.Error("failed to cleanly close consensus data source",
+			"err", err.Error(),
+		)
+	}
 }
 
 // Name returns the name of the Main.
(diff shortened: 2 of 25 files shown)


storage/oasis/nodeapi/file/consensus.go
```diff
@@ -24,12 +24,15 @@ type FileConsensusApiLite struct {
 var _ nodeapi.ConsensusApiLite = (*FileConsensusApiLite)(nil)
 
 func NewFileConsensusApiLite(cacheDir string, consensusApi nodeapi.ConsensusApiLite) (*FileConsensusApiLite, error) {
-	db, err := pogreb.Open(cacheDir, &pogreb.Options{BackgroundSyncInterval: -1})
+	db, err := OpenKVStore(
+		log.NewDefaultLogger("cached-node-api").With("runtime", "consensus"),
```
pro-wh, line 28: api uses the term "layer" for values that can be "consensus" or a runtime name

cmd/analyzer/analyzer.go
```diff
@@ -239,12 +241,37 @@ func NewService(cfg *config.AnalysisConfig) (*Service, error) {
 func (a *Service) Start() {
 	a.logger.Info("starting analysis service")
 
+	var ctx context.Context
+	ctx, a.cancelAnalyzers = context.WithCancel(context.Background())
+
+	// trap Ctrl+C and call cancel on the context
+	signalChan := make(chan os.Signal, 1)
+	signal.Notify(signalChan, os.Interrupt)
```
pro-wh, line 249: I've read that a lot of service managers use SIGTERM

cmd/analyzer/analyzer.go
```diff
@@ -239,12 +241,37 @@ func NewService(cfg *config.AnalysisConfig) (*Service, error) {
...
+	defer func() {
+		signal.Stop(signalChan) // Stop catching Ctrl+C signals.
+		a.Shutdown()            // Shut down gracefully.
+	}()
+
+	go func() {
+		select {
+		case <-signalChan:
+			// We received the first Ctrl+C; try to shut down gracefully.
+			a.Shutdown()
```
pro-wh, line 259: the caller of Start, either runAnalyzer or func in rootMain calls .Shutdown. should only have to return, rather than putting these a.Shutdown calls in various places. are we able to return?

pro-wh: oh are we not able to return because we don't cancel the context until something calls Shutdown thus calling cancelAnalyzers? seems like it would be more natural to receive the signal and only call cancelAnalyzers. no a.Shutdown(). and let us return normally

pro-wh: that might also let us extract this into a helper that only creates a ctx that cancels on ctrl+c

cmd/analyzer/analyzer.go
mitjat: :+1: added

storage/oasis/nodeapi/file/consensus.go
mitjat: I was on the fence what to use. We do use "layer" in the API, (I'm not a fan, even if I'm the one who introduced it :slightly_smiling_face:; I just don't have a better name), but not outside the API. Using "layer" here would make that more cohesive, but it would also mean that you cannot get "all" (most?) logs related to a runtime by filtering for `.runtime = "my_runtime_name"`, which is the log attribute that all/most other loggers use. I think the latter is more important.

pro-wh: nvm I'll just mark this as resolved lol

cmd/analyzer/analyzer.go
mitjat: I don't fully understand this comment. Is your main concern that the caller of `Start()` now needs to be careful to call `Shutdown()`, therefore a clean shutdown is not guaranteed? Not sure what kind of returning you had in mind; where and why would the callers want to return? 

This might be relevant: The analysis `Service` (which owns/controls all the analyzers; the other service we have is the API service) was originally designed with explicit cleanup, i.e. with Start+Shutdown. So the Service's `Shutdown()` is always called when appropriate, e.g. [here](https://github.com/oasisprotocol/oasis-indexer/blob/dc3834f3132e79a9e9b07ce61d41b0dea6b81898/cmd/analyzer/analyzer.go#L70-L72) (and in another spot; there are only two calls to Start() anyway).

pro-wh: yup, referring to that. i.e. it's (or was previously at least) up to the thing using the analyzer `Service` (1) to call its Start and (2) to call its Shutdown.

my understanding of this change is that Shutdown is now something that sometimes gets called by the service itself inside Start, sometimes from outside. in fact, if I'm following this right, both?

I feel it's a smaller change and less complicated to cancel the context and let the subordinate analyzers finish, let the Start return, and let the thing using the analyzer `Service` call Shutdown as before

pro-wh, line 259: is this to say we yank out the database connection on the first ctrl+c?

mitjat: Yes. And and the kv store.

pro-wh: could we wait for the analyzers to return before doing that? this PR is about graceful shutdown, and this sounds like we won't really let analyzers finish what they're doing

cmd/analyzer/analyzer.go
pro-wh: is that pushed yet?

cmd/analyzer/analyzer.go
mitjat: > Shutdown is now something that sometimes gets called by the service itself inside Start, sometimes from outside

That was indeed funky. Thank you for noticing.

Your observation caused me to think about the code more, and I noticed more smells that I then refactored away:
- The `Shutdown()` method of the `Service` _had_ to be called by the caller of `Service::Start()`. Then it makes more sense to just call `defer Shutdown()` from `Start()` and not burden the caller. We do so now.
- The code introduced by this PR repeated that sin (Start+Shutdown) at the Analyzer level. I changed it now to also use `defer`.
- Several `Shutdown()` methods throughout the codebase were IMO misleadingly named. They did not actively cause anything to shut down; they performed resource cleanup and assumed that any active processing had completed. I renamed them appropriately, IIRC to either `cleanup()` or `Close()`.

I also refactored the code around this comment thread so it no longer waits for ctrl+C in a separate goroutine; I find the new approach much easier to grok.

I added all these changes as a separate commit, should you prefer to view them that way, but I think it makes more sense to diff against main.

analyzer/runtime/runtime.go
```diff
@@ -74,8 +74,8 @@ func NewRuntimeAnalyzer(
 	}, nil
 }
 
-func (m *Main) Start() {
-	ctx := context.Background()
+func (m *Main) Start(ctx context.Context) {
+	defer m.cleanup()
 
 	if err := m.prework(); err != nil {
```
ptrus, line 80: Also pass the `ctx` into `prework` so that this cancelable context is also used in `SendBatch`.

cmd/analyzer/analyzer.go
mitjat: Yes, Ctrl+F `SIGTERM`. And make sure you're looking at the full diff, not the diff in the context of which this thread started.

pro-wh: ah ya thanks

cmd/analyzer/analyzer.go
pro-wh: o true. gonna take a look, thanks

analyzer/evmtokenbalances/evm_token_balances.go
```diff
@@ -294,6 +301,10 @@ func (m Main) Start() {
 	}
 }
 
+func (m *Main) cleanup() {
+	m.cfg.Source.Close()
```
pro-wh, line 305: some Main's log error, some don't. this and evm_tokens being ones that don't

analyzer/runtime/runtime.go
```diff
@@ -143,6 +149,14 @@ func (m *Main) Start() {
 			m.cfg.Range.From, m.cfg.Range.To))
 }
 
+func (m *Main) cleanup() {
+	if err := m.cfg.Source.Close(); err != nil {
+		m.logger.Error("failed to cleanly close consensus data source",
```
pro-wh, line 154: consensus -> runtime (I think?)

cmd/analyzer/analyzer.go
pro-wh: oo looking good

little asymmetry, but don't let this distract you: initializing those resources is in a public function `NewMain` etc, while cleaning them up is in a private method `cleanup`. once you make a service, you have no way out other than to Start it and let Start clean things up when it's done

pro-wh: :+1: light touch context cancellation :feather:

analyzer/runtime/runtime.go
pro-wh: thanks, are there any other context.Background's around?

pro-wh (approved):
thanks for the changes

storage/oasis/nodeapi/history/runtime.go
```diff
@@ -43,6 +43,15 @@ func NewHistoryRuntimeApiLite(ctx context.Context, history *config.History, sdkP
 	}, nil
 }
 
+func (rc *HistoryRuntimeApiLite) Close() error {
+	for _, api := range rc.APIs {
+		if err := api.Close(); err != nil {
+			return err
```
Andrew7234, line 49: should this have the same logic as consensus to save the `firstErr` and continue attempting to close the other APIs first?

analyzer/evmtokenbalances/evm_token_balances.go
mitjat: :+1: added logging to those two

Andrew7234 (approved):
thanks!

cmd/analyzer/analyzer.go
```diff
@@ -237,23 +237,57 @@ func NewService(cfg *config.AnalysisConfig) (*Service, error) {
...
+	case <-signalChan:
+		a.logger.Info("received interrupt, shutting down")
+		// Cancel the analyzers' context and wait for them to exit cleanly.
+		cancelAnalyzers()
+		select {
+		case <-analyzersDone:
+			a.logger.Info("all analyzers have exited cleanly")
+			return
+		case <-signalChan:
+			a.logger.Info("received second interrupt, forcing exit")
```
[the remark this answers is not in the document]
mitjat: I like it! Implemented it now. For my taste, the code is about the same complexity to understand though. There's no more explicit handling of a second ctrl+C, which is nicer and simpler (and more agnostic of surrounding code, which might theoretically have its own handlers); that's good. But there is the additional cognitive burden of tracking whether the custom signal interceptor is active or not.

I also feel the image is harder to grok than the code :). Also, not sure it's correct; the far left path takes three ctrl+C to reach "process exited"?

storage/oasis/nodeapi/history/runtime.go
mitjat: good catch, thank you! implemented.

cmd/analyzer/analyzer.go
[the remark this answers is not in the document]
pro-wh: right, in the previous code, the second ctrl+c makes this function return, which transfers control to the `defer`s. first, the defer to unregister the signal notifier. second, to call `cleanup`. `cleanup` I'm pretty sure is fast, it's only closing the target storage postgresql db connection. but what happens after that? the analyzer goroutines are still running. I think go keeps the process alive until those goroutines finish. but with no signal handler installed, a third ctrl+c does make the program exit

PR #396, opened 2023-05-01
cleanup: enable offline cache-based runtimeClient

Andrew7234:
After #390 the clunky `runtimeInfo`/`sdkPT` config was removed, which allows us to easily instantiate a cache-based runtimeClient without attempting to connect to a node (eg in CI).


pro-wh (approved):
:point_up:

PR #397, opened 2023-05-02
Cobalt core API: Support DebondingStartEscrowEvent

mitjat:
This PR allows our cobalt RPC client to parse the DebondingStartEscrowEvent that was introduced mid-Cobalt.
The PR also includes some additional error message details to make this error (and similar ones) easier to pinpoint.

Prompted by https://app.clickup.com/t/866a7412t, and closes that ticket.

We've had a similar problem before: https://github.com/oasisprotocol/oasis-indexer/blob/82e29b83591b75432467ed2303a713a6d3128bc6/coreapi/v21.1.1/staking/api/api.go#L84

I believe the original sin was that when we vendored the cobalt APIs into the indexer, we took v21.1.0, which was the earliest compatible API. We should have instead gone with 21.3.9, the latest compatible one. I was wondering about it at the time but got different recommendations on slack. 
Anyway, I re-vendored v21.3.9 locally and compared it with the manually-updated v21.1.1 that we have in indexer, and it looks like this missing field is the last meaningful difference. So hopefully we can index the rest of Cobalt with no further changes.

**Testing:** Indexed block 4741442 (which was the problematic one in the ticket). Indexed some early blocks to make sure the old format still works well. This also led to ...

**... Change 2:** If the KVStore cache caches a value but then fails to fetch it (e.g. because the definition of the value's type has changed between runs), we should revert to re-fetching the value from the node, rather than sticking with hte cached value and failing the round. This PR implements this change.

### DIFF
--- a/coreapi/v21.1.1/staking/api/api.go
+++ b/coreapi/v21.1.1/staking/api/api.go
@@ -59,9 +59,10 @@ type BurnEvent struct {
 
 // EscrowEvent is an escrow event.
 type EscrowEvent struct {
-	Add     *AddEscrowEvent     `json:"add,omitempty"`
-	Take    *TakeEscrowEvent    `json:"take,omitempty"`
-	Reclaim *ReclaimEscrowEvent `json:"reclaim,omitempty"`
+	Add            *AddEscrowEvent            `json:"add,omitempty"`
+	Take           *TakeEscrowEvent           `json:"take,omitempty"`
+	DebondingStart *DebondingStartEscrowEvent `json:"debonding_start,omitempty"` // Only emitted at _some_ heights; introduced in oasis-core v21.3.0 mid-Cobalt.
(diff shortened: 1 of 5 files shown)


pro-wh (commented):
lg, thanks for the error wrapping.

what was the methodology for "Cobalt core API: Support DebondingStartEscrowEvent"? reimport from a newer version used in cobalt?

pro-wh (approved):

mitjat:
> what was the methodology for "Cobalt core API: Support DebondingStartEscrowEvent"? reimport from a newer version used in cobalt?

`./scripts/vendor-oasis-core/vendor_oasis_core.sh`, with a small local hack to the script so it vendored the newer version (21.3.9) but into the existing `coreapi/v21.1.1` directory. Then `git diff`.

PR #399, opened 2023-05-02
config: allow connecting to a special node for some things

pro-wh:
- one entry in `analysis.source.nodes` is no longer a NodeConfig, it's this new thing ArchiveConfig.
- ArchiveConfig is a struct with a .DefaultNode field, where you'd put the NodeConfig you used to have
- ArchiveConfig additionally has .ConsensusNode and .RuntimeNodes[runtime], which let you specify to connect to a different node for different purposes.

for operators, a config like this:

```yml
analysis:
  source:
    chain_name: mainnet
    nodes:
      damask:
        rpc: grpc.oasis.dev:443
```

becomes:

```yml
analysis:
  source:
    chain_name: mainnet
    nodes:
      damask:
        default:
          rpc: grpc.oasis.dev:443
```

you can then add things:

```yml
analysis:
  source:
    chain_name: mainnet
    nodes:
      damask:
        # the node to consult for most damask-network data, e.g. consensus, emerald, cipher
        default:
          rpc: grpc.oasis.dev:443
        runtimes:
          # a special node with SGX to provide sapphire data
          sapphire:
            rpc: whatever:443
```

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -68,6 +68,11 @@ func (cfg *AnalysisConfig) Validate() error {
 	} else if cfg.Source.ChainName != "" && cfg.Source.CustomChain != nil {
 		return fmt.Errorf("source.chain_name and source.custom_chain specified, can only use one")
 	}
+	for archiveName, archiveConfig := range cfg.Source.Nodes {
+		if archiveConfig.DefaultNode == nil && archiveConfig.ConsensusNode == nil && len(archiveConfig.RuntimeNodes) == 0 {
+			return fmt.Errorf("source.nodes[%v] has none of .default, .consensus, or .runtimes", archiveName)
+		}
+	}
 	if cfg.Analyzers.Consensus != nil {
 		if err := cfg.Analyzers.Consensus.Validate(); err != nil {
 			return err
--- a/config/config.go
+++ b/config/config.go
@@ -131,7 +131,7 @@ type SourceConfig struct {
 	// Nodes describe the oasis-node(s) to connect to. Keys are "archive
 	// names," which are named after mainnet releases, in lowercase e.g.
 	// "cobalt" and "damask."
-	Nodes map[string]*NodeConfig `koanf:"nodes"`
+	Nodes map[string]*ArchiveConfig `koanf:"nodes"`
 
 	// If set, the analyzer will skip some initial checks, e.g. that
 	// `rpc` really serves the chain with the chain context we expect.
@@ -168,10 +168,12 @@ func CustomSingleNetworkSourceConfig(rpc string, chainContext string) *SourceCon
 	}
 }
 
-func SingleNetworkLookup(rpc string) map[string]*NodeConfig {
-	return map[string]*NodeConfig{
+func SingleNetworkLookup(rpc string) map[string]*ArchiveConfig {
+	return map[string]*ArchiveConfig{
 		"damask": {
-			RPC: rpc,
+			DefaultNode: &NodeConfig{
+				RPC: rpc,
+			},
 		},
 	}
 }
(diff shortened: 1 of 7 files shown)


config/config.go
```diff
@@ -200,6 +204,34 @@ type CustomChainConfig struct {
 	SDKNetwork *sdkConfig.Network `koanf:"sdk_network"`
 }
 
+// ArchiveConfig is information about the nodes for a network.
+type ArchiveConfig struct {
+	// DefaultNode is information about the node to get data from by default.
+	DefaultNode *NodeConfig `koanf:"default"`
+	// ConsensusNode is information about the node to get consensus data from,
+	// instead of the default node.
+	ConsensusNode *NodeConfig `koanf:"consensus"`
```
pro-wh, line 213: might not need this, but I added just to be thorough :upside_down_face:

config/config.go
```diff
@@ -131,7 +131,9 @@ type SourceConfig struct {
 	// Nodes describe the oasis-node(s) to connect to. Keys are "archive
 	// names," which are named after mainnet releases, in lowercase e.g.
 	// "cobalt" and "damask."
-	Nodes map[string]*NodeConfig `koanf:"nodes"`
+	Nodes map[string]*ArchiveConfig `koanf:"nodes"`
+
+	RuntimeNodes map[string]map[common.Runtime]*NodeConfig `koanf:"runtime_nodes"`
```
pro-wh, line 136: o shoot forgot to remove this older idea lol

config/config.go
mitjat: I would be happy to either 
- omit it (I don't expect we'll use it, so simple = good; and we wouldn't lose expressive power, just gain verbosity in one extreme case) or
- stick it under `runtimes` (`layers`? :confounded:).

But LGTM whichever way, including leaving this as-is.

config/config.go
```diff
@@ -200,6 +202,34 @@ type CustomChainConfig struct {
...
+	DefaultNode *NodeConfig `koanf:"default"`
+	// ConsensusNode is information about the node to get consensus data from,
+	// instead of the default node.
+	ConsensusNode *NodeConfig `koanf:"consensus"`
+	// RuntimeNodes is the information about the nodes to get runtime data
+	// from, instead of the default node.
+	RuntimeNodes map[common.Runtime]*NodeConfig `koanf:"runtimes"`
+}
+
+func (ac *ArchiveConfig) ResolveConsensus() *NodeConfig {
```
mitjat, line 217: Suggestion: `ResolvedConsensusNode()`?

`ResolveD` to make it clearer that it's returning the resolved value, not causing a side effect in the state of `ArchiveConfig`.

`ResolvedConsensusNODE` for symmetry with the public `ConsensusNode` field, and because node is the root of the noun phrase, so it feels weird/misleading to omit it.

Same for `ResolveRuntime` below.

config/config.go
```diff
@@ -131,7 +131,7 @@ type SourceConfig struct {
 	// Nodes describe the oasis-node(s) to connect to. Keys are "archive
 	// names," which are named after mainnet releases, in lowercase e.g.
 	// "cobalt" and "damask."
-	Nodes map[string]*NodeConfig `koanf:"nodes"`
+	Nodes map[string]*ArchiveConfig `koanf:"nodes"`
```
mitjat, line 134: Remember to update the checked-in .yml files accordingly.

mitjat (approved):
Thank you! That was more painless than I feared.

config/config.go
pro-wh: :scream: the sample files!

config/config.go
pro-wh: ok

pro-wh:
gonna add some more messages to AnalysisConfig.Validate

PR #400, opened 2023-05-03
db: Disable/enable FK constraints on demand

mitjat:
Resolves https://app.clickup.com/t/860qer9hq

Adds methods for disabling and enabling constraints.
I do not plan to add disabling/enabling indexes; it's a little nuanced what to disable (we need some of them for inserts even during fast-sync), plus they support async updates so I hope they won't slow down inserts _that_ much. To be revisited once db insertions become the bottleneck.
I think we do not need to add disabling/enabling non-FK checks, in particular non-negativity checks for balances. In fast-sync, we won't be doing any dead reckoning, so I reckon (heh) we won't produce negative balances or other values.

Very lightly tested; I just called each method a few times and verified FKs were no longer enforced. Will be better tested once we have the calling code (= fast-sync mode in analyzers).

### DIFF
--- a/storage/api.go
+++ b/storage/api.go
@@ -272,4 +272,15 @@ type TargetStorage interface {
 
 	// Wipe removes all contents of the target storage.
 	Wipe(ctx context.Context) error
+
+	// DisableTriggersAndFKConstraints disables all triggers and foreign key constraints
+	// in indexer tables. This is useful when inserting blockchain data out of order,
+	// so that later blocks can refer to (yet unindexed) earlier blocks without violating constraints.
+	DisableTriggersAndFKConstraints(ctx context.Context) error
+
+	// EnableTriggersAndFKConstraints enables all triggers and foreign key constraints
+	// in the given schema.
+	// WARNING: This might enable triggers not explicitly disabled by DisableTriggersAndFKConstraints.
+	// WARNING: This does not enforce/check contraints on rows that were inserted while triggers were disabled.
+	EnableTriggersAndFKConstraints(ctx context.Context) error
 }
--- a/storage/postgres/client.go
+++ b/storage/postgres/client.go
@@ -208,31 +208,45 @@ func (c *Client) Name() string {
 	return moduleName
 }
 
-// Wipe removes all contents of the database.
-func (c *Client) Wipe(ctx context.Context) error {
-	// List, then drop all tables.
+// Returns all tables that are not internal to Postgres. Table names are fully-qualified,
+// i.e. of the form "<schema>.<table>".
+func (c *Client) listIndexerTables(ctx context.Context) ([]string, error) {
 	rows, err := c.Query(ctx, `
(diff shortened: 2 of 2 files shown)


storage/api.go
```diff
@@ -272,4 +272,15 @@ type TargetStorage interface {
...
+
+	// DisableTriggersAndFKConstraints disables all triggers and foreign key constraints
+	// in indexer tables. This is useful when inserting blockchain data out of order,
+	// so that later blocks can refer to (yet unindexed) earlier blocks without violating constraints.
+	DisableTriggersAndFKConstraints(ctx context.Context) error
+
+	// EnableTriggersAndFKConstraints enables all triggers and foreign key constraints
+	// in the given schema.
+	// WARNING: This might enable triggers not explicitly disabled by DisableTriggersAndFKConstraints.
+	// WARNING: This does not enforce/check contraints on rows that were inserted while triggers were disabled.
```
Andrew7234, line 284: Would it be worthwhile to eventually validate the constraints on the inserted data, at some point after the parallel analyzers have all finished? 

It seems like it's not trivial, we'd have to delete all the constraints, then re-add them which would be expensive. Plus, even enumerating all the constarints for a single table is a [bit gnarly](https://dba.stackexchange.com/questions/214863/how-to-list-all-constraints-of-a-table-in-postgresql). But it would be a useful sanity check to have, even as a one-off similar to the statecheck

mitjat: Agreed that it would be nice to have, and agreed that it's not trivial. 
Opened https://app.clickup.com/t/866a7qg4j

storage/postgres/client.go
```diff
@@ -297,3 +311,40 @@ func (c *Client) Wipe(ctx context.Context) error {
...
+func (c *Client) DisableTriggersAndFKConstraints(ctx context.Context) error {
+	// List all tables, then drop their triggers.
+	// FK constraints are implemented via triggers, so this also disables FK constraints.
+	tables, err := c.listIndexerTables(ctx)
+	if err != nil {
+		return err
+	}
+	for _, table := range tables {
+		c.logger.Info("disabling DB triggers on table", "table", table)
+		if _, err = c.pool.Exec(ctx, fmt.Sprintf("ALTER TABLE %s DISABLE TRIGGER ALL;", table)); err != nil {
```
Andrew7234, line 327: here and below: What do you think of batching all of these together so we never have an intermediate state where some tables have triggers and others don't?

In theory it's unnecessary since we're planning to only enable/disable triggers only when no other analyzers are running... but it also seems relatively simple to add

Andrew7234 (approved):

storage/api.go
csillag: @mitjat:

> The current fast-sync solution (still not fully implemented, so things might change, but unlikely) does not actually disable FKs, so this task is not needed.

PR #401, opened 2023-05-03
cache analyzer sources

Andrew7234:
https://app.clickup.com/t/866a6z1fg

todo: choose better var names

### DIFF
--- a/analyzer/consensus/consensus.go
+++ b/analyzer/consensus/consensus.go
@@ -67,27 +67,16 @@ type Main struct {
 var _ analyzer.Analyzer = (*Main)(nil)
 
 // NewMain returns a new main analyzer for the consensus layer.
-func NewMain(sourceConfig *config.SourceConfig, cfg *config.BlockBasedAnalyzerConfig, target storage.TargetStorage, logger *log.Logger) (*Main, error) {
-	ctx := context.Background()
-
-	// Initialize source storage.
-	client, err := source.NewConsensusClient(ctx, sourceConfig)
-	if err != nil {
-		logger.Error("error creating consensus client",
-			"err", err.Error(),
-		)
-		return nil, err
-	}
-
+func NewMain(cfg *config.BlockBasedAnalyzerConfig, genesisChainContext string, sourceClient *source.ConsensusClient, target storage.TargetStorage, logger *log.Logger) (*Main, error) {
 	// Configure analyzer.
 	blockRange := analyzer.BlockRange{
 		From: cfg.From,
 		To:   cfg.To,
 	}
 	ac := analyzer.ConsensusConfig{
-		GenesisChainContext: sourceConfig.History().CurrentRecord().ChainContext,
+		GenesisChainContext: genesisChainContext,
 		Range:               blockRange,
-		Source:              client,
+		Source:              sourceClient,
 	}
 
 	logger.Info("Starting consensus analyzer", "config", ac)
@@ -101,8 +90,6 @@ func NewMain(sourceConfig *config.SourceConfig, cfg *config.BlockBasedAnalyzerCo
 
 // Start starts the main consensus analyzer.
 func (m *Main) Start(ctx context.Context) {
-	defer m.cleanup()
-
 	// Get block to be indexed.
 	var height int64
 
@@ -180,14 +167,6 @@ func (m *Main) Start(ctx context.Context) {
 	m.logger.Info(
 		fmt.Sprintf("finished processing all blocks in the configured range [%d, %d]",
 			m.cfg.Range.From, m.cfg.Range.To))
-}
-
-func (m *Main) cleanup() {
-	if err := m.cfg.Source.Close(); err != nil {
-		m.logger.Error("failed to cleanly close consensus data source",
-			"err", err.Error(),
-		)
-	}
 }
 
 // Name returns the name of the Main.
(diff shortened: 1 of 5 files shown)


cmd/analyzer/analyzer.go
```diff
@@ -136,10 +138,59 @@ func wipeStorage(cfg *config.StorageConfig) error {
 type Service struct {
 	Analyzers map[string]analyzer.Analyzer
 
+	source *AllSources
 	target storage.TargetStorage
 	logger *log.Logger
 }
 
+type AllSources struct {
```
pro-wh, line 146: AbstractServiceClientFactoryAggregator here we come

Andrew7234: yeah... I was also considering `SourceFactory` but it's not quite satisfactory either

analyzer/consensus/consensus.go
```diff
@@ -67,27 +67,16 @@ type Main struct {
...
-	// Initialize source storage.
-	client, err := source.NewConsensusClient(ctx, sourceConfig)
-	if err != nil {
-		logger.Error("error creating consensus client",
-			"err", err.Error(),
-		)
-		return nil, err
-	}
-
+func NewMain(cfg *config.BlockBasedAnalyzerConfig, chainContext string, sourceClient *source.ConsensusClient, target storage.TargetStorage, logger *log.Logger) (*Main, error) {
```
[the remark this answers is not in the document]
Andrew7234: sure, updated!

analyzer/runtime/runtime.go
```diff
@@ -40,30 +40,21 @@ var _ analyzer.Analyzer = (*Main)(nil)
...
-		return nil, err
-	}
 	roundRange := analyzer.RoundRange{
 		From: uint64(cfg.From),
 		To:   uint64(cfg.To),
 	}
 	ac := analyzer.RuntimeConfig{
 		RuntimeName: runtime,
-		ParaTime:    sourceConfig.SDKParaTime(runtime),
+		ParaTime:    sdkPT,
```
mitjat, line 55: Not sure if I commented on this in a recent PR of Warren's ...
WDYT about naming both the struct attribute and the function param `RuntimeInfo` or `RuntimeMeta` or `RuntimeMetadata`? `sdkPT`:
  a) is a little cryptic,
  b) uses "paratime", which none of the indexer codebase does,
  c) uses "sdk" in the name, which IIRC made sense at some point to differentiate the sdk struct from our own internal metadata-like struct, but now just needlessly echoes the type name.

cmd/analyzer/analyzer.go
```diff
@@ -173,52 +232,88 @@ func NewService(cfg *config.AnalysisConfig) (*Service, error) {
...
 		})
 	}
 	if cfg.Analyzers.Emerald != nil {
 		analyzers, err = addAnalyzer(analyzers, err, func() (A, error) {
-			return runtime.NewRuntimeAnalyzer(common.RuntimeEmerald, &cfg.Source, cfg.Analyzers.Emerald, client, logger)
+			sdkPT := cfg.Source.SDKParaTime(common.RuntimeEmerald)
+			sourceClient, err1 := sources.Runtime(ctx, common.RuntimeEmerald)
+			if err1 != nil {
+				return nil, err1
+			}
```
mitjat, line 249: I'm a little unhappy about the code duplication here, but I don't have great suggestions.

Refloating a recent idea: If `RuntimeClient` used lazy initialization, we could inline `sources.Runtime` and not worry about errors. It would also make it so the `query_on_cache_miss` flag is not needed.
Definitely not something we'd do in this PR, and likely ever, just from the effort/benefits ratio. But a man can dream.

cmd/analyzer/analyzer.go
```diff
@@ -283,6 +379,7 @@ func (a *Service) Start() {
 
 // cleanup cleans up resources used by the service.
 func (a *Service) cleanup() {
+	a.sources.Close()
```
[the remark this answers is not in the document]
mitjat: To test, try running with emerald analyzer AND emerald_evm_tokens AND emerald_token_balances all enabled. They all share the source struct. Then ctrl+c it; ideally, try this whole loop a few times.
Very far from a rigorous test, but it's something.

cmd/analyzer/analyzer.go
```diff
@@ -136,8 +138,59 @@ func wipeStorage(cfg *config.StorageConfig) error {
...
 	Analyzers map[string]analyzer.Analyzer
 
-	target storage.TargetStorage
-	logger *log.Logger
+	sources *sourceFactory
+	target  storage.TargetStorage
+	logger  *log.Logger
+}
+
+// sourceFactory stores the sources used by all the analyzers in a Service.
```
mitjat, line 146:
```suggestion
// sourceFactory stores singletons of the sources used by all the analyzers in a Service.
```

cmd/analyzer/analyzer.go
```diff
@@ -136,8 +138,59 @@ func wipeStorage(cfg *config.StorageConfig) error {
...
+// This enables re-use of node connections as well as graceful shutdown.
```
mitjat, line 147:
```suggestion
// This enables re-use of node connections as well as graceful shutdown.
// NOT thread safe.
```

mitjat (approved):
Thank you!
I have only one important comment, it's about closing resources.

analyzer/runtime/runtime.go
pro-wh: :eyes:

Andrew7234: Good points; switched to `RuntimeMetadata`!

cmd/analyzer/analyzer.go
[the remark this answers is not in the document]
Andrew7234: Good catch, thanks for pointing that out. I tested this as you suggested several times with no issue. I also added some logs to better observe the shutdown.

PR #402, opened 2023-05-03
file: don't crash on cleanup with query_on_cache_miss: no

pro-wh:
we leave the delegate api nil in that case. don't try to close it


Andrew7234 (approved):
thanks!

mitjat (approved):
:pray: mea culpa

PR #403, opened 2023-05-03
analyzer: fix evm.Create result type cutting

pro-wh:
the result is the address, not the address embedded in a 256-bit thing. just the address. don't look for 32 bytes, don't cut it.

tested on a recent evm.Create transaction that occurred on mainnet Emerald, verified that the `to` column in the runtime_transactions table was populated, verified eth addresses displayed in API against another block explorer

### DIFF
--- a/analyzer/runtime/extract.go
+++ b/analyzer/runtime/extract.go
@@ -369,9 +369,9 @@ func ExtractRound(blockHeader nodeapi.RuntimeBlockHeader, txrs []nodeapi.Runtime
 				EVMCreate: func(body *evm.Create, ok *[]byte) error {
 					blockTransactionData.Body = body
 					amount = uncategorized.QuantityFromBytes(body.Value)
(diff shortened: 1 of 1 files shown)


Andrew7234 (approved):
o thanks for catching this!

mitjat (approved):
Nice sleuthing! Thank you!

Can you describe testing in the PR description please? I assume you did an e2e check, i.e. looked at the `to` field of a Create tx, [as defined by our docs](https://github.com/oasisprotocol/oasis-indexer/blob/82e29b83591b75432467ed2303a713a6d3128bc6/api/spec/v1.yaml#L2077-L2077)?

In the near future we'll have a regression dataset checked in, and the changes to those files will be proof enough :+1:

pro-wh:
oh I should make sure the API works too. I've only run the analyzer and looked in the database

pro-wh:
_eth addrs fixed in #406

PR #405, opened 2023-05-04
Ability to run analyzers in parallel

mitjat:
Resolves https://app.clickup.com/t/8669qdzbh

This PR
- adds a `fast_sync` key to each block analyzer's config, wherein the block range and parallelism can be specified
- runs those fast-sync analyzers first
- (TODO) changes "slow-sync" consensus analyzer (now at most one) so it first processes the genesis _right before the first block it will analyze_. If that happens to be the chain genesis document, it grabs that; otherwise, it uses StateToGenesis at an appropriate height. If it notices that at least one block _after_ the beginning of the configured "slow" range has already been processed, it assumes genesis was already processed also, and skips genesis processing. This is different from how we've been processing genesis so far (analyze the newest chain's genesis document at most once, regardless of the configured range), but I think the new way makes more sense. It also allows the slow analyzer to do the right thing wrt genesis in any of the three scenarios:
  - it runs after fast-sync analyzers that have processed some range of blocks
  - it runs after a slow-sync analyzer (after k8s redeployment)
  - it is starting from scratch 

This PR does NOT implement fast-sync behavior (= disabling dead reckoning) or parallelism tolerance inside the analyzers themselves; that happens in other PRs. So the only way to test this config in the very short term is to run "fast-sync" analyzers (that are no different from slow ones) with a parallelism level of 1.

### DIFF
--- a/analyzer/consensus/consensus.go
+++ b/analyzer/consensus/consensus.go
@@ -67,7 +67,7 @@ type ConsensusAnalyzer struct {
 var _ analyzer.Analyzer = (*ConsensusAnalyzer)(nil)
 
 // NewConsensusAnalyzer returns a new main analyzer for the consensus layer.
-func NewConsensusAnalyzer(cfg *config.BlockBasedAnalyzerConfig, genesisChainContext string, sourceClient *source.ConsensusClient, target storage.TargetStorage, logger *log.Logger) (*ConsensusAnalyzer, error) {
+func NewConsensusAnalyzer(cfg config.BlockRange, genesisChainContext string, sourceClient *source.ConsensusClient, target storage.TargetStorage, logger *log.Logger) (*ConsensusAnalyzer, error) {
 	// Configure analyzer.
 	blockRange := analyzer.BlockRange{
 		From: cfg.From,
--- a/analyzer/consensus/consensus.go
+++ b/analyzer/consensus/consensus.go
@@ -79,7 +79,6 @@ func NewConsensusAnalyzer(cfg *config.BlockBasedAnalyzerConfig, genesisChainCont
 		Source:              sourceClient,
 	}
 
-	logger.Info("Starting consensus analyzer", "config", ac)
 	return &ConsensusAnalyzer{
 		cfg:     ac,
 		target:  target,
--- a/analyzer/consensus/consensus.go
+++ b/analyzer/consensus/consensus.go
@@ -56,18 +56,18 @@ func OpenSignedTxNoVerify(signedTx *transaction.SignedTransaction) (*transaction
 	return &tx, nil
 }
 
-// Main is the main Analyzer for the consensus layer.
-type Main struct {
+// ConsensusAnalyzer is the main Analyzer for the consensus layer.
+type ConsensusAnalyzer struct {
 	cfg     analyzer.ConsensusConfig
 	target  storage.TargetStorage
 	logger  *log.Logger
 	metrics metrics.DatabaseMetrics
 }
 
-var _ analyzer.Analyzer = (*Main)(nil)
+var _ analyzer.Analyzer = (*ConsensusAnalyzer)(nil)
 
-// NewMain returns a new main analyzer for the consensus layer.
-func NewMain(cfg *config.BlockBasedAnalyzerConfig, genesisChainContext string, sourceClient *source.ConsensusClient, target storage.TargetStorage, logger *log.Logger) (*Main, error) {
+// NewConsensusAnalyzer returns a new main analyzer for the consensus layer.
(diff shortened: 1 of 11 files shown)


cmd/analyzer/analyzer.go
```diff
@@ -234,16 +236,75 @@ func NewService(cfg *config.AnalysisConfig) (*Service, error) {
 		return nil, err
 	}
 
-	// Initialize analyzers.
-	analyzers := map[string]A{}
+	// Initialize fast-sync analyzers.
+	fastSyncAnalyzers := []A{}
+	if cfg.Analyzers.Emerald != nil {
```
ptrus, line 241:
```suggestion
	if cfg.Analyzers.Consensus != nil {
```

cmd/analyzer/analyzer.go
```diff
@@ -345,22 +417,35 @@ func (a *Service) Start() {
...
+	// Prepare slow-sync analyzers (to be started after fast-sync analyzers are done).
 	var wg sync.WaitGroup
-	for _, an := range a.Analyzers {
+	for _, an := range a.analyzers {
 		wg.Add(1)
 		go func(an analyzer.Analyzer) {
 			defer wg.Done()
-			an.Start(ctx)
+			// Wait for fast-sync analyzers to finish.
+			<-fastSyncAnalyzersDone
```
ptrus, line 438: This wait for the `fastSyncAnalyzersDone` closing  should be moved into the switch below (in place of the `default` case).

That way, the gorotine will also exit if the context is canceled, regardless of the state of the `fastSyncAnalyzersDone` channel.

config/config.go
```diff
@@ -245,28 +245,103 @@ type BlockBasedAnalyzerConfig struct {
...
 
-type IntervalBasedAnalyzerConfig struct {
-	// Interval is the time interval at which to run the analyzer.
-	Interval time.Duration `koanf:"interval"`
+type FastSyncConfig struct {
+	// To is the block (inclusive) to which the analyzer
+	// will run in fast sync mode. From this block onwards (exclusive),
+	// the analyzer will run in slow mode. If absent, the analyzer will
+	// only run in slow mode.
+	// Omitting this parameter or setting it to 0 means to use the latest
```
ptrus, line 274: This comment about omitting the parameter is not true at the current state of the code, right? Is the plan to implement it that way?

If yes, need to be careful with things like `From: blockCfg.FastSync.To + 1` below in the code.

cmd/analyzer/analyzer.go
```diff
@@ -329,14 +390,25 @@ func NewService(cfg *config.AnalysisConfig) (*Service, error) {
...
+		analyzers:         analyzers,
 
 		sources: sources,
 		target:  dbClient,
 		logger:  logger,
 	}, nil
 }
 
+// closingChannel returns a channel that closes when the wait group `wg` is done.
+func closingChannel(wg *sync.WaitGroup) chan struct{} {
```
ptrus, line 403:
```suggestion
func closingChannel(wg *sync.WaitGroup) <-chan struct{} {
```

That way a receive-only channel is returned which cannot be (mistakenly) closed.

analyzer/consensus/consensus.go
```diff
@@ -56,18 +56,18 @@ func OpenSignedTxNoVerify(signedTx *transaction.SignedTransaction) (*transaction
 	return &tx, nil
 }
 
-// Main is the main Analyzer for the consensus layer.
-type Main struct {
+// ConsensusAnalyzer is the main Analyzer for the consensus layer.
+type ConsensusAnalyzer struct {
```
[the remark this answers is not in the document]
mitjat: Urgh that's something that's been inconsistent for a while and I never cared enough to unify it. Fixed now. Another reason I didn't want to touch it is that the naming was split between two conventions: Calling every analyzer `Main`, or calling every analyzer in module `mymodule` `MyModuleAnalyzer`. I went with the latter; I expect opinions/preferences will differ. I don't care which one we use, but I don't really want to deal with bikeshedding and editing. If anybody wants to rename further, I welcome PRs (or commits to this PR).

cmd/analyzer/analyzer.go
mitjat: Ah, nice. Fast-sync analyzers all abort upon ctx.Done(), so in practice I don't think the change matters, but it is cleaner and more principled this way. Thank you!

config/config.go
mitjat: Good catch! An outdated comment. I had this (mostly) implemented but then I decided it wasn't worth the complexity. So I ripped it out, and now the `FastSync.To` parameter has to be a fixed value. I now updated the comment too.

cmd/analyzer/analyzer.go
mitjat: sweet, thank you

mitjat:
This was closed automatically when I accidentally deleted the remote (= github) branch.
Will reopen a new PR.

PR #407, opened 2023-05-05
show us the encrypted data

pro-wh:
alright, if you really want to look :shrug: 

no api

philosophy:
- collapse all callformat variants into a looser type with variable size nonce+data fields.
- same for public key, there's just a variable size slot for it.
- store the format alongside, in case anyone wants to make sense of the keys+nonces+data
- the "public key" in the tx is the one provided by the caller. I'm not aware of us capturing the runtime's public key.
- if anything goes wrong, log and leave it nil. the drawback is that NULL in the db could mean either plain or something failed during analysis (e.g. new callformat that we don't support)
- kinda oddly asymmetric, we don't store the data+result when it's unencrypted

### DIFF
--- a/analyzer/queries/queries.go
+++ b/analyzer/queries/queries.go
@@ -254,8 +254,8 @@ const (
       VALUES ($1, $2, $3, $4)`
 
 	RuntimeTransactionInsert = `
-    INSERT INTO chain.runtime_transactions (runtime, round, tx_index, tx_hash, tx_eth_hash, fee, gas_limit, gas_used, size, timestamp, method, body, "to", amount, success, error_module, error_code, error_message)
-      VALUES ($1, $2, $3, $4, $5, $6, $7, $8, $9, $10, $11, $12, $13, $14, $15, $16, $17, $18)`
+    INSERT INTO chain.runtime_transactions (runtime, round, tx_index, tx_hash, tx_eth_hash, fee, gas_limit, gas_used, size, timestamp, method, body, "to", amount, evm_encrypted_format, evm_encrypted_public_key, evm_encrypted_data_nonce, evm_encrypted_data_data, evm_encrypted_result_nonce, evm_encrypted_result_data, success, error_module, error_code, error_message)
+      VALUES ($1, $2, $3, $4, $5, $6, $7, $8, $9, $10, $11, $12, $13, $14, $15, $16, $17, $18, $19, $20, $21, $22, $23, $24)`
 
 	RuntimeEventInsert = `
     INSERT INTO chain.runtime_events (runtime, round, tx_index, tx_hash, type, body, evm_log_name, evm_log_params, related_accounts)
--- a/analyzer/runtime/evm.go
+++ b/analyzer/runtime/evm.go
@@ -8,7 +8,9 @@ import (
 
 	"github.com/ethereum/go-ethereum/accounts/abi"
 	ethCommon "github.com/ethereum/go-ethereum/common"
+	"github.com/oasisprotocol/oasis-core/go/common/cbor"
 	"github.com/oasisprotocol/oasis-core/go/common/errors"
+	sdkTypes "github.com/oasisprotocol/oasis-sdk/client-sdk/go/types"
 
 	"github.com/oasisprotocol/oasis-indexer/analyzer/evmabi"
 	"github.com/oasisprotocol/oasis-indexer/log"
@@ -48,6 +50,15 @@ type EVMTokenBalanceData struct {
 	// integer is *big.Int, and we don't want anyone fainting if they see a
 	// ** in the codebase.
 	Balance *big.Int
+}
+
+type EVMEncryptedData struct {
+	Format      int
+	PublicKey   []byte
+	DataNonce   []byte
+	DataData    []byte
+	ResultNonce []byte
+	ResultData  []byte
 }
 
 type EVMDeterministicError struct {
@@ -289,4 +300,53 @@ func EVMDownloadTokenBalance(ctx context.Context, logger *log.Logger, source sto
 	default:
 		return nil, fmt.Errorf("download stale token balance type %v not handled", tokenType)
 	}
+}
+
+// EVMMaybeUnmarshalEncryptedData breaks down a possibly encrypted data +
+// result into their encryption envelope fields. If the data is not encrypted,
+// it returns nil with no error.
+func EVMMaybeUnmarshalEncryptedData(data []byte, result *[]byte) (*EVMEncryptedData, error) {
+	var encryptedData EVMEncryptedData
+	var call sdkTypes.Call
+	if cbor.Unmarshal(data, &call) != nil {
+		// Invalid CBOR means it's bare Ethereum format data. This is normal.
+		// https://github.com/oasisprotocol/oasis-sdk/blob/runtime-sdk/v0.3.0/runtime-sdk/modules/evm/src/lib.rs#L626
+		return nil, nil
+	}
+	encryptedData.Format = int(call.Format)
+	switch call.Format {
+	case sdkTypes.CallFormatEncryptedX25519DeoxysII:
+		var callEnvelope sdkTypes.CallEnvelopeX25519DeoxysII
+		if err := cbor.Unmarshal(call.Body, &callEnvelope); err != nil {
+			return nil, fmt.Errorf("outer call format %s unmarshal body: %w", call.Format, err)
+		}
+		encryptedData.PublicKey = callEnvelope.Pk[:]
+		encryptedData.DataNonce = callEnvelope.Nonce[:]
+		encryptedData.DataData = callEnvelope.Data
+	default:
+		return nil, fmt.Errorf("outer call format %s (%d) not supported", call.Format, call.Format)
+	}
+	var callResult sdkTypes.CallResult
+	if result != nil {
+		if err := cbor.Unmarshal(*result, &callResult); err != nil {
+			return nil, fmt.Errorf("unmarshal outer call result: %w", err)
+		}
+		if callResult.IsUnknown() {
+			switch call.Format {
+			case sdkTypes.CallFormatEncryptedX25519DeoxysII:
+				var resultEnvelope sdkTypes.ResultEnvelopeX25519DeoxysII
+				if err := cbor.Unmarshal(callResult.Unknown, &resultEnvelope); err != nil {
+					return nil, fmt.Errorf("outer call result unmarshal unknown: %w", err)
+				}
+				encryptedData.ResultNonce = resultEnvelope.Nonce[:]
+				encryptedData.ResultData = resultEnvelope.Data
+			default:
+				// We have already checked when decoding the call envelope,
+				// but I'm keeping this default case here so we don't forget
+				// if this code gets restructured.
+				return nil, fmt.Errorf("outer call format %s (%d) not supported", call.Format, call.Format)
+			}
+		}
+	}
+	return &encryptedData, nil
 }
--- a/analyzer/runtime/evm.go
+++ b/analyzer/runtime/evm.go
@@ -13,6 +13,7 @@ import (
 	sdkTypes "github.com/oasisprotocol/oasis-sdk/client-sdk/go/types"
 
 	"github.com/oasisprotocol/oasis-indexer/analyzer/evmabi"
+	"github.com/oasisprotocol/oasis-indexer/common"
 	"github.com/oasisprotocol/oasis-indexer/log"
 	"github.com/oasisprotocol/oasis-indexer/storage"
 )
@@ -53,7 +54,7 @@ type EVMTokenBalanceData struct {
 }
 
 type EVMEncryptedData struct {
-	Format      int
+	Format      common.CallFormat
 	PublicKey   []byte
 	DataNonce   []byte
 	DataData    []byte
@@ -313,7 +314,7 @@ func EVMMaybeUnmarshalEncryptedData(data []byte, result *[]byte) (*EVMEncryptedD
 		// https://github.com/oasisprotocol/oasis-sdk/blob/runtime-sdk/v0.3.0/runtime-sdk/modules/evm/src/lib.rs#L626
 		return nil, nil
 	}
-	encryptedData.Format = int(call.Format)
+	encryptedData.Format = common.CallFormat(call.Format.String())
 	switch call.Format {
 	case sdkTypes.CallFormatEncryptedX25519DeoxysII:
 		var callEnvelope sdkTypes.CallEnvelopeX25519DeoxysII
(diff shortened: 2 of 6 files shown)


analyzer/queries/queries.go
```diff
@@ -254,8 +254,8 @@ const (
       VALUES ($1, $2, $3, $4)`
 
 	RuntimeTransactionInsert = `
-    INSERT INTO chain.runtime_transactions (runtime, round, tx_index, tx_hash, tx_eth_hash, fee, gas_limit, gas_used, size, timestamp, method, body, "to", amount, success, error_module, error_code, error_message)
-      VALUES ($1, $2, $3, $4, $5, $6, $7, $8, $9, $10, $11, $12, $13, $14, $15, $16, $17, $18)`
+    INSERT INTO chain.runtime_transactions (runtime, round, tx_index, tx_hash, tx_eth_hash, fee, gas_limit, gas_used, size, timestamp, method, body, "to", amount, evm_encrypted_format, evm_encrypted_public_key, evm_encrypted_data_nonce, evm_encrypted_data_data, evm_encrypted_result_nonce, evm_encrypted_result_data, success, error_module, error_code, error_message)
```
pro-wh, line 257: :steam_locomotive::train::train::train::train::train::train::train::train::train::train::train::train::train::train::train::train::train::train::train::train::train::train::train::train::dash:

storage/migrations/02_runtimes.up.sql
```diff
@@ -48,11 +48,19 @@ CREATE TABLE chain.runtime_transactions
...
   "to"        oasis_addr,   -- Exact semantics depend on method. Extracted from body; for convenience only.
   amount      UINT_NUMERIC, -- Exact semantics depend on method. Extracted from body; for convenience only.
 
+  -- Encrypted data in encrypted Ethereum-format transactions.
+  evm_encrypted_format INTEGER,
+  evm_encrypted_public_key BYTEA,
+  evm_encrypted_data_nonce BYTEA,
+  evm_encrypted_data_data BYTEA,
+  evm_encrypted_result_nonce BYTEA,
+  evm_encrypted_result_data BYTEA,
```
mitjat, line 62: Hm ... IDK how much of this we should store in the DB. Especially storing `data_data` and `result_data` will probably about double the size of each row?
@kostko what was your thinking on what we should make easily accessible to the users? Just the encryption format (including NULL, importantly), or more? The encryption public key is [ephemeral](https://github.com/oasisprotocol/oasis-sdk/blob/c823194ed60ad7a922812e6b803d8ff8b9f78c45/client-sdk/go/types/callformat.go#L7-L7), and the nonce is non-sequential, so neither sounds that useful. The encrypted data is only useful to someone who knows how to decode it locally; I have no idea if we expect users of that caliber, and if we want to support them beyond giving them a public gRPC node.

analyzer/runtime/evm.go
```diff
@@ -290,3 +292,50 @@ func EVMDownloadTokenBalance(ctx context.Context, logger *log.Logger, source sto
...
+		}
+	}
+	switch call.Format {
+	case sdkTypes.CallFormatEncryptedX25519DeoxysII:
+		var callEnvelope sdkTypes.CallEnvelopeX25519DeoxysII
+		if err := cbor.Unmarshal(call.Body, &callEnvelope); err != nil {
+			return nil, nil, nil, nil, nil, nil, fmt.Errorf("outer call format %s unmarshal body: %w", call.Format, err)
+		}
+		publicKey = callEnvelope.Pk[:]
+		dataNonce = callEnvelope.Nonce[:]
```
mitjat, line 324: Here and below: Why do we need to copy? Actually, unlike in Python, this [does not copy](https://go.dev/play/p/9fpRLK-i-xd); if the syntax is intentional, what does it achieve?

mitjat (commented):
Thank you for figuring out the deserialization incantations!

analyzer/runtime/evm.go
```diff
@@ -290,3 +292,50 @@ func EVMDownloadTokenBalance(ctx context.Context, logger *log.Logger, source sto
 		return nil, fmt.Errorf("download stale token balance type %v not handled", tokenType)
 	}
 }
+
+// EVMMaybeUnmarshalEncryptedData breaks down a possibly encrypted data +
+// result into their encryption envelope fields. If the data is not encrypted,
+// it returns nils.
+func EVMMaybeUnmarshalEncryptedData(data []byte, result *[]byte) (*int, *[]byte, *[]byte, *[]byte, *[]byte, *[]byte, error) {
```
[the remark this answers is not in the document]
pro-wh: (: yeah that'd be nice

analyzer/runtime/evm.go
pro-wh: the public key and nonce are arrays, a kind of fixed-size value in Go. the envelope is specific to a certain encryption scheme, where these sizes are known, e.g. a 15-byte nonce for deoxys-ii. this makes a slice that we can put in the db along with potentially different schemes, in case we later come up with a new "callformat"

storage/migrations/02_runtimes.up.sql
pro-wh: yup needs discussion

analyzer/runtime/extract.go
```diff
@@ -53,6 +53,12 @@ type BlockTransactionData struct {
 	Body                    interface{}
 	To                      *apiTypes.Address // Extracted from the body for convenience. Semantics vary by tx type.
 	Amount                  *common.BigInt    // Extracted from the body for convenience. Semantics vary by tx type.
+	EVMEncryptedFormat      *int
+	EVMEncryptedPublicKey   *[]byte
+	EVMEncryptedDataNonce   *[]byte
+	EVMEncryptedDataData    *[]byte
+	EVMEncryptedResultNonce *[]byte
+	EVMEncryptedResultData  *[]byte
```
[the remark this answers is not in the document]
pro-wh: yeah might as well, as long as we make one for EVMMaybeUnmarshalEncryptedData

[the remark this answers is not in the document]
pro-wh: it'll be one of those unusual non-pointer struct ):

analyzer/runtime/evm.go
ptrus: > Here and below: Why do we need to copy? Actually, unlike in Python, this [does not copy](https://go.dev/play/p/9fpRLK-i-xd); if the syntax is intentional, what does it achieve?

You should update `a` to be an array in your example (currently it is a slice). Then it might make more sense:
```golang
a := [3]int{1, 2, 3}
b1 := a    // creates a new array that is a copy of a
b2 := a[:] // creates a slice that references a
```

storage/migrations/02_runtimes.up.sql
mitjat: After today's product sync: Nikola thinks we should show everything. Let's go ahead with this as-is, and we can prune the DB columns later if needed.

analyzer/runtime/evm.go
mitjat: Thank you both! I didn't realize the underlying data was array-typed.

analyzer/runtime/evm.go
```diff
@@ -290,3 +292,50 @@ func EVMDownloadTokenBalance(ctx context.Context, logger *log.Logger, source sto
...
+	var call sdkTypes.Call
+	var publicKey []byte
+	var dataNonce []byte
+	var dataData []byte
+	var callResult sdkTypes.CallResult
+	var resultNonce []byte
+	var resultData []byte
```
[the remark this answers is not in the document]
pro-wh: the structure with these declared out here and the return outside of the case is assuming that other callformats will have these same categories of data though

analyzer/runtime/extract.go
```diff
@@ -53,6 +53,12 @@ type BlockTransactionData struct {
 	Body                    interface{}
 	To                      *apiTypes.Address // Extracted from the body for convenience. Semantics vary by tx type.
 	Amount                  *common.BigInt    // Extracted from the body for convenience. Semantics vary by tx type.
+	EVMEncryptedFormat      *int
+	EVMEncryptedPublicKey   *[]byte
+	EVMEncryptedDataNonce   *[]byte
```
Andrew7234, line 58: are we able to decode the format/nonces/publickey into a more concrete type here? Or can it vary?

storage/migrations/02_runtimes.up.sql
```diff
@@ -48,11 +48,19 @@ CREATE TABLE chain.runtime_transactions
...
   
   -- Transaction contents.
-  method      TEXT,         -- accounts.Transter, consensus.Deposit, consensus.Withdraw, evm.Create, evm.Call. NULL for malformed and encrypted txs.
+  method      TEXT,         -- accounts.Transfer, consensus.Deposit, consensus.Withdraw, evm.Create, evm.Call. NULL for malformed and encrypted txs.
   body        JSONB,        -- For EVM txs, the EVM method and args are encoded in here. NULL for malformed and encrypted txs.
   "to"        oasis_addr,   -- Exact semantics depend on method. Extracted from body; for convenience only.
   amount      UINT_NUMERIC, -- Exact semantics depend on method. Extracted from body; for convenience only.
 
+  -- Encrypted data in encrypted Ethereum-format transactions.
+  evm_encrypted_format INTEGER,
```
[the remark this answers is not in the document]
pro-wh: are we sure about string though?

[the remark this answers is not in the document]
pro-wh: > only interpretable with the help of oasis-sdk

we're not building an alternate call decoder though, why wouldn't we use oasis-sdk?

analyzer/runtime/extract.go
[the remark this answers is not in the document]
pro-wh: actually bleh let's have it as an optional pointer on the tx. it'll just be more verbose in the query code, which I don't think we mind

analyzer/runtime/evm.go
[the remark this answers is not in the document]
pro-wh: added EVMEncryptedData

analyzer/queries/queries.go
pro-wh: too scary to do in this PR

pro-wh: #414

analyzer/runtime/evm.go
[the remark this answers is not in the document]
pro-wh: I'm gonna restructure this to unify the per-format 'if result not nil' and 'if result unknown' and have two switches on the callformat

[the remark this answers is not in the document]
pro-wh: restructured, it should be more intuitive why this is declared not locally to the switch

storage/migrations/02_runtimes.up.sql
pro-wh: kept..........................................................................

storage/migrations/02_runtimes.up.sql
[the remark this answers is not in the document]
pro-wh: ok I looked into this some more. the sdk even has a method to stringify the callformat number into a proper looking name, e.g. "encrypted/x25519-deoxysii." so we'll at least have a systematic way to name them. but be advised there's no api to go from name back to sdk callformat number :laughing:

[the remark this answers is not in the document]
pro-wh: ok added as a separate commit, we can decide either way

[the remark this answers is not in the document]
pro-wh: cc @Andrew7234 , the tip of this branch now has common.CallFormat (string) type Format field

storage/migrations/02_runtimes.up.sql
```diff
@@ -3,6 +3,7 @@
 BEGIN;
 
 CREATE TYPE runtime AS ENUM ('emerald', 'sapphire', 'cipher');
+CREATE TYPE call_format AS ENUM ('encrypted/x25519-deoxysii');
```
mitjat, line 6: evm_call_format?

storage/migrations/02_runtimes.up.sql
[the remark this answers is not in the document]
mitjat: Thank you! I'm inclined to leave it. In general, I've been pushing to have strings/enums in the tables so the data representation is a) more self-contained and b) less dependent on oasis-core and oasis-sdk.

Having strings is not a problem. Each enum column is only 4 bytes wide. And they're nice for readability; I think they're the only supported kind in postgres anyway :)

analyzer/runtime/extract.go
```diff
@@ -375,6 +376,18 @@ func ExtractRound(blockHeader nodeapi.RuntimeBlockHeader, txrs []nodeapi.Runtime
 							return fmt.Errorf("created contract: %w", err)
 						}
 					}
+					var evmEncrypted *EVMEncryptedData
+					evmEncrypted, err = EVMMaybeUnmarshalEncryptedData(body.InitCode, ok)
+					if err == nil {
```
mitjat, line 381: Nit:

```suggestion
					if evmEncrypted, err2 := EVMMaybeUnmarshalEncryptedData(body.InitCode, ok); err2 == nil {
```

I find it more readable/idiomatic, but your call. 
If you go with this, you need to also use `err2` in the `logger.Error()` below.

Same in the `EVMCall` branch.

mitjat (approved):

storage/migrations/02_runtimes.up.sql
pro-wh: call formats are actually defined outside of the evm, and they're also used in the whole-transaction "cipher" style encryption. could go with runtime_call_format, if we want to be clear it's not used in consensus. shall we?

storage/migrations/02_runtimes.up.sql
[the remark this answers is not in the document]
pro-wh: :+1:

analyzer/runtime/extract.go
pro-wh: we need to write down some guidelines for this, because there's a lot of shadowing going on in this file and I don't get it

but I'm okay with using err2

pro-wh: what about `to, err = registerRelatedEthAddress` above?

storage/migrations/02_runtimes.up.sql
pro-wh: I'm gonna merge this with the current name

analyzer/runtime/extract.go
mitjat: Hehe I think that's all predominantly your code :P ... From the initial hacky import from the nexus repo.

Yeah I'm also surprised by how much shadowing we have here (I think we avoid it more, in general, hence the original comment), and also surprised that the linter doesn't complain about this instance of shadowing. I know that it sometimes does. Anyway. I'm very OK with sprucing up whatever catches your eye. I didn't want to suggest more stylistic nits because they're squarely outside the PR scope.

pro-wh: rules in nexus were:
- start with `err`
- never shadow, even in cases where the linter would allow shadowing
- clobbering is okay, even through lambda boundary (nexus had not been using goroutines)
- prefer to clobber except when it interferes with a place where `:=` would allow us to omit specifying out an explicit type with `var`
- after `err` is `err1`, `err2`, etc
- use the lowest one that doesn't violate the above rules

PR #409, opened 2023-05-05
bugfix: KVStore (analyzer cache): Fix retrieval and error handling

Andrew7234:
[Mitja] This PR cleans up KVStore and fixes a bug in the process:
- Abstracts away fetching of a known-type value into a separate helper func
- Abstracts away the KVStore interface
- Adds unit tests (that fail the old code)
- Fixes a bug in `GetFromCacheOrCall` where if a value was in the cache, we used to not use it
- Fixes a bug in `GetFromCacheOrCall` where if a value was in the cache but couldn't be retrieved correctly (e.g. because of the wrong type), we logged too many errors

### DIFF
--- a/storage/oasis/nodeapi/file/common.go
+++ b/storage/oasis/nodeapi/file/common.go
@@ -86,6 +86,8 @@ func GetFromCacheOrCall[Value any](cache KVStore, volatile bool, key CacheKey, v
 		if err2 != nil {
 			cache.logger.Warn(fmt.Sprintf("failed to unmarshal key %s from cache into %T: %v", key.Pretty(), result, err2))
 		}
+
+		return result, nil
 	}
 
 	// Otherwise, the value is not cached or couldn't be restored from the cache. Call the backing API to get it.
--- a/storage/oasis/nodeapi/file/common.go
+++ b/storage/oasis/nodeapi/file/common.go
@@ -95,10 +95,10 @@ func GetFromCacheOrCall[Value any](cache KVStore, volatile bool, key CacheKey, v
 	}
 
 	// If the value is cached, return it.
-	var result *Value
-	switch err := fetchTypedValue(cache, key, result); err {
+	var cached Value
+	switch err := fetchTypedValue(cache, key, &cached); err {
 	case nil:
-		return result, nil
+		return &cached, nil
 	case errNoSuchKey: // Regular cache miss; continue below.
 	default:
 		// Log unexpected error and continue to call the backing API.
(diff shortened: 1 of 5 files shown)


storage/oasis/nodeapi/file/common.go
```diff
@@ -86,6 +86,8 @@ func GetFromCacheOrCall[Value any](cache KVStore, volatile bool, key CacheKey, v
 		if err2 != nil {
 			cache.logger.Warn(fmt.Sprintf("failed to unmarshal key %s from cache into %T: %v", key.Pretty(), result, err2))
 		}
+
+		return result, nil
```
ptrus, line 90: Shouldn't the non-cached value be fetched and returned in case errors happen during unmarshaling of the cached value (lines 82 and 87)?

mitjat: :heavy_plus_sign: : 
Plus I had another objection to my own code (a few lines up, if `cache.Has()` returns an error, we just errored out instead of going to the backing store). Then got carried away :disappointed:. Pushed a new commit. Andy, thank you for finding this and doing the initial fix. What a Wild West, I really should have added tests :/

ptrus (approved):

pro-wh (commented):
what does this do?

storage/oasis/nodeapi/file/common.go
```diff
@@ -72,24 +95,19 @@ func GetFromCacheOrCall[Value any](cache KVStore, volatile bool, key CacheKey, v
...
-		raw, err2 := cache.Get(key)
-		if err2 != nil {
-			cache.logger.Warn(fmt.Sprintf("failed to fetch key %s from cache: %v", key.Pretty(), err2))
-		}
-		var result *Value
-		err2 = cbor.Unmarshal(raw, &result)
-		if err2 != nil {
-			cache.logger.Warn(fmt.Sprintf("failed to unmarshal key %s from cache into %T: %v", key.Pretty(), result, err2))
-		}
+	var result *Value
```
pro-wh, line 98: you have it unmarshal into a nil pointer? :raised_eyebrow:

storage/oasis/nodeapi/file/kvstore_test.go
```diff
@@ -0,0 +1,111 @@
...
+	"os"
+	"testing"
+
+	"github.com/oasisprotocol/oasis-indexer/log"
+	"github.com/stretchr/testify/require"
+)
+
+// Returns an open KVStore, and a function to clean it up.
+func openTestKVStore(t *testing.T) (KVStore, func()) {
+	const path = "/tmp/indexer-kv-test"
```
ptrus, line 14: Small nit:
```suggestion
	path, err := os.MkdirTemp("", "indexer-kv-test")
```

to ensure that the tests won't use a potentially left-over data if a previous test run process exited uncleanly and didn't got to clean the folder.

mitjat (approved):
selfhighfive.gif :roll_eyes: 

I have to formally approve since `ptrus` is not in CODEOWNERS.

PR #410, opened 2023-05-05
Andrew7234/ci e2e

Andrew7234:
https://app.clickup.com/t/866a5pprn

Note: I'm holding off on making this step required in CI for a week or longer in order to verify that these tests are as reliable as we hope they are.

* Timestamps are displayed differently in CI `"2022-04-11T11:12:31Z"` vs local machine `"2022-04-11T04:12:31-07:00"` [x]
* Also this one: 
``` diff --git a/home/runner/work/oasis-indexer/oasis-indexer/tests/e2e_regression/expected/spec.headers b/home/runner/work/oasis-indexer/oasis-indexer/tests/e2e_regression/actual/spec.headers
index 35266bd..a49fe6d 100644
--- a/home/runner/work/oasis-indexer/oasis-indexer/tests/e2e_regression/expected/spec.headers
+++ b/home/runner/work/oasis-indexer/oasis-indexer/tests/e2e_regression/actual/spec.headers
@@ -1,7 +1,7 @@
 HTTP/1.1 200 OK
 Accept-Ranges: bytes
 Content-Length: UNINTERESTING
-Content-Type: text/plain; charset=utf-8
+Content-Type: application/x-yaml``` [x]

### CODE REVIEW

File: .github/workflows/ci-test.yaml [hunk 1]
```diff
@@ -113,3 +113,26 @@ jobs:
...
+          make codegen-go
+      - name: Build Go
+        run: |
+          make oasis-indexer
+      - name: Start db
+        run: |
+          make postgres
+      - name: Run e2e regression tests
+        run : |
+          make test-e2e-regression
```
mitjat, line 138: I suggest you also create and upload artifacts for easier uploading: Lines 103-115 from the non-regression e2e test. You can add `(SOME_ABS_PATH?)/tests/e2e_regression/actual` to the paths for `actions/upload-artifact`.

tests/e2e_regression/e2e_config.yml
```diff
@@ -2,23 +2,24 @@ analysis:
   source:
     cache:
       cache_dir: tests/e2e_regression/data
```
mitjat, line 4: can we pls rename `data` to `rpc-cache` (similar to [what we have in k8s](https://github.com/oasislabs/private-charts/blob/14b978e81ccf2928e0ac2b0188f290c1cf61e558/charts/oasis-indexer/templates/deployment.yaml#L72-L72)), or else something more informative?

mitjat (commented):
Looking good!

Looks like `run.sh` is giving false positives?
- I added suggestions for easier debugging of CI failures (in this PR and in the future).
- IIRC we might still have some queries that return results in a nondeterministic order; that could explain the failures.

tests/e2e_regression/run.sh
```diff
@@ -103,14 +105,14 @@ for (( i=0; i<nCases; i++ )); do
...
-  sed -E -i='' 's/^(Date|Content-Length|Last-Modified): .*/\1: UNINTERESTING/g' "$outDir/$name.headers"
+  sed -i -E 's/^(Date|Content-Length|Last-Modified): .*/\1: UNINTERESTING/g' "$outDir/$name.headers"
 done
 
 diff --recursive "$SCRIPT_DIR/expected" "$outDir" >/dev/null || {
   echo
   echo "NOTE: $SCRIPT_DIR/expected and $outDir differ."
   echo "Press enter see the diff, or Ctrl-C to abort."
-  read -r
+  read -r -t 5
```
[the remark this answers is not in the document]
Andrew7234: thanks! updated

.github/workflows/ci-test.yaml
Andrew7234: good idea, updated!

tests/e2e_regression/e2e_config.yml
Andrew7234: sure, done

tests/e2e/e2e_test.go
```diff
@@ -34,8 +34,8 @@ const (
 	// The amount of time to wait after a blockchain tx to make sure that the a block with the
 	// tx was produced and the indexer has indexed the block.
 	// The node seems to generate a block per second.
-	// 2 seconds is experimentally enough, 1 second was flaky.
-	indexerDelay = 2 * time.Second
+	// 6 seconds is experimentally enough.
+	indexerDelay = 6 * time.Second
```
Andrew7234, line 38: Since the indexer now tries to index past the current height (see changes to block.go), the indexer fails often and thus hits the max backoff during the e2e tests while waiting for the node to produce blocks. Changed to 6 seconds here to be in line with the max backoff.

analyzer/block/block.go
```diff
@@ -192,20 +192,9 @@ func (b *blockBasedAnalyzer) Start(ctx context.Context) {
...
-		if err != nil {
-			b.logger.Error("failed to query latest block height on source",
-				"err", err,
-			)
-			backoff.Failure()
-			continue
-		}
-		to = latestBlockHeight
-
+		var to uint64 = util.MaxBlockHeight
```
Andrew7234, line 195: Per discussion from slack with mitja; 

When the analyzers are running in fast sync mode, we expect to run them only to some height h < current_height anyway. Then in slow sync mode there's only one analyzer, so it can lock all it wants -- it will just discover early on in the batch that the block(s) don't exist, it will unlock the rest of the blocks, and go again.

One (expected) side effect of this is that the backoff failure is now triggered repeatedly when the indexer is synced. See `e2e_test.go` for more

ptrus: > One (expected) side effect of this is that the backoff failure is now triggered repeatedly when the indexer is synced. See e2e_test.go for more

I think it would be great if this could be avoided, otherwise this will polute the error log and also unnecessarly increase the backoff.

There used to be this `ErrOutOfRange` (which is still in code, but not actually used anywhere since #389). Maybe it can be introduced back, and in case the block-producer returns it, the block analyzer doesn't log the error (but a debug/info log instead).

tests/e2e/e2e_test.go
ptrus: It's not totaly clear to me why this happens in case `slowSync=true` - which is currently always the case. I think the backoff-logic is roughly same as it was before #389: https://github.com/oasisprotocol/oasis-indexer/blob/849bf5a8d91406a846de96b41d7c44c934518b5a/analyzer/consensus/consensus.go#L137-L166

In slow-sync mode it will stop processing the batch after the first failure so I would assume there would be a similar number of `backoff.Failure` calls as in the old code. 

I probably missed something.

(it is clear why this would be problematic in fast-sync mode, however that's not used at the moment).

analyzer/block/block.go
ptrus: > I think it would be great if this could be avoided, otherwise this will unnecessarily pollute the error log.

Oh, by _"this"_ i only ment logging the error, should have worded it better. All clear on wanting to have cached calls only.

---
 
Also please move the configuring of the `to` value out of the loop, one level higher, eg. to line 167.

tests/e2e/e2e_test.go
ptrus: > To me, it looks like we're not calling backoff.Success() frequently enough

I think we call it after every successfully processed block, as before?

> In any case, 2 seconds consistently failed during my testing, and I think bumping it up to 6 seconds is the most future-proof value for now since it's in line with the max backoff. (Also since the e2e_test takes 4-5min to run anyway the additional 4 seconds is relatively minor).

Agreed that increasing the test timeout is minor, but this also indicates that the indexer will be ~6 seconds behind when processing at the tip of the chain. Which previously was not the case.

---

Anyway, I found out the reason for the regression. In old code, we waited for "current-timeout" after processing every block. 
With the updated code, we only wait for the timeout period when picking a batch. And because we pick 1000 future blocks we are almost guaranteed to fail mid batch. This, together with failures being more punishable than successes rewarded then leads to this error.

I would suggest doing this:

```golang
		for _, height := range heights {
			// If running in slow-sync, we are likely at the tip of the chain and are picking up
			// blocks that are not yet available. In this case, wait before processing every block,
			// so that the backoff mechanism can tweak the per-block wait time as needed.
			if b.slowSync {
				select {
				case <-time.After(backoff.Timeout()):
					// Process another batch of blocks.
				case <-ctx.Done():
					b.logger.Warn("shutting down block analyzer", "reason", ctx.Err())
					return
				}
			}

			b.logger.Info("processing block", "height", height)
			// Rest of the code...


```

I'm almost certain that with this change you should be able to revert the `indexerDelay` change, but have not explicitly tested it myself.

No hard feelings if you decide to disregard the suggestion, though.

Thanks!

Andrew7234: Thanks for digging into that! Happy to include your suggestion; will test it and update

mitjat: Nice. Just in case it doesn't work out, an alternative (and probably equivalent) approach is to move `blocksBatchSize` from being a constant to being a property of `blockAnalyzer`, and setting it to 1 in the constructor when `slowSync` is enabled.

Andrew7234: Looks like @ptrus 's suggestion worked; I'll leave the `blocksBatchSize` change for a potential future PR :)

analyzer/block/block.go
Andrew7234: Got it; reintroduced `ErrOutOfRange` and moved `to` out

analyzer/block/block.go
```diff
@@ -223,13 +214,40 @@ func (b *blockBasedAnalyzer) Start(ctx context.Context) {
...
+			//
+			// Note: If the batch size is greater than 50, the time required to process the blocks
+			// in the batch will exceed the current lock expiry of 5min. The analyzer will terminate
+			// the batch early and attempt to refresh the locks for a new batch.
+			if b.slowSync {
+				select {
+				case <-time.After(backoff.Timeout()):
+					// Process the next block
+				case <-batchCtx.Done():
+					b.logger.Info("batch locks expiring; refreshing batch")
```
ptrus, line 229: Can `break` here, to avoid going through remaining blocks if the context is canceled anyway.

analyzer/block/block.go
ptrus: Thanks!

tests/e2e/e2e_test.go
ptrus: Thanks!

ptrus (approved):

tests/e2e_regression/e2e_config.yml
```diff
@@ -1,24 +1,25 @@
...
     fast_startup: true
   analyzers:
-    metadata_registry:
-      interval: 5m
-    aggregate_stats:
-      tx_volume_interval: 5m
+    # metadata_registry:
+    #   interval: 5m
+    # aggregate_stats:
+    #   tx_volume_interval: 5m
```
mitjat, line 16: Note: Our non-block-based analyzers also have a fair bit of complexity in them, and it would be good to test them also. That's a separate task though.

The git-based `metadata_registry` is an exception, but everything else we could fit into this framework. I think what we want to do is have a way to run all the block analyzers to the end, then run all the `evm_tokens_*` and `evm_token_balances_*` analyzers and `aggregate_stats`. We just don't have a way yet of telling them "exit when there is no more work for you".

mitjat (approved):
Thank you! I'm genuinely excited to try this out in the next PRs.

Q: I vaguely remembered (and IIRC you confirmed) that some API outputs have a non-deterministic order. But I don't see any changes in the server code that fixes this. So is everything deterministic after all?

tests/e2e_regression/e2e_config.yml
mitjat: Created https://app.clickup.com/t/866aak668

PR #411, opened 2023-05-09
storage: expose test postgres client and add migrations test

ptrus:
This introduces a postgres `testutils` package, which allows connecting to the test database from other tests in the codebase. An analyzer test that executes the migrations is also included as an example of usage (the test ensures migrations are applied without errors).

This was needed/useful in https://github.com/oasisprotocol/oasis-indexer/pull/389 and was extracted into a separate PR.

### DIFF
--- a/.github/workflows/ci-test.yaml
+++ b/.github/workflows/ci-test.yaml
@@ -34,7 +34,7 @@ jobs:
         ports:
           - 5432:5432
     env:
-      CI_TEST_CONN_STRING: "postgresql://postgres:postgres@127.0.0.1:5432/postgres"
+      CI_TEST_CONN_STRING: "postgresql://postgres:postgres@127.0.0.1:5432/postgres?sslmode=disable"
     steps:
       - name: Checkout code
         uses: actions/checkout@v3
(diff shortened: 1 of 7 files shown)


.github/workflows/ci-test.yaml
```diff
@@ -34,7 +34,7 @@ jobs:
         ports:
           - 5432:5432
     env:
-      CI_TEST_CONN_STRING: "postgresql://postgres:postgres@127.0.0.1:5432/postgres"
+      CI_TEST_CONN_STRING: "postgresql://postgres:postgres@127.0.0.1:5432/postgres?sslmode=disable"
```
mitjat, line 37: What changed so that this switch is needed?

mitjat (approved):
Nice cleanup, thank you!

.github/workflows/ci-test.yaml
ptrus: The `golang-migrate` (library used for migrations) requires it (it tries to use a ssl connection by default). 

Before, migrations were not run in uni tests. Other parts of the code (e.g. e2e-tests) already have this configured: https://github.com/oasisprotocol/oasis-indexer/blob/4d3a6adc8765e2414fe82fa2e403771e96535118/tests/e2e_regression/e2e_config.yml#L26

PR #412, opened 2023-05-09
node config: Support testnet

mitjat:
Tested locally with port-forwarding that this allows us to index testnet consensus and sapphire at the same time.

### DIFF
--- a/config/history.go
+++ b/config/history.go
@@ -129,7 +129,6 @@ var DefaultChains = map[common.ChainName]*History{
 	common.ChainNameTestnet: {
 		Records: []*Record{
 			// TODO: coalesce compatible records
-			// TODO: rename archives to match compatible API
 			// TODO: fill in chain context
 			{
 				// https://github.com/oasisprotocol/testnet-artifacts/releases/tag/2022-03-03
--- a/storage/oasis/nodeapi/history/history.go
+++ b/storage/oasis/nodeapi/history/history.go
@@ -20,21 +20,34 @@ var _ nodeapi.ConsensusApiLite = (*HistoryConsensusApiLite)(nil)
 
 type APIConstructor func(ctx context.Context, chainContext string, archiveConfig *config.ArchiveConfig, fastStartup bool) (nodeapi.ConsensusApiLite, error)
 
+func damaskAPIConstructor(ctx context.Context, chainContext string, archiveConfig *config.ArchiveConfig, fastStartup bool) (nodeapi.ConsensusApiLite, error) {
+	sdkConn, err := connections.SDKConnect(ctx, chainContext, archiveConfig.ResolvedConsensusNode(), fastStartup)
+	if err != nil {
+		return nil, err
+	}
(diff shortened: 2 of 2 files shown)


storage/oasis/nodeapi/history/history.go
```diff
@@ -20,21 +20,31 @@ var _ nodeapi.ConsensusApiLite = (*HistoryConsensusApiLite)(nil)
...
-		if err != nil {
-			return nil, fmt.Errorf("indexer RawConnect: %w", err)
-		}
-		return cobalt.NewCobaltConsensusApiLite(rawConn), nil
-	},
+	// mainnet
+	"damask": DamaskAPIConstructor,
+	"cobalt": CobaltAPIConstructor,
+	// testnet
+	"2022-03-03": DamaskAPIConstructor,
```
Andrew7234, line 47: huh i never noticed we named testnets just by date

Andrew7234 (approved):
looks good!

storage/oasis/nodeapi/history/history.go
pro-wh: mainnet releases are by date too, at least on the mainnet-artifacts repo

storage/oasis/nodeapi/history/history.go
```diff
@@ -20,21 +20,31 @@ var _ nodeapi.ConsensusApiLite = (*HistoryConsensusApiLite)(nil)
 
 type APIConstructor func(ctx context.Context, chainContext string, archiveConfig *config.ArchiveConfig, fastStartup bool) (nodeapi.ConsensusApiLite, error)
 
+var DamaskAPIConstructor = func(ctx context.Context, chainContext string, archiveConfig *config.ArchiveConfig, fastStartup bool) (nodeapi.ConsensusApiLite, error) {
```
pro-wh, line 23: could we do the normal looking `func DamaskAPIConstructor(...) ... {`

config/history.go
```diff
@@ -129,7 +129,6 @@ var DefaultChains = map[common.ChainName]*History{
 	common.ChainNameTestnet: {
 		Records: []*Record{
 			// TODO: coalesce compatible records
-			// TODO: rename archives to match compatible API
```
pro-wh, line 132: lol k :shrug:

pro-wh (approved):

config/history.go
mitjat: I didn't want to name the most recent testnet "damask" because with testnet, we'll probably (if we ever want to support the whole testnet history) end up having more than one archive that can be accessed with DamaskConsensusApiLite. Then we'd need "damask" as a map key for 2+ archives.

pro-wh: o ok

storage/oasis/nodeapi/history/history.go
mitjat: yeah oasis-core does that already. they don't deserve zingy codenames :)

PR #415, opened 2023-05-10
Andrew7234/feature/envelope api

Andrew7234:
updates `{runtime}/transactions` endpoint post #407 

```
curl "localhost:8008/v1/sapphire/transactions?limit=10&block=411186" | jq                                                                                                                                                                 andrewlow@valinor
100  4059    0  4059    0     0   220k      0 --:--:-- --:--:-- --:--:--  220k
{
  "total_count": 1,
  "transactions": [
    {
      "amount": "0",
      "body": {
        "address": "jAZLz3wNo7OwkLq/6PMyNTTYTWg=",
        "data": "omRib2R5o2Jwa1gglAo9bAaeN/hIdanHuSzKR0ZKdCmxyD50GvD/wk26J3ZkZGF0YVkEXbyLYde4OekicawaA3pQUfjj8xdT1WyVZOr4Yyheuu4sW4LBTYvubx3P9hwSy1JjzQjIg8+/vil29bq7FhI2P32N/UifnHJaTKNgGhGnWfonaxPEhGi7pduHCLDXcy26HXrWQPOFuBtabhfJrrt1AHQJZMjlu2tJJow5YgpxEsHNTp2iAqcKz6vSLX7lAzzzWvCPsA75Ll/rsPxkEwxsLI5jEYZt8REOAWBnQKTXkuLV8INNMjNGV3VepGQnSvSWZNzxmLyN+I5p3hJHMgzhluSsxv0SE9CWNAfYHL8YC4y/VVR9Xxlgxn2qmwfQ+vf2uRayQelwjG53kPRKhBDCkax/iQVS9qRDTMJv84jz/+vQkckRTx+VE11VDSjCcTp1WqYpaCiY0p//rIXT1wI98s6nLcDaxPQLQ5v4eRRq9uNuvyP+ZnS+wsrx123wA3t6uQmG9QjLpNjjvO5FxdzenfY/kXX45LUybClEQBo7YHVJ6EQsuc1RSigws9GvkJ+HOlAmYpwfc83y9J7dmRehikCCsqAWAt0pIkkB3R7KJCLWrbP5UVR9n4PFhoyEb/yQolaxE+T4PE0sFKKttdTfVYBBV2h+orIJvZ+Xy7O8bp/NcVzsZ/PFLMdeAU/mSHPrtZiO8PpWNVmViVMVzQXckqwKplL3OvC6zx947HbYFT6R8m7NsgglFZWkplbm7MZ5qT1uRC7jJYuzIO5HpmsgaVFMPphcIDZRek6Qn8BPmMeyI8CZXqOModT3mWzYsxeKB613Q9D84GbRxly0xgYE12GZ/niYTFwqxyja54NUL1KOkEoH4pyUZLL0ro1KeAFPY9khcoTKa6jxvDE6O89udOgfWDhwU2VNFjSs96qe1PEan4Nd+TwhtvfXLP1LuMXRzp8IDuiUtHFlI7mCisxdSczk37NaqTbo2QOI3WCyb0czi2pLLWmg3hgv/3YLU4oLz737kfAPDg3FmOg3DKpIPDgKHqXDCatg3LHUPQvbyI9A14nJO95uLDGva8+GipjBc4jFrSSNXyNmC/RCYp6iTEndUbTh/7/cOro2GrNoLpfOEYUyeU8RMzMqpY2L4hvebwCdDuV+DuCmxf7skel7mUiKsTYYEvBE0V18A2ZFmETk0Uk4DKaRpZQ+t2GyECv2wjZ+VaQANn5Zm43bmCSXfbI+e+UuensHWCyexqwyZ7NYBFYov3mPS5SSNwjbqJhXCRCaco+Zj6FK+aNLLHHdYseE3UPgTfMuC7gztE3Mz/bcBVW6zjNRfle2HPGEyfazZt+6cZ1Awxdsjzb4E6ExSMBhDXuKMYYPz3TuFoRuO264JEL13dUxRso9Z7SmhWEjSwf9Na01f/Tiqr/uvDFOlL1hxAlIelgUi6CvI+Xy6gckNnaWzejHXymLXWV7diFkt7Yr3fvLlcO5ciyx28s9UMDyIL5d2Ocop82sIzKEdtRTe+1D2Qf1TTTy6T/7Ph8sA98YJ1U7fqyyXfgKJ8dlbm9uY2VPN/74MYNyPMhAUy3EC3vDZmZvcm1hdAE=",
        "value": ""
      },
      "encryption_envelope": {
        "data": "vIth17g56SJxrBoDelBR+OPzF1PVbJVk6vhjKF667ixbgsFNi+5vHc/2HBLLUmPNCMiDz7++KXb1ursWEjY/fY39SJ+cclpMo2AaEadZ+idrE8SEaLul24cIsNdzLbodetZA84W4G1puF8muu3UAdAlkyOW7a0kmjDliCnESwc1OnaICpwrPq9ItfuUDPPNa8I+wDvkuX+uw/GQTDGwsjmMRhm3xEQ4BYGdApNeS4tXwg00yM0ZXdV6kZCdK9JZk3PGYvI34jmneEkcyDOGW5KzG/RIT0JY0B9gcvxgLjL9VVH1fGWDGfaqbB9D69/a5FrJB6XCMbneQ9EqEEMKRrH+JBVL2pENMwm/ziPP/69CRyRFPH5UTXVUNKMJxOnVapiloKJjSn/+shdPXAj3yzqctwNrE9AtDm/h5FGr2426/I/5mdL7CyvHXbfADe3q5CYb1CMuk2OO87kXF3N6d9j+RdfjktTJsKURAGjtgdUnoRCy5zVFKKDCz0a+Qn4c6UCZinB9zzfL0nt2ZF6GKQIKyoBYC3SkiSQHdHsokItats/lRVH2fg8WGjIRv/JCiVrET5Pg8TSwUoq211N9VgEFXaH6isgm9n5fLs7xun81xXOxn88Usx14BT+ZIc+u1mI7w+lY1WZWJUxXNBdySrAqmUvc68LrPH3jsdtgVPpHybs2yCCUVlaSmVubsxnmpPW5ELuMli7Mg7kemayBpUUw+mFwgNlF6TpCfwE+Yx7IjwJleo4yh1PeZbNizF4oHrXdD0PzgZtHGXLTGBgTXYZn+eJhMXCrHKNrng1QvUo6QSgfinJRksvSujUp4AU9j2SFyhMprqPG8MTo7z2506B9YOHBTZU0WNKz3qp7U8Rqfg135PCG299cs/Uu4xdHOnwgO6JS0cWUjuYKKzF1JzOTfs1qpNujZA4jdYLJvRzOLakstaaDeGC//dgtTigvPvfuR8A8ODcWY6DcMqkg8OAoepcMJq2DcsdQ9C9vIj0DXick73m4sMa9rz4aKmMFziMWtJI1fI2YL9EJinqJMSd1RtOH/v9w6ujYas2gul84RhTJ5TxEzMyqljYviG95vAJ0O5X4O4KbF/uyR6XuZSIqxNhgS8ETRXXwDZkWYROTRSTgMppGllD63YbIQK/bCNn5VpAA2flmbjduYJJd9sj575S56ewdYLJ7GrDJns1gEVii/eY9LlJI3CNuomFcJEJpyj5mPoUr5o0sscd1ix4TdQ+BN8y4LuDO0TczP9twFVbrOM1F+V7Yc8YTJ9rNm37pxnUDDF2yPNvgToTFIwGENe4oxhg/PdO4WhG47brgkQvXd1TFGyj1ntKaFYSNLB/01rTV/9OKqv+68MU6UvWHECUh6WBSLoK8j5fLqByQ2dpbN6MdfKYtdZXt2IWS3tivd+8uVw7lyLLHbyz1QwPIgvl3Y5yinzawjMoR21FN77UPZB/VNNPLpP/s+HywD3xgnVTt+rLJd+Aonxw==",
        "data_nonce": "N/74MYNyPMhAUy3EC3vD",
        "format": "encrypted/x25519-deoxysii",
        "public_key": "lAo9bAaeN/hIdanHuSzKR0ZKdCmxyD50GvD/wk26J3Y=",
        "result": "kLwKdO9D4lZoY4zFEe86bt9VZ60W",
        "result_nonce": "AAAAAAAGRjEAAAAAAAAA"
      },
      "eth_hash": "12106a139c9eb799f25b2fd3879827c23c737debefa63da8569560ba556e163c",
      "fee": "100000000000000000",
      "gas_limit": 1000000,
      "gas_used": 59190,
      "hash": "c79a1e8f7ea8e585018ac7cd5a7e0f2fe52243917ae620170ec56f90d3765cff",
      "index": 0,
      "method": "evm.Call",
      "nonce_0": 678,
      "round": 411186,
      "sender_0": "oasis1qzdl5cfmux9k6le7rc6r2sw5yrqgq4924gs40d7s",
      "sender_0_eth": "0x6fA03516728d4a483578D63ea124C4Ec4A00e2bB",
      "size": 1338,
      "success": true,
      "timestamp": "2023-03-03T23:09:43-08:00",
      "to": "oasis1qzhg6r0cxtm0mv7w2udcyq8fcfcsz7vvrcktr98e",
      "to_eth": "0x8c064bCf7C0DA3B3b090BAbFE8f3323534D84d68"
    }
  ]
}
```

testing: ran the indexer locally on sapphire and tested that encrypted tx information appeared correctly in the api call (see above)


storage/client/queries/queries.go
```diff
@@ -310,6 +310,12 @@ const (
 			to_preimage.context_version AS to_preimage_context_version,
 			to_preimage.address_data AS to_preimage_data,
 			txs.amount,
+			txs.evm_encrypted_format,
+			txs.evm_encrypted_public_key,
+			txs.evm_encrypted_data_nonce,
+			txs.evm_encrypted_data_data,
+			txs.evm_encrypted_result_nonce,
+			txs.evm_encrypted_result_data,
```
pro-wh, line 318: :steam_locomotive::train::train::train::train::train::train::train::train::train::train::train::train::train::train::train::train::train::train::train::train::train::train::train::train::train::train::train::train::train::train::train::dash:

pro-wh (approved):
lg thanks

api/spec/v1.yaml
```diff
@@ -2089,6 +2089,10 @@ components:
             applicable. The meaning varies based on the transaction method.
             Usually in native denomination, ParaTime units. As a string.
           example: "100000001666393459"
+        encryption_envelope:
```
mitjat, line 2092: I remember now that I saw you were asking on Slack whether people think of "the encryption envelope" as just the encryption metadata (format, key, nonce) or also the encrypted data itself. Sorry I never answered. Unlike this PR, I myself thought of it as metadata only. But that doesn't matter much, what matters is what the internet thinks, and that's ... nothing. It thinks it's a misspelling of "envelope encryption". After more sleuthing, "digital envelope" seems to be well established. In both cases, the expected contents of the envelope (which does seem to include the data, btw :)) are pretty universally understood to be the asymmetrically encrypted DEK (data encryption key; itself symmetric), and the data itself (encrypted with DEK).

Since our envelope is a little different, maybe it's best to avoid a term with such a strongly defined narrow meaning. #407 uses `EVMEncryptedData`. We could go with that ... or `evm_encrypted_contents` (to avoid having `data` inside `[...]_data`?

api/spec/v1.yaml
```diff
@@ -2089,6 +2089,10 @@ components:
...
+          description: |
+            The encrypted transaction envelope.
```
mitjat, line 2094:
```suggestion
            The encrypted data and result of the transaction. Only present for encrypted transactions in confidential EVM runtimes like Sapphire.
```

@pro-wh this is specific to EVM, right? Cipher would have slightly different fields, and we'd want to store/present them differently?

api/spec/v1.yaml
```diff
@@ -2100,6 +2104,34 @@ components:
       description: |
         A runtime transaction.
 
+    RuntimeTransactionEncryptionEnvelope:
+      type: object
+      required: [format, public_key, data_nonce, data, result_nonce, result]
+      properties:
+        format:
+          type: string
```
mitjat, line 2112: Introduce an openapi enum please. Warren introduced it as `common.CallFormat` in #407; you can refer to it with `x-go-type` (see existing examples in v1.yaml).

@pro-wh what do we do for `CallFormatPlain`-encoded txs? (https://github.com/oasisprotocol/oasis-sdk/blob/c823194ed60ad7a922812e6b803d8ff8b9f78c45/client-sdk/go/callformat/callformat.go#L33-L33) Do we not support them, or do we support them because that's just how a nonencrypted tx is always represented, or something else?

api/spec/v1.yaml
```diff
@@ -2100,6 +2104,34 @@ components:
       description: |
         A runtime transaction.
 
+    RuntimeTransactionEncryptionEnvelope:
+      type: object
+      required: [format, public_key, data_nonce, data, result_nonce, result]
```
mitjat, line 2109: Just thinking out loud: When we're returning a list of txs, we might want to keep the response a little more lightweight and not return anything but the encryption format; the body and result are on the hefty side. Some options:
- Ignore this possibility, keep everything as-is. If we ever need to implement these lighter responses, we'll break backwards compatilibiltiy and mark all fields except `format` as optional; or we'll introduce another field `is_encrypted: bool` parallel to the envelop field, and drop the entire envelope field in list queries.
- Mark only `format` as required, so that if we start dropping the other fields in the future, we're not breaking the API.
- Move `format` one level up, so we'd have `encryption_format` (could be then extended to non-evm txs too) and `evm_encrypted_contents`. In list queries, we could omit `evm_encrypted_contents` in the future if we wanted.

We should realistically probably go for #2? Minimal and quick change. A little dirty because the resulting API is sloppy (no guarantees of when the fields will/won't be present), but still fine IMO.

On the topic of optional fields: line 2067 should now probably be `May be null if the transaction was malformed or encrypted.`

mitjat (approved):
Thank you!

api/spec/v1.yaml
```diff
@@ -2100,6 +2104,34 @@ components:
...
+          format: byte
+          description: The base64-encoded public key used to encrypt the transaction.
+        data_nonce:
+          type: string
+          format: byte
+          description: The base64-encoded nonce used to encrypt the transaction data.
+        data:
+          type: string
+          format: byte
+          description: The base64-encoded transaction data.
```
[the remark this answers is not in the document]
Andrew7234: Updated!

api/spec/v1.yaml
Andrew7234: discussed offline. We will use `encryption_envelope` but update the description to note the difference between our encryption envelopes and the traditional [digital envelope](https://nordvpn.com/cybersecurity/glossary/digital-envelope/)

api/spec/v1.yaml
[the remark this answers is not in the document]
pro-wh: 'ciphertext' may also have been appropriate https://github.com/oasisprotocol/deoxysii-js/blob/v0.0.5/deoxysii.js#L680

api/spec/v1.yaml
pro-wh: feels like it serves the same purpose as a digital envelope, it uses an asymmetric scheme to set up a symmetric key and it has the data encrypted with that symmetric key

api/spec/v1.yaml
pro-wh: oh wait, it would be possible to determine that by examining the body

api/spec/v1.yaml
mitjat: :+1: also after what we discussed about Cipher, having `format` a level higher would be potentially confusing if we ever wanted to expand to Cipher, or to other confidential runtimes that encrypt a different subset of the runtime's tx.

api/spec/v1.yaml
```diff
@@ -2089,6 +2094,13 @@ components:
             applicable. The meaning varies based on the transaction method.
             Usually in native denomination, ParaTime units. As a string.
           example: "100000001666393459"
+        encryption_envelope:
+          description: |
+            The data relevant to the encrypted transaction. Only present for encrypted 
+            transactions in confidential EVM runtimes like Sapphire.
+            Note: The term "envelope" in this context refers to the [Oasis-style encryption envelopes](https://github.com/oasisprotocol/oasis-sdk/blob/c36a7ee194abf4ca28fdac0edbefe3843b39bf69/runtime-sdk/src/types/callformat.rs)
+            which differ slightly from [digital envelopes](https://nordvpn.com/cybersecurity/glossary/digital-envelope/).
```
mitjat, line 2102: Nit: Prefer to link to [Wikipedia](https://en.wikipedia.org/wiki/Hybrid_cryptosystem#Envelope_encryption) or an .edu domain (here's a [slightly random one](http://security.nknu.edu.tw/crypto/faq/html/2-2-4.html), probably too high-levle) over NordVPN.

storage/client/client.go
```diff
@@ -1080,6 +1080,13 @@ func (c *StorageClient) RuntimeTransactions(ctx context.Context, p apiTypes.GetR
 		t := RuntimeTransaction{
 			Error: &TxError{},
 		}
+		var encryptionEnvelope RuntimeTransactionEncryptionEnvelope
+		var encryptionEnvelopeFormat *common.CallFormat
+		// var encryptionEnvelopePublicKey []byte
+		// var encryptionEnvelopeDataNonce []byte
+		// var encryptionEnvelopeData []byte
+		// var encryptionEnvelopeResultNonce []byte
+		// var encryptionEnvelopeResult []byte
```
mitjat, line 1089: forgot to remove? same a few lines lower.

mitjat (approved):
Thank you! Re-LGTM now that all discussion threads are resolved, with two last tiny comments.

PR #416, opened 2023-05-11
evm_tokens: Explicitly tracks unsupported tokens in the DB

mitjat:
After this PR, every smart contract we encounter has an entry in the `evm_tokens` table, with the `type` column explicitly set to `0` (signifyin "not a token") when appropriate.

**Background:**
Previously, "tokens" (smart contracts, really) that were found to not be a known/supported token type were tracked in the `evm_token_analysis` table, in that their `last_dowload_round` was nonzero ... but they were not inserted into the `evm_tokens` table, because that table only stored "true" tokens.

This led to a the `evm_token_balances` (a separate analyzer) needing access to the `evm_token_analysis` table.

In https://app.clickup.com/t/861mn3n0e, we're extending `evm_token_balances` to also query native balances. This will require some special-casing of the native token. This PR makes it so that the special-casing stays limited to just the `evm_token_balances` analyzer, and does not spill over into `evm_tokens`.

**Testing**:
Dumped the evm_* tables (`docker exec -it indexer-postgres pg_dump -U rwuser indexer --table chain.evm_token_balance_analysis --table chain.evm_token_analysis --table chain.evm_tokens --table chain.evm_token_balances`) after indexing the first 1000 emerald blocks in Damask, both before and after the PR. There were no differences. To be fair, also no (!) new entries in `evm_token_balances`, I guess because the smart contracts that we heuristically determine _might_ be tokens (= we see a Transfer log coming from them) end up being actual ERC-20 tokens. So for a complete test, I should run with (potentially many) more rounds.

### DIFF
--- a/analyzer/evmtokens/evm_tokens.go
+++ b/analyzer/evmtokens/evm_tokens.go
@@ -101,8 +101,10 @@ func (m Main) processStaleToken(ctx context.Context, batch *storage.QueryBatch, 
 	if err != nil {
 		return fmt.Errorf("token address: %w", err)
 	}
-	//nolint:nestif
-	if staleToken.LastDownloadRound == nil {
+	// These two conditions should be equivalent; only a never-before-downloaded token
+	// can have a nil type, because downloading a token always sets the type.
+	// We check both just in case, because we later dereference the .Type pointer.
+	if staleToken.LastDownloadRound == nil || staleToken.Type == nil {
 		tokenData, err := runtime.EVMDownloadNewToken(
 			ctx,
 			m.logger,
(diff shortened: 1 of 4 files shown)


pro-wh:
probably the easiest way to find an 'unsupported' thing is to find an ERC-721 NFT. it looks like that uses the same Transfer signature. although I'm not sure if in practice they tend to use the same number of indexed event fields

analyzer/evmtokens/evm_tokens.go
```diff
@@ -102,7 +102,7 @@ func (m Main) processStaleToken(ctx context.Context, batch *storage.QueryBatch,
 		return fmt.Errorf("token address: %w", err)
 	}
 	//nolint:nestif
-	if staleToken.LastDownloadRound == nil {
+	if staleToken.LastDownloadRound == nil || staleToken.Type == nil {
```
pro-wh, line 105: how can this be nil now? it's an inner JOIN with evm_tokens having token_type NOT NULL

pro-wh: oh wait I was looking at the wrong query

pro-wh: okay so the .Type can be nil when the evm tokens analysis stale query's left join on the evm_tokens table has no row in the latter. but we currently don't have that situation other than when we've never attempted to scan that token. that is, you've changed the scanner always to create an entry with non-null type 0 (unsupported).

is this in preparation for later having this situation, where staleToken.LastDownloadRound != nil and staleToken.Type == nil?

pro-wh (approved):
yeah

analyzer/evmtokens/evm_tokens.go
mitjat: Ha sorry for sending you on a walk through the code. I was even thinking it might be good to call out this change in the PR :)

I added a comment now:
```
	// These two conditions should be equivalent; only a never-before-downloaded token
	// can have a nil type, because downloading a token always sets the type.
	// We check both just in case, because we later dereference the .Type pointer.
```

In the future (once we track native tokens), this will continue to be the case. Native tokens are special and won't have an entry in the `evm_tokens` table.

mitjat:
> probably the easiest way to find an 'unsupported' thing is to find an ERC-721 NFT. it looks like that uses the same Transfer signature. although I'm not sure if in practice they tend to use the same number of indexed event fields

Yeah I was also hoping for ERC-721; IIRC they do have exactly matching event signatures. But since we're limited to recent heights only ... huh ... should have tried with emerald, not sapphire. Oh well. I can do a little more testing after I add native token tracking.

PR #417, opened 2023-05-15
analyzer/block: configurable batch size

ptrus:
Adds a config parameter for configuring the batch size of the block analyzer.

The default value of 100 is likely unreasonably small for production (especially if reading blocks from cache).

I think a reasonable value for production is such that the batch processing takes 30s-1min in the expected/average case (the batch processing and lock timeouts are both 5 min). Otherwise, the "mega-query" is run unnecessarily often.

### DIFF
--- a/analyzer/block/block.go
+++ b/analyzer/block/block.go
@@ -22,8 +22,8 @@ import (
 const (
 	// Timeout to process a block.
 	processBlockTimeout = 61 * time.Second
-	// Number of blocks to be processed in a batch.
-	blocksBatchSize = 100
+	// Default number of blocks to be processed in a batch.
+	defaultBatchSize = 1_000
 	// Lock expire timeout for blocks (in minutes). Locked blocks not processed within
 	// this time can be picked again.
 	lockExpiryMinutes = 5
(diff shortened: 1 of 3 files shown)


analyzer/block/block.go
```diff
@@ -22,8 +22,8 @@ import (
 const (
 	// Timeout to process a block.
 	processBlockTimeout = 61 * time.Second
-	// Number of blocks to be processed in a batch.
-	blocksBatchSize = 100
+	// Default number of blocks to be processed in a batch.
+	defaultBatchSize = 100
```
mitjat, line 26: Should we change this to 1000 or even 5000? Now that we have the cache in place (and it's prewarmed and we can copy it around with some ops busywork), it's more likely that we'll read from the cache than from the node. It's also fine if the batch is too large (for the batch timeout); it just means we'll lock some blocks too many, but the lock will auto-release soon enough.

config/config.go
```diff
@@ -250,6 +250,14 @@ type BlockBasedAnalyzerConfig struct {
 	// continue processing new blocks until the next breaking
 	// upgrade.
 	To uint64 `koanf:"to"`
+
+	// BatchSize determines the number of blocks the block analyzer
+	// processes per batch.
```
mitjat, line 255:
```suggestion
	// BatchSize determines the maximum number of blocks the block analyzer
	// processes per batch. This is relevant only when the analyzer is still catching
	// up to the latest block.
```

mitjat (approved):
Nice!

analyzer/block/block.go
ptrus: Updated to 1000.

PR #418, opened 2023-05-15
consensus: make malformed transaction not fail

pro-wh:
as they say, the proposer can propose anything. the committee decides if it's valid, but it's still part of the block

### DIFF
--- a/storage/oasis/nodeapi/damask/node.go
+++ b/storage/oasis/nodeapi/damask/node.go
@@ -14,6 +14,8 @@ import (
 	genesis "github.com/oasisprotocol/oasis-core/go/genesis/api"
 	governance "github.com/oasisprotocol/oasis-core/go/governance/api"
 	scheduler "github.com/oasisprotocol/oasis-core/go/scheduler/api"
+
+	"github.com/oasisprotocol/oasis-indexer/log"
 	"github.com/oasisprotocol/oasis-indexer/storage/oasis/nodeapi"
 
 	// data types for Damask gRPC APIs.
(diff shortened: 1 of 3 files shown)


pro-wh:
do we have an example of this happening? we don't have the right testing tools right now to simulate this.

I've only loosely looked through the downstream code, which roughly looks like it'll survive an uninitialized SignedTransaction object, but we should check

storage/oasis/nodeapi/damask/node.go
```diff
@@ -75,7 +77,13 @@ func (c *DamaskConsensusApiLite) GetTransactionsWithResults(ctx context.Context,
 	for i, txBytes := range rsp.Transactions {
 		var tx consensusTx.SignedTransaction
 		if err := cbor.Unmarshal(txBytes, &tx); err != nil {
-			return nil, fmt.Errorf("GetTransactionsWithResults(%d): transaction %d (%x): %w", height, i, txBytes, err)
+			log.NewDefaultLogger("damask-consensus-api-lite").Error("malformed consensus transaction, leaving empty",
```
mitjat, line 80: I guess we donćt have to do the same for Cobalt because it's set in stone, and we know that there were no malformed blocks?

mitjat (approved):
Thank you!

For testing, do you mind just doing a local test where you set `tx consensusTx.SignedTransaction{}` unconditionally (or based on a hardcoded hash, or something), and see if anything goes horribly wrong?

storage/oasis/nodeapi/damask/node.go
pro-wh: let's add it there too

pro-wh:
> tx.Fee.Amount.String()

oop

mitjat:
Thanks for testing. Hrm those two :-1: are indeed sucky, as is the fact that we don't retain the raw tx bytes (which are kinda meaningful in this rare case). But given how rare this case is (0 known occurrences, and for it to happen in the future, one of the nodes with lots of stake would have to go rogue IIUC), let's leave it at this. We might not show the best info on this kind of txs, but at least the indexer won't halt, which I think is the more important aspect.

pro-wh:
let's go with this.

PR #420, opened 2023-05-16
/{runtime}/transaction(s): Expose actual fee used

ptrus:


api/spec/v1.yaml
```diff
@@ -2051,6 +2051,12 @@ components:
           type: integer
           format: uint64
           description: The total gas used by the transaction.
+        charged_fee:
+          type: string
+          description: |
+            The fee that was charged for the transaction execution (total, native denomination,
+            ParaTime base units, as a string).
+            Calculated from gas price (=fee/gas_limit) and gas used.
```
mitjat, line 2059:
```suggestion
            Calculated as `gas_price * gas_used`, where `gas_price = fee / gas_limit`.
```

Let's be explicit since there's so much related terminology here.

mitjat (approved):
Thank you!

PR #421, opened 2023-05-19
EVM parsing functions with ABI

pro-wh:
This PR has just the basics for some code to parse Ethereum ABI format things, including tx input data, tx result data, and log data. Later PRs build upon this


pro-wh:
```
=== RUN   TestEVMParseData
method approve
arg address _spender = 0x250d48C5E78f1E85F7AB07FEC61E93ba703aE668
arg uint256 _value = 4000000000000000000
--- PASS: TestEVMParseData (0.00s)
=== RUN   TestEVMParseEvent
event Approval
arg address indexed _owner = 0x0000000000000000000000000ecf5262e5b864e1612875f8fc18f151315b5e91
arg address indexed _spender = 0x000000000000000000000000250d48c5e78f1e85f7ab07fec61e93ba703ae668
arg uint256 _value = 4000000000000000000
--- PASS: TestEVMParseEvent (0.00s)
PASS
```

api/logic/evm.go
```diff
@@ -0,0 +1,40 @@
...
+	ethCommon "github.com/ethereum/go-ethereum/common"
+)
+
+func EVMParseData(data []byte, contractABI *abi.ABI) (*abi.Method, []interface{}, error) {
+	if len(data) < 4 {
+		return nil, nil, fmt.Errorf("data (%dB) too short to have method ID", len(data))
+	}
+	methodID := data[:4]
+	packedArgs := data[4:]
+	method, err := contractABI.MethodById(methodID)
```
mitjat, line 16: dang are things convenient once we speak the lingua franca of smart contracts ...

api/logic/evm.go
```diff
@@ -0,0 +1,40 @@
+package logic
+
+import (
+	"fmt"
+
+	"github.com/ethereum/go-ethereum/accounts/abi"
+	ethCommon "github.com/ethereum/go-ethereum/common"
+)
+
+func EVMParseData(data []byte, contractABI *abi.ABI) (*abi.Method, []interface{}, error) {
```
mitjat, line 10: It's not clear what Go types will be used in the elements of `[]interface{}`.

Looking at the ethereum-go tests, it's "whatever makes sense" https://github.com/ethereum/go-ethereum/blob/9231770811cda0473a7fa4e2bccc95bf62aae634/accounts/abi/abi_test.go#L810-L813 but it's not clearly documented. Going more into the weeds, it is one of the many possible return values of [toGoType()](https://github.com/ethereum/go-ethereum/blob/ea9e62ca3db5c33aa7438ebf39c189afd53c6bf8/accounts/abi/unpack.go#L222).

Perhaps we don't need to care much, as long as serializes cleanly into JSON (which is the format we use in the DB, and in our API).
Please test at least with large integers; ethereum-go uses `*big.Int`, which is reasonable, but does not support pgx-compatible serialization. We'll be converting everything into JSON manually before throwing it at pgx, so we're probably fine?
Anyway, it would be good to have some EVM parsing tests not just into Go-native types, but into JSON. If you can find/generate example inputs for wacky types (e.g. multi-dimensional array of tuples of (bigint, address)), all the better.

mitjat (commented):
Super neat! IMO this needs some JSON compatibility tests (see thread inline) and `fmt.Printf` replaced with asserts ... and it's ready.

I'm assuming you were planning to integrate these funcs into the analyzer in a separate PR?

pro-wh:
we don't have the ABIs being collected yet, so I don't have a place to integrate this yet

api/logic/evm.go
pro-wh: yeah it's *big.Int and it tries to put it into json as just the digits :scream:

mitjat:
> we don't have the ABIs being collected yet, so I don't have a place to integrate this yet

You can integrate with the existing ERC-20 parsing; it will make it easier to then later generalize from ERC-20 to [any contract whose ABI we have].

pro-wh:
we have the ERC-20 ABI, but we don't have a piece of data telling us which contract to use which ABI for

pro-wh:
```sol
struct O {
    uint16 n;
    string s;
}

interface Varied {
    function test(
        int8 i8,
        uint8 u8,
        int256 i,
        uint256 u,
        bool b,
        // not supported in go-ethereum
        // fixed128x18 f18,
        // ufixed128x18 uf18,
        bytes32 b32,
        address a,
        function (uint16) external returns (uint16) f,
        uint16[2] calldata xy,
        bytes calldata buf,
        string calldata s,
        uint16[] calldata l,
        O calldata o
    ) external;
}
```

`0xa85acffaffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffff0000000000000000000000000000000000000000000000000000000000000001ffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffff000000000000000000000000000000000000000000000000000000000000000100000000000000000000000000000000000000000000000000000000000000010101010101010101010101010101010101010101010101010101010101010101000000000000000000000000010101010101010101010101010101010101010101010101010101010101010101010101010101010202020200000000000000000000000000000000000000000000000000000000000000000000000000000001000000000000000000000000000000000000000000000000000000000000000100000000000000000000000000000000000000000000000000000000000001c00000000000000000000000000000000000000000000000000000000000000200000000000000000000000000000000000000000000000000000000000000024000000000000000000000000000000000000000000000000000000000000002800000000000000000000000000000000000000000000000000000000000000001010000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000016100000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000100000000000000000000000000000000000000000000000000000000000000010000000000000000000000000000000000000000000000000000000000000001000000000000000000000000000000000000000000000000000000000000004000000000000000000000000000000000000000000000000000000000000000016100000000000000000000000000000000000000000000000000000000000000`

```json
[-1,1,-1,1,true,[1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1],"0x0101010101010101010101010101010101010101",[1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,2,2,2,2],[1,1],"AQ==","a",[1],{"n":1,"s":"a"}]
```
eth (solidity) -> go (go-ethereum) -> JSON

pro-wh:
current complaints:

- large integers are in json as numbers, we want strings
- function is array of numbers, we probably want base64 bytes, or maybe hex
- bytes32 (and similar sizes I assume) is array of numbers, we probably want base64 bytes, or maybe hex

pro-wh:
```json
[-1,1,"-1","1",true,"AQEBAQEBAQEBAQEBAQEBAQEBAQEBAQEBAQEBAQEBAQE=","0x0101010101010101010101010101010101010101","AQEBAQEBAQEBAQEBAQEBAQEBAQECAgIC",[1,1],"AQ==","a",[1],{"n":1,"s":"a"}]
```

custom converting

api/logic/evm_test.go
```diff
@@ -0,0 +1,93 @@
...
+	require.Equal(t, Varied.Methods["test"], *method)
+	jsonExpected := []string{
+		"-1",     // int8
+		"1",      // uint8
+		"\"-1\"", // int256
+		"\"1\"",  // uint256
+		"true",   // bool
+		"\"AQEBAQEBAQEBAQEBAQEBAQEBAQEBAQEBAQEBAQEBAQE=\"", // bytes32
+		"\"0x0101010101010101010101010101010101010101\"",   // address
+		"\"AQEBAQEBAQEBAQEBAQEBAQEBAQECAgIC\"",             // function (uint16) external returns (uint16)
```
mitjat, line 45: Not important (because fns won't get passed/emitted much, I reckon), but out of curiosity, is this the fn ... body? address of some kind? signature of a hash that includes the name?
My guess is the latter, in which case it would be nicer to display the fn name. But if it involves indpendent lookups into the ABI, let's not go down that rabbithole and just format this in hex.

api/logic/evm_test.go
```diff
@@ -0,0 +1,93 @@
...
+	require.NoError(t, err)
+	method, args, err := EVMParseData(data, Varied)
+	require.Equal(t, Varied.Methods["test"], *method)
+	jsonExpected := []string{
+		"-1",     // int8
+		"1",      // uint8
+		"\"-1\"", // int256
+		"\"1\"",  // uint256
+		"true",   // bool
+		"\"AQEBAQEBAQEBAQEBAQEBAQEBAQEBAQEBAQEBAQEBAQE=\"", // bytes32
```
mitjat, line 43: You mentioned somewhere else: "possibly weird philosophically: uint8[] and uint8[n] I believe will also be base64"

I think the philosophy as we discussed it is pretty straightforward: This JSON formatting is primarily intended to look good to a human reader, and we don't really expect it to be re-parsed often.
In that vein, I'd say that we should render any array/sequence of bytes (types `bytes`, `bytes1..bytes32`, `uint8[]` (maybe this is the same as `bytes`?), `uint8[N]`) as a hex string. I don't think eth devs ever expect to see their byte blobs rendered as base64.

mitjat (approved):
Thank you! Largely LGTM for the current changes.

Also, to revisit the deeper-integration conversation (stupid github for preventing PR-level threads):

>>>> I'm assuming you were planning to integrate these funcs into the analyzer in a separate PR?

>>> we don't have the ABIs being collected yet, so I don't have a place to integrate this yet

>> You can integrate with the existing ERC-20 parsing; it will make it easier to then later generalize from ERC-20 to [any contract whose ABI we have].

> we have the ERC-20 ABI, but we don't have a piece of data telling us which contract to use which ABI for

We do have `token_type` in `chain.evm_tokens`. Though doing that properly will require a new analyzer that will parse events and txs post-hoc, as ABIs become available, right? Created #872 to track this.

What you can do short-term though is replace the slicing [here](https://github.com/oasisprotocol/oasis-indexer/blob/f33b4a9558b4d5736af5c137592ed951cf1c0fff/analyzer/runtime/visitors.go#L176-L176). We'll still only heuristically decide the event is an ERC20 Transfer, but then we'll parse it with the ABI. Kinda silly, but might be a worthy early use case for the new code, and a way of testing it out gradually.

pro-wh:
note for later: see BoundContract.UnpackLog for how to parse the topics. filter .Inputs by their .Indexed, then use abi.ParseTopics

api/logic/evm_test.go
pro-wh: it's contract addr and that 4-byte solidity function selector

api/logic/evm_test.go
pro-wh: displaying byte string as hex is for the frontend :upside_down_face:. this is the separate problem of getting a little document over some format that fits in an HTTP response, and we've selected JSON for that. JSON doesn't support byte strings which this document contains, and for whatever reason, the very conventional Go and/or oasis-core thing to do is to use base64 for that :skull:

pro-wh: actually do we already use hex in most of our byte-strings-in-json for nexus?

api/logic/evm.go
pro-wh: alright tests are added

api/logic/evm_test.go
mitjat: > displaying byte string as hex is for the frontend

But if frontend wanted to do that, it would need to traverse the whole JSON object, and either a) understand and follow the ABI, replicating much of the work you're doing here (in which case, why even bother here?), or b) heuristically decide which values are likely easiest undertood when displayed as hex but were relayed in base64 (ugly). It's not a lot of work to re-encode here, right? A few extra cases in `evmPreMarshal()`?

The main reason we're constructing this JSON is so frontend (any frontend, not just ours), or humans in general, can make better sense of the values. So let's take it all the way?

This is perhaps inconsistent with the base64 examples I posted above ... But I'd argue that those are more about offering a uniformly-, predictably-encoded machine-readable representation of the events, rather than producing eye candy. Also, speaking of consistency, I see this PR already hex-encodes `address` types (courtesy of `go-ethereum`).

mitjat (approved):
Thank you!
I'll leave it to you to decide how to show `bytes` and `bytes1..bytes32` to the end users. I'd go for hex, but I can see how it's a matter of opinion. I'm quite convinced the final format is mostly the domain of this code though, not the frontend.

api/logic/evm_test.go
pro-wh: message from 3 weeks in the future: lmao fine

pro-wh:
oh update: uint8[] and uint8[n] don't become hex/base64. they're proper arrays of numbers

pro-wh:
ok bytes and bytes[n] are now like 0x1234

PR #422, opened 2023-05-19
evm_token_balances: Refresh balance of native tokens

mitjat:
Closes https://app.clickup.com/t/861mn3n0e

I don't like how a lot of logic for native tokens is very similar to EVM tokens, so it feels like we're duplicating code, but it's just different enough that when I try to reuse snippets (or here, the whole analyzer), it ends up having a lot of parallel codepaths internally, one for EVM tokens and one for natives. Oh well. This PR is one possible balance to strike.

**Testing: **
Indexed the first 10k blocks of emerald, then ran `evm_tokens` and `evm_token_balances` analyzers. Did this both pre- and post- PR and diffed the DBs. No changes except for the new entries in `evm_token_balances_analysis` table for the native tokens.

### DIFF
--- a/analyzer/evmtokenbalances/evm_token_balances.go
+++ b/analyzer/evmtokenbalances/evm_token_balances.go
@@ -19,6 +19,7 @@ import (
 	"github.com/oasisprotocol/oasis-indexer/storage"
 	"github.com/oasisprotocol/oasis-indexer/storage/client"
 	source "github.com/oasisprotocol/oasis-indexer/storage/oasis"
+	"github.com/oasisprotocol/oasis-indexer/storage/oasis/nodeapi"
 )
 
 // Imagine a timeline starting from a `balanceOf` output `v0` followed by
@@ -121,7 +122,10 @@ type StaleTokenBalance struct {
 
 func (m Main) getStaleTokenBalances(ctx context.Context, limit int) ([]*StaleTokenBalance, error) {
 	var staleTokenBalances []*StaleTokenBalance
-	rows, err := m.target.Query(ctx, queries.RuntimeEVMTokenBalanceAnalysisStale, m.runtime, limit)
+	rows, err := m.target.Query(ctx, queries.RuntimeEVMTokenBalanceAnalysisStale,
+		m.runtime,
+		nativeTokenSymbol(&m.runtimeMetadata),
+		limit)
 	if err != nil {
 		return nil, fmt.Errorf("querying stale token balances: %w", err)
 	}
@@ -151,17 +155,49 @@ func (m Main) getStaleTokenBalances(ctx context.Context, limit int) ([]*StaleTok
 }
 
 func (m Main) processStaleTokenBalance(ctx context.Context, batch *storage.QueryBatch, staleTokenBalance *StaleTokenBalance) error {
-	m.logger.Info("downloading", "stale_token_balance", staleTokenBalance)
-	// todo: assert that token addr and account addr contexts are secp256k1
-	tokenEthAddr, err := client.EVMEthAddrFromPreimage(staleTokenBalance.TokenAddrContextIdentifier, staleTokenBalance.TokenAddrContextVersion, staleTokenBalance.TokenAddrData)
-	if err != nil {
-		return fmt.Errorf("token address: %w", err)
-	}
 	accountEthAddr, err := client.EVMEthAddrFromPreimage(staleTokenBalance.AccountAddrContextIdentifier, staleTokenBalance.AccountAddrContextVersion, staleTokenBalance.AccountAddrData)
 	if err != nil {
 		return fmt.Errorf("account address: %w", err)
 	}
-	if staleTokenBalance.Type != nil {
+	switch *staleTokenBalance.Type {
+	case evm.EVMTokenTypeUnsupported:
+		// Do nothing; we'll just mark this token as processed so we remove it from the queue.
+	case evm.EVMTokenTypeNative:
+		// Query native balance.
+		addr := nodeapi.Address{}
+		if err := addr.UnmarshalText([]byte(staleTokenBalance.AccountAddr)); err != nil {
+			m.logger.Error("invalid account address bech32 '%s': %w", staleTokenBalance.AccountAddr, err)
+			break
+		}
+		balance, err := m.source.GetNativeBalance(ctx, staleTokenBalance.DownloadRound, addr)
+		if err != nil || balance == nil {
+			return fmt.Errorf("getting native runtime balance: %w", err)
+		}
+		if balance.Cmp(staleTokenBalance.Balance) != 0 {
+			correction := (&common.BigInt{}).Sub(&balance.Int, staleTokenBalance.Balance)
+			// Native token balance changes do not produce events, so our dead reckoning
+			// is expected to often be off. Log discrepancies at Info level only.
+			m.logger.Info("correcting reckoned native runtime balance to downloaded balance",
+				"account_addr", staleTokenBalance.AccountAddr,
+				"download_round", staleTokenBalance.DownloadRound,
+				"reckoned_balance", staleTokenBalance.Balance.String(),
+				"downloaded_balance", balance.String(),
+			)
+			batch.Queue(queries.RuntimeNativeBalanceUpdate,
+				m.runtime,
+				staleTokenBalance.AccountAddr,
+				nativeTokenSymbol(&m.runtimeMetadata),
+				correction,
+			)
+		} else {
+			m.logger.Debug("native balance: nothing to correct", "account_addr", staleTokenBalance.AccountAddr, "balance", balance)
+		}
+	default:
+		// All other ERC-X tokens. Query the token contract.
+		tokenEthAddr, err := client.EVMEthAddrFromPreimage(staleTokenBalance.TokenAddrContextIdentifier, staleTokenBalance.TokenAddrContextVersion, staleTokenBalance.TokenAddrData)
+		if err != nil {
+			return fmt.Errorf("token address: %w", err)
+		}
 		balanceData, err := evm.EVMDownloadTokenBalance(
 			ctx,
 			m.logger,
@@ -178,15 +214,12 @@ func (m Main) processStaleTokenBalance(ctx context.Context, batch *storage.Query
 			if balanceData.Balance.Cmp(staleTokenBalance.Balance) != 0 {
 				correction := &big.Int{}
 				correction.Sub(balanceData.Balance, staleTokenBalance.Balance)
-				// Note: This will happen because we currently don't scan
-				// before the beginning of the Dasmask upgrade, so the
-				// reckoning will be wrong about any balances from before
-				// then. It can also happen when contracts misbehave.
+				// Can happen when contracts misbehave, or if we haven't indexed all the runtime blocks from the very first one on.
 				m.logger.Warn("correcting reckoned balance of token to downloaded balance",
 					"token_addr", staleTokenBalance.TokenAddr,
 					"account_addr", staleTokenBalance.AccountAddr,
 					"download_round", staleTokenBalance.DownloadRound,
-					"reckoned_balance", staleTokenBalance.Balance,
+					"reckoned_balance", staleTokenBalance.Balance.String(),
 					"downloaded_balance", balanceData,
 					"correction", correction,
 				)
@@ -197,6 +230,8 @@ func (m Main) processStaleTokenBalance(ctx context.Context, batch *storage.Query
 					correction.String(),
 				)
 			}
+		} else {
+			m.logger.Debug("EVM token balance: nothing to correct", "token_addr", staleTokenBalance.TokenAddr, "account_addr", staleTokenBalance.AccountAddr, "balance", balanceData.Balance)
 		}
 	}
 	batch.Queue(queries.RuntimeEVMTokenBalanceAnalysisUpdate,
--- a/analyzer/evmtokenbalances/evm_token_balances.go
+++ b/analyzer/evmtokenbalances/evm_token_balances.go
@@ -171,7 +171,7 @@ func (m Main) processStaleTokenBalance(ctx context.Context, batch *storage.Query
 			break
 		}
 		balance, err := m.source.GetNativeBalance(ctx, staleTokenBalance.DownloadRound, addr)
-		if err != nil || balance == nil {
+		if err != nil {
 			return fmt.Errorf("getting native runtime balance: %w", err)
 		}
 		if balance.Cmp(staleTokenBalance.Balance) != 0 {
--- a/analyzer/evmtokenbalances/evm_token_balances.go
+++ b/analyzer/evmtokenbalances/evm_token_balances.go
@@ -109,7 +109,7 @@ func NewMain(
 type StaleTokenBalance struct {
 	TokenAddr                    string
 	AccountAddr                  string
-	Type                         *evm.EVMTokenType
+	Type                         evm.EVMTokenType
 	Balance                      *big.Int
 	TokenAddrContextIdentifier   string
 	TokenAddrContextVersion      int
@@ -159,7 +159,7 @@ func (m Main) processStaleTokenBalance(ctx context.Context, batch *storage.Query
 	if err != nil {
 		return fmt.Errorf("account address: %w", err)
 	}
-	switch *staleTokenBalance.Type {
+	switch staleTokenBalance.Type {
 	case evm.EVMTokenTypeUnsupported:
 		// Do nothing; we'll just mark this token as processed so we remove it from the queue.
 	case evm.EVMTokenTypeNative:
@@ -205,7 +205,7 @@ func (m Main) processStaleTokenBalance(ctx context.Context, batch *storage.Query
 			staleTokenBalance.DownloadRound,
 			tokenEthAddr,
 			accountEthAddr,
-			*staleTokenBalance.Type,
+			staleTokenBalance.Type,
 		)
 		if err != nil {
 			return fmt.Errorf("downloading token balance %s %s: %w", staleTokenBalance.TokenAddr, staleTokenBalance.AccountAddr, err)
--- a/analyzer/evmtokenbalances/evm_token_balances.go
+++ b/analyzer/evmtokenbalances/evm_token_balances.go
@@ -167,6 +167,7 @@ func (m Main) processStaleTokenBalance(ctx context.Context, batch *storage.Query
 		addr := nodeapi.Address{}
 		if err := addr.UnmarshalText([]byte(staleTokenBalance.AccountAddr)); err != nil {
 			m.logger.Error("invalid account address bech32 '%s': %w", staleTokenBalance.AccountAddr, err)
+			// Do not return; mark this token as processed later on in the func, so we remove it from the DB queue.
 			break
 		}
 		balance, err := m.source.GetNativeBalance(ctx, staleTokenBalance.DownloadRound, addr)
--- a/analyzer/evmtokenbalances/evm_token_balances.go
+++ b/analyzer/evmtokenbalances/evm_token_balances.go
@@ -10,7 +10,7 @@ import (
 
 	"github.com/oasisprotocol/oasis-indexer/analyzer"
 	"github.com/oasisprotocol/oasis-indexer/analyzer/queries"
-	"github.com/oasisprotocol/oasis-indexer/analyzer/runtime"
+	"github.com/oasisprotocol/oasis-indexer/analyzer/runtime/evm"
 	"github.com/oasisprotocol/oasis-indexer/analyzer/util"
 	"github.com/oasisprotocol/oasis-indexer/common"
 	"github.com/oasisprotocol/oasis-indexer/log"
@@ -103,7 +103,7 @@ func NewMain(
 type StaleTokenBalance struct {
 	TokenAddr                    string
 	AccountAddr                  string
-	Type                         *runtime.EVMTokenType
+	Type                         *evm.EVMTokenType
 	Balance                      *big.Int
 	TokenAddrContextIdentifier   string
 	TokenAddrContextVersion      int
@@ -157,7 +157,7 @@ func (m Main) processStaleTokenBalance(ctx context.Context, batch *storage.Query
 		return fmt.Errorf("account address: %w", err)
 	}
 	if staleTokenBalance.Type != nil {
-		balanceData, err := runtime.EVMDownloadTokenBalance(
+		balanceData, err := evm.EVMDownloadTokenBalance(
 			ctx,
 			m.logger,
 			m.source,
--- a/analyzer/evmtokenbalances/evm_token_balances.go
+++ b/analyzer/evmtokenbalances/evm_token_balances.go
@@ -82,7 +82,7 @@ const (
 
 type Main struct {
 	runtime         common.Runtime
-	runtimeMetadata sdkConfig.ParaTime
+	runtimeMetadata *sdkConfig.ParaTime
 	source          storage.RuntimeSourceStorage
 	target          storage.TargetStorage
 	logger          *log.Logger
@@ -92,7 +92,7 @@ var _ analyzer.Analyzer = (*Main)(nil)
 
 func NewMain(
 	runtime common.Runtime,
-	runtimeMetadata sdkConfig.ParaTime,
+	runtimeMetadata *sdkConfig.ParaTime,
 	sourceClient *source.RuntimeClient,
 	target storage.TargetStorage,
 	logger *log.Logger,
@@ -124,7 +124,7 @@ func (m Main) getStaleTokenBalances(ctx context.Context, limit int) ([]*StaleTok
 	var staleTokenBalances []*StaleTokenBalance
 	rows, err := m.target.Query(ctx, queries.RuntimeEVMTokenBalanceAnalysisStale,
 		m.runtime,
-		nativeTokenSymbol(&m.runtimeMetadata),
+		nativeTokenSymbol(m.runtimeMetadata),
 		limit)
 	if err != nil {
 		return nil, fmt.Errorf("querying stale token balances: %w", err)
@@ -186,7 +186,7 @@ func (m Main) processStaleTokenBalance(ctx context.Context, batch *storage.Query
 			batch.Queue(queries.RuntimeNativeBalanceUpdate,
 				m.runtime,
 				staleTokenBalance.AccountAddr,
-				nativeTokenSymbol(&m.runtimeMetadata),
+				nativeTokenSymbol(m.runtimeMetadata),
 				correction,
 			)
 		} else {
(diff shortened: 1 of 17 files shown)


cmd/analyzer/analyzer.go
```diff
@@ -294,21 +294,23 @@ func NewService(cfg *config.AnalysisConfig) (*Service, error) {
...
 			sourceClient, err1 := sources.Runtime(ctx, common.RuntimeEmerald)
 			if err1 != nil {
 				return nil, err1
 			}
-			return evmtokenbalances.NewMain(common.RuntimeEmerald, sourceClient, dbClient, logger)
+			return evmtokenbalances.NewMain(common.RuntimeEmerald, *runtimeMetadata, sourceClient, dbClient, logger)
 		})
 	}
 	if cfg.Analyzers.SapphireEvmTokenBalances != nil {
+		runtimeMetadata := cfg.Source.SDKParaTime(common.RuntimeEmerald)
```
ptrus, line 307: `common.RuntimeSapphire` likely

go.sum
```diff
@@ -880,6 +880,8 @@ github.com/kevinburke/ssh_config v0.0.0-20190725054713-01f96b0aa0cd/go.mod h1:CT
 github.com/kisielk/errcheck v1.1.0/go.mod h1:EZBBE59ingxPouuu3KfxchcWSUPOHkagtvWXihfKN4Q=
 github.com/kisielk/errcheck v1.2.0/go.mod h1:/BMXB+zMLi60iA8Vv6Ksmxu/1UDYcXs4uQLJ+jE2L00=
 github.com/kisielk/errcheck v1.5.0/go.mod h1:pFxgyoBC7bSaBwPgfKdkLd5X25qrDl4LWUI2bnpBCr8=
+github.com/kisielk/godepgraph v0.0.0-20221115040737-2d0831789458 h1:QHffewzTTAhdV2FQ0X6hxp/tHpRYXE3STwzaD94N1qw=
```
ptrus, line 883: Where did this come from?

analyzer/runtime/extract.go
```diff
@@ -404,7 +408,25 @@ func ExtractRound(blockHeader nodeapi.RuntimeBlockHeader, txrs []nodeapi.Runtime
...
+					// Dead-reckon native token balances.
+					// Native token transfers do not generate events. Theoretically, any call can change the balance of any account,
+					// and we do not have a good way of tracking them; we just query them with the evm_token_balances analyzer.
+					// Heuristically, a call is most likely to change the balances of the sender and the receiver, so we create
+					// a (quite possibly incorrect) dead-reckoned change of 0 for those accounts, which will cause the evm_token_balances analyzer
+					// to re-query their real balance.
+					reckonedAmount := amount.ToBigInt() // Calls with an empty body represent a transfer of the native token.
+					if len(body.Data) != 0 || len(blockTransactionData.SignerData) > 1 {
+						// Calls with a non-empty body have no standard impact on native balance. Better to dead-reckon a 0 change (and keep stale balances)
+						// than to reckon a wrong change (and have a "random" incorrect balance until it is re-queried).
```
pro-wh, line 418: might be okay to reckon it fully into the contract for now. the evm does that, and the transaction can send some back, send some elsewhere, etc.

oh possibly also check that the transaction succeeded

analyzer/runtime/extract.go
```diff
@@ -20,9 +20,10 @@ import (
 	"github.com/oasisprotocol/oasis-sdk/client-sdk/go/modules/accounts"
 	"github.com/oasisprotocol/oasis-sdk/client-sdk/go/modules/consensusaccounts"
 	"github.com/oasisprotocol/oasis-sdk/client-sdk/go/modules/core"
-	"github.com/oasisprotocol/oasis-sdk/client-sdk/go/modules/evm"
+	evm_sdk "github.com/oasisprotocol/oasis-sdk/client-sdk/go/modules/evm"
```
pro-wh, line 23: I've been using "sdk" as a prefix, e.g. in sdkTypes. would sdkEVM be appropriate here?

storage/oasis/nodeapi/universal_runtime.go
```diff
@@ -136,3 +140,16 @@ func (rc *UniversalRuntimeApiLite) GetEventsRaw(ctx context.Context, round uint6
 func (rc *UniversalRuntimeApiLite) EVMSimulateCall(ctx context.Context, round uint64, gasPrice []byte, gasLimit uint64, caller []byte, address []byte, value []byte, data []byte) ([]byte, error) {
 	return evm.NewV1(rc.sdkClient).SimulateCall(ctx, round, gasPrice, gasLimit, caller, address, value, data)
 }
+
+func (rc *UniversalRuntimeApiLite) GetNativeBalance(ctx context.Context, round uint64, addr Address) (*common.BigInt, error) {
+	balances, err := accounts.NewV1(rc.sdkClient).Balances(ctx, round, sdkTypes.Address(addr))
+	if err != nil {
+		return nil, err
+	}
+	nativeBalance, ok := balances.Balances[sdkTypes.NativeDenomination]
```
pro-wh, line 148: oh right, because we have support for multiple denominations in the accounts module :neutral_face: 

need to ask some questions about this

1. architecturally what would it be like to support other denominations? more reserved addresses and token type integers?
2. will the indexer team ever put pressure on paratime developers including OPF never to use multiple denominations?
3. does the indexer understand that the native denomination of all paratimes and consensus staking token are the same (modulo decimals changes)? this is incidentally true of the networks and paratimes we currently support, but paratimes can be configured otherwise

analyzer/evmtokenbalances/evm_token_balances.go
```diff
@@ -78,25 +80,28 @@ const (
...
+	source          storage.RuntimeSourceStorage
+	target          storage.TargetStorage
+	logger          *log.Logger
 }
 
 var _ analyzer.Analyzer = (*Main)(nil)
 
 func NewMain(
 	runtime common.Runtime,
+	runtimeMetadata sdkConfig.ParaTime,
```
pro-wh, line 95: most of the codebase uses *sdkConfig.ParaTime. do we specifically want a copy here? or if copying is always better, should we change many things over? there are lots of places where there's a structure that's not especially large or mutable and where we neither want separate copies nor a shared copy.

analyzer/evmtokenbalances/evm_token_balances.go
```diff
@@ -151,17 +155,49 @@ func (m Main) getStaleTokenBalances(ctx context.Context, limit int) ([]*StaleTok
...
-	// todo: assert that token addr and account addr contexts are secp256k1
-	tokenEthAddr, err := client.EVMEthAddrFromPreimage(staleTokenBalance.TokenAddrContextIdentifier, staleTokenBalance.TokenAddrContextVersion, staleTokenBalance.TokenAddrData)
-	if err != nil {
-		return fmt.Errorf("token address: %w", err)
-	}
 	accountEthAddr, err := client.EVMEthAddrFromPreimage(staleTokenBalance.AccountAddrContextIdentifier, staleTokenBalance.AccountAddrContextVersion, staleTokenBalance.AccountAddrData)
 	if err != nil {
 		return fmt.Errorf("account address: %w", err)
 	}
-	if staleTokenBalance.Type != nil {
```
pro-wh, line 159: if this is no longer possible, could we change staleTokenBalance.Type not to be pointer

analyzer/runtime/extract.go
```diff
@@ -453,7 +453,7 @@ func ExtractRound(blockHeader nodeapi.RuntimeBlockHeader, txrs []nodeapi.Runtime
 			// Early versions of runtimes didn't emit a GasUsed event.
 			if (txr.Result.IsUnknown() || txr.Result.IsSuccess()) && tx != nil {
 				// Treat as if it used all the gas.
-				logger.Info("tx didn't emit a core.GasUsed event, assuming it used max allowed gas", "tx_hash", txr.Tx.Hash(), "assumed_gas_used", tx.AuthInfo.Fee.Gas)
+				logger.Debug("tx didn't emit a core.GasUsed event, assuming it used max allowed gas", "tx_hash", txr.Tx.Hash(), "assumed_gas_used", tx.AuthInfo.Fee.Gas)
```
pro-wh, line 456: :hear_no_evil:

analyzer/evmtokenbalances/evm_token_balances.go
```diff
@@ -286,3 +291,7 @@ func (m Main) Start(ctx context.Context) {
 func (m Main) Name() string {
 	return EvmTokenBalancesAnalyzerPrefix + string(m.runtime)
 }
+
+func nativeTokenSymbol(sdkPT *sdkConfig.ParaTime) string {
+	return sdkPT.Denominations[sdkConfig.NativeDenominationKey].Symbol
```
pro-wh, line 331: did we really not have this anywhere?

pro-wh:
okay that is a lot of things in the balance analyzer that are rewritten. let me try to summarize what's in common

- shared db table, evm_token_balance_analysis
- the big loop around the outside of the balance analyzer
- the LIMIT clause on the stale tokens query

analyzer/evmtokenbalances/evm_token_balances.go
```diff
@@ -178,15 +214,12 @@ func (m Main) processStaleTokenBalance(ctx context.Context, batch *storage.Query
...
-				// before the beginning of the Dasmask upgrade, so the
-				// reckoning will be wrong about any balances from before
-				// then. It can also happen when contracts misbehave.
+				// Can happen when contracts misbehave, or if we haven't indexed all the runtime blocks from the very first one on.
 				m.logger.Warn("correcting reckoned balance of token to downloaded balance",
 					"token_addr", staleTokenBalance.TokenAddr,
 					"account_addr", staleTokenBalance.AccountAddr,
 					"download_round", staleTokenBalance.DownloadRound,
-					"reckoned_balance", staleTokenBalance.Balance,
+					"reckoned_balance", staleTokenBalance.Balance.String(),
```
pro-wh, line 222: oh shoot did this break?

cmd/analyzer/analyzer.go
mitjat: Thank you! Fixed.

go.sum
mitjat: A visualization tool that I meant to install globally. I removed it but forgot to push the change, sorry. Fixed now.

analyzer/runtime/extract.go
mitjat: yeah that's more in line with the other imports, thanks, renamed

storage/oasis/nodeapi/universal_runtime.go
```diff
@@ -136,3 +140,16 @@ func (rc *UniversalRuntimeApiLite) GetEventsRaw(ctx context.Context, round uint6
 func (rc *UniversalRuntimeApiLite) EVMSimulateCall(ctx context.Context, round uint64, gasPrice []byte, gasLimit uint64, caller []byte, address []byte, value []byte, data []byte) ([]byte, error) {
 	return evm.NewV1(rc.sdkClient).SimulateCall(ctx, round, gasPrice, gasLimit, caller, address, value, data)
 }
+
+func (rc *UniversalRuntimeApiLite) GetNativeBalance(ctx context.Context, round uint64, addr Address) (*common.BigInt, error) {
+	balances, err := accounts.NewV1(rc.sdkClient).Balances(ctx, round, sdkTypes.Address(addr))
```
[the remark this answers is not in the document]
mitjat: nice, switched both

analyzer/runtime/extract.go
pro-wh: ya ok thanks

storage/oasis/nodeapi/universal_runtime.go
mitjat: I'm working with the assumption that no EVM (!) runtime will ever actually introduce multiple runtime-levle tokens. Or that if they do, then the non-native runtime-level tokens (because even among the denominations specified by the `accounts` module, one is first among equals!) will be trackable via runtime events and won't require EVM querying.

If this assumption is proven wrong, these are good questions and I think we'd have to rethink/redesign things a bit at that point.
Re 1, yeah it's an option as a hack to support one or two known hardcoded denominations, but it gets hairy fast.
Re 2, I don't think that's our place, we have to adapt to the chain, not the other way around. (We did discuss the node adding a handy API extension here or there at our behest, but that's different.)
Re 3, I don't think we make that assumption anywhere.

analyzer/evmtokenbalances/evm_token_balances.go
mitjat: Changed to pointers for consistency. I tend to prefer copies over pointers, because pointers imply mutability or nullability, or (unfortunately) the need to hand-hold the compiler and prevent expensive copies. I'm hoping that the compiler can get it right most of the time and avoid an actual copy, but I don't know how true that is.

analyzer/runtime/extract.go
mitjat: It's expected prior to a certain height (was thinking of hardcoding it, but it depends on the runtime and chain), and nearly impossible after the height/version that introduced core.GasUsed events. So it's not providing much value. Should be `Debug` in the first case and `Error` or `Warn` in the latter, really.

analyzer/evmtokenbalances/evm_token_balances.go
mitjat: We have it once in test code, and once somewhere else but it's a method of a struct that contains `sdkConfig.ParaTime` :)

analyzer/evmtokenbalances/evm_token_balances.go
mitjat: IIRC I was seeing scientific notation in my logs

analyzer/evmtokenbalances/evm_token_balances.go
```diff
@@ -146,25 +155,57 @@ func (m Main) getStaleTokenBalances(ctx context.Context, limit int) ([]*StaleTok
...
-		balanceData, err := runtime.EVMDownloadTokenBalance(
+	switch staleTokenBalance.Type {
+	case evm.EVMTokenTypeUnsupported:
+		// Do nothing; we'll just mark this token as processed so we remove it from the queue.
+	case evm.EVMTokenTypeNative:
+		// Query native balance.
+		addr := nodeapi.Address{}
+		if err := addr.UnmarshalText([]byte(staleTokenBalance.AccountAddr)); err != nil {
+			m.logger.Error("invalid account address bech32 '%s': %w", staleTokenBalance.AccountAddr, err)
+			break
```
ptrus, line 170: Should this be a `return err`?

analyzer/evmtokenbalances/evm_token_balances.go
```diff
@@ -146,25 +155,57 @@ func (m Main) getStaleTokenBalances(ctx context.Context, limit int) ([]*StaleTok
...
+		}
+		balance, err := m.source.GetNativeBalance(ctx, staleTokenBalance.DownloadRound, addr)
+		if err != nil || balance == nil {
```
ptrus, line 173: Is the additional balance nil check redundant? I'm not used to seeing such a pattern in go code, unless the function returning nil is an actual expected case.

analyzer/queries/queries.go
```diff
@@ -437,35 +443,76 @@ const (
...
+        analysis.account_address,
+        evm_tokens.token_type,
+        evm_token_balances.balance,
+        token_preimage.context_identifier,
+        token_preimage.context_version,
+        token_preimage.address_data,
+        account_preimage.context_identifier,
+        account_preimage.context_version,
+        account_preimage.address_data,
+        (SELECT MAX(height) FROM chain.processed_blocks WHERE analyzer = analysis.runtime::text AND processed_time IS NOT NULL) AS download_round
```
ptrus, line 460: Small nit: since this query now already uses CTE it might make it a bit nicer if the download round query is moved out into a separate CTE.

analyzer/runtime/extract.go
```diff
@@ -367,7 +368,7 @@ func ExtractRound(blockHeader nodeapi.RuntimeBlockHeader, txrs []nodeapi.Runtime
 					}
 					return nil
 				},
-				EVMCreate: func(body *evm.Create, ok *[]byte) error {
+				EVMCreate: func(body *sdkEVM.Create, ok *[]byte) error {
```
ptrus, line 371: I believe that also in a `EVMCreate` transaction, the caller can send tokens into the contracts, so likely the caller and the resulting contract addresses should be dead-recokned here as well?

Side-note: actually also all other calls can change the runtime native balance as well right? (withdraw/deposit/transfer). 
I don't fully understand/know about all the account balance tracking code, so this might already be handled elsewhere.

storage/oasis/nodeapi/universal_runtime.go
pro-wh: thanks for the analysis. you're right, the evm does only touch one accounts module denomination at a time. https://github.com/oasisprotocol/emerald-paratime/blob/v10.0.0/src/lib.rs#L69 that's configured here, which for the current paratimes we support, it's the native token.

analyzer/runtime/extract.go
pro-wh: maybe someday we can have a field saying what upgrade each block is from

analyzer/evmtokenbalances/evm_token_balances.go
pro-wh: > it's a method of a struct that contains `sdkConfig.ParaTime` :)

oh no that was me! :see_no_evil:

analyzer/evmtokenbalances/evm_token_balances.go
pro-wh: oh it actually went in as a number? haha then yeah stringify

analyzer/evmtokenbalances/evm_token_balances.go
mitjat: In practice it probably doesn't matter because we're very unlikely to hit this case.

In theory, we want this to behave much like `case evm.EVMTokenTypeUnsupported:` -- this is a "weird" token-account pair that we aren't ever able to process, but we still want to mark them as "processed" (rather, "looked at") at this height so that they get removed from the queue (implemented in the DB as an index) so it doesn't grow without bounds. It's a somewhat brittle setup :/ and incurs some busywork, but it's fine for now.

I added a comment inline to clarify.

analyzer/evmtokenbalances/evm_token_balances.go
mitjat: It's redundant in that we don't expect `GetNativeBalance()` to ever return `nil, nil`. I removed the check now for consistency. Man, Go's cavalier approach to possible nilness ... :rage: ... like Java in the 90s taught us nothing.

analyzer/queries/queries.go
mitjat: ok, done

analyzer/runtime/extract.go
mitjat: I'm wary of dead-reckoning where we're not close to 100% sure that we have the correct delta. See https://github.com/oasisprotocol/oasis-indexer/pull/422#discussion_r1199269233 and the code comment that the thread refers to.

For completeness: I now also ran an experiment where we do no dead reckoning at all; we only enqueue the involved accounts to be queried by `evm_token_balances` analyzer later. (We do the latter always, and that's the more important part of the PR, IMO.) That produces 849 corrections. So we might even want to just do that? IDK, leaving as-is for now.

analyzer/evmtokenbalances/evm_token_balances.go
ptrus: Ok thanks!

analyzer/evmtokenbalances/evm_token_balances.go
ptrus: :shrug:

analyzer/runtime/extract.go
ptrus: The dead-reckoning part makes sense. 

> we only enqueue the involved accounts to be queried by evm_token_balances analyzer later. (We do the latter always, and that's the more important part of the PR, IMO.)

I thought it would make sense to also do this for `EVMCreate` (it can send funds into contract address), `AccountsTransfer` (it can transfer native runtime balance between accounts), `ConsensusAccountsWithdraw` (it can move native funds out of the runtime account) and `ConsensusAccountsDeposit` (it can move funds into the runtime account)?

Sorry if this is already implemented and I missed it.

storage/oasis/nodeapi/file/runtime.go
```diff
@@ -73,6 +73,14 @@ func (r *FileRuntimeApiLite) GetEventsRaw(ctx context.Context, round uint64) ([]
 	)
 }
 
+func (r *FileRuntimeApiLite) GetNativeBalance(ctx context.Context, round uint64, addr nodeapi.Address) (*common.BigInt, error) {
```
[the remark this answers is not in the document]
pro-wh: there's no coordination to make sure the requests come at the same round when both analyzers run concurrently ):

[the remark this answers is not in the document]
mitjat: The current "solution" is to not run the balance analyzer in e2e tests :(
The proposed more real solution (though testing the balance analyzer in a slightly unnatural regime) is https://app.clickup.com/t/866aak668

analyzer/runtime/extract.go
```diff
@@ -404,7 +405,27 @@ func ExtractRound(blockHeader nodeapi.RuntimeBlockHeader, txrs []nodeapi.Runtime
...
+						// to re-query their real balance.
+						reckonedAmount := amount.ToBigInt() // Calls with an empty body represent a transfer of the native token.
+						if len(body.Data) != 0 || len(blockTransactionData.SignerData) > 1 {
+							// Calls with a non-empty body have no standard impact on native balance. Better to dead-reckon a 0 change (and keep stale balances)
+							// than to reckon a wrong change (and have a "random" incorrect balance until it is re-queried).
+							reckonedAmount = big.NewInt(0)
+						}
+						registerTokenIncrease(blockData.TokenBalanceChanges, evm.NativeRuntimeTokenAddress, to, reckonedAmount)
+						for _, signer := range blockTransactionData.SignerData {
+							registerTokenDecrease(blockData.TokenBalanceChanges, evm.NativeRuntimeTokenAddress, signer.Address, reckonedAmount)
```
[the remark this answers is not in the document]
mitjat: Neither. In practice, I don't think we have any txs with multiple signers. If we did, maybe this will do the wrong thing in terms of dead reckoning (or maybe not; IDK what the semantics of `value` are in that case), but at least all the signers will be marked dirty.

analyzer/runtime/extract.go
mitjat: Ohh I see. Thanks for clarifying. Very good point. I added dead reckoning to EVMCreate. (Not really, technically; I just mark them dirty. I ran the tests and doing dead reckoning resulted in 657 balance corrections overall, as opposed to NOT doing dead reckoning, which needed 656 balance corrections. These are both relative to the 643 corrections we needed when we ignored EVMCreate.)

I did not do anything for the non-evm txs. Those are fully controlled by the SDK/modules part of the runtime, so they will emit (non-evm) runtime events like `accounts.Transfer` that we [already handle](https://github.com/oasisprotocol/oasis-indexer/blob/9ae106d3a38c31cd21d83408518b772d22a0ef85/analyzer/runtime/accounts.go#L75-L75)

ptrus: > I did not do anything for the non-evm txs. Those are fully controlled by the SDK/modules part of the runtime, so they will emit (non-evm) runtime events like accounts.Transfer that we [already handle](https://github.com/oasisprotocol/oasis-indexer/blob/9ae106d3a38c31cd21d83408518b772d22a0ef85/analyzer/runtime/accounts.go#L75-L75)

Thanks, I was missing this part.

ptrus (approved):

analyzer/runtime/extract.go
ptrus: Out-of-scope for this PR, but I have one more question after looking at the native balance tracking in the accounts module. Is the potential fee charge tracked anywhere?
Technically, every transaction could additionally reduce the sender's balance for the charged fee amount. I assume this is already handled somewhere, but I couldn't find it.
Thanks!

mitjat: In my understanding, we track fees (and fee disbursements to validators) the same way we track all other transfers because these fee-related micro-transfers also generate `staking.Transfer` events. See [here](https://github.com/oasisprotocol/oasis-core/blob/595061512e534a79edfd92f9ae84a264cda78459/go/consensus/tendermint/apps/staking/fees.go#L79-L79) and all other mentions of `staking.TransferEvent` in that file.

ptrus: Thanks, that makes sense for the consensus layer. I'm a bit out of date on the runtime-sdk details, but as far as I recall, I don't think we emit any such events there - will double-check this (mid-writing edit: did double-check, we don't emit such events: https://github.com/oasisprotocol/oasis-sdk/blob/75e9e523c62b2857c3c1b4aee27c86ecdf2a671d/runtime-sdk/src/modules/accounts/mod.rs#L927). 

This PR now queues callers for all EVM transactions, so fees paid by those will be handled correctly. 

---
Out of scope of this PR:

I think that the code that handles [`Accounts` module](https://github.com/oasisprotocol/oasis-indexer/blob/9ae106d3a38c31cd21d83408518b772d22a0ef85/analyzer/runtime/accounts.go#L75-L75)  will miss any fees paid by the caller?

mitjat: I see. I didn't have a good mental model of runtime fees, and how they're completely separate from consensus fees. 

This is pretty messy and worrying :(
For now, since fees are often 0 (*), and since we have the "self-healing" process (= this PR marking senders' addresses as dirty, and their balance then being queried from the EVM), and since we're only supporting EVM-based runtimes right now ... We're OK in the short term.

Longer term, I created a writeup in https://app.clickup.com/t/866abwwaf. If you have time to read through it and see if I made any wrong assumptions, I'll appreciate it.

(*) NOT REALLY! In the first 10k blocks of Emerald (7259 txs), 459 txs have non-zero fees. Some of them are huge, like 199120300000000000000 Wei! So this probably (hopefully) explains a large part of the 600+ corrected balances that I saw in https://github.com/oasisprotocol/oasis-indexer/pull/422#discussion_r1199269233.

ptrus: The writeup looks good, thanks! 

> So this possibly explains a large part of the 600+ corrected balances that I saw in https://github.com/oasisprotocol/oasis-indexer/pull/422#discussion_r1199269233

Yeah, likely.

> And IIRC the old versions of EVM maybe didn't return the unspent portion of the fee?

IIRC yes indeed.

pro-wh (approved):
ya

the complexity of that stale tokens query is going to sit in the back of my mind for a few days

mitjat:
> the complexity of that stale tokens query is going to sit in the back of my mind for a few days

yeah it's one of the side effects of trying (a little too?) hard to shoehorn native tokens into the EVM-tokens framework :(
Luckily it decomposes pretty trivially into two only-borderline-complex queries.
IDK IDK

PR #423, opened 2023-05-22
evm: only store log signatures for parsed events

ptrus:
Log signatures are now stored only for parsed events. This is because only in case of parsed events, we can be sure that the thing in `topic[0]` is actually the event signature (otherwise it could be random data inserted by an anonymous event).

This makes the field a bit less useful until more EVM events will be parsed (once contract verification is supported). If the use-case for filtering arbitrary EVM events by topic is needed (regardless if verified or not), support for filtering by `topic[0]` value directly could be added to API.

Side-note:
In old code, the `evm_log_signature` was actually stored base64 encoded in the DB, which was probably unexpected. The API filtering by `evm_log_signature` did not work as described in the API docs (where hex string is used in the example). 

Testing:
```
select distinct(evm_log_signature) from chain.runtime_events where evm_log_signature != '';
                         evm_log_signature
--------------------------------------------------------------------
 0x8c5be1e5ebec7d5bd14f71427d1e84f3dd0314c0f7b2291e5b200ac8c7c3b925  # ERC-20 Approval
 0xddf252ad1be2c89b69c2b068fc378daa952ba7f163c4a11628f55a4df523b3ef # ERC-20 Transfer
(2 rows)

```

API test:
```
curl "localhost:8008/v1/emerald/events?evm_log_signature=0xddf252ad1be2c89b69c2b068fc378daa952ba7f163c4a11628f55a4df523b3ef&limit=1"
{"events":[{"body":{"address":"I4OW1NAbpWIeZolKAij2s2UfFWY=","data":"AAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAA=","topics":["3fJSrRviyJtpwrBo/DeNqpUrp/FjxKEWKPVaTfUjs+8=","AAAAAAAAAAAAAAAAjZzJ7hGq+GWRPe7JOe6y3Hg4q3s=","AAAAAAAAAAAAAAAARUyUgag32PmIFkKUTpoi1aP3kl0="]},"eth_tx_hash":"5cb3deb32fd4e3ca66c7af4dbb0b986e9b516b8ed06e4bbced34cc1b7911ca5e","evm_log_name":"Transfer","evm_log_params":[{"evm_type":"address","name":"from","value":"0x8d9cc9ee11aaf865913deec939eeb2dc7838ab7b"},{"evm_type":"address","name":"to","value":"0x454c9481a837d8f9881642944e9a22d5a3f7925d"},{"evm_type":"uint256","name":"value","value":"0"}],"round":1019379,"tx_hash":"dea14048b683bca15863465aafbecdfdd09c4aed8910635643ad9c7f4278311e","tx_index":0,"type":"evm.log"}],"is_total_count_clipped":true,"total_count":1000}
```

### DIFF
--- a/analyzer/runtime/extract.go
+++ b/analyzer/runtime/extract.go
@@ -76,6 +76,7 @@ type EventData struct {
 	Body             EventBody
 	WithScope        ScopedSdkEvent
 	EvmLogName       string
+	EvmLogSignature  ethCommon.Hash
 	EvmLogParams     []*apiTypes.EvmEventParam
 	RelatedAddresses map[apiTypes.Address]bool
 }
@@ -645,6 +646,7 @@ func extractEvents(blockData *BlockData, relatedAccountAddresses map[apiTypes.Ad
 						blockData.PossibleTokens[eventAddr].Mutated = true
 					}
 					eventData.EvmLogName = apiTypes.Erc20Transfer
+					eventData.EvmLogSignature = ethCommon.BytesToHash(event.Topics[0])
 					eventData.EvmLogParams = []*apiTypes.EvmEventParam{
 						{
 							Name:    "from",
@@ -687,6 +689,7 @@ func extractEvents(blockData *BlockData, relatedAccountAddresses map[apiTypes.Ad
 					amount := &big.Int{}
 					amount.SetBytes(amountU256)
 					eventData.EvmLogName = apiTypes.Erc20Approval
+					eventData.EvmLogSignature = ethCommon.BytesToHash(event.Topics[0])
 					eventData.EvmLogParams = []*apiTypes.EvmEventParam{
 						{
 							Name:    "owner",
--- a/storage/migrations/02_runtimes.up.sql
+++ b/storage/migrations/02_runtimes.up.sql
@@ -113,15 +113,16 @@ CREATE TABLE chain.runtime_events
   -- TODO: add link to openapi spec section with runtime event types.
   type TEXT NOT NULL,
   -- The raw event, as returned by the oasis-sdk runtime client.
-  -- `evm.log` events are further parsed into known event types,
-  -- e.g. (ERC20) Transfer, to populate the `evm_log_name` and
-  -- `evm_log_params` fields below.
   body JSONB NOT NULL,
+  related_accounts TEXT[],
+
+  -- The events of type `evm.log` are further parsed into known event types, e.g. (ERC20) Transfer,
+  -- to populate the `evm_log_name`, `evm_log_params`, and `evm_log_signature` fields.
+  -- These fields are populated only when the ABI for the event is known. Typically, this is the
+  -- case for events emitted by verified contracts.
   evm_log_name TEXT,
-  -- The event signature, if it exists, will be the first topic.
-  evm_log_signature TEXT GENERATED ALWAYS AS (body->'topics'->>0) STORED,
   evm_log_params JSONB,
-  related_accounts TEXT[]
+  evm_log_signature BYTEA CHECK (octet_length(evm_log_signature) = 32)
 );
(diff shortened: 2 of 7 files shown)


pro-wh:
are these changes desired for something?

> otherwise it could be random data inserted by an anonymous event

we've generally not done much to filter out unusual smart contract behavior. do we especially want it here rather than to mark those events as unpraseable when we later query it?



> This makes the field a bit less useful until more EVM events will be parsed (once contract verification is supported).

it'll also make it more complex to fill in the field for existing events when we add support for new event signatures, which I'm against



> In old code, the `evm_log_signature` was actually stored base64 encoded in the DB, which was probably unexpected.

with this change, there's now some inconsistency with evm_log_signature being a formatted ethereum hash, fully with the 0x prefix in my understanding. meanwhile, the body is full of un-'typed' byte arrays shoehorned into JSON, which as a rule means base64 strings. this kind of discussion seems to come up a lot, about what we want to store in the database, whether it's _data_ or what I'm calling "rendered" stuff.

storage/migrations/02_runtimes.up.sql
```diff
@@ -117,8 +117,8 @@ CREATE TABLE chain.runtime_events
   -- `evm_log_params` fields below.
   body JSONB NOT NULL,
   evm_log_name TEXT,
-  -- The event signature, if it exists, will be the first topic.
-  evm_log_signature TEXT GENERATED ALWAYS AS (body->'topics'->>0) STORED,
+  -- The event signature, present only for known event types.
```
mitjat, line 120: Maybe we can get rid of this comment now and instead incorporate it into line 117.
And/or somehow visually separate the `evm_*` fields and not that they are only populated for events whose `type='evm.log'` AND their ABI is known, which happens roughly when the emitting contract is verified.

analyzer/runtime/extract.go
```diff
@@ -654,6 +656,7 @@ func extractEvents(blockData *BlockData, relatedAccountAddresses map[apiTypes.Ad
 					amount := &big.Int{}
 					amount.SetBytes(amountU256)
 					eventData.EvmLogName = apiTypes.Erc20Approval
+					eventData.EvmEventSignature = ethCommon.BytesToHash(event.Topics[0]).Hex()
```
mitjat, line 659: I would prefer to store the signature in a more normalized representation, i.e. raw bytes (`BYTEA` db type).

If we store a string
- we have to be more careful to support variants of search strings, e.g. lowercase vs uppercase hex, or present/absent leading `0x`
- we use up more space in the db
- it's easier to unknowingly store the wrong thing (like we were storing base64 until this PR -- thank you for finding that!)

If we want to the API to return the hex (more customary in eth) as opposed to base64 (recommended/prescribed by openapi for []byte fields), we can do the serialization in SQL in the SELECT to avoid type gymnastics in Go.

mitjat (commented):
Looks good in general! I just have a suggestion about the storage format.

analyzer/runtime/extract.go
ptrus: Agreed, will update.

storage/migrations/02_runtimes.up.sql
ptrus: Done

analyzer/runtime/extract.go
ptrus: Updated!

mitjat (approved):
Thank you! We also discussed the PR in-person with @pro-wh and I think he's also OK the changes as-is (the discussion got a little spread over the threads and clickup, sorry), but he should weigh in himself.

pro-wh (approved):
as @mitjat posted above :+1:

PR #424, opened 2023-05-23
API: add endpoints for querying (debonding)delegations to an address

ptrus:
Closes: https://github.com/oasisprotocol/oasis-indexer/issues/391

TODO: testing
- [ ] sanity check some delegations by comparing delegations and delegantions-to API responses


api/spec/v1.yaml
```diff
@@ -594,6 +594,27 @@ paths:
                 $ref: '#/components/schemas/DelegationList'
         <<: *common_error_responses
 
+  /consensus/accounts/{address}/delegations_to:
```
ptrus, line 597: Used `(debonding)delegations_to` for the method and API names. This is the same terminology as used in oasis-core. 

I didn't like using `delegators` because that would imply (at least to me) that the endpoint only returns the delegators (just a list of addresses) to an address. However, the endpoint returns the full delegations (delegator-delegatee pair, and amount), same as the `delegations` API counterpart.

tests/e2e_regression/run.sh
```diff
@@ -36,6 +36,9 @@ testCases=(
   'account                  /v1/consensus/accounts/oasis1qp0302fv0gz858azasg663ax2epakk5fcssgza7j'
   'runtime-only_account     /v1/consensus/accounts/oasis1qphyxz5csvprhnn09r49nuyzl0jdw0wsj5xpvsg2'
   'delegations              /v1/consensus/accounts/oasis1qp0302fv0gz858azasg663ax2epakk5fcssgza7j/delegations'
+  'delegations_to           /v1/consensus/accounts/oasis1qp0j5v5mkxk3eg4kxfdsk8tj6p22g4685qk76fw6/delegations_to'
```
ptrus, line 39: Only these three lines were updated.

Other changes are autofixes done by my editor (_i think_ that by `prettier`, can revert if desired).

pro-wh (approved):
lg

I see the debonding delegations table has no indices. maybe we're fine because it tends to be small

tests/e2e_regression/run.sh
mitjat: :shrug: I think line 109 was clearer pre-`prettier` because it visually separated `cp` and `jq`, but I don't care enough. SGTM to leave as-is.

Thanks for adding the tests!

storage/client/queries/queries.go
```diff
@@ -170,6 +170,20 @@ const (
...
 		OFFSET $3::bigint`
 
+	DelegationsTo = `
+		-- delegatee_info is used to calculate the escrow amount of the delegators in base units
+		WITH delegatee_info AS (
+			SELECT escrow_balance_active, escrow_total_shares_active
+			FROM chain.accounts
+			WHERE address = $1::text
+		)
+	        SELECT delegator, shares, delegatee_info.escrow_balance_active, delegatee_info.escrow_total_shares_active
```
mitjat, line 180: nit: weird indentation (tabs vs spaces)

storage/client/queries/queries.go
```diff
@@ -179,6 +193,20 @@ const (
...
+		-- delegatee_info is used to calculate the debonding escrow amount of the delegators in base units
+		WITH delegatee_info AS (
+			SELECT escrow_balance_debonding, escrow_total_shares_debonding
+			FROM chain.accounts
+			WHERE address = $1::text
+		)
+		SELECT delegator, shares, debond_end, delegatee_info.escrow_balance_debonding, delegatee_info.escrow_total_shares_debonding
+			FROM chain.debonding_delegations, delegatee_info
+			WHERE delegatee = $1::text
+		ORDER BY debond_end
```
mitjat, line 206: Since this is given in epochs, it's quite coarse and bound to yield sorting ties, so API responses might not be fully deterministic. Let's add a secondary ORDER BY, probably `shares DESC, delegator` (like in the active delegations query).

storage/client/queries/queries.go
```diff
@@ -170,6 +170,20 @@ const (
...
+		WITH delegatee_info AS (
+			SELECT escrow_balance_active, escrow_total_shares_active
+			FROM chain.accounts
+			WHERE address = $1::text
+		)
+	        SELECT delegator, shares, delegatee_info.escrow_balance_active, delegatee_info.escrow_total_shares_active
+			FROM chain.delegations, delegatee_info
+			WHERE delegatee = $1::text
+		ORDER BY delegator, shares
+		LIMIT $2::bigint
```
mitjat, line 184: Note: This does require a sequential scan on the table. We currently have 60k delegations and the query takes ~250ms for the top delegatee (with ~60k delegators). I think we're far enough from a catastrophically slow query (nothing quadratic going on) that we can leave as-is for now.

api/spec/v1.yaml
mitjat: Excellent, I had the same misgiving about the proposed `delegators` endpoint name, and the same idea/wish for the `Delegation` struct.

I was also hoping we'd unify delegations and debonding delegations more :), but realistically the hassle/benefit ratio is probably not worth it. Thank you!

storage/client/client.go
```diff
@@ -685,6 +687,49 @@ func (c *StorageClient) Delegations(ctx context.Context, address staking.Address
...
+			&d.Address,
+			&shares,
+			&escrowBalanceActive,
+			&escrowTotalSharesActive,
+		); err != nil {
+			return nil, wrapError(err)
+		}
+		amount := new(big.Int).Mul(&shares.Int, &escrowBalanceActive.Int)
+		amount.Quo(amount, &escrowTotalSharesActive.Int)
+		d.Amount = BigInt{Int: *amount}
```
mitjat, line 724: We do this ugly dance in four places now; WDYT about extracting it into a helper func?

tests/e2e_regression/expected/spec.body
```diff
@@ -1109,6 +1151,10 @@ components:
           type: string
           description: The delegatee address.
           example: *staking_address_1
+        address:
```
mitjat, line 1154: Here and in `DebondingDelegation`: I don't think that the `address` of a delegation is obviously the delegator address, as opposed to the delegatee address.

Let's change this to `delegator_address`; or alternatively (and IMO better), just `delegator`, and rename `validator_address` to `validator`. (Nobody uses the consensus API yet, we can change it freely)

storage/client/queries/queries.go
```diff
@@ -170,6 +170,20 @@ const (
...
+		-- delegatee_info is used to calculate the escrow amount of the delegators in base units
+		WITH delegatee_info AS (
+			SELECT escrow_balance_active, escrow_total_shares_active
+			FROM chain.accounts
+			WHERE address = $1::text
+		)
+	        SELECT delegator, shares, delegatee_info.escrow_balance_active, delegatee_info.escrow_total_shares_active
+			FROM chain.delegations, delegatee_info
+			WHERE delegatee = $1::text
+		ORDER BY delegator, shares
```
mitjat, line 183: Kinda silly, but `delegator, shares DESC` is probably more natural? Or even `shares DESC, delegator`. Until we have user-controlled sorting in the UI, this is the order they're bound to want to see.

mitjat (approved):
Great, thank you!

api/spec/v1.yaml
ptrus: > I was also hoping we'd unify delegations and debonding delegations more :), but realistically the hassle/benefit ratio is probably not worth it. Thank you!

Maybe it would make sense unifying the debonding delegations with delegations more, but I'm not super sure about that.

It's not only that the debonding delegations are delegations with an end period, the debonding delegations also have a semantic difference - the delegator doesn't earn staking rewards for those shares, so these are actually "locked" shares.

So:
- delegations -> escrowed shares
- debonding delegations -> locked shares

So in some sense it makes sense to me to have these separated.

---

However, maybe with less in-depth logic of oasis-core (which is likely the majority of consumers of the API) this doesn't make all that much sense and it would be clearer to just merge these two. No strong opinions about it tbh.

storage/client/queries/queries.go
ptrus: Agreed that `shares DESC, delegator` makes most sense. Also updated for `Delegations` query.

tests/e2e_regression/expected/spec.body
ptrus: :+1:  Updated to `validator` and `delegator`.

storage/client/queries/queries.go
ptrus: If this ever gets problematic, one simple fix would be to include `shrares` in the "delegator"/"delegatee" indexes, since at least for now we always sort by shares first.

Let me know if I should add it in a follow-up PR.

storage/client/client.go
ptrus: Good idea, added

PR #425, opened 2023-05-23
API: support filtering (runtime) transactions by timestamp

ptrus:
Closes: https://github.com/oasisprotocol/oasis-indexer/issues/335

TODO: test


storage/migrations/02_runtimes.up.sql
```diff
@@ -70,6 +70,7 @@ CREATE TABLE chain.runtime_transactions
 );
 CREATE INDEX ix_runtime_transactions_tx_hash ON chain.runtime_transactions USING hash (tx_hash);
 CREATE INDEX ix_runtime_transactions_tx_eth_hash ON chain.runtime_transactions USING hash (tx_eth_hash);
+CREATE INDEX ix_runtime_transactions_timestamp ON chain.runtime_transactions (runtime, timestamp);
```
ptrus, line 73: Could avoid adding this index if `RuntimeTransactions` query is updated to also `LEFT JOIN on chain.runtime_blocks` and use the timestamp from there, since that one already has an index.

mitjat: That's one scary query already, let's not add more JOINs. Even as-is, it seems somewhat likely that it's possible to come up with a very resource-intensive query by combining just the right filters.

tests/e2e_regression/expected/spec.body
```diff
@@ -283,6 +283,20 @@ paths:
...
+            type: string
+            format: date-time
+          description: A filter on minimum transaction time, inclusive.
+          example: *iso_timestamp_1
+        - in: query
+          name: before
+          schema:
+            type: string
+            format: date-time
+          description: A filter on maximum transaction time, inclusive.
```
mitjat, line 298: I see we already have the "inclusive" part for blocks filtering (both consensus and runtime). How do you feel about changing it to "exclusive", just for the `before` arg, both for blocks and txs?

The web UI won't particularly care because it can specify the timestamps with arbitrary precision. But for other clients (and e.g. manual use), it feel very unintuitive to say `before=2022-03-15`, read that `before` is inclusive, and still get no results for `2022-03-15`. That will happen because `2022-03-15` expands to `2022-03-15T00:00:00`.

mitjat (approved):

tests/e2e_regression/expected/spec.body
ptrus: Agreed, will update, thanks!

storage/migrations/02_runtimes.up.sql
ptrus: :see_no_evil:

PR #426, opened 2023-05-23
stats: Update materialized views concurrently

mitjat:
Refreshing the materialized view takes ~5 min. Before this PR, the view was locked (even for read queries) during the refresh, making the related `/stats/tx_volume` endpoint unavailable for ~5min every ~1h, 1h being the refresh interval.

The changes (CREATE INDEX) have been applied to mainnet prod + staging indexers already.

Testing: Locally.

### DIFF
--- a/analyzer/queries/queries.go
+++ b/analyzer/queries/queries.go
@@ -477,11 +477,11 @@ const (
       account_address = $3`
 
 	RefreshDailyTxVolume = `
-    REFRESH MATERIALIZED VIEW stats.daily_tx_volume
+    REFRESH MATERIALIZED VIEW CONCURRENTLY stats.daily_tx_volume
   `
 
 	RefreshMin5TxVolume = `
-    REFRESH MATERIALIZED VIEW stats.min5_tx_volume
+    REFRESH MATERIALIZED VIEW CONCURRENTLY stats.min5_tx_volume
   `
 
 	// LatestDailyAccountStats is the query to get the timestamp of the latest daily active accounts stat.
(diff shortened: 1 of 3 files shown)


analyzer/queries/queries.go
```diff
@@ -477,11 +477,11 @@ const (
       account_address = $3`
 
 	RefreshDailyTxVolume = `
-    REFRESH MATERIALIZED VIEW stats.daily_tx_volume
+    REFRESH MATERIALIZED VIEW CONCURRENTLY stats.daily_tx_volume
```
pro-wh, line 480: :information_source: https://www.postgresql.org/docs/current/sql-refreshmaterializedview.html

> Without this option a refresh which affects a lot of rows will tend to use fewer resources and complete more quickly, but could block other connections which are trying to read from the materialized view.

pro-wh (approved):
index looks good too

analyzer/queries/queries.go
ptrus, line 480: Thanks!

From docs:

> This option may not be used when the materialized view is not already populated.

Does this mean that the first refresh on a fresh DB should be done non-concurrently?

ptrus:
Might make sense opening an issue for future, to refactor this to not constantly recompute all data. I think the logic from `dailyActiveAccountsWorker` could be extracted and reused.

analyzer/queries/queries.go
mitjat: Good find. I tested now on a synthetic example and it's not the case (= we're ok), but I too found it hard to parse the docs. The view is populated at creation time. I think they're talking about the corner case of having created the view with `CREATE MATERIALIZED VIEW ... WITH NO DATA` [(docs)](https://www.postgresql.org/docs/current/sql-creatematerializedview.html) and then trying to do a `REFRESH ... CONCURRENTLY`.

mitjat:
> Might make sense opening an issue for future, to refactor this to not constantly recompute all data. I think the logic from `dailyActiveAccountsWorker` could be extracted and reused.

Created https://app.clickup.com/t/866abpjhd

analyzer/queries/queries.go
ptrus: Good :+1: thanks for checking!

PR #427, opened 2023-05-25
api: Recognize sapphire, cipher as valid Layer values

mitjat:
The `IsValid()` function is used to validate APIs; before this bugfix PR, https://index-staging.oasislabs.com/v1/sapphire/stats/active_accounts returned `"Invalid format for parameter layer: not a valid enum value: sapphire"`

Reported by @csillag -- thanks!


Makefile
```diff
@@ -65,12 +65,23 @@ fill-cache-for-e2e-regression: oasis-indexer
 
 # Run the api tests locally, assuming the environment is set up with an oasis-node that is 
 # accessible as specified in the config file.
-test-e2e-regression: export TZ=UTC
 test-e2e-regression: oasis-indexer
-	./oasis-indexer --config tests/e2e_regression/e2e_config.yml analyze
+	TZ=UTC ./oasis-indexer --config tests/e2e_regression/e2e_config.yml analyze
```
mitjat, line 69: Unrelated. No-op change, just makes the command easier to copy-paste out of the Makefile

Makefile
```diff
@@ -65,12 +65,23 @@ fill-cache-for-e2e-regression: oasis-indexer
...
 
+# Accept the outputs of the e2e tests as the new expected outputs.
+accept-e2e-regression:
+	@# Delete all old expected files first, in case any test cases were renamed or removed.
+	rm -rf ./tests/e2e_regression/expected
+	@# Copy the actual outputs to the expected outputs.
+	cp -r  ./tests/e2e_regression/{actual,expected}
+	@# The result of the "spec" test is a special case. It should always match the
+	@# current openAPI spec file, so we symlink it to avoid having to update the expected
+	@# output every time the spec changes.
```
mitjat, line 81: The reason I abstracted the inline command form `run.sh` into this `accept-e2e-regressions` rule in the Makefile. The PR caused e2e regressions to fail just because it changed the openapi JSON (but not the e2e expected file). It would be awkward if every change to the spec produced two equivalent diffs: one in the actual spec, and one in the e2e expected file.

Makefile
```diff
@@ -65,12 +65,23 @@ fill-cache-for-e2e-regression: oasis-indexer
...
 # accessible as specified in the config file.
-test-e2e-regression: export TZ=UTC
 test-e2e-regression: oasis-indexer
-	./oasis-indexer --config tests/e2e_regression/e2e_config.yml analyze
+	TZ=UTC ./oasis-indexer --config tests/e2e_regression/e2e_config.yml analyze
 	@$(ECHO) "$(CYAN)*** Indexer finished; starting api tests...$(OFF)"
 	./tests/e2e_regression/run.sh
 
+# Accept the outputs of the e2e tests as the new expected outputs.
+accept-e2e-regression:
```
pro-wh, line 74: add to .PHONY?

pro-wh (approved):
as described

Makefile
mitjat: :shrug: We don't really use/need it. Added, and half-heartedly added several other e2e-related targets that were missing, but didn't make a careful pass to ensure _every_ relevant target is listed under .PHONY.

Makefile
Andrew7234: I think we actually need to move this to the line that instantiates the api server in `run.sh`, since the timestamp is formatted by the api server.

Andrew7234 (approved):
Thanks! One comment but lgtm otherwise!

PR #431, opened 2023-06-06
tests: assert.Nil -> require.Nil in some places

ptrus:
Use `require` instead of `assert` in places where continuing with tests does not make sense due to the failure (e.g. connecting to the db, querying the node etc...).  `require` is similar to `assert` but it also stops the execution of the test case, while `assert` continues.

This makes `statecheck` output a bit more readable (and meaningful) in case it fails early. 

Concrete example: if `oasisRuntimeClient.Accounts.Addresses`  failed, the `statecheck` reported a discrepancy for every account found in the postgresql db, which masked the actual error - loading the state in the first place.

### DIFF
--- a/storage/oasis/client_test.go
+++ b/storage/oasis/client_test.go
@@ -5,7 +5,6 @@ import (
 	"os"
 	"testing"
 
-	"github.com/stretchr/testify/assert"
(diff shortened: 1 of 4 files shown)


Andrew7234 (approved):
nice!

Issue #432, opened 2023-06-06
file consensus api lite: parameterize GetGenesisDocument key


pro-wh:
currently:

```go
generateCacheKey("GetGenesisDocument")
```

going forward, we'll probably want the genesis document of later networks, but we'll always see it as cached. the cache key probably needs something to specify which archive, e.g. damask

related, there's this current complication where we are indexing from cobalt but the consensus analyzer always 'starts' by getting the latest genesis doc, which is from damask. that leads to consistency errors. this might also be a reason to parameterize the overall GetGenesisDocument call so that the analyzer can specify which one it wants

PR #433, opened 2023-06-07
storage: add genesis document chain context parameter

pro-wh:
this allows (1) the file impl to cache different genesis documents and (2) the history impl to choose which archive node to use.

hack alert: the parameter is irrelevant to actual archive implementations. they ignore this parameter.

part of #432


storage/oasis/nodeapi/history/history.go
```diff
@@ -104,14 +104,24 @@ func (c *HistoryConsensusApiLite) APIForHeight(height int64) (nodeapi.ConsensusA
...
 
-func (c *HistoryConsensusApiLite) GetGenesisDocument(ctx context.Context) (*genesis.Document, error) {
-	// Use latest.
-	height := consensus.HeightLatest
-	api, err := c.APIForHeight(height)
+func (c *HistoryConsensusApiLite) APIForChainContext(chainContext string) (nodeapi.ConsensusApiLite, error) {
+	record, err := c.History.RecordForChainContext(chainContext)
 	if err != nil {
-		return nil, fmt.Errorf("getting api for height %d: %w", height, err)
+		return nil, fmt.Errorf("dertermining archive: %w", err)
```
Andrew7234, line 110:
```suggestion
		return nil, fmt.Errorf("determining archive: %w", err)
```

Andrew7234 (approved):

Andrew7234:
Note: the e2e_regression tests are failing bc the cache needs to be updated with `make fill-cache-for-e2e-regression` and `make accept-e2e-regression`. This requires setting up an oasis-node connection and starting `make postgres` beforehand.

However, I have an incoming update to a lot of the e2e_regression test anyway, so feel free to skip :)

pro-wh:
tried updating the cache. I have no way to delete the old genesis doc though

but why would it have failed with this message?

```
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x38 pc=0x10ab4b0]

goroutine 48 [running]:
github.com/oasisprotocol/oasis-indexer/storage/oasis/nodeapi/file.(*FileConsensusApiLite).GetGenesisDocument.func1()
	github.com/oasisprotocol/oasis-indexer/storage/oasis/nodeapi/file/consensus.go:56 +0x30
```

Andrew7234:
>  why would it have failed with this message?

I think since the cache is keyed by method name + args, it's looking for a cache entry that doesn't exist, hence the nil pointer exception.

pro-wh:
ah yeah, when that's false, it doesn't set up the FileConsensusApiLite .consensusApi field, allowing the `valueFunc` to fail with a nil pointer deref

PR #438, opened 2023-06-08
tests: don't run tests concurrently

ptrus:
Fixes: https://github.com/oasisprotocol/oasis-indexer/issues/435

By default, `go test` will run tests from different packages concurrently in separate instances of the test binary. Since we have tests that use the local postgresql instance in multiple packages, make sure these are not run concurrently.

The CI `Go Build` step time increases from 3 to 5 minutes in CI.


pro-wh (approved):
we're taking mitigations this week. as described

interesting, `-p` is documented in the `go build` section rather than in `go test` :skull: 

https://pkg.go.dev/cmd/go#hdr-Compile_packages_and_dependencies

Issue #439, opened 2023-06-08
ERROR: invalid byte sequence for encoding \"UTF8\": 0x00 (SQLSTATE 22021)


ptrus:
Staging mainnet analyzer
```
{
    "args": [
        "emerald",
        5251077,
        0,
        "6304eff38a3d8118fa391c20356bc0210dcd0695769b83c68de4120b15d0c78f",
        "8d979f57c0118d9442ab97a162a1dc8644096df2b4b1bb35b05fe24c8686edb6",
        "75000000000000000",
        750000,
        213775,
        752,
        "2023-04-10T11:42:22Z",
        "evm.Call",
        {
            "address": "FxBUibW4yYZHRsGhb3GJMAGCWdQ=",
            "value": "",
[...]

PR #440, opened 2023-06-08
runtime: error message does not need to be a valid utf-8

ptrus:
Fixes: https://github.com/oasisprotocol/oasis-indexer/issues/439

An alternative would be to change the error_message field to `bytea`, but it seems that non-utf8 encodable error messages are the exception rather than the norm, so maybe it makes more sense handling it this way?

Could also prefix it with "hex-encoded" or something.

### DIFF
--- a/analyzer/runtime/runtime.go
+++ b/analyzer/runtime/runtime.go
@@ -2,6 +2,7 @@ package runtime
 
 import (
 	"context"
+	"encoding/hex"
 	"fmt"
 	"math/big"
 	"strings"
@@ -193,6 +194,14 @@ func (m *processor) queueDbUpdates(batch *storage.QueryBatch, data *BlockData) {
 			error_module = transactionData.Error.Module
 			error_code = transactionData.Error.Code
 			error_message = transactionData.Error.Message
+		}
+		if error_message != nil {
+			// Apparently the message does need to be valid UTF-8.
+			// In the rare case it's not, hex encode it.
+			// https://github.com/oasisprotocol/oasis-indexer/issues/439
+			if !storage.IsValidText(*error_message) {
+				*error_message = hex.EncodeToString([]byte(*error_message))
+			}
 		}
 		batch.Queue(
 			queries.RuntimeTransactionInsert,
--- a/storage/api.go
+++ b/storage/api.go
@@ -3,7 +3,9 @@ package storage
 
 import (
 	"context"
+	"strings"
 	"sync"
+	"unicode/utf8"
 
 	"github.com/jackc/pgx/v5"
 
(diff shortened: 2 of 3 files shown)


ptrus:
Yes, although cbor also requires strings to be valid utf-8 it does allow NUL bytes, which postgres TEXT does not.

pro-wh (approved):
we're taking mitigations this week. hex-encodes as described

feel free to adjust the behavior of this later

thoughts on this:

- related to your idea about adding a prefix, it does sound nice to have a way to distinguish. but maybe something more systematic. probably have a prefix for both cases, one to introduce valid utf-8 string, one to introduce hex-encoded binary
- or maybe two separate columns :skull: 
- I was searching around, and the Go blog said that strings are UTF-8. I think CBOR also says text is UTF-8. so maybe we ought to fix this in the runtime layer someday. still gonna need support for this in the indexer now that these blocks exist though.
- and due to Go and CBOR saying it ought to be UTF-8, I wonder if anything has happened to this message since it left the EVM. maybe the bytes aren't fully preserved, yet still are invalid. people ought to be aware that even though it's presented in hex, the actual bytes we don't really know are accurate or not

ptrus:
> related to your idea about adding a prefix, it does sound nice to have a way to distinguish. but maybe something more systematic. probably have a prefix for both cases, one to introduce valid utf-8 string, one to introduce hex-encoded binary
> or maybe two separate columns skull

Agreed, however it does seem there was only one such case in first 5251077 emerald blocks, so I'm kinda meh on properly supporting this case further.

> I was searching around, and the Go blog said that strings are UTF-8. I think CBOR also says text is UTF-8. so maybe we ought to fix this in the runtime layer someday. still gonna need support for this in the indexer now that these blocks exist though.

Go string is a slice of bytes and it can contain arbitrary bytes. See test examples in this PR: e.g. `"\xc5z"` is not a valid UTF-8 string.

CBOR does require UTF-8 strings, though. Unmarshaling into a string will fail otherwise (both by definition and in practice in the `go` and `rust` libraries we use).

> ought to fix this in the runtime layer someday
> and due to Go and CBOR saying it ought to be UTF-8, I wonder if anything has happened to this message since it left the EVM. maybe the bytes aren't fully preserved, yet still are invalid. people ought to be aware that even though it's presented in hex, the actual bytes we don't really know are accurate or not

Yeah, we do this already in oasis-sdk, but actually due to the rust `String` type, which does require UTF-8 :-) : https://github.com/oasisprotocol/oasis-sdk/blob/ca630e7d48986bd102d7aa477a48f8966a3d1d23/runtime-sdk/modules/evm/src/lib.rs#L259

---

The actual problem in this specific case #439 are NUL bytes in the strings. NUL is technically valid UTF-8 byte, but Postgresql does not allow it in TEXT (i guess due to being used as string terminator in C).

---


So all in all it's like:
- go -> string can be whatever
- rust & cbor -> string needs to be UTF-8
- postgresql -> TEXT needs to be UTF-8 without NUL bytes

Due to rust and cbor in practice we would only need to check for NUL bytes, but since go strings can also be not UTF-8 I implemented the `IsValidText` helper that checks a go string for both.

PR #447, opened 2023-06-14
ERC-721 support

pro-wh:
functionality:
- token scanner can detect ERC-721 using ERC-165 supportsInterface
- block scanner can detect ERC-721 events and schedule token scanning
- schedules 'mutable' parts rescan (i.e. total supply) when tokens are minted
- registers transaction related accounts when seeing events

not included:
- approvals tracking (similar to that not being supported for ERC-20)
- tracking ownership of tokens
- holders count

tested on recent blocks on emerald, where a BlockchainCuties transfer had occurred

### DIFF
--- a/analyzer/runtime/evm/client.go
+++ b/analyzer/runtime/evm/client.go
@@ -160,22 +160,28 @@ func evmCallWithABI(
 	return evmCallWithABICustom(ctx, source, round, DefaultGasPrice, DefaultGasLimit, DefaultCaller, contractEthAddr, DefaultValue, contractABI, result, method, params...)
 }
 
+// logDeterministicError is for if we know how to handle a deterministic
+// error--in those cases you can use this to make a note of the error. Just in
+// case someone wasn't expecting it, you know?
+func logDeterministicError(logger *log.Logger, round uint64, contractEthAddr []byte, interfaceName string, method string, err error, keyvals ...interface{}) {
+	keyvals = append([]interface{}{
+		"round", round,
+		"contract_eth_addr_hex", hex.EncodeToString(contractEthAddr),
+		"interface_name", interfaceName,
+		"method", method,
+		"err", err,
+	}, keyvals...)
+	logger.Info("call failed", keyvals...)
+}
+
 func evmDownloadTokenERC20Mutable(ctx context.Context, logger *log.Logger, source nodeapi.RuntimeApiLite, round uint64, tokenEthAddr []byte) (*EVMTokenMutableData, error) {
 	var mutable EVMTokenMutableData
-	logError := func(method string, err error) {
-		logger.Info("ERC20 call failed",
-			"round", round,
-			"token_eth_addr_hex", hex.EncodeToString(tokenEthAddr),
-			"method", method,
-			"err", err,
-		)
-	}
 	// These mandatory methods must succeed, or we do not count this as an ERC-20 token.
 	if err := evmCallWithABI(ctx, source, round, tokenEthAddr, evmabi.ERC20, &mutable.TotalSupply, "totalSupply"); err != nil {
-		logError("totalSupply", err)
 		if !errors.Is(err, EVMDeterministicError{}) {
 			return nil, fmt.Errorf("calling totalSupply: %w", err)
 		}
+		logDeterministicError(logger, round, tokenEthAddr, "ERC20", "totalSupply", err)
 		return nil, nil
 	}
 	return &mutable, nil
@@ -185,17 +191,8 @@ func evmDownloadTokenERC20(ctx context.Context, logger *log.Logger, source nodea
 	tokenData := EVMTokenData{
 		Type: EVMTokenTypeERC20,
 	}
-	logError := func(method string, err error) {
-		logger.Info("ERC20 call failed",
-			"round", round,
-			"token_eth_addr_hex", hex.EncodeToString(tokenEthAddr),
-			"method", method,
-			"err", err,
-		)
-	}
 	// These optional methods may fail.
 	if err := evmCallWithABI(ctx, source, round, tokenEthAddr, evmabi.ERC20, &tokenData.Name, "name"); err != nil {
-		logError("name", err)
 		// Propagate the error (so that token-info fetching may be retried
 		// later) only if the error is non-deterministic, e.g. network
 		// failure. Otherwise, accept that the token contract doesn't expose
@@ -203,18 +200,19 @@ func evmDownloadTokenERC20(ctx context.Context, logger *log.Logger, source nodea
 		if !errors.Is(err, EVMDeterministicError{}) {
 			return nil, fmt.Errorf("calling name: %w", err)
 		}
+		logDeterministicError(logger, round, tokenEthAddr, "ERC20", "name", err)
 	}
 	if err := evmCallWithABI(ctx, source, round, tokenEthAddr, evmabi.ERC20, &tokenData.Symbol, "symbol"); err != nil {
-		logError("symbol", err)
 		if !errors.Is(err, EVMDeterministicError{}) {
 			return nil, fmt.Errorf("calling symbol: %w", err)
 		}
+		logDeterministicError(logger, round, tokenEthAddr, "ERC20", "symbol", err)
 	}
 	if err := evmCallWithABI(ctx, source, round, tokenEthAddr, evmabi.ERC20, &tokenData.Decimals, "decimals"); err != nil {
-		logError("decimals", err)
 		if !errors.Is(err, EVMDeterministicError{}) {
 			return nil, fmt.Errorf("calling decimals: %w", err)
 		}
+		logDeterministicError(logger, round, tokenEthAddr, "ERC20", "decimals", err)
 	}
 	mutable, err := evmDownloadTokenERC20Mutable(ctx, logger, source, round, tokenEthAddr)
 	if err != nil {
@@ -276,21 +274,14 @@ func EVMDownloadMutatedToken(ctx context.Context, logger *log.Logger, source nod
 
 func evmDownloadTokenBalanceERC20(ctx context.Context, logger *log.Logger, source nodeapi.RuntimeApiLite, round uint64, tokenEthAddr []byte, accountEthAddr []byte) (*EVMTokenBalanceData, error) {
 	var balanceData EVMTokenBalanceData
-	logError := func(method string, err error) {
-		logger.Info("ERC20 call failed",
-			"round", round,
-			"token_eth_addr_hex", hex.EncodeToString(tokenEthAddr),
-			"account_eth_addr_hex", hex.EncodeToString(accountEthAddr),
-			"method", method,
-			"err", err,
-		)
-	}
 	accountECAddr := ethCommon.BytesToAddress(accountEthAddr)
 	if err := evmCallWithABI(ctx, source, round, tokenEthAddr, evmabi.ERC20, &balanceData.Balance, "balanceOf", accountECAddr); err != nil {
-		logError("balanceOf", err)
 		if !errors.Is(err, EVMDeterministicError{}) {
 			return nil, fmt.Errorf("calling balanceOf: %w", err)
 		}
+		logDeterministicError(logger, round, tokenEthAddr, "ERC20", "balanceOf", err,
+			"account_eth_addr_hex", hex.EncodeToString(accountEthAddr),
+		)
 		return nil, nil
 	}
 	return &balanceData, nil
--- a/analyzer/runtime/evm/client.go
+++ b/analyzer/runtime/evm/client.go
@@ -12,7 +12,6 @@ import (
 	"github.com/oasisprotocol/oasis-core/go/common/errors"
 	sdkTypes "github.com/oasisprotocol/oasis-sdk/client-sdk/go/types"
 
-	"github.com/oasisprotocol/nexus/analyzer/evmabi"
 	apiTypes "github.com/oasisprotocol/nexus/api/v1/types"
 	"github.com/oasisprotocol/nexus/common"
 	"github.com/oasisprotocol/nexus/log"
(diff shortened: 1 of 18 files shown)


pro-wh:
>   Error: `if ok && supportsMetadata` has complex nested blocks (complexity: 6) (nestif)

:angry:

analyzer/runtime/evm/client.go
```diff
@@ -149,90 +156,48 @@ func evmCallWithABI(
...
 // with the `Type` field set to `EVMTokenTypeUnsupported`.
 func EVMDownloadNewToken(ctx context.Context, logger *log.Logger, source nodeapi.RuntimeApiLite, round uint64, tokenEthAddr []byte) (*EVMTokenData, error) {
-	// todo: check ERC-165 0xffffffff compliance
-	// todo: try other token standards based on ERC-165
-	// see https://github.com/oasisprotocol/oasis-indexer/issues/225
+	useERC165, err := detectERC165(ctx, logger, source, round, tokenEthAddr)
+	if err != nil {
+		return nil, fmt.Errorf("detect ERC165: %w", err)
+	}
+	if useERC165 {
```
mitjat, line 185:
```suggestion
	// Note: Per spec, every ERC-721 token has to support ERC-165.
```

analyzer/runtime/evm/client.go
```diff
@@ -149,90 +156,48 @@ func evmCallWithABI(
...
+				return nil, fmt.Errorf("download token ERC-721: %w", err2)
+			}
+			return tokenData, nil
+		}
+
+		// todo: add support for other token types
+		// see https://github.com/oasisprotocol/oasis-indexer/issues/225
+	}
 
 	// Check ERC-20.
```
mitjat, line 202:
```suggestion
	// Check ERC-20. These tokens typically do not implement ERC-165, so do not bother checking for
	// explicit declaration of support, detect ERC-20 support with heuristics instead.
```

analyzer/runtime/evm/erc165.go
```diff
@@ -0,0 +1,79 @@
...
+	return id
+}
+
+var (
+	InvalidInterfaceID             = []byte{0xff, 0xff, 0xff, 0xff}
+	ERC165InterfaceID              = InterfaceID(evmabi.ERC165)
+	ERC721InterfaceID              = InterfaceID(evmabi.ERC721)
+	ERC721TokenReceiverInterfaceID = InterfaceID(evmabi.ERC721TokenReceiver)
+	ERC721MetadataInterfaceID      = InterfaceID(evmabi.ERC721Metadata)
+	ERC721EnumerableInterfaceID    = InterfaceID(evmabi.ERC721Enumerable)
```
mitjat, line 33: Consider providing the actual hex value in a comment, for ctrl+F debugging

analyzer/runtime/evm/erc165.go
```diff
@@ -0,0 +1,79 @@
...
+)
+
+const ERC165GasLimit uint64 = 30_000
```
mitjat, line 36: why do we need this instead of `DefaultGasLimit`?

analyzer/runtime/evm/erc165.go
```diff
@@ -0,0 +1,79 @@
...
+	var result bool
+	var interfaceIDFixed [4]byte
+	copy(interfaceIDFixed[:], interfaceID)
+	if err := evmCallWithABICustom(ctx, source, round, DefaultGasPrice, ERC165GasLimit, DefaultCaller, contractEthAddr, DefaultValue, evmabi.ERC165, &result, "supportsInterface", interfaceIDFixed); err != nil {
+		return false, err
+	}
+	return result, nil
+}
+
+func detectInterface(ctx context.Context, logger *log.Logger, source nodeapi.RuntimeApiLite, round uint64, contractEthAddr []byte, interfaceID []byte) (bool, bool, error) {
```
mitjat, line 48: Can you document the second return variable please?

It's weird to have `ok` and `err`. Can we fold the functionality of `ok` into the `err` somehow, perhaps by using `EVMDeterministicError`? (I haven't thought this through, just noting a weirdness.)

mitjat (commented):
Looking good, and close to victory!

analyzer/runtime/evm/client.go
```diff
@@ -24,7 +23,7 @@ type EVMTokenType int
 const (
 	EVMTokenTypeNative      EVMTokenType = -1 // A placeholder type to represent the runtime's native token. No contract should be assigned this type.
 	EVMTokenTypeUnsupported EVMTokenType = 0  // A smart contract for which we're confident it's not a supported token kind.
-	EVMTokenTypeERC20       EVMTokenType = 20
```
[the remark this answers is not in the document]
pro-wh: adding a doc comment

analyzer/runtime/evm/client.go
```diff
@@ -149,90 +156,48 @@ func evmCallWithABI(
...
 
 // EVMDownloadNewToken tries to download the data of a given token. If it
 // transiently fails to download the data, it returns with a non-nil error. If
 // it deterministically cannot download the data, it returns a struct
 // with the `Type` field set to `EVMTokenTypeUnsupported`.
 func EVMDownloadNewToken(ctx context.Context, logger *log.Logger, source nodeapi.RuntimeApiLite, round uint64, tokenEthAddr []byte) (*EVMTokenData, error) {
-	// todo: check ERC-165 0xffffffff compliance
-	// todo: try other token standards based on ERC-165
-	// see https://github.com/oasisprotocol/oasis-indexer/issues/225
+	useERC165, err := detectERC165(ctx, logger, source, round, tokenEthAddr)
```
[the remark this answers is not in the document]
pro-wh: huh, hadn't thought of that. ok

analyzer/runtime/evm/client.go
pro-wh: ok. adding manually so that we don't actually delete the `if`

analyzer/runtime/evm/client.go
pro-wh: ok

pro-wh: or rather I'll have it say that ERC-20 doesn't require ERC-165 support

analyzer/runtime/evm/erc165.go
pro-wh: let's have that association in the tests so that they're programmatically verified

analyzer/runtime/evm/erc165.go
pro-wh: erc-165 says they must work with this amount of gas and puts forth an algorithm in which you use this amount of gas. the default gas used in other calls is a guess that could theoretically be even lower than this, which could cause us to miss things

analyzer/runtime/evm/erc165.go
pro-wh: ok on adding doc

about the design of this though, folding the functionality of `ok` into `err` I'm opposed to. the purpose of this is to handle a kind of error. handling is all of these things, and if we do something other than all or none of them, we'll harm the composability, break consistency, give programmers stomach problems, etc

1. make sure errors have a way to get to eyeballs -- detectInterface uses logDeterministicError
2. take us out of the idiomatic error workflow -- detectInterface returns nil err
3. substitute a consistent value/state/action/whatever -- detectInterface returns supports=indeterminate, ok=false

pro-wh: oh you know what, it'll be simpler if I make a wrapper that &&'s the ok

analyzer/runtime/evm/client.go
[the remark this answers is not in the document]
pro-wh: added to EVMTokenType

analyzer/runtime/extract.go
```diff
@@ -523,7 +524,7 @@ func sumGasUsed(events []*EventData) (sum uint64, foundGasUsedEvent bool) {
 	return
 }
 
-func extractEvents(blockData *BlockData, relatedAccountAddresses map[apiTypes.Address]bool, eventsRaw []nodeapi.RuntimeEvent) ([]*EventData, error) {
+func extractEvents(blockData *BlockData, relatedAccountAddresses map[apiTypes.Address]bool, eventsRaw []nodeapi.RuntimeEvent) ([]*EventData, error) { //nolint:gocyclo
```
pro-wh, line 527: weird that it thinks the complexity of extractEvents has increased, there's only entries added to a map literal :upside_down_face:

analyzer/runtime/extract.go
```diff
@@ -727,6 +728,144 @@ func extractEvents(blockData *BlockData, relatedAccountAddresses map[apiTypes.Ad
...
+						eventData.RelatedAddresses[fromAddr] = true
+					}
+					if !toZero {
+						toAddr, err2 := registerRelatedEthAddress(blockData.AddressPreimages, relatedAccountAddresses, toEthAddr)
+						if err2 != nil {
+							return fmt.Errorf("to: %w", err2)
+						}
+						eventData.RelatedAddresses[toAddr] = true
+					}
+					// todo: reckon ownership
```
pro-wh, line 750: this PR doesn't add token ownership tracking. there are no 'balances' in this current state of ERC-721 support

analyzer/runtime/extract.go
```diff
@@ -727,6 +728,144 @@ func extractEvents(blockData *BlockData, relatedAccountAddresses map[apiTypes.Ad
...
+						operatorAddr, err2 := registerRelatedEthAddress(blockData.AddressPreimages, relatedAccountAddresses, operatorEthAddr)
+						if err2 != nil {
+							return fmt.Errorf("operator: %w", err2)
+						}
+						eventData.RelatedAddresses[operatorAddr] = true
+					}
+					if _, ok := blockData.PossibleTokens[eventAddr]; !ok {
+						blockData.PossibleTokens[eventAddr] = &evm.EVMPossibleToken{}
+					}
+					approved := approvedBool[31] != 0
```
pro-wh, line 845: these topic+data parsing are all very ad-hoc, I'm hoping to do something easier in #421 with the go-ethereum libraries

analyzer/runtime/extract.go
```diff
@@ -727,6 +728,144 @@ func extractEvents(blockData *BlockData, relatedAccountAddresses map[apiTypes.Ad
...
+					if fromZero != toZero && tokenID.Cmp(&big.Int{}) != 0 {
+						blockData.PossibleTokens[eventAddr].Mutated = true
+					}
+					eventData.EvmLogName = evmabi.ERC721.Events["Transfer"].Name
+					eventData.EvmLogSignature = ethCommon.BytesToHash(event.Topics[0])
+					eventData.EvmLogParams = []*apiTypes.EvmEventParam{
+						{
+							Name:    "from",
+							EvmType: "address",
+							Value:   ethCommon.BytesToAddress(fromEthAddr),
```
mitjat, line 767: Is this format compatible with what https://github.com/oasisprotocol/oasis-indexer/pull/421 (or rather, an analyzer that makes use of it) will eventually produce?
I guess even if it's not, we can leave it in and later change the parse; this is intended primarily for human consumption, and stability doesn't matter that much. But it would be nice if we don't have to change it.

analyzer/runtime/extract.go
mitjat: :+1: 
I expect that eventually, all this ERC-20 and ERC-721 handling from extract.go will move into a separate analyzer (that will use #421). Because right now, we trigger e.g. the ERC721Transfer handler just based on the fact that the event signature matches the ABI of ERC721 Transfer. We do not look at the token, and what we've learned about it.
Hrm, that will make dead reckoning trickier.
Anyway, not really important for this PR.

storage/client/queries/queries.go
```diff
@@ -452,6 +452,7 @@ const (
 			tokens.total_supply,
 			CASE -- NOTE: There are two queries that use this CASE via copy-paste; edit both if changing.
```
mitjat, line 453: :point_up:

mitjat (approved):
Noice!

analyzer/runtime/extract.go
pro-wh: > all this ERC-20 and ERC-721 handling from extract.go will move into a separate analyzer

I think not, this reckoning is special, which we won't do in general for other interfaces

analyzer/runtime/evm/erc165.go
```diff
@@ -0,0 +1,91 @@
...
+	ERC721TokenReceiverInterfaceID = InterfaceID(evmabi.ERC721TokenReceiver)
+	ERC721MetadataInterfaceID      = InterfaceID(evmabi.ERC721Metadata)
+	ERC721EnumerableInterfaceID    = InterfaceID(evmabi.ERC721Enumerable)
+)
+
+const ERC165GasLimit uint64 = 30_000
+
+func callERC165SupportsInterface(ctx context.Context, source nodeapi.RuntimeApiLite, round uint64, contractEthAddr []byte, interfaceID []byte) (bool, error) {
+	var result bool
+	var interfaceIDFixed [4]byte
```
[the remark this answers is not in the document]
pro-wh: right, they won't take a slice for bytes4

analyzer/runtime/extract.go
pro-wh: > Is this format compatible with what #421

ya, the go-ethereum library's own address type

storage/client/queries/queries.go
pro-wh: omg there are?

analyzer/runtime/evm/erc165.go
pro-wh: won't provide. marking as resolved :stuck_out_tongue:

analyzer/runtime/evm/erc165.go
pro-wh: keeping this. marking as resolved

analyzer/runtime/extract.go
pro-wh: the extract function must grow. marking as resolved

analyzer/runtime/extract.go
pro-wh: gonna all-caps the todo. marking as resolved

analyzer/runtime/extract.go
pro-wh: keeping it ad-hoc, marking as resolved

PR #448, opened 2023-06-15
Analyzer for fetching bytecode of EVM contracts

mitjat:
This PR adds a new analyzer that fetches the runtime bytecode of EVM contracts. Some details:
- Adds (and populates) a `runtime_bytecode` column to the `evm_contracts` table.
- Code is fetched with a call into the runtime (`getCode()`). Even for contracts created via `evm.Create` (which is [not all of them](https://github.com/oasisprotocol/oasis-indexer/pull/428#discussion_r1206741225)), we only learn the deployment code, which the EVM then interprets to produce the runtime code.
- Like the evm_tokens analyzer, calls `getCode()` at the height of the most recent indexer-processed block. This is to keep the call cacheable (which it would not be if we called at the `latestBlock` pseudo-height).
- Adds a new `evm_contract_code` table to track accounts for which we already tried calling `getCode()` and realized the address was not a contract, and also to track accounts that we've encountered so far that look contract-like (=they were the target of an `evm.Call` tx). We need to track non-contracts because we need to try getCode() on every contract-like address because otherwise we won't know about contracts created by other contracts.
    - This new table is created in a new `analysis` DB schema. The goal is to eventually have all `*_analysis` tables live there; they're not about the contents of the chain, they're about the internal state of the analyzer(s).

Testing:
Ran for 10000 blocks of Emerald. Compared with Blockscout as sanity check. Two contracts had the same runtime bytecode reported by both explorers. Interestingly, I also found an address that we identify as a contract, and show its runtime bytecode (localhost:8008/v1/emerald/accounts/oasis1qrg90d4qlelg5zg4q4sd4y0z8j2lpjpvuspzjlyl) but blockscout does not (https://explorer.emerald.oasis.dev/address/0x6Cb9750a92643382e020eA9a170AbB83Df05F30B/)

Resolves https://app.clickup.com/t/866adz948


analyzer/evmcontractcode/evm_contract_code.go
```diff
@@ -0,0 +1,198 @@
...
+}
+
+var _ analyzer.Analyzer = (*Main)(nil)
+
+func NewMain(
+	runtime common.Runtime,
+	sourceClient nodeapi.RuntimeApiLite,
+	target storage.TargetStorage,
+	logger *log.Logger,
+) (*Main, error) {
```
ptrus, line 49: Nit: If this is changed to return the `Analyzer` interface `Main` can be unexported as well.

cmd/analyzer/analyzer.go
```diff
@@ -318,6 +320,24 @@ func NewService(cfg *config.AnalysisConfig) (*Service, error) {
 			return evmtokenbalances.NewMain(common.RuntimeSapphire, runtimeMetadata, sourceClient, dbClient, logger)
 		})
 	}
+	if cfg.Analyzers.EmeraldContractCode != nil {
```
ptrus, line 323: Random thought:

Does it ever make sense to not enable the "ContractCode" analyzer for EVM-like runtimes? Could probably always configure the analyzer if `cfg.Analyzers.Emerald != nil`.

Same for `EvmTokens` & `EvmTokenBalances`? Could simplify the configurations a bit this way.

ptrus (commented):
Looks good, left some minor/nit comments.

analyzer/evmcontractcode/evm_contract_code.go
```diff
@@ -0,0 +1,198 @@
...
+// Candidate addresses are inserted into the DB by the block analyzer.
+// Each candidate address only needs to be checked once.
+
+const (
+	EvmContractCodeAnalyzerPrefix = "evm_contract_code_"
+	MaxDownloadBatch              = 20
+	DownloadTimeout               = 61 * time.Second
+)
+
+type EthAddress []byte
```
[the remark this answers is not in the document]
mitjat: I noticed you used that yeah ... this analyzer doesn't interact with go-ethereum in any way, whereas it does interact with our internal sdk where an eth address is simply a naked `[]byte`. So pulling in go-ethereum felt kinda wrong ... but maybe it's nice to have uniformity across the project. I switched the type over now but I'm pretty ambivalent.

analyzer/evmcontractcode/evm_contract_code.go
```diff
@@ -0,0 +1,198 @@
...
+)
+
+// This analyzer checks the list of addresses with an unknown is_contract status,
+// and determines it by calling `getCode()` on the address.
+// If the code is returned, it is also stored in the DB.
+// Every address that is the recipient of a call is a potential contract address.
+// Candidate addresses are inserted into the DB by the block analyzer.
+// Each candidate address only needs to be checked once.
+
+const (
```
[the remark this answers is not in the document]
mitjat: :+1: 
fixed here and in the two analyzers this was cargo-culted from

cmd/analyzer/analyzer.go
mitjat: I've thought about it before too, yes. Why I haven't gone along with it so far:
- I find it handy for local debugging to have fine-grained control
- If these analyzers ever need a configurable parameter, there's won't an obvious way to put them
- It would make it so that entries under the `analyzers` YAML key no longer have a 1:1 mapping to actual analyzers, which is potentially confusing

storage/migrations/05_evm_runtime_bytecode.up.sql
```diff
@@ -0,0 +1,39 @@
+BEGIN;
+
+-- A schema for keeping track of analyzers' internal state/progess.
+-- TODO: Move chain.processed_blocks and chain.*_analysis tables here.
+CREATE SCHEMA IF NOT EXISTS analysis;
```
mitjat, line 5: @pro-wh We discussed DB wiping. This new schema will be covered automatically; we wipe [all schemas except the built-in ones](https://github.com/oasisprotocol/oasis-indexer/blob/435fe753b227b8c8036b881cd8b838321baf6ab6/storage/postgres/client.go#L218-L218)

storage/migrations/05_evm_runtime_bytecode.up.sql
```diff
@@ -0,0 +1,39 @@
...
+CREATE INDEX ix_evm_contract_code_todo ON analysis.evm_contract_code (runtime, contract_candidate) WHERE is_contract IS NULL;
+
+-- Bootstrap the table with the set of addresses we known are contracts because they are the result of an evm.Create tx.
+INSERT INTO analysis.evm_contract_code (runtime, contract_candidate, is_contract)
+  SELECT runtime, contract_address, NULL
+  FROM chain.evm_contracts
+ON CONFLICT (runtime, contract_candidate) DO NOTHING;
+
+ALTER TABLE chain.evm_contracts
+  ADD COLUMN runtime_bytecode BYTEA;
```
mitjat, line 33: Our first real migration! Sucks to have users ;)

storage/migrations/05_evm_runtime_bytecode.up.sql
pro-wh: o good

storage/migrations/05_evm_runtime_bytecode.up.sql
```diff
@@ -0,0 +1,39 @@
...
+-- which talks about a contract's "runtime bytecode" as opposed to "creation bytecode".)
+CREATE TABLE analysis.evm_contract_code (
+    runtime runtime NOT NULL,
+    contract_candidate oasis_addr NOT NULL,
+    PRIMARY KEY (runtime, contract_candidate),
+    -- Meaning of is_contract:
+    --   TRUE:  downloaded runtime bytecode
+    --   FALSE: download failed because `contract_candidate` is not a contract (= does not have code)
+    --   NULL:  not yet attempted
+    is_contract bool
```
Andrew7234, line 21:
```suggestion
    is_contract BOOL
```

Andrew7234 (approved):
Looks good!

analyzer/evmcontractcode/evm_contract_code.go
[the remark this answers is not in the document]
ptrus: Thanks!  Agreed, I do not have strong feelings about it in this case either, would be fine either way.

cmd/analyzer/analyzer.go
ptrus: Good points :+1:

Maybe in future the config could be rafactred a bit so that it would look like:
```yaml
analyzers:
  emerald:
     base: ...
     evm_tokens: ...
     contract_code: ...
  sapphire:
      base: ...
      ...
```

But deferentially a really low priority, probably not even worth it, as it might just cause additional work fixing all the deployments at this time. Maybe if there are ever a lot more runtimes.

config/config.go
```diff
@@ -112,10 +112,12 @@ type AnalyzersList struct {
 	Sapphire  *BlockBasedAnalyzerConfig `koanf:"sapphire"`
 	Cipher    *BlockBasedAnalyzerConfig `koanf:"cipher"`
 
-	EmeraldEvmTokens         *EvmTokensAnalyzerConfig `koanf:"evm_tokens_emerald"`
-	SapphireEvmTokens        *EvmTokensAnalyzerConfig `koanf:"evm_tokens_sapphire"`
-	EmeraldEvmTokenBalances  *EvmTokensAnalyzerConfig `koanf:"evm_token_balances_emerald"`
-	SapphireEvmTokenBalances *EvmTokensAnalyzerConfig `koanf:"evm_token_balances_sapphire"`
+	EmeraldEvmTokens         *EvmTokensAnalyzerConfig       `koanf:"evm_tokens_emerald"`
```
ptrus, line 115: Would it make sense to enable these "evm subanalyzers" in the e2e test config for regressions? I see that only the basic emerald block analyzer is enabled there.

ptrus (approved):

config/config.go
mitjat: Yes, it's strongly on my wishlist. It sucks that we don't have visibility into most changes via e2e regressions, especially now that we're working primarily on non-block analyzers.

It's not just a matter of enabling those analyzers though, there's a gotcha. They need to know when to self-terminate, and they need to be run in a separate phase after the block analyzers. See https://app.clickup.com/t/866aak668

PR #449, opened 2023-06-15
Remove RuntimeSourceStorage abstraction

mitjat:
This PR removes the `RuntimeSourceStorage` interface in favor of directly using the underlying `nodeapi.RuntimeApiLite`.

As a TODO in the code said, this wrapper had no utility any more.

My motivation in removing it was that I needed to introduce one more method to the runtime API, and it sucks to have one more layer of boilerplate to pipe the results through.

This PR does not remove the similar `ConsensusSourceStorage`. It is similarly useless, but it does restructure the data from `ConsensusApiLite` a little instead of just passing it through, which makes it more work to rip out.


mitjat:
Thank you for the review. I accidentally merged PRs out of order, so this change went in as part of #448

PR #454, opened 2023-06-16
e2e_regression: Fix expected output after #452

mitjat:
I realized after #452 went in that its e2e regression test was broken, and that we do not block PRs on that.
This PR fixes the test; the change in expected output is expected :)

I separately changed the repo settings so we now require e2e tests to pass before merging the PR.

The PR also updates run.sh so that diffs show more cleany.


pro-wh (approved):
ok it do pass

PR #459, opened 2023-06-23
analyzer/runtime: enable balance tracking for ERC-721

pro-wh:
the "balance" is the number of the tokens they have, i.e. exactly one moves per Transfer event. this will help us show the number of holders

tested with the same block range as #447


mitjat (approved):
:+1: elegant

PR #460, opened 2023-06-24
 api: allow filter EVM tokens by name

pro-wh:
tested locally on ~500 blocks from Emerald. very few tokens though

### DIFF
--- a/api/spec/v1.yaml
+++ b/api/spec/v1.yaml
@@ -977,6 +977,11 @@ paths:
         - *limit
         - *offset
         - *runtime
+        - in: query
+          name: name
+          schema:
+            type: string
+          description: A filter on the name, the name must contain this value as a substring.
       responses:
         '200':
           description: The requested tokens.
--- a/storage/client/queries/queries.go
+++ b/storage/client/queries/queries.go
@@ -466,10 +466,11 @@ const (
 		WHERE
 			(tokens.runtime = $1) AND
 			($2::oasis_addr IS NULL OR tokens.token_address = $2::oasis_addr) AND
+			($3::text IS NULL OR tokens.token_name ILIKE '%' || $3 || '%' OR tokens.symbol ILIKE '%' || $3 || '%') AND
 			tokens.token_type != 0 -- exclude unknown-type tokens; they're often just contracts that emitted Transfer events but don't expose the token ticker, name, balance etc.
 		ORDER BY num_holders DESC
-		LIMIT $3::bigint
-		OFFSET $4::bigint`
+		LIMIT $4::bigint
+		OFFSET $5::bigint`
 
 	//nolint:gosec // Linter suspects a hardcoded credentials token.
 	EvmTokenHolders = `
--- a/storage/client/client.go
+++ b/storage/client/client.go
@@ -1448,6 +1448,7 @@ func (c *StorageClient) RuntimeTokens(ctx context.Context, p apiTypes.GetRuntime
 		queries.EvmTokens,
 		runtimeFromCtx(ctx),
 		address,
+		p.Name,
 		p.Limit,
 		p.Offset,
 	)


storage/migrations/08_evm_tokens_name_index.up.sql
```diff
@@ -0,0 +1,9 @@
+-- Add a trigram index for evm_tokens token names for search.
+
+BEGIN;
+
+CREATE EXTENSION pg_trgm;
+
+CREATE INDEX ix_evm_tokens ON chain.evm_tokens USING GIST (token_name gist_trgm_ops);
```
pro-wh, line 7: https://www.postgresql.org/docs/current/pgtrgm.html

can someone check if this is right

storage/client/queries/queries.go
```diff
@@ -462,10 +462,11 @@ const (
 		WHERE
 			(tokens.runtime = $1) AND
 			($2::oasis_addr IS NULL OR tokens.token_address = $2::oasis_addr) AND
+			($3::text IS NULL OR tokens.token_name ILIKE '%' || $3 || '%') AND
```
pro-wh, line 465: the name input is not escaped, searches with `%` and `_` will do special things

ptrus: The clickup issue also mentioned filtering by ticker (`symbol` in the db).

pro-wh: oh let me add that

storage/migrations/08_evm_tokens_name_index.up.sql
```diff
@@ -0,0 +1,9 @@
+-- Add a trigram index for evm_tokens token names for search.
+
+BEGIN;
+
+CREATE EXTENSION pg_trgm;
```
pro-wh, line 5: this breaks our wipe db script :skull:

pro-wh: added some code to drop extensions too :skull: :skull:

storage/postgres/client.go
```diff
@@ -250,9 +250,32 @@ func (c *Client) Wipe(ctx context.Context) error {
...
 
+	// list, then drop some extensions.
+	rows, err := c.Query(ctx, `
+		SELECT extname
+		FROM pg_extension
+	`)
+	if err != nil {
+		return fmt.Errorf("failed to list extensions: %w", err)
+	}
+	defer rows.Close()
```
pro-wh, line 261: we use multiple `defer rows.Close()` in this function, reassigning `rows` :skull: :skull: :skull:

storage/postgres/client.go
```diff
@@ -250,9 +250,32 @@ func (c *Client) Wipe(ctx context.Context) error {
...
+	for rows.Next() {
+		var extension string
+		if err = rows.Scan(&extension); err != nil {
+			return err
+		}
+		// This one was here when I got here.
+		if extension == "plpgsql" {
```
pro-wh, line 268: never heard of this extension before, so not touching it??? :skull: :skull: :skull: :skull:

storage/client/queries/queries.go
pro-wh: added same stuff for symbol

pro-wh:
rebased on main

api/spec/v1.yaml
```diff
@@ -977,6 +977,11 @@ paths:
         - *limit
         - *offset
         - *runtime
+        - in: query
+          name: name
+          schema:
+            type: string
+          description: A filter on the name, the name must contain this value as a substring.
```
pro-wh, line 984: ugh this should be updated to say that the symbol can match too

storage/migrations/08_evm_tokens_name_index.up.sql
pro-wh: oh the index name was not good. fixed to be more descriptive

storage/postgres/client.go
mitjat: I'm surprised this is even an extension, technically. It's the de-facto default procedural language for writing functions in postgres. And yes, good call not to remove, I've used it in ad-hoc ways before, and we might use it later.

WDYT about avoiding this `if` and just doing `SELECT ... WHERE extname != 'plpgsql'` instead? It's how most other filtering in Wipe() is done.

storage/postgres/client.go
mitjat: I think `defer` captures enough context that it's all OK? Example: https://go.dev/play/p/6g8xQemALDi

If not, please do fix :pray:

storage/migrations/08_evm_tokens_name_index.up.sql
mitjat: You could also just go with `CREATE EXTENSION IF NOT EXISTS pg_trgm`, and avoid the wiping. I'd prefer that, actually, for simplicity. It's extremely unlikely that we'll want to redefine the extension at some point :)

mitjat: Alternatively, I think it would also not be irresponsible to just omit the index. We have 2k EVM tokens in mainnet emerald. Even at 10k, scanning through their short names should be very fast.

Edit: I tried in prod mainnet now, with and without the index. The index _does_ get used. Make of the time difference what you will:
```sql
                                                                                                                                                                                                                                                                                 [0/21466]
oasisindexer=> explain analyze select count(*) from chain.evm_tokens where token_name ilike '%eth%';                                                                                                                                                                                      
                                                 QUERY PLAN                                                                                  
-------------------------------------------------------------------------------------------------------------                                
 Aggregate  (cost=68.17..68.18 rows=1 width=8) (actual time=1.491..1.493 rows=1 loops=1)                                                     
   ->  Seq Scan on evm_tokens  (cost=0.00..68.04 rows=53 width=0) (actual time=0.016..1.481 rows=58 loops=1)                                 
         Filter: (token_name ~~* '%eth%'::text)                                                                                              
         Rows Removed by Filter: 1865                        
 Planning Time: 0.325 ms                                              
 Execution Time: 1.544 ms         
(6 rows)                                                 
                                                                      
oasisindexer=> CREATE EXTENSION pg_trgm;                                                                                                     
CREATE EXTENSION                                                                                                                                                                                                                                                                          
oasisindexer=>                                                        
oasisindexer=> CREATE INDEX ix_evm_tokens_name ON chain.evm_tokens USING GIST (token_name gist_trgm_ops);                                    
CREATE INDEX                                                                                                                                                                                                                                                                              
oasisindexer=> CREATE INDEX ix_evm_tokens_symbol ON chain.evm_tokens USING GIST (symbol gist_trgm_ops);                                      
CREATE INDEX                             
oasisindexer=> explain analyze select count(*) from chain.evm_tokens where token_name ilike '%eth%';                                                                                                                                                                                      
                                                            QUERY PLAN                                                                       
-----------------------------------------------------------------------------------------------------------------------------------          
 Aggregate  (cost=51.69..51.70 rows=1 width=8) (actual time=0.455..0.456 rows=1 loops=1)                                                                                                                                                                                                  
   ->  Bitmap Heap Scan on evm_tokens  (cost=4.56..51.56 rows=53 width=0) (actual time=0.360..0.447 rows=58 loops=1)                                                                                                                                                                      
         Recheck Cond: (token_name ~~* '%eth%'::text)                                                                                                                                                                                                                                     
         Heap Blocks: exact=30                                 
         ->  Bitmap Index Scan on ix_evm_tokens_name  (cost=0.00..4.54 rows=53 width=0) (actual time=0.344..0.345 rows=58 loops=1)                                                                                                                                                        
               Index Cond: (token_name ~~* '%eth%'::text)
 Planning Time: 0.538 ms                            
 Execution Time: 0.651 ms   
```

mitjat (approved):
Thanks! I have some ideas for simplifying, see what you think.

storage/postgres/client.go
pro-wh: struct, pointer receiver
https://go.dev/play/p/fWui7UAvo7N -> 2 2 :x: 

pointer to struct, pointer receiver
https://go.dev/play/p/FNsKXdCQ7Fs -> 2 1 :heavy_check_mark: 

interface with pointer to struct, pointer receiver
https://go.dev/play/p/Nywd7MMYyvd -> 2 1 :heavy_check_mark: 

interface with struct, struct receiver
https://go.dev/play/p/E4f6lT8zO5o -> 2 1 :heavy_check_mark:

storage/postgres/client.go
pro-wh: oh ok

although let's go with the create extension if not exists

storage/postgres/client.go
pro-wh: this stuff is too... it's like the phrase "sensitive to initial conditions" that you might say of a physical system. I'm just gonna wrap these in iifes.

pro-wh: although I'm going to get rid of this and go with the create extension if not exists way

storage/migrations/08_evm_tokens_name_index.up.sql
pro-wh: on 58 entries?

well I'll move this to a separate PR, let's hold off until it becomes slower

this query is going to run ~on keystroke, so

pro-wh: oh but
```
{"args":[],"caller":"client.go:60","db":"indexer","err":"ERROR: cannot drop type gtrgm because extension pg_trgm requires it (SQLSTATE 2BP01)","level":"error","module":"postgres","msg":"Query","pid":188,"sql":"DROP TYPE public.gtrgm CASCADE;","time":"158.785µs","ts":"2023-06-27T18:11:29.493112027Z"}
```
with the current wipe code

mitjat: We can manually exclude `gtrgm` from the list of types to be wiped?

> on 58 entries?
58 that match `'%eth%'`, but ~2k entries total (`Rows Removed by Filter: 1865`)

pro-wh: then there's a bunch of functions that it depends on

storage/postgres/client.go
pro-wh: hm there's a listIndexerTables in its own method, so I'll go with methods

PR #466, opened 2023-06-27
 storage: add chain.evm_tokens token_name index

pro-wh:
from #460 

This speeds up searching tokens by name. For simplicity, we're not merging this for now (2024-06) until search proves problematically slow.

### DIFF
--- a/.changelog/466.feature.md
+++ b/.changelog/466.feature.md
@@ -1,0 +1,1 @@
+storage: add chain.evm_tokens name and symbol indices
--- a/storage/migrations/01_runtimes.up.sql
+++ b/storage/migrations/01_runtimes.up.sql
@@ -333,6 +333,10 @@ CREATE TABLE chain.evm_tokens
   -- Token analyzer bumps this when it downloads info about the token.
   last_download_round UINT63
 );
+-- Added in 30_evm_tokens_name_index.up.sql.
+-- -- Add indexes for evm_tokens token names and symbols for search.
+-- CREATE INDEX ix_evm_tokens_name ON chain.evm_tokens USING GIST (token_name gist_trgm_ops);
+-- CREATE INDEX ix_evm_tokens_symbol ON chain.evm_tokens USING GIST (symbol gist_trgm_ops);
 
 CREATE TABLE analysis.evm_token_balances
 (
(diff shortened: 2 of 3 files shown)


storage/postgres/client.go
```diff
@@ -309,6 +309,10 @@ func (c *Client) listIndexerMaterializedViews(ctx context.Context) ([]string, er
 
 // Wipe removes all contents of the database.
 func (c *Client) Wipe(ctx context.Context) error {
+	if _, err := c.pool.Exec(ctx, "DROP EXTENSION IF EXISTS pg_trgm CASCADE;"); err != nil {
```
pro-wh, line 312: this works out okay when it's at the beginning, because this way dropping the extension with cascade cleans up its types and functions

storage/postgres/client.go
```diff
@@ -322,6 +326,7 @@ func (c *Client) Wipe(ctx context.Context) error {
 
 	// List, then drop all custom types.
 	// Query from https://stackoverflow.com/questions/3660787/how-to-list-custom-types-using-postgres-information-schema
+	// TODO: Don't delete extensions' types.
```
mitjat, line 329: Is this really a TODO, now that we're dropping the extension?

pro-wh: we can switch course to dropping all extensions (except for some)

mitjat: Oh I see, so IIUC it's a TODO in the sense that if we added another extension, and that extension defined a custom type, and we chose not to use `DROP EXTENSION IF EXISTS` on it, we'd have to amend this code here so that it doesn't delete the extension's types.
In that hypothetical world, the same goes for other objects I guess? I don't know enough about postgres extensions but I imagine they can introduce functions and tables and whatnot.

Maybe this line can be something a little milder? Like `NOTE: If you introduce postgres extensions with their own custom types, this will try to delete those types too.`

I don't see it as a realistic TODO, i.e. I don't expect us to ever need to work on it in a principled way.

mitjat (approved):
Thanks!

I don't see us adding and/or removing extensions very frequently, so I think it's no big deal if there's a little friction in Wipe() around that.

ptrus:
Is this still relevant?

pro-wh:
I think we put it off until the search got slower

mitjat (approved):
Not sure how this fell through the cracks unapproved for so long. Thank you, LGTM! 

Remaining request: I did find the [TODO in the code](https://github.com/oasisprotocol/nexus/pull/466/files#r1244416532) very confusing again (I was reading this with fresh eyes now, a year after the first review), and would strongly prefer to see it reworded the way I suggested in the comment thread, or similarly.

PR #467, opened 2023-06-27
(WIP) Rename `oasis-indexer` to `nexus`

mitjat:
# WORK IN PROGRESS; testing CI

Marketing directive.

Renames:
- all references to github.com paths (notably, Go imports)

Does NOT rename:
- references to the docker image repo. That one remains as oasislabs/oasis-indexer (that's Labs!)
-


.github/workflows/ci-test.yaml
```diff
@@ -46,7 +46,7 @@ jobs:
         run: go install github.com/deepmap/oapi-codegen/cmd/oapi-codegen@v1.12
       - name: Build Go
         run: |
-          make oasis-indexer
+          make nexus
```
Andrew7234, line 49: (no-op) I'd prefer `oasis-nexus` over `nexus` since we refer to it as Oasis Nexus in the docs. It'd also be consistent with Oasis Explorer and how most of our services historically were prefixed with `oasis-`. But since this rename is requested by marketing it's not up to us so I'll just lodge my objection here for posterity and we can leave as-is.

Andrew7234 (approved):

.github/workflows/ci-test.yaml
mitjat: I actually prefer `nexus` here because it's more of a token/identifier, and in that role, we never use `oasis-nexus` (with a hyphen); for consistency and simplicity, I like that we always have one or the other.
We mostly use "Oasis Nexus" in external docs when there is little external context and we want to additional qualify the name. We also use just "Nexus" in externally facing docs plenty of time.

PR #469, opened 2023-06-29
/{runtime}/transactions: Add heuristic field for "is tx a native token transfer?"

mitjat:
Resolves https://app.clickup.com/t/866ahv3h0

We cannot know if an `evm.Call` tx transfers native runtime tokens or not, but we can expose a heuristic that all clients will consume.

This PR adds an `is_likely_native_token_transfer` field to the runtime tx output, with that heuristic.

### DIFF
--- a/api/spec/v1.yaml
+++ b/api/spec/v1.yaml
@@ -2303,6 +2303,16 @@ components:
           type: object
           description: The method call body. May be null if the transaction was malformed.
           example: {"address": "t1mAPucIdVnrYBpJOcLV2nZoOFo=", "data": RBo+cAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAA=", "value": ""}
+        is_likely_native_token_transfer:
+          type: boolean
+          description: |
+            Whether this transaction likely represents a native token transfer.
+            This is based on a heuristic, and can change at any time without warning and possibly without updating the documentation.
+            The current heuristic sets this to `true` for:
+             - Transactions with method "accounts.Transfer". Those are always native token transfers.
+             - Transactions with method "evm.Call" that have no `data` field in their `body`. Those tend to be transfers, but the runtimes provides no reliable visibility into whether a transfer happened.
+            Note: Other transactions with method "evm.Call", and possibly "evm.Create", may also be (or include) native token transfers. The heuristic will be `false` for those.
+          example: true
         to:
           <<: *AddressType
           description: |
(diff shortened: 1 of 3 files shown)


csillag (approved):
As far as I can tell, this makes sense. (But someone with more GO experience should also review this.)

api/spec/v1.yaml
```diff
@@ -2303,6 +2303,16 @@ components:
           type: object
           description: The method call body. May be null if the transaction was malformed.
           example: {"address": "t1mAPucIdVnrYBpJOcLV2nZoOFo=", "data": RBo+cAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAA=", "value": ""}
+        is_likely_native_token_transfer:
```
Andrew7234, line 2306: should we add this field to `required`?

storage/client/client.go
```diff
@@ -1233,16 +1233,30 @@ func (c *StorageClient) RuntimeTransactions(ctx context.Context, p apiTypes.GetR
...
 		}
 		if toPreimageContextIdentifier != nil && toPreimageContextVersion != nil {
 			t.ToEth = EthChecksumAddrFromPreimage(*toPreimageContextIdentifier, *toPreimageContextVersion, toPreimageData)
 		}
 
+		// Heuristically decide if this is a native runtime token transfer.
+		// TODO: Similarly to above, this application logic doesn't belong here (= the DB layer);
+		// move it out if we establish a separate app/logic layer.
+		if t.Method != nil {
+			if *t.Method == "accounts.Transfer" {
```
pro-wh, line 1251: put this on the list of places to think about if a paratime ever uses multiple denominations in the accounts system

pro-wh (approved):
description is very detailed. could fall out of sync with implementation. please keep an eye out for changes to the heuristic

storage/client/client.go
mitjat: Yeah :/, I thought about it.
The alternative is to add the requirement that `body.amount.Denomination == ""`, where `body` is the unstructured JSON column in the DB, and that's a little brittle in its own right. So I just went with the less-code option.

mitjat:
> description is very detailed. could fall out of sync with implementation.

I agree. The description does say "This is based on a heuristic, and can change at any time without warning and possibly without updating the documentation."
I added the detailed description because I feel it also give a bit of a rationale for why this field is even desirable (i.e. evm.Call txs with their murky semantics)

api/spec/v1.yaml
mitjat: Maybe. I do have arguments for keeping it optional, but they are relatively weak: 1) We currently just omit it if it's `false`. 2) It's such an unusual field (in that it's a heuristic, rather than just restructured blockchain contents) that I kinda like keeping the door ajar to removing the field later if desired.

I don't need much of a push to promote it to `required` later.

PR #472, opened 2023-07-02
Update parameter description to better match behavior

csillag:
Looking at the code [here](https://github.com/oasisprotocol/nexus/commit/f29b2db3a6cc1bdb5ddc358de51e52cc34ba5855#diff-c50bc5e7bf70c6477e2cfbc66d3d6130ffcf0dd9060ba20bc63a129edb477b29R469), I think we are returning matches in the token symbol, not only in the token name. (Which is perfect.) I think the description of the parameter should reflect that.

### DIFF
--- a/api/spec/v1.yaml
+++ b/api/spec/v1.yaml
@@ -981,7 +981,7 @@ paths:
           name: name
           schema:
             type: string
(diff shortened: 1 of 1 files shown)


mitjat (approved):
Thanks!

PR #475, opened 2023-07-04
statecheck: select latest block that is processed

ptrus:


Andrew7234 (approved):

pro-wh (approved):
ah yeah we started having null processed_time didn't we

PR #477, opened 2023-07-05
tx_volume stats: Produce windowed results, compute more efficiently

ptrus:
https://app.clickup.com/t/866abpjhd

### Changes:
- tx_volume stats computation logic is now unified with active_accounts, where stats for each window are only computed once and historic data is not recomputed every hour.  
- daily tx_volume data is now available for every 5-minute interval (analogous to active_accounts)

NOTE for reviewers: the `aggregateStatsWorker` which is the main/biggest part of `aggreagte_stats.go` file is more or less the same as the old code used to compute `active_accounts`, just that it now uses the new `statsComputation` struct. No other changes in the computation logic there.

### API changes:

`tx_volume` endpoint:
-  all occurrences of "bucket" (in parameter names, responses) are replaced with "window" - that way the endpoint is more unified with the "active accounts" endpoint (see examples below)

- "bucket_start" in responses is replaced with "window_end". This now indicates the end of the interval, analogous to the `active_accounts` endpoint (see examples below)

- `window_step_seconds` parameter is added to the endpoint. The endpoint now has both `window_step_seconds` and `window_size_seconds`. Both support two values `300` & `86400`. This enables the following query possibilities:

|window_size_seconds|window_step_seconds|result|
|----------|-----------|------------|
|300|300|returns 5min tx_volume intervals for every 5 minutes|
|300|86400|unsupported (doesn't really make sense / isn't all that useful)|
|86400|300| returns daily tx_volume sums ending at every 5 minute interval (overlapping)|
|86400|86400| returns daily tx_volume sums at every midnight (non-overlapping)|

Examples:
```bash
curl "localhost:8008/v1/consensus/stats/tx_volume?window_size_seconds=300&window_step_seconds=300&limit=5"
{"window_size_seconds":300,"windows":[{"tx_volume":368,"window_end":"2022-04-13T03:10:00Z"},{"tx_volume":477,"window_end":"2022-04-13T03:05:00Z"},{"tx_volume":444,"window_end":"2022-04-13T03:00:00Z"},{"tx_volume":485,"window_end":"2022-04-13T02:55:00Z"},{"tx_volume":476,"window_end":"2022-04-13T02:50:00Z"}]}

curl "localhost:8008/v1/consensus/stats/tx_volume?window_size_seconds=86400&window_step_seconds=300&limit=5"
{"window_size_seconds":86400,"windows":[{"tx_volume":143643,"window_end":"2022-04-13T03:05:00Z"},{"tx_volume":143605,"window_end":"2022-04-13T03:00:00Z"},{"tx_volume":143719,"window_end":"2022-04-13T02:55:00Z"},{"tx_volume":143570,"window_end":"2022-04-13T02:50:00Z"},{"tx_volume":143584,"window_end":"2022-04-13T02:45:00Z"}]}

curl "localhost:8008/v1/consensus/stats/tx_volume?window_size_seconds=86400&window_step_seconds=86400&limit=5"
{"window_size_seconds":86400,"windows":[{"tx_volume":143596,"window_end":"2022-04-13T00:00:00Z"},{"tx_volume":66321,"window_end":"2022-04-12T00:00:00Z"}]}

```

### Migration test:
```
before:
select * from stats.daily_tx_volume ;
   layer   |      window_start      | tx_volume
-----------+------------------------+-----------
 consensus | 2022-04-11 00:00:00+00 |     61241
 emerald   | 2022-04-11 00:00:00+00 |     16806
 emerald   | 2022-04-12 00:00:00+00 |      5076

after:
-----------+------------------------+-----------
 consensus | 2022-04-12 00:00:00+00 |     61241
 emerald   | 2022-04-12 00:00:00+00 |     16806
 emerald   | 2022-04-13 00:00:00+00 |      5076
```

```
before:
select * from stats.min5_tx_volume ORDER BY window_start LIMIT 5;
   layer   |      window_start      | tx_volume
-----------+------------------------+-----------
 consensus | 2022-04-11 11:00:00+00 |       105
 consensus | 2022-04-11 11:05:00+00 |         9
 consensus | 2022-04-11 11:10:00+00 |        16
 consensus | 2022-04-11 11:15:00+00 |        10
 consensus | 2022-04-11 11:20:00+00 |         7

after:
select * from stats.min5_tx_volume ORDER BY window_end LIMIT 5;
   layer   |       window_end       | tx_volume
-----------+------------------------+-----------
 consensus | 2022-04-11 11:05:00+00 |       105
 consensus | 2022-04-11 11:10:00+00 |         9
 consensus | 2022-04-11 11:15:00+00 |        16
 consensus | 2022-04-11 11:20:00+00 |        10
 consensus | 2022-04-11 11:25:00+00 |         7
```

```
before:
select * from stats.min5_tx_volume WHERE layer = 'emerald' ORDER BY window_start LIMIT 5;
  layer  |      window_start      | tx_volume
---------+------------------------+-----------
 emerald | 2022-04-11 14:50:00+00 |       112
 emerald | 2022-04-11 14:55:00+00 |       157
 emerald | 2022-04-11 15:00:00+00 |       206
 emerald | 2022-04-11 15:05:00+00 |       218
 emerald | 2022-04-11 15:10:00+00 |       204

after:
select * from stats.min5_tx_volume WHERE layer = 'emerald' ORDER BY window_end LIMIT 5;
  layer  |       window_end       | tx_volume
---------+------------------------+-----------
 emerald | 2022-04-11 14:55:00+00 |       112
 emerald | 2022-04-11 15:00:00+00 |       157
 emerald | 2022-04-11 15:05:00+00 |       206
 emerald | 2022-04-11 15:10:00+00 |       218
 emerald | 2022-04-11 15:15:00+00 |       204

```


analyzer/aggregate_stats/aggregate_stats.go
```diff
@@ -0,0 +1,282 @@
...
+				latestPossibleWindow = floorWindow(latestBlockTs)
+			} else {
+				// Compute the latest available window using the provided custom function.
+				ts, err := fn(ctx, a.target)
+				if err != nil {
+					// Log this as info as this can be expected when there's no indexed blocks yet, or if any of the layers are not being indexed.
+					logger.Info("failed querying the latest possible window, skipping iteration", "err", err)
+					continue
+				}
+				latestPossibleWindow = floorWindow(ts)
```
mitjat, line 209: Look like we can dedupe this block with its equivalent from `fn == nil`? 

Optionally, we could simplify away the whole `if ... customLatestAvailableTs == nil` if we forced every `statsComputation` to specify a fn. It does introduce a little more abstraction though, so your call on what you prefer.

analyzer/aggregate_stats/aggregate_stats.go
```diff
@@ -0,0 +1,282 @@
...
+
+func (a *aggregateStatsAnalyzer) aggregateStatsWorker(ctx context.Context) {
+	statsComputations := []*statsComputation{}
+
+	// Compute 5-minute tx volume stats every 5 minutes for all layers.
+	for _, layer := range statsLayers {
+		statsComputations = append(statsComputations, &statsComputation{
+			name:                "min5_tx_volume_" + layer,
+			layer:               layer,
+			dbTable:             "stats.min5_tx_volume",
```
mitjat, line 104: Can we call this `outputTable` and `outputColumn`? To differentiate from the input side of things (which also uses the DB).

analyzer/aggregate_stats/aggregate_stats.go
```diff
@@ -0,0 +1,282 @@
...
+					logger.Info("failed querying latest indexed block timestamp, skipping iteration", "err", err)
+					continue
+				}
+				latestPossibleWindow = floorWindow(latestBlockTs)
+			} else {
+				// Compute the latest available window using the provided custom function.
+				ts, err := fn(ctx, a.target)
+				if err != nil {
+					// Log this as info as this can be expected when there's no indexed blocks yet, or if any of the layers are not being indexed.
+					logger.Info("failed querying the latest possible window, skipping iteration", "err", err)
```
mitjat, line 206: Might be nice to distinguish between `ErrNoRows` and other errors -- logging with different severities, and different messages. ("failed" looks alarming, even though an ErrNoRows is not really an error here)

mitjat (approved):
Thanks Peter! I keep being dismayed at how much code we need for the measly few stats we support :/ ... But this moves us nicely towards a more generalizable solution.

I only have very minor comments.

analyzer/aggregate_stats/aggregate_stats.go
```diff
@@ -0,0 +1,282 @@
...
+)
+
+// Layers that are tracked for aggregate stats.
+// XXX: Instead of hardcoding, these could also be obtained
+// by (periodically) querying the `runtime_related_transactions` table.
+var statsLayers = []string{
+	layerConsensus,
+	string(common.RuntimeEmerald),
+	string(common.RuntimeSapphire),
+	// RuntimeCipher.String(), // Enable once Cipher is supported by the indexer.
```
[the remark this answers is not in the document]
ptrus: I think there's no harm. Will uncomment.

storage/client/client.go
```diff
@@ -1586,15 +1586,23 @@ func (c *StorageClient) RuntimeStatus(ctx context.Context) (*RuntimeStatus, erro
...
-		query = queries.TxVolumes
+	case *p.WindowSizeSeconds == 86400 && *p.WindowStepSeconds == 86400:
+		// 1 day window, 1 day step.
+		query = queries.DailyTxVolumes
+	case *p.WindowSizeSeconds == 86400 && *p.WindowStepSeconds == 300:
+		// 1 day window, 5 minute step.
+		query = queries.FineDailyTxVolumes
+	default:
+		// Unsupported: case *p.WindowSizeSeconds == 300 && *p.WindowStepSeconds == 86400:
+		return nil, fmt.Errorf("invalid window size parameters: %w", apiCommon.ErrBadRequest)
```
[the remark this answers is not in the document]
ptrus: Yep. I wanted to open a separate PR for that but forgot. Now that this one is already approved, i'll just move the change into a separate commit.

storage/client/queries/queries.go
```diff
@@ -528,22 +528,35 @@ const (
...
 		ORDER BY
-			window_start DESC
+			window_end DESC
+		LIMIT $2::bigint
+		OFFSET $3::bigint
+	`
+
+	// DailyTxVolumes returns the query for daily sampled daily tx volume windows.
+	DailyTxVolumes = `
+		SELECT date_trunc('day', window_end) as window_end, tx_volume
```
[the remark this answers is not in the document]
ptrus: Yeah true :+1:

analyzer/aggregate_stats/aggregate_stats.go
ptrus: Good idea, used `Debug` and reworded the message a bit in case the error is `ErrNoRows`. Also refactored this block a bit to reuse some code.

analyzer/aggregate_stats/aggregate_stats.go
ptrus: Good suggestion, thanks! Updated.

analyzer/aggregate_stats/aggregate_stats.go
ptrus: > It does introduce a little more abstraction though, so your call on what you prefer.

Maybe makes sense, yeah, will see how it looks.

storage/migrations/08_agg_stats_refactor.up.sql
```diff
@@ -0,0 +1,49 @@
+-- Drop stats materialized views and replace with tables.
+-- Also migrate the data from the old materialized views to the new tables.
```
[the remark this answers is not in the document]
ptrus: I think it's nice to support already computed values and since it's pretty straightforward i'd be keen in keeping it in.

> re about greppability:

Agree, it reduces it, but I think we should aim for a general solution to improve the greppability of the DB schema and migrations. One possible approach would be that every so many migrations (e.g. maybe at 10), we would flatten and collapse all migrations into the initial (3? consensus/runtime/stats?) migrations and remove the other migration files.

analyzer/aggregate_stats/aggregate_stats.go
ptrus: Updated this as suggested. Also, after taking another look at this PR after some time has passed, I did some other (pretty minor) code updates that I think increase the readability a bit.

mitjat: Thank you! Since there is only one squashed and rebased commit, I don't think there's a good way for me to see the diff, so I'll take your word for it :)

mitjat:
I think we're good to merge this! Let's just wait for https://github.com/oasisprotocol/nexus/pull/497 to go in first since that is a hotfix. If we wait, it will be easier to cut a release (without special branches etc). We'll need to be a little careful about deploying this particular PR because it breaks the API, so we'll need to coordinate with the frontenders.

storage/migrations/08_agg_stats_refactor.up.sql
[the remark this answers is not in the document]
ptrus: > Whoever wanted to deploy the "flattened" migrations would first have to make sure they've deployed the release just before that so that their DB schema is the newest version

I actually imagined this differently. The "flattened" migrations would work with empty database. And existing deployments would just skip the "flattened" migrations.

ptrus:
> We'll need to be a little careful about deploying this particular PR because it breaks the API, so we'll need to coordinate with the frontenders.

I'll notify frontend about the changes once this is merged, so that they can prepare a PR. Should be simple fixes on frontend.

storage/migrations/08_agg_stats_refactor.up.sql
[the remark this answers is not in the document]
ptrus: Regardless, I decided to drop the migration because I figured that the migration should also handle the case where the last daily window can actually be incomplete. Also didn't test the time it would take to do the migration on staging/prod.

Instead, I updated the main aggregate stats worker code to detect the case where it is catching up on the stats (if the per-iteration batch limit is reached) and in that case use a very short timeout before recomputing the stats again (instead of waiting 1 minute). See last commit.

PR #478, opened 2023-07-06
Re-enable ERC-721 token type in the API specs

csillag:
Support for this token type has been re-enabled by #447 and #459, but the API specification needs to be updated, too.


mitjat (approved):
Thanks!

lukaw3d:
I thought part of go code is generated from api spec. If yes: should it catch this? If no: should it be :D

pro-wh:
it doesn't translate these constraints other than the type

PR #479, opened 2023-07-06
Fix typo on name of token standard in comment

csillag:


pro-wh (approved):
that looks right, thanks

PR #480, opened 2023-07-07
analyzer/evm: handle downloading ERC-721 mutable data

pro-wh:
had only hooked up the method for downloading new tokens. mint/burn will cause us to reconcile the "mutable" total supply, which needs a case here


csillag:
Does this mean that the analyzer will have to be re-executed from the start, in order to collect all the data?

Andrew7234 (approved):

pro-wh:
> Does this mean that the analyzer will have to be re-executed from the start, in order to collect all the data?

no

csillag:
> > Does this mean that the analyzer will have to be re-executed from the start, in order to collect all the data?
> 
> no

Sweet

PR #485, opened 2023-07-13
[analyzer] dead reckon erc721 total supply

Andrew7234:
https://app.clickup.com/t/863h6t2mc

todo:
[] testing (preferably find some similar sequence of blocks to what warren used to test 459)

### DIFF
--- a/analyzer/evmtokens/evm_tokens.go
+++ b/analyzer/evmtokens/evm_tokens.go
@@ -5,6 +5,7 @@ import (
 	"fmt"
 	"time"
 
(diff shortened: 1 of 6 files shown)


ptrus (approved):

Andrew7234:
testing uncovered an issue: mints/burns that were indexed before the evm_tokens analyzer processed the contract were being dropped. I updated the analyzer to store a dead-reckoned totalSupply in the `analysis.evm_tokens` table which is then used by the evm_tokens analyzer as the default totalSupply. 

I'll leave it up for a bit for re-review before merging this in.

PR #486, opened 2023-07-13
pogreb: non-blocking recovery

ptrus:
`pogreb` , the backing KV store for the cache, blocks the entire indexer for potentially multiple hours if it needs to reindex the cache (which happens whenever the database is not closed cleanly).

This PR updates the `KVStore` to open the backing DB in the background. The `OpenKVStore` will block for a maximum of 30 seconds. If the database is not opened within that period, it is likely doing the reindexing, and it will continue to do so in the background. Meanwhile, the store will work and behave as if it had an empty backing DB.

### Testing:
Artificially added timeouts to the pogreb initialization and tested on a local instance. It works as expected, the indexer starts while the pogreb is initializing in the background. And after the db is initialized, the cache is used.

### DIFF
--- a/storage/oasis/nodeapi/file/kvstore.go
+++ b/storage/oasis/nodeapi/file/kvstore.go
@@ -2,6 +2,8 @@ package file
 
 import (
 	"fmt"
+	"sync/atomic"
+	"time"
 
 	"github.com/akrylysov/pogreb"
 	"github.com/oasisprotocol/oasis-core/go/common/cbor"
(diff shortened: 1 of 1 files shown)


storage/oasis/nodeapi/file/kvstore.go
```diff
@@ -27,28 +29,95 @@ type KVStore interface {
...
 	path   string
 	logger *log.Logger
+
+	initialized uint32
 }
 
 var _ KVStore = (*pogrebKVStore)(nil)
 
+func (s *pogrebKVStore) isInitialized() bool {
+	return atomic.LoadUint32(&s.initialized) == 1
```
mitjat, line 43: Is there a scenario where using just plain value assignment and reading (instead of `atomic.*`) would get us into trouble?
If not, I'd slightly prefer avoiding `atomic` because it otherwise misleads the reader into thinking they're missing a complex scenario (which is sucky for refactors/edits later), and the complexity is unlikely to increase.

mitjat (approved):
Thank you!

storage/oasis/nodeapi/file/kvstore.go
ptrus: We need an atomic (or a different form of synchronization, e.g. a lock) since the initialization is done in a separated go-routine. Without it, concurrent reads and writes can lead to undefined behavior. 

I can add a comment to the `initialzied` field to make it clear.

PR #491, opened 2023-07-19
add timestamp and token details to runtime events

Andrew7234:
## Task

**api: add timestamp and the ERC-xx token symbol as one of the fields for each ERC-xx Transfer event**

1) Add timestamp to all runtime events

2) Add the EVM token symbol for events that look like a ERC-20 or ERC-721 Transfer

Edit\*\*: 3) Add the evm token decimals and token type too

Both of these are very useful when rendering events in isolation.

From Slack, regarding token symbols:

> I also don't know how can I determine the ticker / symbol used in token transfers.

Currently you'd have to make another query. In the example above, `pksO2XiYJMvaLgqXEZDsKDR7vE4=` (`data.address`) is the token address. You can then look up the token info.For the above `address`: base64-decoded, it equals `0xa64bed9789824cbda2ea971190ec28347bbc4e`, the oasis-equivalent address is `oasis1qz9k3xk8h59p34svu2l2yusza25dqdpqzcd442hk`, and [https://index.oasislabs.com/v1/sapphire/evm\_tokens/oasis1qz9k3xk8h59p34svu2l2yusza25dqdpqzcd442hk](https://index.oasislabs.com/v1/sapphire/evm_tokens/oasis1qz9k3xk8h59p34svu2l2yusza25dqdpqzcd442hk)  gives you `AKN`.We can add the token symbol as one of the fields for each event. That would probably be helpful?

## Discussion:

@csillag:

> We have already created a workaround (in explorer) which gets the token symbol from external sources (as described above), so having it as part of the event is no longer a priority for us. (This doesn't impact the timestamp, which is still missing.)

@csillag:

> In order to display the data properly, besides the token symbol, we also need to know the decimals. Currently we are getting it from external sources (as part of the separately loaded token info), which works for now. But in order to eliminate the individual round-trips for each list element, we will also need this field, alongside the token symbol.

## This PR

Implements that.

**Open questions:**
- [x] Do we want to display the additional token details for only "Transfer" events? Or would we want to also display this for any event where `body.address` is a known token contract? **Update:** All events

### DIFF
--- a/api/spec/v1.yaml
+++ b/api/spec/v1.yaml
@@ -1345,7 +1345,7 @@ components:
           type: string
           format: date-time
           description: |
-            The second-granular consensus time this tx's block, i.e. roughly when the
+            The second-granular consensus time of this tx's block, i.e. roughly when the
             [block was proposed](https://github.com/tendermint/tendermint/blob/v0.34.x/spec/core/data_structures.md#header).
           example: *iso_timestamp_3
         hash:
@@ -2060,7 +2060,7 @@ components:
 
     RuntimeEvent:
       type: object
-      required: [round, type, body]
+      required: [round, timestamp, type, body]
       properties:
         round:
           type: integer
@@ -2085,6 +2085,12 @@ components:
           description: |
             Ethereum trasnsaction hash of this event's originating transaction.
             Absent if the event did not originate from an EVM transaction.
+        timestamp:
+          type: string
+          format: date-time
+          description: |
+            The second-granular consensus time of this event's block.
+          example: *iso_timestamp_1
         type:
           description: The type of the event.
           $ref: '#/components/schemas/RuntimeEventType'
@@ -2112,6 +2118,8 @@ components:
             The decoded `evm.log` event data. We currently support only two types of evm events, ERC20 `Transfer`
             and `Approve`.
             Absent if the event type is not `evm.log`.
+        evm_token:
+          $ref: '#/components/schemas/EvmEventToken'
       description: An event emitted by the runtime layer
 
     RuntimeEventType:
@@ -2144,6 +2152,25 @@ components:
         Values of EVM type `address` and `address payable` are represented as lowercase hex strings with a "0x" prefix.
         Values of EVM type `bytes` and `bytes<N>` are represented as base64 strings.
         Values of other EVM types (integer types, strings, arrays, etc.) are represented as their JSON counterpart.
+
+    EvmEventToken:
+      type: object
+      properties:
+        type:
+          $ref: '#/components/schemas/EvmTokenType'
+        symbol:
+          type: string
+          description: Symbol of the token, as provided by token contract's `symbol()` method.
+          example: USDT
+        decimals:
+          type: integer
+          description: |
+            The number of least significant digits in base units that should be displayed as
+            decimals when displaying tokens. `tokens = base_units / (10**decimals)`.
+            Affects display only. Often equals 18, to match ETH.
+          example: 18
+      description: |
+        Details about the EVM token involved in the event, if any.
 
     RuntimeEvmContract:
       type: object
--- a/storage/client/queries/queries.go
+++ b/storage/client/queries/queries.go
@@ -394,18 +394,40 @@ const (
 		`
 
 	RuntimeEvents = `
-		SELECT evs.round, evs.tx_index, evs.tx_hash, evs.tx_eth_hash, evs.type, evs.body, evs.evm_log_name, evs.evm_log_params
-			FROM chain.runtime_events as evs
-			WHERE (evs.runtime = $1) AND
-					($2::bigint IS NULL OR evs.round = $2::bigint) AND
-					($3::integer IS NULL OR evs.tx_index = $3::integer) AND
-					($4::text IS NULL OR evs.tx_hash = $4::text OR evs.tx_eth_hash = $4::text) AND
-					($5::text IS NULL OR evs.type = $5::text) AND
-					($6::bytea IS NULL OR evs.evm_log_signature = $6::bytea) AND
-					($7::text IS NULL OR evs.related_accounts @> ARRAY[$7::text])
-			ORDER BY evs.round DESC, evs.tx_index, evs.type, evs.body::text
-			LIMIT $8::bigint
-			OFFSET $9::bigint`
+		SELECT
+			evs.round,
+			evs.tx_index,
+			evs.tx_hash,
+			evs.tx_eth_hash,
+			evs.timestamp,
+			evs.type,
+			evs.body,
+			evs.evm_log_name,
+			evs.evm_log_params,
+			tokens.symbol,
+			CASE -- NOTE: There are three queries that use this CASE via copy-paste; edit both if changing.
+				WHEN tokens.token_type = 20 THEN 'ERC20'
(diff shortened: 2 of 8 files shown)


csillag:
I can't comment on the implementation, but I can confirm that the changes in `api/spec` look good!

storage/migrations/02_runtimes.up.sql
```diff
@@ -111,6 +111,7 @@ CREATE TABLE chain.runtime_events
 
   tx_hash HEX64,
   tx_eth_hash HEX64,
+  timestamp TIMESTAMP WITH TIME ZONE NOT NULL,
```
ptrus, line 114: I think that recently we started doing migrations the "proper" way. e.g. not updating existing migrations and instead creating a new migration that migrates the DB schema.  Since this added field is `NOT NULL` I assume there will also be the need to populate it when migrating live DBs. An added benefit of having a separate migration is that that SQL query for populating the timestamps will be committed and reviewed as well.
 
--- 

But I guess that for such simple cases, it's fine to also just update the old migration and manually do the migration when deploying. Not sure what the stance of the team is on this.

Andrew7234: Good point; I'll add the appropriate migration. Agreed that this can lead to a lot of tiny migrations that are hard to track - a good middle ground might be to always create "proper" migrations during development, but consolidate them into existing migrations prior to any full reindex deployment.

csillag:
> Do we want to display the additional token details for only "Transfer" events? Or would we want to also display this for any event where `body.address` is a known token contract?

So, how does the current implementation behave? What are the conditions for this new data to be present?
Do we still need to keep this field optional?

api/spec/v1.yaml
```diff
@@ -2112,6 +2118,8 @@ components:
             The decoded `evm.log` event data. We currently support only two types of evm events, ERC20 `Transfer`
             and `Approve`.
             Absent if the event type is not `evm.log`.
+        evm_log_details:
```
mitjat, line 2121: I think we can name this better.
1) "event" is the more generic, high-level, preferred term; "log" is specific to the serialized structure that an EVM contract emits. So we have `log_name` and `log_params` because they directly describe/reflect that structure, but `event` elsewhere. So, `evm_event_details`. It also matches the struct name better.
2) The above two fields are already "EVM details" of the event, making it a little weird to have a `details` field with only some of the EVM details included there, seemingly arbitrarily. Since the current `EvmEventDetails` is exclusively about the token involved in the event, I suggest `evm_token`, with a description like `Details about the EVM token involved in the event, if any.` (and the struct can be maybe `EvmEventToken`? Less important.).
3) I wonder if an event can involve multiple tokens, e.g. in the case of DEXes. If so, we should perhaps make this field into an array or map. But looking at an example like https://explorer.emerald.oasis.dev/tx/0x9233f42f9cc487abd92586e9320a48efd1cac097df317100cca701f0702ec4bc/logs makes me think that no, there will typically be a separate event for each token involved. So my vote here would be to keep things as-is (= a single token) and if there ever is an event with multiple tokens, then the frontend (or any other client) will have to issue a separate query to retrieve the details of the 2nd, 3rd etc tokens involved.

TLDR: I suggest we rename this to `evm_token`, and adjust the struct name and its description accordingly.

storage/migrations/02_runtimes.up.sql
mitjat: Yeah short-term we used to be hackier, and now we're trying to be more proper and write actual migrations ... so thank you. Even if not strictly needed, it's "nice", and it also helps us build discipline.

For our own internal use, I'd still be fine with doing the migration manually (because it's low stakes and easy to revert), but now that we've launched more publicly we should ideally operate under the assumption that other people are running their own deployments of the indexer, and we cannot control those deployments and don't want to force a reindex.
That said, the number of external deployments is still very low and possibly zero at this point, so we can be a little pragmatic/flexible and rethink this on a case by case basis. 

**One request though:** Can you add a comment here about the column that you're adding via a separate migration? Like [we did here](https://github.com/oasisprotocol/nexus/blob/774ab49875f16315c9cd6afac855c21e0009fd56/storage/migrations/02_runtimes.up.sql#L224-L224) and in a few other places. It will make future consolidations, if we ever get to that, easier. It also makes it easier to glance the schema of any single table.

mitjat (commented):
Thank you! Looks good, just some small comments.

storage/client/queries/queries.go
```diff
@@ -394,18 +394,37 @@ const (
...
+			tokens.symbol,
+			tokens.token_type,
+			tokens.decimals
+		FROM chain.runtime_events as evs
+		LEFT JOIN chain.address_preimages AS preimages ON
+			DECODE(evs.body ->> 'address', 'base64')=preimages.address_data
+		LEFT JOIN chain.evm_tokens as tokens ON
+			(evs.runtime=tokens.runtime) AND
+			(preimages.address=tokens.token_address) AND
+			(evs.evm_log_name='Transfer')
```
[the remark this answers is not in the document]
Andrew7234: Yep, sounds good!

api/spec/v1.yaml
Andrew7234: Agreed on #1; however I'm leaning slightly more towards `evm_event_details` if/when we start parsing more event types(or contracts, eg DEXes) in the future. In my mind this field is a holding area for any event-related information that isn't included in the event itself. 

OTOH I think `evm_token` would also work for now since we only have token information, and I don't have a clear idea of what other types of contracts we'd eventually support. Maybe DEXes? In any case, I'd prefer to use `evm_event_details` but open to discuss further.

mitjat: Ah, I see you point now. Yeah, valid. It's a judgement call whether `details` is premature generalization (leading to an abstract, harder-to-navigate API) or `token` is us painting ourselves into a corner. I'm still leaning towards `evm_tokens` because
- I can't think of other details either; for better or worse, dapps are still very token-centric
- If other relevant details come up, it's likely they will appear infrequently, and so we'll expect clients to perform separate follow-up queries for those details, rather than perform a slightly expensive JOIN for every `/events` request
- If more details come up that we do want to include in every response, we can still add a new top-level `evm_*` field to contain them.

All of that said, it's not like `_details` is a terrible name, and it's all pretty subjective, so your call, ultimately.

Andrew7234: Makes sense, I still can't think of any other related non-token info that we'd include. Changed to `evm_tokens`.

storage/migrations/02_runtimes.up.sql
Andrew7234: Sure, added!

storage/client/queries/queries.go
```diff
@@ -455,7 +476,7 @@ const (
 			tokens.symbol,
 			tokens.decimals,
 			tokens.total_supply,
-			CASE -- NOTE: There are two queries that use this CASE via copy-paste; edit both if changing.
+			CASE -- NOTE: There are three queries that use this CASE via copy-paste; edit both if changing.
```
mitjat, line 479: :grinning: :see_no_evil:  this is ripe for a tiny refactor; by the book, we should be doing this conversion (which happens for the sake of openapi) in Go anyway. Not in this PR though :)

mitjat (approved):
Thank you!

PR #492, opened 2023-07-20
[api] return evm address instead of oasis addr in RuntimeEvmBalance

Andrew7234:
According to the [openapi spec](https://github.com/oasisprotocol/nexus/blob/main/api/spec/v1.yaml#L1748), the token_contract_address in `RuntimeEvmBalance` should be the EVM address and not the oasis address. 

Need to confirm with explorer team that the currently returned oasis address can be replaced by the evm address. If not, we can return both the oasis and evm addresses (though the oasis address can be derived from the eth address).

### DIFF
--- a/api/spec/v1.yaml
+++ b/api/spec/v1.yaml
@@ -1740,15 +1740,19 @@ components:
     RuntimeEvmBalance:
       description: Balance of an account for a specific runtime and EVM token.
       type: object
-      required: [balance, token_contract_addr, token_decimals, token_type]
+      required: [balance, token_contract_addr, token_contract_addr_eth, token_decimals, token_type]
       properties:
         balance:
           <<: *BigIntType
           description: Number of tokens held, in base units.
         token_contract_addr:
           type: string
-          description: The EVM address of this token's contract. Encoded as a lowercase hex string.
-          example: 'dc19a122e268128b5ee20366299fc7b5b199c8e3'
+          description: The oasis address of this token's contract.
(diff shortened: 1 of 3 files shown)


ptrus (approved):

csillag:
> Need to confirm with explorer team that the currently returned oasis address can be replaced by the evm address. I

If we do that, it's a breaking change, which would need synchronized Nexus and Explorer releases to avoid breakage.

I think adding `token_contract_addr_eth` besides the unchanged `token_contract_addr` would provide a much smoother upgrade path.

Also, we have a switch on Explorer UI (well, at least for some parts) that supports choosing displaying Oasis and ETH addresses, according to the user's preferences. So having the Oasis address ready is definitely a plus.

csillag:
Thanks, I like this version much better!

csillag:
This will be used by https://github.com/oasisprotocol/explorer/pull/762.

api/spec/v1.yaml
```diff
@@ -1740,15 +1740,19 @@ components:
...
         balance:
           <<: *BigIntType
           description: Number of tokens held, in base units.
         token_contract_addr:
+          type: string
+          description: The oasis address of this token's contract.
+          example: 'oasis1qzk5pr2x8ah04lgjee3lv06fmyvqvz45egjv4ps0'
+        token_contract_addr_eth:
           type: string
           description: The EVM address of this token's contract. Encoded as a lowercase hex string.
```
mitjat, line 1754:
```suggestion
          description: The EVM address of this token's contract. 
```

Thumbs up for proper stringification.

mitjat (approved):
Thank you for this patch, and all the others!

PR #496, opened 2023-08-02
Andrew7234/status info

Andrew7234:
https://app.clickup.com/t/8669zx20d

The `Status` now includes the current height of the node as seen by the analyzer. We fetch this in a background goroutine in the consensus analyzer. Note that if the analyzer crashes, Nexus will have an out-of-date view of what the latest on-chain block is. However, `latest_update` will indicate that Nexus has not processed updates and is behind.


Andrew7234:
hmm the e2e api tests break because we now call `getLatestBlock`...

analyzer/consensus/consensus.go
```diff
@@ -175,6 +180,52 @@ func (m *processor) processGenesis(ctx context.Context, chainContext string) err
...
+		requestCtx       context.Context
+		requestCtxCancel context.CancelFunc = func() {}
+	)
+	for {
+		requestCtxCancel()
+		select {
+		case <-ctx.Done():
+			m.logger.Warn("shutting down node height goroutine", "reason", ctx.Err())
+			return
+		case <-time.After(3 * time.Second):
```
Andrew7234, line 195: we don't use the backoff here because it defaults to running as fast as possible.

analyzer/consensus/consensus.go
```diff
@@ -80,6 +82,9 @@ func NewAnalyzer(cfg *config.BlockBasedAnalyzerConfig, genesisChainContext strin
 
 // Implements BlockProcessor interface.
 func (m *processor) PreWork(ctx context.Context) error {
+	// Start background goroutine to fetch current node height.
+	go m.fetchNodeHeights(ctx)
```
ptrus, line 86: Does it make more sense to start this later, e.g. at the end of `PreWork` after genesis is processed?

storage/migrations/12_analysis_node_height.up.sql
```diff
@@ -0,0 +1,14 @@
+BEGIN;
+
+-- Tracks the current (consensus) height of the node. TODO: expand to paratime heights.
+CREATE TABLE chain.node_heights
+(
+  analyzer TEXT NOT NULL PRIMARY KEY,
```
ptrus, line 6: I think it makes more sense this being `layer` instead of `analyzer`? As there can be multiple consensus related (or runtime related) analyzers, but the latest node_height depends on the layer.

storage/migrations/12_analysis_node_height.up.sql
```diff
@@ -0,0 +1,14 @@
+BEGIN;
+
+-- Tracks the current (consensus) height of the node. TODO: expand to paratime heights.
+CREATE TABLE chain.node_heights
```
ptrus, line 4: `chain.latest_node_heights` to make it clearer that this only tracks the latest node heights, not all heights.

analyzer/consensus/consensus.go
```diff
@@ -80,6 +82,9 @@ func NewAnalyzer(cfg *config.BlockBasedAnalyzerConfig, genesisChainContext strin
 
 // Implements BlockProcessor interface.
 func (m *processor) PreWork(ctx context.Context) error {
+	// Start background goroutine to fetch current node height.
```
ptrus, line 85: If/when this is expended to include other layers it should (as the comment on the DB table states) likely be moved into the `blockBasedAnalyzer` or maybe even into a new, `nodeStats` analyzer, to avoid duplication. 

Maybe just include a short TODO here.

storage/client/client.go
```diff
@@ -176,23 +176,31 @@ func (c *StorageClient) withTotalCount(ctx context.Context, sql string, args ...
 // Status returns status information for Oasis Nexus.
 func (c *StorageClient) Status(ctx context.Context) (*Status, error) {
 	var s Status
+	var latest_block_update time.Time
```
ptrus, line 179: Use camel case

analyzer/consensus/consensus.go
mitjat: +1 to moving this out of the consensus analyzer. It has little or nothing to do with the analysis of consensus blocks.

[Here](https://github.com/oasisprotocol/nexus/blob/d937c9bc9ca31ac8d39e521b84696863fc35e55d/analyzer/evmcontractcode/evm_contract_code.go) is a recent minimal analyzer. (Aside: The code can be made even shorter with https://app.clickup.com/t/866afuw9w, but we're not working on that yet.) It should be relatively easy to adapt it to the needs of this PR.

I see now that the ticket suggests to implement this as a goroutine in the consensus analyzer, as you did. Sorry that was misleading; the ticket is outdated. Apart from the aesthetic/architectural considerations (first paragraph above), there is a recent practical consideration that makes a simple goroutine less palatable: With fast-sync and parallel analysis, we'll be running potentially multiple consensus analyzers at once. Although maybe in fast sync we'd want to disable this goroutine altogether, IDK.

analyzer/consensus/consensus.go
Andrew7234: moved to separate analyzer!

storage/migrations/12_analysis_node_height.up.sql
Andrew7234: Good point; updated!

analyzer/consensus/consensus.go
Andrew7234: got it, moved to a separate analyzer!

storage/migrations/13_analysis_node_height.up.sql
```diff
@@ -0,0 +1,16 @@
+BEGIN;
+
+-- Tracks the current (consensus) height of the node. TODO: expand to paratime heights.
+CREATE TABLE chain.latest_node_heights
+(
+  layer TEXT NOT NULL PRIMARY KEY,
+  height UINT63 NOT NULL
+);
+
+INSERT INTO chain.latest_node_heights (layer, height) VALUES ('consensus', 0);
```
Andrew7234, line 10: This is a temporary kludge. The larger issue here is that we currently do not have a way to cleanly test interval-based analyzers in the e2e tests because they run indefinitely and the e2e_test waits for all analyzers to complete. Planning to work on this in a follow-up PR so that we can test more endpoints and also remove this line. 

Note: The `node_stats` analyzer is disabled in the e2e tests for now, but I'm leaving the changes to add the `cache_volatile_endpoints` flag because it will be used in the aforementioned changes to support all analyzers in the e2e regression tests.

analyzer/consensus/consensus.go
Andrew7234: ok now it's configurable but defaults to 3 seconds

storage/migrations/13_analysis_node_height.up.sql
Andrew7234: Filed a clickup issue for it: https://app.clickup.com/t/8692jqubb

storage/oasis/nodeapi/file/consensus.go
```diff
@@ -16,14 +16,20 @@ import (
...
+// If cacheVolatileEndpoints is `true`, FileConsensusApiLite will cache responses
+// from volatile endpoints and return the cached data on subsequent queries.
+// This means that the data fed to Nexus will always be consistent, but may be outdated,
+// which is acceptable in testing scenarios.
 type FileConsensusApiLite struct {
-	db           KVStore
-	consensusApi nodeapi.ConsensusApiLite
+	db                     KVStore
+	consensusApi           nodeapi.ConsensusApiLite
+	cacheVolatileEndpoints bool
```
mitjat, line 27: Uhh ... I don't have a specific scenario to point to, but I expect pain and suffering lay down this path. At best, every time we recreate the e2e cache we'll have to manually decide which changes in outputs are expected. If that happens often, it's annoying; if it happens rarely, it's surprising.
Implementing the flag also requires quite a few little changes in the codebase, which is a smell given that the flag is used for testing only.

We'll have a more stable situation if we just ignore the (very few) volatile fields in e2e tests. Example for the already-volatile `latest_update` field is here: https://github.com/oasisprotocol/nexus/blob/5cdf94ebc0cfa704a5f0efb4d91a2655446ab327/tests/e2e_regression/run.sh#L114-L114
It does slightly decrease the power of the test, but not much.

mitjat (commented):
Thanks Andy! Looks good overall, but I do have some concerns about API stability, and about robustness of e2e tests. See below.

storage/oasis/nodeapi/file/consensus.go
Andrew7234: > every time we recreate the e2e cache we'll have to manually decide which changes in outputs are expected

Yeah, but I'm hoping that this is a rare enough occurrence that it'll still be palatable. If recent memory serves, I haven't had to do it too often 

I think overall the tradeoff is worth it since we can start including other analyzers (like evm_tokens + evm_contracts) that depend on volatile endpoints in the e2e regression tests. But open to other suggestions as well that might be more elegant 😅

mitjat: Sorry, I still strongly think that caching volatile endpoints is asking for trouble, and we shouldn't do it.

It's easy enough enough to contaminate your local cache (or, god forbid, a production cache) by using the wrong flag accidentally, during testing or such. A single short run suffices, and there's no undoing it (unless we write new tools.) Once the cache is contaminated, there will be no indication of it, and no immediate failure, just hard-to-explain incorrect results that we discover possibly days later. As a similar pattern: Remember how we had false bug reports that stemmed from the incorrect caching of genesis documents (cobalt and damask were not distinguishable). The root case was quite non-obvious because the cache is so opaque.

Cache invalidation is a famously hard problem that's prone to bugs. We have the luxury of being able to have a cache that never goes stale, if we only make small accommodations elsewhere. Let's hold onto that good fortune.

To your other point: Adding other analyzers to e2e will require some thought/hacks/compromises, yes. But I actually don't expect trouble from the analyzers that pull data from the node (like `evm_tokens` and `evm_contract_code`). They fetch data from at the latest block-analyzer height, not at the latest node height; this was implemented that way partially with e2e tests in mind. So the analyzers will be hitting non-volatile endpoints. (*)
In e2e, the plan is to run those analyzers _after_ the block analyzers finish; this way, they will always fetch data at the same height. If we ran them in parallel with block analyzers, they'd still be hitting non-volatile endpoints, but at unpredictable (and thus uncacheable) heights. 

Non-node analyzers are the bigger problem: Sourcify, CoingGecko, metadata registry. At first, we'll just not include them in e2e tests. Then later we can think of mocking those endpoints (and the mock can even be cache-based).

(*) The special analyzer from this PR is a necessary exception, but is likely to remain a one-off.

Andrew7234: > easy enough enough to contaminate your local cache (or, god forbid, a production cache) 

Good points, agreed that it's not worth the risk of that happening. Updated accordingly

> They fetch data from at the latest block-analyzer height,

Oh nice, this will make expanding the tests way easier

mitjat (commented):
Thank you for the PR and all the follow-ups on it.
This LGTM, pending either FE approval or a change such that the API stays backwards compatible.

config/sapphire-local-dev.yml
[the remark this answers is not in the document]
Andrew7234: oops didn't mean to check this in. removed

analyzer/consensus/consensus.go
mitjat: Hm I just realized we can generate a response where `latest_indexed_block > latest_block`, because the block analyzer and this fetcher (regardless of whether it lives in its own analyzer) fetch data independently.

Maybe we can work around that by generating the `latest_indexed_block` number as
```sql
SELECT GREATER(
  (SELECT ... FROM chain.latest_node_heights),
  (SELECT ... FROM analysis.processed_blocks WHERE processed_time IS NOT NULL)
)
```
?

mitjat (approved):
Thank you!
Approving under the assumption that @lukaw3d 's response in Slack is tacit approval from frontend of us breaking the API (see unresolved threads). I'd also appreciate if you coordinate a release+deploy with them.

Alternatively, you can tweak the API to be backwards compatible :)

api/spec/v1.yaml
```diff
@@ -1153,23 +1155,29 @@ components:
...
+          format: int64
+          description: The height of the most recently produced block on-chain as seen by Nexus.
+          example: *block_height_2
         latest_block_time:
           type: string
           format: date-time
-          description: The RFC 3339 formatted consensus time of when the most recent block was produced.
+          description: The RFC 3339 formatted consensus time of when the most recently indexed block was produced.
           example: *iso_timestamp_2
         latest_update:
```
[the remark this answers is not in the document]
lukaw3d: Oh, you tagged latest_update, not latest_block_time. Yep yep

analyzer/consensus/consensus.go
Andrew7234: Good idea; implemented this in the go client instead of the db query though to keep the queries straightforward.

api/spec/v1.yaml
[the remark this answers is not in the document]
Andrew7234: Yep the only 'breaking change' now is in `latest_update`.

storage/client/client.go
```diff
@@ -215,6 +215,9 @@ func (c *StorageClient) Status(ctx context.Context) (*Status, error) {
 	).Scan(&s.LatestNodeBlock)
 	switch err {
 	case nil:
+		if s.LatestNodeBlock < s.LatestBlock {
```
mitjat, line 218: Nit: Maybe a comment since it's not obvious why this is needed?
```
// Current node height is fetched in a separate goroutine, so it's possible for it
// to be more out of date than the height of the most recently processed block.
```

api/spec/v1.yaml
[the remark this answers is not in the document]
mitjat: Oy vey, we had so much back and forth around the semantics of this field and others that I missed the _format_ of this field. Valid values seem to include `'72h3m0.5s'` and `'370.318844ms'`. That's ... machine-readable, I guess, but needlessly a PITA to parse and weirdly human-formatted for an API. And technically, nothing prevents Go from changing this human-oriented `String()` representation across Go versions (e.g. add `d` for days).
I suggest we sneak in another quick change where this becomes a float (or int) representing the (approx) number of seconds? Sorry for the late comment.

PR #497, opened 2023-08-07
bugfix: initialize possibleToken fields

Andrew7234:
The analyzer crashed with 

```
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x0 pc=0x64bc9d]

goroutine 127299 [running]:
math/big.(*Int).Add(0x0, 0x1?, 0xc0000ab2f0?)
	/usr/local/go/src/math/big/int.go:116 +0x1d
github.com/oasisprotocol/nexus/analyzer/runtime.extractEvents.func4.3({0xc000644b0c, 0x14, 0x14}, {0xc000644b2c, 0x14, 0x14}, {0xc000644b40?, 0xc000602858?, 0x69fab4?})
	/code/go/analyzer/runtime/extract.go:764 +0x5e5
github.com/oasisprotocol/nexus/analyzer/runtime.VisitEVMEvent(0xc0000a5b30, 0xc000602e08)
	/code/go/analyzer/runtime/visitors.go:215 +0x7b9
github.com/oasisprotocol/nexus/analyzer/runtime.extractEvents.func4(0xc0000a5b30)
	/code/go/analyzer/runtime/extract.go:633 +0x456
github.com/oasisprotocol/nexus/analyzer/runtime.VisitSdkEvent(0x8?, 0xc000603220)
	/code/go/analyzer/runtime/visitors.go:148 +0x531
github.com/oasisprotocol/nexus/analyzer/runtime.VisitSdkEvents({0xc000456e00, 0x2, 0x0?}, 0xc000603218?)
	/code/go/analyzer/runtime/visitors.go:158 +0x68
github.com/oasisprotocol/nexus/analyzer/runtime.extractEvents(0xc00054be48?, 0xc000311900?, {0xc000456e00?, 0x2?, 0x6?})
	/code/go/analyzer/runtime/extract.go:529 +0x135
github.com/oasisprotocol/nexus/analyzer/runtime.ExtractRound({0x0, {0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, ...}, ...}, ...)
	/code/go/analyzer/runtime/extract.go:472 +0x13e7
github.com/oasisprotocol/nexus/analyzer/runtime.(*processor).ProcessBlock(0xc000526230, {0x18e98c0, 0xc000b0c1e0}, 0xc00048e700?)
	/code/go/analyzer/runtime/runtime.go:116 +0x259
github.com/oasisprotocol/nexus/analyzer/block.(*blockBasedAnalyzer).Start(0xc0005262d0, {0x18e9850, 0xc00030f7c0})
	/code/go/analyzer/block/block.go:240 +0x951
github.com/oasisprotocol/nexus/cmd/analyzer.(*Service).Start.func1({0x18e3318?, 0xc0005262d0?})
	/code/go/cmd/analyzer/analyzer.go:387 +0x71
created by github.com/oasisprotocol/nexus/cmd/analyzer.(*Service).Start
```

The root cause was that the totalSupplyChange could be nil if a previously processed ERC721Approval or ERC721ApprovalForAll event in the block initialized `possibleToken.TotalSupplyChange` to `nil`. 

Todo:
Might be better to change `possibleToken.TotalSupplyChange` to not be a pointer since we [default to 0 anyway](https://github.com/oasisprotocol/nexus/blob/803ca11501cfbf0c75917afc6ee620f4e228d9c6/analyzer/runtime/runtime.go#L289)... but big.Int uses pointer receiver methods.
```
type EVMPossibleToken struct {
	Mutated           bool
	TotalSupplyChange big.Int
}
```

Testing... wasn't able to connect to a mainnet sapphire grpc endpoint yet locally (only testnet).

### DIFF
--- a/analyzer/runtime/extract.go
+++ b/analyzer/runtime/extract.go
@@ -808,7 +808,9 @@ func extractEvents(blockData *BlockData, relatedAccountAddresses map[apiTypes.Ad
 						eventData.RelatedAddresses[approvedAddr] = true
 					}
 					if _, ok := blockData.PossibleTokens[eventAddr]; !ok {
-						blockData.PossibleTokens[eventAddr] = &evm.EVMPossibleToken{}
+						blockData.PossibleTokens[eventAddr] = &evm.EVMPossibleToken{
+							TotalSupplyChange: &big.Int{},
+						}
 					}
 					tokenID := &big.Int{}
 					tokenID.SetBytes(tokenIDU256)
@@ -851,7 +853,9 @@ func extractEvents(blockData *BlockData, relatedAccountAddresses map[apiTypes.Ad
 						eventData.RelatedAddresses[operatorAddr] = true
 					}
 					if _, ok := blockData.PossibleTokens[eventAddr]; !ok {
-						blockData.PossibleTokens[eventAddr] = &evm.EVMPossibleToken{}
+						blockData.PossibleTokens[eventAddr] = &evm.EVMPossibleToken{
+							TotalSupplyChange: &big.Int{},
+						}
 					}
 					approved := approvedBool[31] != 0
 					eventData.EvmLogName = evmabi.ERC721.Events["ApprovalForAll"].Name
--- a/analyzer/runtime/extract.go
+++ b/analyzer/runtime/extract.go
@@ -751,9 +751,7 @@ func extractEvents(blockData *BlockData, relatedAccountAddresses map[apiTypes.Ad
 					}
 					// TODO: Reckon ownership.
 					if _, ok := blockData.PossibleTokens[eventAddr]; !ok {
-						blockData.PossibleTokens[eventAddr] = &evm.EVMPossibleToken{
-							TotalSupplyChange: &big.Int{},
-						}
+						blockData.PossibleTokens[eventAddr] = &evm.EVMPossibleToken{}
 					}
 					// Mark as mutated if transfer is between zero address
 					// and nonzero address (either direction) and nonzero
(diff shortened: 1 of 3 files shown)


analyzer/runtime/extract.go
```diff
@@ -808,7 +808,9 @@ func extractEvents(blockData *BlockData, relatedAccountAddresses map[apiTypes.Ad
 						eventData.RelatedAddresses[approvedAddr] = true
 					}
 					if _, ok := blockData.PossibleTokens[eventAddr]; !ok {
-						blockData.PossibleTokens[eventAddr] = &evm.EVMPossibleToken{}
+						blockData.PossibleTokens[eventAddr] = &evm.EVMPossibleToken{
+							TotalSupplyChange: &big.Int{},
```
mitjat, line 812: This creates/encourages long-range dependencies: All the code that initializes values in `PossibleTokens`, present and future, needs to be aware of [the code that expects `.TotalSupplyChange` to be non-nil](https://github.com/oasisprotocol/nexus/blob/43a480c419552aaf6d6499139e3b79f6fda54e22/analyzer/runtime/extract.go#L764-L764). We're bound to mess up again, sooner or later.

I suggest one of these alternative fixes:
- Make the code linked above more robust, i.e. make it initialize `.TotalSupplyChange` if it's `nil`.
- Change the type of `.TotalSupplyChange` so it's `big.Int`, not `*big.Int`. This will prevent future nilness problems for good.

mitjat (commented):
Thank you for the quick fix, Andy! I have a suggestion for making it more future-proof.

analyzer/runtime/extract.go
Andrew7234: > Change the type of .TotalSupplyChange so it's big.Int, not *big.Int

Slightly prefer this because it simplifies the other parts as well; going with this

analyzer/runtime/runtime.go
```diff
@@ -286,10 +286,7 @@ func (m *processor) queueDbUpdates(batch *storage.QueryBatch, data *BlockData) {
 
 	// Insert EVM token addresses.
 	for addr, possibleToken := range data.PossibleTokens {
-		totalSupply := "0"
-		if possibleToken.TotalSupplyChange != nil {
-			totalSupply = possibleToken.TotalSupplyChange.String()
-		}
+		totalSupply := possibleToken.TotalSupplyChange.String()
```
mitjat, line 289: `totalSupplyChange` please, or `totalSupplyChangeStr`. Or pretty much anything that has `change` or `delta` in the name :)

analyzer/runtime/extract.go
mitjat: Thanks Andy.

:facepalm: to the `big.Int` API. I guess give how/where we use `totalSupplyChange`, we could have gone with `common.BigInt`, where API is either more sane or we're free to make it more sane. Oh well, enough bikeshedding I guess.

mitjat (approved):

analyzer/runtime/extract.go
Andrew7234: > big.Int API

TIL that even though `big.Int`'s [methods are pointer receiver](https://pkg.go.dev/math/big#Int.Add), it doesn't matter if it's in a struct field like `totalSupply big.Int` ¯\_(ツ)_/¯

PR #498, opened 2023-08-07
add index for chain.addres_preimages.address_data

Andrew7234:
@lukaw3d pointed out that the `/events` endpoint was slow. Root cause was the left join on `address_preimages` in order to look up token-specific information for events.

### DIFF
--- a/storage/client/queries/queries.go
+++ b/storage/client/queries/queries.go
@@ -413,7 +413,9 @@ const (
 			tokens.decimals
 		FROM chain.runtime_events as evs
 		LEFT JOIN chain.address_preimages AS preimages ON
-			DECODE(evs.body ->> 'address', 'base64')=preimages.address_data
+			DECODE(evs.body ->> 'address', 'base64')=preimages.address_data AND
+			preimages.context_identifier = 'oasis-runtime-sdk/address: secp256k1eth' AND
+			preimages.context_version = 0
 		LEFT JOIN chain.evm_tokens as tokens ON
 			(evs.runtime=tokens.runtime) AND
 			(preimages.address=tokens.token_address)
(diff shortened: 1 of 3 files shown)


pro-wh (approved):
I always approve these sort of changes that describe what's already done

:smiling_face_with_tear: here's us using a database table to compute a hash

storage/client/queries/queries.go
```diff
@@ -413,7 +413,9 @@ const (
 			tokens.decimals
 		FROM chain.runtime_events as evs
 		LEFT JOIN chain.address_preimages AS preimages ON
```
mitjat, line 415: Maybe a comment like this to prevent future confusion?

```suggestion
		-- Look up the oasis-style address derived from evs.body.address.
		-- The derivation is just a keccak hash and we could theoretically compute it instead of looking it up,
		-- but the implementing/importing the right hash function in postgres would take some work.
		LEFT JOIN chain.address_preimages AS preimages ON
```

mitjat (approved):
Thank you!

Just stating what we've discussed between the lines: The cleaner alternative would be to compute the hash in postgres.

My understanding is that the hash in question is keccak-256 [here](https://github.com/oasisprotocol/oasis-sdk/blob/c823194ed60ad7a922812e6b803d8ff8b9f78c45/client-sdk/go/types/address.go#L139-L139). It is potentially accessible in postgres via the no-root-required `pgcrypto` [extension](https://www.postgresql.org/docs/current/pgcrypto.html), which "supports all digest algorithms supported by OpenSSL". OpenSSL [added support](https://github.com/openssl/openssl/issues/13033) for keccak-256 back in Sept 2020, but at least my local install doesn't support it yet (`openssl list -digest-algorithms` and `openssl version`), nor does the one in our docker images. So this relatively ugly index is probably the right call :)

storage/client/queries/queries.go
Andrew7234: thanks; added this and updated other documentation as well

PR #501, opened 2023-08-15
Mark charged_fee as required

lukaw3d:
Is it ever missing?


Andrew7234 (approved):
It's [always set](https://github.com/oasisprotocol/nexus/blob/main/storage/client/queries/queries.go#L345), so lgtm

PR #502, opened 2023-08-15
nodeapi: Make CommitteeKind compatbile across cobalt and damask

mitjat:
Bugfix for accessing cobalt state. I don't know how we didn't hit this problem before.

**Problem**
The CommitteeKind type (which is used inside committee updates at the consensus layer) is an enum, with **three** possible values in cobalt, but only **two** possible values in damask. Before this PR, we were using the damask type. So when trying to serialize an enum value that comes from cobalt and is valid in cobalt but not in damask, an error was raised.

**Solution**
This PR creates a new type that's independent of cobalt and damask (even though it's technically just copy-paste from cobalt), but is compatible with both. 
No data changes are needed (caches, db, etc). The new type is byte-compatible with the type we've been using. Just the serde methods are able to accept the extra enum value.

**Testing**
Not much because the change is minimal. Previously, some blocks just before damask failed to get indexed (because the CommitteeKind value gets serialized for insertion into the DB). After this PR, those blocks can be indexed.

### DIFF
--- a/storage/oasis/nodeapi/api.go
+++ b/storage/oasis/nodeapi/api.go
@@ -190,7 +190,12 @@ type (
 
 type (
 	Validator scheduler.Validator
-	Committee scheduler.Committee
+	Committee struct {
+		Kind      CommitteeKind
+		Members   []*scheduler.CommitteeNode
+		RuntimeID coreCommon.Namespace
+		ValidFor  beacon.EpochTime
+	}
 )
 
 // .................... Governance ....................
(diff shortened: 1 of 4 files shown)


pro-wh (approved):
explanation makes sense.

that we didn't encounter this before: had we not reached the point on chain where paratimes are registered? I remember there being a complaint that the explorer showing zero compute nodes for paratimes because the consensus analyzer had not caught up to when the paratime first had a committee elected

mitjat:
> consensus analyzer had not caught up to when the paratime first had a committee elected

IIRC that was only after the accidental cache deletion + starting full reindex. But maybe.

PR #504, opened 2023-08-16
Andrew7234/evm token transfers

Andrew7234:
https://app.clickup.com/t/8692hqtf5

Adds support for dead-reckoning `evmToken.NumTransfers` without using triggers.

The accounting in this is almost identical to how we track `evmToken.TotalSupply`, in that we maintain two dead-reckoned sums in `analysis.evmTokens` and `chain.evmTokens` for possible and confirmed evm tokens, respectively. Once the evmTokens analyzer downloads a token, it copies the value from `analysis` to `chain`.

### DIFF
--- a/analyzer/evmtokens/evm_tokens.go
+++ b/analyzer/evmtokens/evm_tokens.go
@@ -70,6 +70,7 @@ type StaleToken struct {
 	Addr                  string
 	LastDownloadRound     *uint64
 	TotalSupply           common.BigInt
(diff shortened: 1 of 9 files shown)


mitjat (approved):
Thank you, looks good!
It's unfortunate that we need the kludge of tracking state in both `analysis` and `chain` `evm_tokens` tables. As discussed in #456, I'll try to make away with that soon.

PR #505, opened 2023-08-21
Andrew7234/account num txs

Andrew7234:
https://app.clickup.com/t/863h5r3dt

tldr: the `/{runtime}/accounts/{addr}` endpoint is slow because we recalculate the account_stats.num_txs on the fly (scanning entire `runtime_related_transactions` table) for every request. 
```
InitPlan 3 (returns $2)
     ->  Aggregate  (cost=393.60..393.61 rows=1 width=8) (actual time=2636.657..2636.658 rows=1 loops=1)
           ->  Index Only Scan using ix_runtime_related_transactions_address on runtime_related_transactions  (cost=0.57..387.13 rows=2586 width=0) (actual time=5.619..2611.100 rows=235330 loops=1)
                 Index Cond: ((runtime = 'sapphire'::runtime) AND (account_address = 'oasis1qru38k694sl27t399kgn5dptdj2atn3u6v4qvkyw'::text))
                 Heap Fetches: 3000
```

This PR creates a new `stats.runtime_accounts` table, backfills it, and populates it via a trigger. Note that the trigger function is an upsert so it should be safe with the fast-sync analyzer mode.

Testing:
- Tested locally and confirmed no outward api changes.

### DIFF
--- a/analyzer/queries/queries.go
+++ b/analyzer/queries/queries.go
@@ -337,6 +337,12 @@ var (
     INSERT INTO chain.runtime_related_transactions (runtime, account_address, tx_round, tx_index)
       VALUES ($1, $2, $3, $4)`
 
+	RuntimeAccountNumTxsUpsert = `
+    INSERT INTO chain.runtime_accounts as accounts (runtime, address, num_txs)
+      VALUES ($1, $2, $3)
+    ON CONFLICT (runtime, address) DO UPDATE
+      SET num_txs = accounts.num_txs + $3;`
+
 	RuntimeTransactionInsert = `
     INSERT INTO chain.runtime_transactions (runtime, round, tx_index, tx_hash, tx_eth_hash, fee, gas_limit, gas_used, size, timestamp, method, body, "to", amount, evm_encrypted_format, evm_encrypted_public_key, evm_encrypted_data_nonce, evm_encrypted_data_data, evm_encrypted_result_nonce, evm_encrypted_result_data, success, error_module, error_code, error_message)
       VALUES ($1, $2, $3, $4, $5, $6, $7, $8, $9, $10, $11, $12, $13, $14, $15, $16, $17, $18, $19, $20, $21, $22, $23, $24)`
--- a/storage/client/queries/queries.go
+++ b/storage/client/queries/queries.go
@@ -464,8 +464,8 @@ const (
 				(SELECT sum(amount) from chain.runtime_transfers where runtime = $1 AND receiver = $2::text)
 				, 0) AS total_received,
 			COALESCE (
-				(SELECT count(*) from chain.runtime_related_transactions where runtime = $1 AND account_address = $2::text)
-				, 0) AS num_txns`
+				(SELECT num_txs from chain.runtime_accounts where runtime = $1 AND address = $2::text)
+				, 0) AS num_txs`
 
 	//nolint:gosec // Linter suspects a hardcoded access token.
 	EvmTokens = `
--- a/analyzer/runtime/runtime.go
+++ b/analyzer/runtime/runtime.go
@@ -178,6 +178,7 @@ func (m *processor) queueDbUpdates(batch *storage.QueryBatch, data *BlockData) {
 		}
 		for addr := range transactionData.RelatedAccountAddresses {
 			batch.Queue(queries.RuntimeRelatedTransactionInsert, m.runtime, addr, data.Header.Round, transactionData.Index)
+			batch.Queue(queries.RuntimeAccountNumTxsUpsert, m.runtime, addr, 1)
 		}
 		var (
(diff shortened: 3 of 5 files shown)


storage/client/queries/queries.go
```diff
@@ -464,7 +464,7 @@ const (
 				(SELECT sum(amount) from chain.runtime_transfers where runtime = $1 AND receiver = $2::text)
 				, 0) AS total_received,
 			COALESCE (
-				(SELECT count(*) from chain.runtime_related_transactions where runtime = $1 AND account_address = $2::text)
+				(SELECT num_txs from stats.runtime_accounts where runtime = $1 AND address = $2::text)
 				, 0) AS num_txns`
```
Andrew7234, line 468:
```suggestion
				, 0) AS num_txs`
```

I'd prefer to standardize this but this would be a breaking change in the api...

storage/migrations/14_runtime_account_stats.up.sql
```diff
@@ -0,0 +1,37 @@
+BEGIN;
+
+CREATE TABLE stats.runtime_accounts
```
Andrew7234, line 3: not sure if it belongs in this schema but it seemed to fit better than in `analysis`

storage/migrations/14_runtime_account_stats.up.sql
```diff
@@ -0,0 +1,37 @@
...
+    num_txs UINT63 NOT NULL DEFAULT 0
+);
+
+-- Backfill stats.runtime_accounts
+INSERT INTO stats.runtime_accounts (runtime, address, num_txs)
+    SELECT runtime, account_address, COUNT(*) 
+    FROM chain.runtime_related_transactions
+    GROUP BY (runtime, account_address);
+
+CREATE OR REPLACE FUNCTION calculate_runtime_account_num_txs() RETURNS TRIGGER
```
mitjat, line 18: I thought we decided to avoid triggers? Let's move this into Go, please. We only perform the corresponding `INSERT INTO chain.runtime_related_transactions` in [one spot](https://github.com/oasisprotocol/nexus/blob/087ac66d6df6d1f9783b042057f0a7eddd57998d/analyzer/runtime/runtime.go#L180), so Go should be roughly the same amount of code.

Using just one mechanism makes code reading, debugging, and maintainance easier.
For example: For fast sync, we will need to disable certain triggers. It's easier to do that either in just Go or just in postgres.
Or: To understand how a table is updated, one has to code-spelunk (grep or otherwise). If all updates follow the same pattern, that's easier to do.

storage/migrations/14_runtime_account_stats.up.sql
mitjat: At least so far, `stats` has been for statistics that are:
- Less time sensitive; we do not calculate them in real time
- Calculated in dedicated analyzers (largely because of the previous point). We might even want to move them out of analyzers into an off-the-shelf data analysis product in the future; it's a common problem in the industry.
- Keyed by time, and available for any point in history

So they're quite a different beast.

I'd put this into `chain.runtime_accounts`; it's just another "live" dead-reckoned value, like total_gas_used, or arguably even like the balances.

storage/migrations/14_runtime_account_stats.up.sql
Andrew7234: Sounds good - updated

storage/migrations/14_runtime_account_stats.up.sql
Andrew7234: Good points; updated!

analyzer/queries/queries.go
```diff
@@ -337,6 +337,12 @@ var (
     INSERT INTO chain.runtime_related_transactions (runtime, account_address, tx_round, tx_index)
       VALUES ($1, $2, $3, $4)`
 
+	RuntimeAccountUpsert = `
```
mitjat, line 340: Nit/optional: RuntimeAccountNumTxsUpsert, so it's clearer what we're doing from the callsite. Also, if this table ever acquires more columns, it's unlikely we'll be updating all of them in a single upsert, so this one will likely stay dedicated to num_txs.

mitjat (approved):
Thank you!

api/spec/v1.yaml
```diff
@@ -2558,15 +2558,15 @@ components:
...
+      required: [total_sent, total_received, num_txs]
       properties:
         total_sent:
           description: The total number of tokens sent, in base units.
           <<: *BigIntType
         total_received:
           description: The total number of tokens received, in base units.
           <<: *BigIntType
-        num_txns:
+        num_txs:
```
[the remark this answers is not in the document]
Andrew7234: d'oh, I misread it. Fixed

PR #506, opened 2023-08-21
NFT analyzer

pro-wh:
an analyzer for NFTs. the term "NFT" is proposed to refer to specific instances managed by a contract.

this code highly experimental and missing some important things.

- NFT analysis jobs involve downloading a "metadata" file from a URI given by a smart contract. currently those jobs may fail. we do not have a system for scheduling those jobs to be retried
- a contract may point to a server that is arbitrarily slow. currently we set a timeout of 30 seconds for downloading this file. but yeah, anyone could make the analyzer as slow as 30s/batch
- we have no way to start on more jobs if only a few downloads are straggling.
- we have no policy for judging what errors to treat as permanent, when it comes to downloading this metadata file

conflicts:

- ~~there's ongoing work #511 to build a superclass for analyzers like this. can we use that here? in anticipation, I have not written much to go around the actual download logic, e.g. it's a single-threaded download loop in this PR~~ integrated, thanks!
- ~~I think I saw there's something in the reckoning fixes #512 for unifying certain analysis.something and chain.something tables. might the principle leading to that apply here too?~~ unified woo


analyzer/runtime/evm/erc721.go
```diff
@@ -62,6 +84,57 @@ func evmDownloadTokenERC721(ctx context.Context, logger *log.Logger, source node
...
+			)
+		}
+		if nftData.MetadataURI != "" {
+			logger.Info("downloading metadata",
+				"token_eth_addr", hex.EncodeToString(tokenEthAddr),
+				"token_id", id,
+				"uri", nftData.MetadataURI,
+			)
+			nftData.MetadataAccessed = time.Now()
+			rc, err1 := drawerfullofprotocols.Get(ctx, ipfsClient, nftData.MetadataURI)
```
pro-wh, line 109: we're starting to bleed out of evm-related things in this file :skull:

analyzer/runtime/evm/erc721.go
```diff
@@ -62,6 +84,57 @@ func evmDownloadTokenERC721(ctx context.Context, logger *log.Logger, source node
...
+			if err1 != nil {
+				// todo: retry on some errors?
```
pro-wh, line 111: big implications here, from the possibility that the metadata request goes to something other than our own node

1. the server could be slow. even maliciously slow. there's a 30s timeout, but like that means someone could make every batch take 30 seconds
2. some jobs can now fail without it being a horrible "our node is down and nothing else will ever work" situation, as was the case before. we'll want a way to continue with other work
3. some failures probably will be intermittent. after all, these requests will go over the internet, right? we'll want to hang on to a job so we can retry it later
4. oh and "later" will best be one of those backoff systems
5. all that, while we shouldn't assume that the list of all NFTs fits in memory

pro-wh:
opening this for "review," although it might be easier not to merge this for quite some time. there's no API for the data this collects, and the analyzer is very un-robust, and it's missing some features that will affect the tables added here

analyzer/runtime/evm/erc721.go
mitjat: I'm trying to foresee what the architecture will be like, longer term. I expect there will be several async phases:
- Token id -> token metadata URL (by issuing an EVM query to the runtime).
- Metadata fetching (URL -> JSON) 
- Metadata analysis. This is almost trivial; just extracting the relevant field(s) from the metadata JSON.
- Image fetching (URL -> blob)

The reason I think it probably makes sense for each of these to be async from other is that the URL fetching is a) fallible and b) a reusable component.

Phases 2 and 4 are pretty much identical when it comes to requirements: URLs are untrusted, the fetched resource might become stale (I'm referring to non-constant NFTs here), fetching might need to be retried multiple times with backoff. So one possible solution would be to have a "resource fetching" analyzer: Stages 1 and 3 give jobs to it, and tag the job somehow to signify the table into which the result should be output. So the analyzer has the understanding of those tags built in. Or, it could be two separate analyzers, with a largely shared codebase, but each instantiated with a separate `dataInserter` struct (much like `processor` for blockAnalyzers) that knows what to do with the fetched data.

I don't think we need all of the above in this PR (or that it'd be an appropriate amount of new code for one PR, at that :)), but it would probably be helpful to align things with the architecture we want to move towards. For example, if we go with what I outlined above, I think it would make sense to add a minimal `resourceFetchAnalyzer` now. No retries, cache expiration, etc, just a naive one-off fetch.

Depending on how much our thought (mis)align, this might be worth a short doc, actually. I think neither the structure of the NFT analyzer(s) nor the structure of a resource fetcher/cache is clear yet.

mitjat (commented):
Lots of work! Thank you.
Also, good points about future work in the PR description. I started a thread to hear your thinking on how we could address those, and to add thoughts of my own.

analyzer/runtime/evm/erc721.go
mitjat: Also wanted to point out that only phase 1 is EVM specific, which would be nice to reflect in analyzer names. (Just `nft`, no `evm_nft`)

analyzer/drawerfullofprotocols/drawer_full_of_protocols.go
```diff
@@ -0,0 +1,42 @@
+package drawerfullofprotocols
```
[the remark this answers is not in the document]
mitjat: :raised_eyebrow:  how did you get into my house?

storage/migrations/18_evm_nfts.up.sql
```diff
@@ -0,0 +1,21 @@
...
+    nft_id uint_numeric NOT NULL,
+    PRIMARY KEY (runtime, token_address, nft_id),
+    metadata_uri TEXT,
+    metadata_accessed TIMESTAMP,
+    name TEXT,
+    description TEXT,
+    image TEXT
+);
+
+CREATE TABLE analysis.evm_nfts (
```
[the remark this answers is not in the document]
pro-wh: ok I'll be trying that

storage/migrations/18_evm_nfts.up.sql
```diff
@@ -0,0 +1,21 @@
+CREATE TABLE chain.evm_nfts (
+    runtime runtime NOT NULL,
+    token_address oasis_addr NOT NULL,
+    nft_id uint_numeric NOT NULL,
+    PRIMARY KEY (runtime, token_address, nft_id),
+    metadata_uri TEXT,
+    metadata_accessed TIMESTAMP,
+    name TEXT,
+    description TEXT,
+    image TEXT
```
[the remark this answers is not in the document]
pro-wh: just the url

can we put that image fetch in the frontend at first?

cmd/analyzer/analyzer.go
```diff
@@ -207,6 +210,10 @@ func (s *sourceFactory) Runtime(ctx context.Context, runtime common.Runtime) (no
 	return s.runtimes[runtime], nil
 }
 
+func (s *sourceFactory) IPFS(_ context.Context) (ipfsc.Client, error) {
+	return ipfsc.NewGateway(s.cfg.IPFS.Gateway)
```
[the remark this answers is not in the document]
pro-wh: ok

analyzer/drawerfullofprotocols/drawer_full_of_protocols.go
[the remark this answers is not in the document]
pro-wh: huh yeah, makes sense not to lump them together so much. leme put these inside the one analyzer where they're used for now.

analyzer/ipfsc/gateway.go
```diff
@@ -0,0 +1,42 @@
...
+	url := g.base + "/ipfs/" + path
+	resp, err := neededforhttp.GetWithContextWithClient(ctx, gatewayHTTPClient, url)
+	if err != nil {
+		return nil, fmt.Errorf("requesting from gateway: %w", err)
+	}
+	rc, err := neededforhttp.ResponseOK(resp)
+	if err != nil {
+		return nil, err
+	}
+	return rc, nil
```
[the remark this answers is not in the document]
pro-wh: the problem with doing that here is checking the response status code is a step. writing out the separate error return on the last fallible step is in the same vein as using the trailing commas in multiline lists

analyzer/queries/queries.go
```diff
@@ -593,6 +593,60 @@ var (
...
+          processed_time IS NOT NULL
+      ) AS download_round
+    FROM analysis.evm_nfts
+    JOIN chain.evm_tokens USING
+      (runtime, token_address)
+    LEFT JOIN chain.address_preimages ON
+      chain.address_preimages.address = analysis.evm_nfts.token_address
+    WHERE
+      runtime = $1 AND
+      (last_download_round IS NULL OR last_want_download_round > last_download_round)
```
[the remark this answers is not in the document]
pro-wh: at this point I want to get this in and work on retry logic later

storage/migrations/18_evm_nfts.up.sql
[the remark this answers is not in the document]
pro-wh: combined

analyzer/runtime/evm/erc721.go
pro-wh: oh I forgot to remove the "evm" part

analyzer/neededforhttp/needed_for_http.go
```diff
@@ -0,0 +1,50 @@
+package neededforhttp
```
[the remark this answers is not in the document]
pro-wh: the chunk of opinions encoded in this file are common to (i) that package formerly called pubclient, for making requests to an externally provided URL without SSRF; and (ii) accessing IPFS content over an IPFS HTTP gateway. unfortunately, (i) and (ii) can't be in common because it's intended that we use a local IPFS node, and SSRF prevention would prevent us from connecting to a local service. so (i) and (ii) are separate, and this code is not particularly closer to either of them, so it's in a third package now :smiling_face_with_tear:

analyzer/ipfsc/gateway.go
```diff
@@ -0,0 +1,42 @@
+package ipfsc
```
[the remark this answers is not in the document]
pro-wh: renamed

analyzer/drawerfullofprotocols/drawer_full_of_protocols.go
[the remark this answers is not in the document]
pro-wh: renamed

analyzer/ipfsc/gateway.go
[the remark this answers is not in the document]
pro-wh: okay something unrelated made me have to change the way this goes (omg was that a pun). the `bodyclose` lint can't detect that the resp.Body lives on if ResponseOK extracts it for us. so now ResponseOK no longer 'returns' anything (ok it still returns an error :roll_eyes:) and we do `return resp.Body, nil` instead

analyzer/runtime/evm/erc721.go
pro-wh: okay counter proposal: as this analyzer still has the EVM part, let's leave the "evm" in the name. we can separate out the network download later, along with tackling server errors

analyzer/evmnfts/evm_nfts.go
```diff
@@ -0,0 +1,127 @@
...
+		cfg,
+		p,
+		target,
+		logger,
+	)
+}
+
+type StaleNFT struct {
+	Addr                  string
+	ID                    *big.Int
```
mitjat, line 62: Why not use common.BigInt here? The main point of introducing our own BigInt was that it interops more smoothly with pgx and other (de)serializations. Also, doesn't have funky pointer semantics.

analyzer/runtime/evm/erc721.go
```diff
@@ -62,6 +84,58 @@ func evmDownloadTokenERC721(ctx context.Context, logger *log.Logger, source node
 	return &tokenData, nil
 }
 
+func evmDownloadNFTERC721(ctx context.Context, logger *log.Logger, source nodeapi.RuntimeApiLite, ipfsClient ipfsclient.Client, round uint64, tokenEthAddr []byte, id *big.Int) (*EVMNFTData, error) {
+	var nftData EVMNFTData
+	supportsMetadata, err := detectInterface(ctx, logger, source, round, tokenEthAddr, ERC721MetadataInterfaceID)
+	if err != nil {
+		return nil, fmt.Errorf("checking ERC721Metadata interface: %w", err)
+	}
+	//nolint:nestif
```
mitjat, line 93: Nit: It _is_ a little hard to read. Can we do an early exit instead? `if !supportMetadata { return ... }`

Actually, looks like we could do early return on several more ifs below, and have this function be pretty much linear. That would help readability a fair bit.

analyzer/runtime/evm/erc721.go
```diff
@@ -62,6 +84,58 @@ func evmDownloadTokenERC721(ctx context.Context, logger *log.Logger, source node
...
+	if supportsMetadata {
+		if err = evmCallWithABI(ctx, source, round, tokenEthAddr, evmabi.ERC721Metadata, &nftData.MetadataURI, "tokenURI", id); err != nil {
+			if !errors.Is(err, EVMDeterministicError{}) {
+				return nil, fmt.Errorf("calling tokenURI: %w", err)
+			}
+			logDeterministicError(logger, round, tokenEthAddr, "ERC721Metadata", "tokenURI", err,
```
mitjat, line 99: Nit: Can you add "deterministically" to the log message inside this helper func, and/or some other reassuring phrase? Right now if I saw that log, I'd be a little alarmed and felt like I needed to investigate.

analyzer/runtime/evm/erc721.go
mitjat: sgtm for the rename.

For the network download: if I understand the code correctly, any failed downloads currently will just cause the token to stay in the analyzer queue, so if we add a more complex `downloadAnalyzer` later, it will "just work" in that we won't need any complex db migrations to construct a work queue for it. If so, SGTM. Can you create a ticket in that case, please?

analyzer/evmnfts/evm_nfts.go
```diff
@@ -0,0 +1,127 @@
...
+		staleNFT.DownloadRound,
+		tokenEthAddr,
+		*staleNFT.Type,
+		staleNFT.ID,
+	)
+	if err != nil {
+		return fmt.Errorf("downloading NFT %s %v: %w", staleNFT.Addr, staleNFT.ID, err)
+	}
+	batch.Queue(queries.RuntimeEVMNFTUpdate, p.runtime, staleNFT.Addr, staleNFT.ID, nftData.MetadataURI, nftData.MetadataAccessed, nftData.Name, nftData.Description, nftData.Image)
+	batch.Queue(queries.RuntimeEVMNFTAnalysisUpdate, p.runtime, staleNFT.Addr, staleNFT.ID, staleNFT.DownloadRound)
```
mitjat, line 117: this can probably be a single query now?

storage/migrations/18_evm_nfts.up.sql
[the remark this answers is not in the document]
mitjat: Yeah IIRC @lukaw3d said there's no csrf in place because the explorer is read-only. So for the short term, letting them fetch directly from untrusted sources should be fine.

cmd/analyzer/analyzer.go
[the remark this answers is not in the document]
mitjat: not implemented, I think

analyzer/evmnfts/httpmisc/http_misc.go
```diff
@@ -0,0 +1,49 @@
...
+	if _, ok := target.(NotOKError); ok {
+		return true
+	}
+	return false
+}
+
+func ResponseOK(resp *http.Response) error {
+	if resp.StatusCode >= 500 {
+		if err := resp.Body.Close(); err != nil {
+			return fmt.Errorf("HTTP closing body: %w", err)
```
mitjat, line 38: Here and below: Let's include the HTTP status code in the error.

analyzer/evmnfts/httpmisc/http_misc.go
```diff
@@ -0,0 +1,49 @@
...
+		}
+		return fmt.Errorf("HTTP %d", resp.StatusCode)
+	}
+	if resp.StatusCode != 200 {
+		if err := resp.Body.Close(); err != nil {
+			return fmt.Errorf("HTTP closing body: %w", err)
+		}
+		return NotOKError{fmt.Errorf("HTTP %d", resp.StatusCode)}
```
mitjat, line 46: What is the point of NotOKError? It's an "enum" with a single value, it's produced seemingly inconsistently (I don't understand why the ability to read the response body of e.g. a HTTP 404 would differentiate between a `NotOKError` and a `fmt.Errorf`) and nobody "reads" it, i.e. checks for its presence.
Can we get rid of it and reintroduce when/if requirements are clearer?

analyzer/evmnfts/httpmisc/http_misc.go
```diff
@@ -0,0 +1,49 @@
...
+
+func (err NotOKError) Is(target error) bool {
+	if _, ok := target.(NotOKError); ok {
+		return true
+	}
+	return false
+}
+
+func ResponseOK(resp *http.Response) error {
+	if resp.StatusCode >= 500 {
```
mitjat, line 36: I don't understand how this is different from `StatusCode != 200`; the handler seems identical, except for the subtle difference in producing the `NotOKError` wrapper or not.

analyzer/neededforhttp/needed_for_http.go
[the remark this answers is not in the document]
mitjat: Hrm yeah I wanted these funcs to be embedded in `multiproto`, but that would cause circular dependencies with (i) and (ii) :/
Can you add top-level descriptions to these new modules? They're kinda vaguely/broadly named, and at the same time contain so little functionality that it's hard to see at a glance what their intended scope is.

analyzer/queries/queries.go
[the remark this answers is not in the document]
mitjat: OK. To me, it looks easier (and fast enough) to change it now that to deal with migrations in the future. But whatever is your preference.

mitjat (approved):
Thank you Warren!

I only have small comments left at this point, so LGTM pending resolved comment threads.

storage/migrations/18_evm_nfts.up.sql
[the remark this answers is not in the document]
lukaw3d: Yep yep, we don't have CSP yet
(also no CSRF, but that's different)

cmd/analyzer/analyzer.go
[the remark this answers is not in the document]
pro-wh: woops yeah, gotta add that

analyzer/neededforhttp/needed_for_http.go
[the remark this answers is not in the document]
pro-wh: ok adding comments

analyzer/queries/queries.go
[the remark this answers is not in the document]
pro-wh: let me file something for that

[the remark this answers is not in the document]
pro-wh: #532

analyzer/evmnfts/evm_nfts.go
pro-wh: lately I've been using a lot more go-ethereum types, which it uses *big.Int for its large integers. common.BigInt does come in handy when scanning from the database though.

analyzer/runtime/evm/erc721.go
pro-wh: early exits are more suitable for "something went wrong, bail." this is a section where we progressively obtain more info if it does support a certain interface. not supporting that interface is fine, and we may theoretically do other things to get information. following these guidelines for when to use early exit and when to do something inside an `if` block helps make code more conveniently composable

analyzer/runtime/evm/erc721.go
pro-wh: ya adding

analyzer/runtime/evm/erc721.go
pro-wh: I'll break this up into other methods tho. would be good to early-exit the metadata part if calling tokenURI fails for example

analyzer/evmnfts/evm_nfts.go
pro-wh: oh yeah

analyzer/runtime/evm/erc721.go
pro-wh: > any failed downloads currently will just cause the token to stay in the analyzer queue

no, there's more fiddliness to it. right now:

"DONE" means download job is successfully completed, possibly with no or partial info collected; "BREAK" means something really bad is happening and abort the whole batch. currently there's no third option for "just this one has a problem"

- problem checking for ERC721Metadata ->
  - deterministic -> ERC165 says this means it's not supported. (continue as if not supported)
  - nondeterministic -> something's wrong with our access to the network. BREAK
- doesn't support ERC721Metadata -> that's fine, no data from this source. DONE
- supports ERC721Metadata -> (query tokenURI)
  - problem querying tokenURI ->
    - deterministic -> contract's fault. DONE
    - nondeterministic -> something's wrong with our access to the network. BREAK
  - got token URI -> (save token URI and "accessed" date)
    - URI empty or malformed etc. -> contract's fault. DONE
    - URI unsupported scheme -> uh... just give up. DONE
    - URI parseable and supported -> (try to download)
      - just about anything goes wrong -> not necessarily NFT's fault, but also don't want to let this starve the queue. DONE :arrow_left: **this thing**
      - download OK -> (try to parse the info JSON doc)
        - problem parsing JSON -> NFT's fault. DONE
        - JSON OK -> close download stream
          - problem closing download stream -> Go runtime integrity problem?? BREAK
          - close stream OK -> (save name + description + image URL) DONE

so the main thing labeled "this thing" above, that's where currently we treat that as "DONE" and it's arguably bad. this PR has the beginnings of some things that'll help us tell if it's our fault or the NFT's fault, but it's not in use yet. we'll probably have to do a migration to repopulate the queue with jobs that ended up this way. a superset of those jobs, really

but still yeah, filed in #532.

analyzer/evmnfts/httpmisc/http_misc.go
pro-wh: ok

analyzer/evmnfts/httpmisc/http_misc.go
pro-wh: meant for a future update to separate whether to retry. statuses 5xx indicate server error, which would be a good candidate for retry.

4xx indicates the server isn't satisfied with our request, so it's probably better to give up. Come to think of it, 429 might be better in the other category.

3xx is supposed to be a redirect, but the HTTP client ought to handle that for us. if for whatever reason it didn't, then it's not like we can handle it out here, so it's probably better to give up.

2xx other than 200 ("OK") indicates some kind of successful request but where it doesn't actually give us a file. we're always looking for a file, so it's probably better to give up.

1xx is some message used in the HTTP protocol but aren't formally the response. I suspect the HTTP client handles these internally, but if it somehow gets out here, then we can't handle it, so it's probably better to give up.

These "give up" cases are lumped into a `NotOKError`

analyzer/evmnfts/httpmisc/http_misc.go
pro-wh: answering here https://github.com/oasisprotocol/nexus/pull/506#discussion_r1339330308

analyzer/runtime/evm/erc721.go
pro-wh: added

analyzer/runtime/evm/erc721.go
pro-wh: refactored

analyzer/runtime/evm/erc721.go
```diff
@@ -3,18 +3,40 @@ package evm
...
+// https://eips.ethereum.org/EIPS/eip-721
+type ERC721AssetMetadata struct {
+	// Name identifies the asset to which this NFT represents
+	Name string `json:"name"`
+	// Description describes the asset to which this NFT represents
+	Description string `json:"description"`
+	// Image is A URI pointing to a resource with mime type image/*
+	// representing the asset to which this NFT represents. Consider making
+	// any images at a width between 320 and 1080 pixels and aspect ratio
+	// between 1.91:1 and 4:5 inclusive.
```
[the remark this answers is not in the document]
pro-wh: removed

cmd/analyzer/analyzer.go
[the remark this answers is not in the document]
pro-wh: added

analyzer/drawerfullofprotocols/drawer_full_of_protocols.go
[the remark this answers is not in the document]
pro-wh: moved

analyzer/evmnfts/httpmisc/http_misc.go
pro-wh: included

analyzer/evmnfts/evm_nfts.go
mitjat: Yeah looking at the DB code was what prompted this comment.
I think this is the right layer to convert from whatever types go-ethereum gives us to the types we prefer to use internally, if the two are different.
Obviously a minor comment though, your call in the end, I won't cry either way.

analyzer/evmnfts/httpmisc/http_misc.go
mitjat: Alright thanks for the explanation. (And great catch with `HTTP 429 Too Many Requests`! I agree. Wanna encode that now?)
In that case, two small followups:
- Let's rename `NotOKError` please. The name misled me when reading because I think of "error" and "not ok" as pretty synonymous. Maybe `NonRetriableError`? Or we could reuse the name from the RPC cache: `DeterministicError`. Or similar.
- In the non-5xx case, even if we fail to `Close()` the body, I'd say that's still a `DeterministicError`. The HTTP status code is all we need to determine (heh) that. We can just silently log the `Close()` problem maybe? Or even ignore it altogether, assuming some resource cleanup happens behind the scenes.

analyzer/runtime/evm/erc721.go
pro-wh: :skull: resolving this to focus on active discussions

storage/migrations/18_evm_nfts.up.sql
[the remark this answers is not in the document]
pro-wh: resolving this to focus on active discussions :skull:

analyzer/drawerfullofprotocols/drawer_full_of_protocols.go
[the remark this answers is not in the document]
pro-wh: haven't moved the other non-analyzer subdirectories though. so it continues :skull:

analyzer/evmnfts/evm_nfts.go
pro-wh: should we just say it? do you want us to prefer using common.BigInt and converting to *big.Int locally where needed?

analyzer/evmnfts/httpmisc/http_misc.go
pro-wh: closing this, let's discuss in the other conversation :skull:

analyzer/evmnfts/evm_nfts.go
mitjat: oh. Not sure what "should we just say it" means exactly, but yes, I do prefer what you describe. I think it makes for more consistent code and smoother interop between various parts of the indexer. (EVM work that often needs *big.Int because of go-ethereum is an unfortunate exception.)
So inasmuch as I have decision power, I choose to wield it in that direction :)

pro-wh: k going with common.BigInt here :tipping_hand_person:

analyzer/evmnfts/httpmisc/http_misc.go
pro-wh: changed to ResourceError, to convey that the server successfully answered about the Resource that the URL Locates, but we still find it unsatisfactory

PR #507, opened 2023-08-22
bugfix: db: grant permissions on new tables

mitjat:
New tables were introduced in https://github.com/oasisprotocol/nexus/pull/477/files, but we forgot to set permissions for them.

This has already been fixed/applied manually in our existing deploys by @ptrus


pro-wh (approved):
ok

Andrew7234 (approved):
looks good. just checking - it was applied to all deploys, right? Iirc there was a slack message that said it was only applied to prod mainnet at the time

mitjat:
> looks good. just checking - it was applied to all deploys, right? Iirc there was a slack message that said it was only applied to prod mainnet at the time

Huh not sure, my understanding of Peter Us's message was that he had applied it everywhere. I reapplied everywhere just now, to make sure.

PR #508, opened 2023-08-28
Bump client-sdk to 0.6.0 (for Sapphire 0.6.0)

mitjat:
In support of Sapphire 0.6.0.

The new Sapphire runtime was recently deployed on the testnet. It introduces some new enum values that client-sdk v0.2.1 (currently in use by the nexus) does not recognize, and therefore chokes on a recent block:
```
{"analyzer":"sapphire","caller":"block.go:250","err":"extract non-tx events: event 2: decode consensus accounts: invalid consensus accounts event code: 3","height":2378823,"level":"error","module":"analysis_service","msg":"error processing block","slow_sync":true,"ts":"2023-08-28T10:49:24.559964731Z"}
```

**Compatibility:**
I tried to look for any backward-incomptible changes in `client-sdk/go/types` and `client-sdk/go/modules/*/types.go`:
```git diff --ignore-all-space client-sdk/go/v0.2.0..client-sdk/go/v0.6.0 client-sdk/go/types client-sdk/go/modules/*/types.go```

Manual inspection shows mostly only new fields and new `PrettyPrint()` methods. There is a single backwards-incompatible change:
Some struct fields are renamed [here](https://github.com/oasisprotocol/oasis-sdk/blob/eb97a8162f84ae81d11d805e6dceeeb016841c27/client-sdk/go/modules/contracts/types.go#L303-L307): `wasm_public_storage_get_base` used to be `wasm_storage_get_base` etc, all fields gained a `public_` in the name. But this only surfaces inside the return type of the [contracts.Paramters() API endpoint](https://github.com/oasisprotocol/oasis-sdk/blob/eb97a8162f84ae81d11d805e6dceeeb016841c27/client-sdk/go/modules/contracts/contracts.go#L113-L113), which the indexer never invokes.

In other words, **indexer should be able to use client-sdk v0.6.0 smoothly.**

**Testing:**
Ran against 1000 blocks of mainnet just in case, to smoke out any obvious incompats. Nothing came up.
Testnet: No testing yet because our internal sapphire testnet node is currently inaccessible.

### CODE REVIEW

File: analyzer/runtime/extract.go [hunk 1]
```diff
@@ -338,7 +338,7 @@ func ExtractRound(blockHeader nodeapi.RuntimeBlockHeader, txrs []nodeapi.Runtime
 				blockTransactionData.Success = nil
 			}
 
-			blockTransactionData.Method = tx.Call.Method
+			blockTransactionData.Method = string(tx.Call.Method)
```
pro-wh, line 341: :information_source: oh it's a MethodName now

pro-wh (approved):
that's us!

analyzer/runtime/extract.go
mitjat: Ah yes thank you, I was going to comment. It's a `MethodName`, which is defined as `type MethodName string`.

analyzer/uncategorized/evm_raw_tx_test.go
```diff
@@ -60,8 +60,7 @@ func decodeExpectCall(
 	require.Equal(t, expectedDataBytes, body.Data)
 	require.Len(t, tx.AuthInfo.SignerInfo, 1)
 	from0xChecksummed := helpers.EthAddressFromPubKey(*tx.AuthInfo.SignerInfo[0].AddressSpec.Signature.Secp256k1Eth)
-	fromChecksummed := from0xChecksummed[2:]
-	from := strings.ToLower(fromChecksummed)
```
mitjat, line 64: `EthAddressFromPubKey` used to return a `string` (with a leading `0x`, but now returns `ethCommon.Address`, which is `[20]byte`.
I checked, and the rest of indexer is not using `EthAddressFromPubKey`, or even the `helpers` package.

pro-wh: I gotta rename from0xChecksummend now :skull:

PR #510, opened 2023-08-30
analyzer/runtime/consensusaccounts: Handle new client-sdk 0.6.0 event codes

mitjat:
This PR adds suppport for parsing (the new event codes present in) Sapphire 0.6.0 blocks; it is a follow-up to #508 which was not possible to test at the time.
It was not enough to just import the newer type definitions (as #508 did), we also had to add some handling for the new types (as this PR does).

Testing: Indexed the offending block 2_378_823 of sapphire testnet. (Now the internal testnet node is working again, so I was able to test properly.)


Andrew7234 (approved):
noice

PR #511, opened 2023-08-30
add itemBasedAnalyzer; refactor evm analyzers accordingly

Andrew7234:
https://app.clickup.com/t/866afuw9w

Todo:
- itemBasedAnalyzer tests 
- update metadata registry analyzer
- update node stats analyzer
- consolidate analyzer args into cfg struct


analyzer/item/item.go
```diff
@@ -0,0 +1,211 @@
...
+	if err != nil {
+		a.logger.Warn("error fetching queue length", "err", err)
+	}
+	a.queueLengthMetric.Set(float64(queueLength))
+}
+
+// processBatch fetches the next batch of work items, processes them in parallel, and
+// commits the resulting state changes to the database. Note that if the analyzer fails
+// to process one or more items in a batch, the entire batch is discarded and no database
+// updates are applied.
```
Andrew7234, line 111: This behavior mirrors the current batch/retry logic in the existing evm analyzers. However, we likely want to update this soon/now to apply db updates from items that were processed successfully, even if other items in the batch weren't, in order to make progress. Otherwise, a single bad/malicious item could stall the analyzer.

analyzer/metadata_registry/metadata_registry.go
```diff
@@ -0,0 +1,90 @@
...
+		stopOnEmptyBatch,
+		fixedInterval,
+		p,
+		target,
+		logger,
+	)
+}
+
+func (p *processor) GetItems(ctx context.Context, limit uint64) ([]common.ChainName, error) {
+	return []common.ChainName{common.ChainNameMainnet}, nil
```
Andrew7234, line 60: note: for this and the node_stats analyzer, this fixed queue means that these analyzers are not yet compatible with the e2e_regression tests. This will likely need to be slightly reworked with`stopOnEmptyBatch` but I'd prefer to leave it for a future PR as this one is already somewhat long, and the regression tests do not currently test those analyzers so there is no loss of functionality.

pro-wh: does this fully coincide with the use of `fixedInterval`?

analyzer/item/item.go
mitjat: :+1: let's update this now

analyzer/evmtokens/evm_tokens.go
```diff
@@ -110,29 +113,29 @@ func (m main) getStaleTokens(ctx context.Context, limit int) ([]*StaleToken, err
 	return staleTokens, nil
 }
 
-func (m main) processStaleToken(ctx context.Context, batch *storage.QueryBatch, staleToken *StaleToken) error {
-	m.logger.Info("downloading", "stale_token", staleToken)
-	tokenEthAddr, err := client.EVMEthAddrFromPreimage(staleToken.AddrContextIdentifier, staleToken.AddrContextVersion, staleToken.AddrData)
+func (p *processor) ProcessItem(ctx context.Context, batch *storage.QueryBatch, item *StaleToken) error {
```
mitjat, line 116: FWIW, the type checker only cares about the types of args, not their names. So you could rename `item` back to `staleToken` to make the body of the method more readable.

mitjat (commented):
lotsa red! 🤘
and with tests :heart: 

Looks good by and large, I have only one or two nontrivial comments.

Despite the new unit tests, can you please also run e2e tests? It's a pretty big systemic change.
When running locally, you don't need to worry about caches and/or analyzers supporting `StopOnEmptyBatch` ... just enable all analyzers in config, then manually run analysis, kill analyzers when you can tell from the logs they're done, and run `run.sh` (which you possibly want to extend with an endpoint or two to exercise all the analyzers).

analyzer/item/item.go
Andrew7234: got it - updated!

analyzer/metadata_registry/metadata_registry.go
Andrew7234: > does this fully coincide with the use of fixedInterval?

currently yes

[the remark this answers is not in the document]
Andrew7234: makes sense; updated

analyzer/item/item.go
```diff
@@ -0,0 +1,211 @@
...
+	prometheus.MustRegister(a.queueLengthMetric)
+
+	return a, nil
+}
+
+// sendQueueLength reports the current number of items in the work queue to Prometheus.
+func (a *itemBasedAnalyzer[Item]) sendQueueLengthMetric(ctx context.Context) {
+	queueLength, err := a.processor.QueueLength(ctx)
+	if err != nil {
+		a.logger.Warn("error fetching queue length", "err", err)
```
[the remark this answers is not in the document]
Andrew7234: good catch, fixed

analyzer/evmverifier/evmverifier.go
```diff
@@ -76,127 +114,108 @@ func (a *evmContractsVerifier) getUnverifiedContracts(ctx context.Context, runti
...
+		p.runtime,
+		item.Addr,
+		abi.Output.ABI,
+		metadata,
+		sourceFiles,
+	)
+	if err != nil {
+		p.logger.Error("failed to update verified contract", "err", err, "eth_address", item.EthAddr, "address", item.Addr)
+	}
+	defer rows.Close()
```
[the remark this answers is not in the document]
Andrew7234: yup, changed

analyzer/evmverifier/evmverifier.go
```diff
@@ -76,127 +114,108 @@ func (a *evmContractsVerifier) getUnverifiedContracts(ctx context.Context, runti
...
+	}
+	return verifiableContracts[:limit], nil
 }
 
-func NewEVMVerifierAnalyzer(cfg *config.EVMContractsVerifierConfig, target storage.TargetStorage, logger *log.Logger) (analyzer.Analyzer, error) {
-	logger = logger.With("analyzer", evmContractsVerifierName)
+func (p *processor) ProcessItem(ctx context.Context, batch *storage.QueryBatch, item *unverifiedContract) error {
+	p.logger.Debug("verifying contract", "address", item.Addr, "eth_address", item.EthAddr)
+
+	// Timeout a bit between API calls.
```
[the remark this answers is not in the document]
Andrew7234: good catch! I added the `interItemDelay` flag you suggested so that we don't have to modify all our http clients to throttle requests.

cmd/analyzer/analyzer.go
```diff
@@ -326,7 +328,7 @@ func NewService(cfg *config.AnalysisConfig) (*Service, error) { //nolint:gocyclo
 			if err1 != nil {
 				return nil, err1
 			}
-			return evmtokens.NewMain(common.RuntimeEmerald, sourceClient, dbClient, logger)
+			return evmtokens.NewAnalyzer(common.RuntimeEmerald, cfg.Analyzers.EmeraldEvmTokens.BatchSize, cfg.Analyzers.EmeraldEvmTokens.StopOnEmptyBatch, cfg.Analyzers.EmeraldEvmTokens.Interval, cfg.Analyzers.EmeraldEvmTokens.InterItemDelay, sourceClient, dbClient, logger)
```
mitjat, line 331: I see now why you were asking about grouping constructor params into struct :), this is quite verbose.

I'm in favor of going with ItemBasedAnalyzerConfig here, but your call whether to do it. I think your TODO in the PR description is about that?

mitjat (approved):
Thank you! LGTM assuming e2e tests pass.

cmd/analyzer/analyzer.go
Andrew7234: yeah it's getting hairy - updated to use the cfg struct

analyzer/item/item.go
```diff
@@ -0,0 +1,228 @@
...
+	var wg sync.WaitGroup
+	batches := make([]*storage.QueryBatch, len(items))
+
+	for i, it := range items {
+		wg.Add(1)
+		go func(idx int, item Item) {
+			defer wg.Done()
+			batch := storage.QueryBatch{}
+			if err := a.processor.ProcessItem(batchCtx, &batch, item); err != nil {
+				return
```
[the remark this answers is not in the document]
Andrew7234: updated!

analyzer/item/item.go
```diff
@@ -0,0 +1,228 @@
...
+			}
+			batches[idx] = &batch
```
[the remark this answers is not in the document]
Andrew7234: updated!

analyzer/item/item.go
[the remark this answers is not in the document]
mitjat: I don't see it, maybe I'm missing something? One possible solution is
```suggestion
				batches[idx] = storage.QueryBatch{} // avoid having a nil batch
				return
```

analyzer/node_stats/node_stats.go
```diff
@@ -2,96 +2,81 @@ package nodestats
...
+		layer,
+		height,
+	)
+
+	return nil
 }
 
-func (m main) Name() string {
-	return nodeStatsAnalyzerName
+func (p *processor) QueueLength(ctx context.Context) (int, error) {
```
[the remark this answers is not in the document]
Andrew7234: Good idea - should be simple enough now that we have the `node_stats` analyzer. I'll file a ticket :)

PR #513, opened 2023-09-01
[api] change format of latest_update to int64

Andrew7234:
Follow up of https://github.com/oasisprotocol/nexus/pull/496#discussion_r1309539679

### DIFF
--- a/api/spec/v1.yaml
+++ b/api/spec/v1.yaml
@@ -1155,7 +1155,7 @@ components:
 
     Status:
       type: object
-      required: [latest_block, latest_node_block, latest_block_time, latest_update]
+      required: [latest_block, latest_node_block, latest_block_time, latest_update_age_ms]
       properties:
         latest_block:
           type: integer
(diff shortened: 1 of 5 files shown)


lukaw3d (approved):

api/spec/v1.yaml
```diff
@@ -1175,9 +1175,10 @@ components:
           description: The RFC 3339 formatted consensus time of when the most recently indexed block was produced.
           example: *iso_timestamp_2
         latest_update:
```
mitjat, line 1177: Optional: The field implies an absolute timestamp to me. You might want to consider `latest_update_age[_ms]` or similar.

mitjat (approved):
Thank you!

api/spec/v1.yaml
lukaw3d: I agree. Naming-wise I vote "time_since_latest_block"

lukaw3d: oh, too easily confused with time since chain produced a block.

Andrew7234: updated to `latest_update_age_ms`

PR #514, opened 2023-09-06
NFT APIs

pro-wh:
this first part has a way to list NFTs by contract address


mitjat:
When this gets closer to stable, let's loop in the FE folks, and possibly Matevz as effectively our developer advocate.

api/spec/v1.yaml
```diff
@@ -2572,6 +2596,59 @@ components:
...
+          required: [evm_nfts]
+          properties:
+            evm_nfts:
+              type: array
+              items:
+                $ref: '#/components/schemas/EvmNft'
+              description: A list of L2 EVM NFT (ERC-721, ...) instances.
+          description: A list of NFT instances.
+
+    EvmNft:
```
mitjat, line 2612: It feels awkward that we have some info here that's specific to an individual tokenID (like metadata_uri and image), and some that is a property of the token contract as a whole (most fields).

ERC721 are also already listed under `/{runtime}/evm_tokens`. How would it be if we expanded that response type to optionally include any contract-level fields that this PR newly introduces? Then this type could focus on just the per-ID properties (and maybe include the per-contract properties as a subtree, if we want to be verbose).

api/spec/v1.yaml
```diff
@@ -1036,6 +1036,30 @@ paths:
                 $ref: '#/components/schemas/TokenHolderList'
         <<: *common_error_responses
 
+  /{runtime}/evm_nfts/{address}:
```
mitjat, line 1039: `/{runtime}/evm_nfts/{address}` and `/{runtime}/evm_tokens/{address}` will have a fair amount of overlap in terms of info provided. Which is not the end of the world, but it is potentially confusing.

WDYT about having this under `/{runtime}/evm_tokens/{address}/nfts`? Especially in combination with my comment below.

mitjat (commented):
:+1: 
Just trying to figure out how the relationship between individual NFTs and ERC721 contracts should be reflected in our API.

csillag:
I am getting SWAGGER OpenAPI velidation errors.

> 🍻 Start orval v6.17.0 - A swagger client generator for typescript
⚠  SyntaxError: Swagger schema validation failed.
  #/paths/~1{runtime}~1evm_nfts~1{address}/get/responses/200/content/application~1json/schema/allOf/1/properties/evm_nfts/items/required must NOT have fewer than 1 items
  #/paths/~1{runtime}~1evm_nfts~1{address}/get/responses/200/content/application~1json/schema/allOf/1/properties/evm_nfts/items must have required property '$ref'
  #/paths/~1{runtime}~1evm_nfts~1{address}/get/responses/200/content/application~1json/schema/allOf/1/properties/evm_nfts/items must match exactly one schema in oneOf
  #/paths/~1{runtime}~1evm_nfts~1{address}/get/responses/200/content/application~1json/schema/allOf/1/properties/evm_nfts must have required property '$ref'
  #/paths/~1{runtime}~1evm_nfts~1{address}/get/responses/200/content/application~1json/schema/allOf/1/properties/evm_nfts must match exactly one schema in oneOf
  #/paths/~1{runtime}~1evm_nfts~1{address}/get/responses/200/content/application~1json/schema/allOf/1 must have required property '$ref'
  #/paths/~1{runtime}~1evm_nfts~1{address}/get/responses/200/content/application~1json/schema/allOf/1 must match exactly one schema in oneOf
  #/paths/~1{runtime}~1evm_nfts~1{address}/get/responses/200/content/application~1json/schema must have required property '$ref'
  #/paths/~1{runtime}~1evm_nfts~1{address}/get/responses/200/content/application~1json/schema must match exactly one schema in oneOf
  #/paths/~1{runtime}~1evm_nfts~1{address}/get/responses/200 must have required property '$ref'
  #/paths/~1{runtime}~1evm_nfts~1{address}/get/responses/200 must match exactly one schema in oneOf
  #/components/schemas/EvmNftList/allOf/1/properties/evm_nfts/items/required must NOT have fewer than 1 items
  #/components/schemas/EvmNftList/allOf/1/properties/evm_nfts/items must have required property '$ref'
  #/components/schemas/EvmNftList/allOf/1/properties/evm_nfts/items must match exactly one schema in oneOf
  #/components/schemas/EvmNftList/allOf/1/properties/evm_nfts must have required property '$ref'
  #/components/schemas/EvmNftList/allOf/1/properties/evm_nfts must match exactly one schema in oneOf
  #/components/schemas/EvmNftList/allOf/1 must have required property '$ref'
  #/components/schemas/EvmNftList/allOf/1 must match exactly one schema in oneOf
  #/components/schemas/EvmNftList must have required property '$ref'
  #/components/schemas/EvmNftList must match exactly one schema in oneOf
  #/components/schemas/EvmNft/required must NOT have fewer than 1 items
  #/components/schemas/EvmNft must have required property '$ref'
  #/components/schemas/EvmNft must match exactly one schema in oneOf


These all seem to be related to the new additions; the version on the master branch passes without any issues.

api/spec/v1.yaml
pro-wh: this is for later when there's get NFT instances owned by an account

pro-wh:
yeah these look like they're related to the additions. I'll check on these tomorrow

csillag:
> yeah these look like they're related to the additions. I'll check on these tomorrow

If there will be an update, it would be very nice to sneak it the `owner` field, too.

pro-wh:
gonna add the owner field in #529

pro-wh:
oh it's because I forgot to fill in the `required`. required fields are the contract address and id. other fields' availability is:

- eth contract addr: available if we have an ethereum address preimage. it's expected that we have it though, as we only scan EVM NFTs currently. just not architecturally required
- token name, symbol: available if the contract answers these metadata queries
- token type: available after token analyzer runs
- token total supply: probably always available, but not architecturally required for some reason
- metadata URI: if contract answers a query for it
- metadata accessed: if we've tried to download the metadata doc
- name, description, image: if we successfully get them from the metadata doc

pro-wh:
required fields added

api/spec/v1.yaml
```diff
@@ -2572,6 +2596,60 @@ components:
...
+          <<: *BigIntType
+        id:
+          <<: *BigIntType
+        metadata_uri:
+          type: string
+        metadata_accessed:
+          type: string
+        name:
+          type: string
+          description: Identifies the asset to which this NFT represents
```
mitjat, line 2642: Nit here and a few lines down: No "to"

api/spec/v1.yaml
mitjat: Right. I didn't express myself clearly. I was talking about the _datatype_ of the `/{runtime}/evm_tokens` response, which is `#/components/schemas/EvmToken[List]`. Would it make sense to reuse `#/components/schemas/EvmToken` as a subtree of the new `EvmNft` type so that the other fields of `EvmNft` are mostly just things that are specific to that NFT ID?

api/spec/v1.yaml
pro-wh: https://github.com/ethereum/EIPs/blob/master/EIPS/eip-721.md?plain=1#L206

can you file upstream :pray:

pro-wh: https://github.com/ethereum/EIPs/commit/223e46c7ce39de62c54bb6a2f9d1e630d3aff079#diff-93c4038537e3c4794bc42d0cd3e75d12876b722a743e0932c28549b23ea28edeL208 interesting, the "to" comes from when it was "DAR" that granted access to an asset

pro-wh: huh, but even back then, shouldn't it have been ownership "of" an asset? :thinking:

mitjat: Filed https://github.com/ethereum/EIPs/pull/7795. We can fix here independently tho.

api/spec/v1.yaml
csillag: > Would it make sense to reuse `#/components/schemas/EvmToken` as a subtree of the new `EvmNft` type so that the other fields of `EvmNft` are mostly just things that are specific to that NFT ID?

I think that's a great idea.

> maybe include the per-contract properties as a subtree, if we want to be verbose

Yes I definitely think that we want to be verbose, otherwise we will potentially have to make multiple subsequent follow-up queries from the client side to get the missing data.

pro-wh: oh is that structure bigger and does it contain other stuff you want for the UI?

api/spec/v1.yaml
pro-wh: oh yeah that'd be good

api/spec/v1.yaml
csillag: Not bigger, but different. In same cases, we need one, on some cases, the other, in same cases, both.

api/spec/v1.yaml
```diff
@@ -1036,6 +1036,30 @@ paths:
                 $ref: '#/components/schemas/TokenHolderList'
         <<: *common_error_responses
 
+  /{runtime}/evm_nfts/{address}:
+    get:
+      summary: |
+        Returns the list of non-fungible token (NFT) instances of an EVM (ERC-721, ...) token.
+        This endpoint does not verify that `address` is actually an EVM token; if it is not, it will simply return an empty list.
+      parameters:
```
csillag, line 1044: Would it be possible to have an optional parameter for specifying the instance ID, for getting data about a specific NFT instance?

(Or is there / will there be a separate API endpoint for that?)

pro-wh: let's add another route with /id at the end

csillag: That is also fine

pro-wh: oh or have we more commonly done it by parameter?

csillag: No, I think having it in the path is perfect

api/spec/v1.yaml
```diff
@@ -2572,6 +2596,60 @@ components:
...
+      type: object
+      required:
+        - contract_addr
+        - id
+      properties:
+        contract_addr:
+          type: string
+          description: The Oasis address of this NFT's contract.
+          example: 'oasis1qp2hssandc7dekjdr6ygmtzt783k3gn38uupdeys'
+        eth_contract_addr:
```
pro-wh, line 2622: inconsistent, should be contract_eth_addr, but it's like this for consistency with EvmToken

api/spec/v1.yaml
pro-wh: I added a commit to fix it here. too bad about the lint hell in the EIPs repository

api/spec/v1.yaml
pro-wh: moved "token" (terminology disputed) fields to a `token` subobject

pro-wh:
"token" info moved to subobject, matching evm tokens json schema. however, note that that api was designed earlier, where more fields are required. those fields will have the Go default value when the data is not available. e.g. the zero ethereum address if we somehow don't know the preimage or token type unknown instead of null if we haven't checked the contract yet

api/spec/v1.yaml
pro-wh: moved

api/spec/v1.yaml
pro-wh: now uses the EvmToken struct directly

pro-wh:
moved to /{runtime}/evm_tokens/{address}/nfts

api/spec/v1.yaml
pro-wh: adding in #538

pro-wh: gonna resolve here. let's discuss there

storage/client/queries/queries.go
```diff
@@ -494,6 +494,47 @@ const (
...
+			chain.address_preimages.address = chain.evm_nfts.token_address
+		LEFT JOIN chain.evm_tokens USING (runtime, token_address)
+		LEFT JOIN token_holders USING (token_address)
+		LEFT JOIN chain.evm_contracts ON
+			chain.evm_contracts.runtime = chain.evm_tokens.runtime AND
+			chain.evm_contracts.contract_address = chain.evm_tokens.token_address
+		WHERE
+			chain.evm_nfts.runtime = $1::runtime AND
+			chain.evm_nfts.token_address = $2::oasis_addr
+		LIMIT $3::bigint
```
mitjat, line 535: I noticed this in some other PR that I reviewed first, but repeating here just in case: Needs `ORDER BY token_address, nft_id`

api/spec/v1.yaml
```diff
@@ -2572,6 +2596,47 @@ components:
...
+      type: object
+      required:
+        - token
+        - contract_addr
+        - id
+      properties:
+        token:
+          $ref: '#/components/schemas/EvmToken'
+        id:
+          <<: *BigIntType
```
mitjat, line 2622:
```suggestion
          <<: *BigIntType
          description: The instance ID of this NFT within the collection represented by `token`.
```

WDYT about this as an attempt to tie together the terminology / make our choices a little more explicit?

api/spec/v1.yaml
```diff
@@ -1036,6 +1036,30 @@ paths:
                 $ref: '#/components/schemas/TokenHolderList'
         <<: *common_error_responses
 
+  /{runtime}/evm_tokens/{address}/nfts:
```
mitjat, line 1039: Can you add this to `e2e_regresion/run.sh` please?

mitjat (approved):
Thank you for the additional changes!

A reply to your comment:

> "token" info moved to subobject, matching evm tokens json schema.
> note that that api was designed earlier, where more fields are required.

Now is the "best" time to change that in a backwards-incompatible way -- best compared to any future time. I'm sure @csillag can accommodate optionality on the FE, and we don't have other clients for now.
OTOH if the only two potentially-defaulted fields are eth_address and token_type, I wouldn't worry. token_type _has_ to be 721 for us to even list the NFT, presumably? And not knowing the preimage would pretty much imply a bug in nexus, because all events via which we observe NFTs are encoded with eth addresses, so we observe preimages all the time. So I'm not worried about that corner case either.

pro-wh:
metadata_accessed lmao

api/spec/v1.yaml
pro-wh: might have to do this at the end of the PR train, as this'll take more work to normalize the metadata_accessed field (it's a timestamp). should probably also cache network items for the regression test x_x

pro-wh: ok, discussed today, there's a way to blank it out in jq. added that. no cache. passes locally

storage/client/queries/queries.go
pro-wh: added, thanks

api/spec/v1.yaml
pro-wh: sounds good, thanks

pro-wh:
changes worth looking at:

1. new e2e regression test hits network resources. time bomb warning
2. db fixes, let me know if it's right

storage/migrations/22_evm_nfts_grant.up.sql
```diff
@@ -0,0 +1,7 @@
+BEGIN;
+
+-- Grant others read-only use.
+GRANT SELECT ON ALL TABLES IN SCHEMA chain TO PUBLIC;
+GRANT EXECUTE ON ALL FUNCTIONS IN SCHEMA chain TO PUBLIC;
```
Andrew7234, line 5: nit: If `21_evm_nfts.up.sql` hasn't been applied yet to prod, I'd also be fine with merging this with that migration. This way is also completely fine too.

pro-wh: it was in 0.1.16 :grimacing:

PR #515, opened 2023-09-12
analyzer/runtime: Index (un)delegate events, txs

mitjat:
Adds indexing of txs that call the `consensus.Delegate` and `consensus.Undelegate` methods, and of `consensus_accounts` SDK events of type `Delegate`, `UndelegateStart`, `UndelegateDone`.

TODO: Testing. For starters, I'll run this on the last few weeks of data on testnet, which is where these txs and events first appeared.

### DIFF
--- a/analyzer/runtime/extract.go
+++ b/analyzer/runtime/extract.go
@@ -364,13 +364,38 @@ func ExtractRound(blockHeader nodeapi.RuntimeBlockHeader, txrs []nodeapi.Runtime
 					blockTransactionData.Body = body
 					amount = body.Amount.Amount
 					if body.To != nil {
-						// Beware, this is the address of an account in the consensus
-						// layer, not an account in the runtime that generated this event.
+						// This is the address of an account in the consensus
+						// layer, not an account in the runtime that generated this tx.
 						// We do not register it as a preimage.
 						if to, err = uncategorized.StringifySdkAddress(body.To); err != nil {
 							return fmt.Errorf("to: %w", err)
 						}
 					}
+					return nil
+				},
+				ConsensusAccountsDelegate: func(body *consensusaccounts.Delegate) error {
+					blockTransactionData.Body = body
+					amount = body.Amount.Amount
+					// This is the address of an account in the consensus
+					// layer, not an account in the runtime that generated this tx.
+					// We do not register it as a preimage.
+					if to, err = uncategorized.StringifySdkAddress(&body.To); err != nil {
+						return fmt.Errorf("to: %w", err)
+					}
+					return nil
+				},
+				ConsensusAccountsUndelegate: func(body *consensusaccounts.Undelegate) error {
+					blockTransactionData.Body = body
+					// NOTE: The `from` and `to` addresses have swapped semantics compared to most other txs:
+					// Assume R is a runtime address and C is a consensus address. The inverse of Delegate(from=R, to=C) is Undelegate(from=C, to=R).
+					if _, err := registerRelatedSdkAddress(blockTransactionData.RelatedAccountAddresses, &body.From); err != nil {
+						return fmt.Errorf("from: %w", err)
+					}
+					// The `to` is implicitly the account that created this tx. Ref: https://github.com/oasisprotocol/oasis-sdk/blob/eb97a8162f84ae81d11d805e6dceeeb016841c27/runtime-sdk/src/modules/consensus_accounts/mod.rs#L465-L465
+					to = blockTransactionData.SignerData[0].Address
+					// The `amount` (of tokens) is not contained in the body, only `shares` is. We do not have sufficient information
+					// to convert `shares` to `amount`; we'd need the total amount of shares and tokens in the `from` account.
+					// Assign nothing to `amount`.
 					return nil
 				},
 				EVMCreate: func(body *sdkEVM.Create, ok *[]byte) error {
@@ -455,6 +480,10 @@ func ExtractRound(blockHeader nodeapi.RuntimeBlockHeader, txrs []nodeapi.Runtime
 					}
 
 					// TODO: maybe parse known token methods (ERC-20 etc)
+					return nil
+				},
+				UnknownMethod: func(methodName string) error {
+					logger.Warn("unknown tx method, skipping tx-specific analysis", "tx_method", methodName)
 					return nil
 				},
 			}); err != nil {
@@ -548,15 +577,11 @@ func extractEvents(blockData *BlockData, relatedAccountAddresses map[apiTypes.Ad
 				if err1 != nil {
 					return fmt.Errorf("to: %w", err1)
 				}
-				eventRelatedAddresses := map[apiTypes.Address]bool{}
-				for _, addr := range []apiTypes.Address{fromAddr, toAddr} {
-					eventRelatedAddresses[addr] = true
-				}
 				eventData := EventData{
 					Type:             apiTypes.RuntimeEventTypeAccountsTransfer,
 					Body:             event.Transfer,
 					WithScope:        ScopedSdkEvent{Accounts: event},
-					RelatedAddresses: eventRelatedAddresses,
+					RelatedAddresses: map[apiTypes.Address]bool{fromAddr: true, toAddr: true},
 				}
 				extractedEvents = append(extractedEvents, &eventData)
 			}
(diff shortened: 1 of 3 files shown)


analyzer/runtime/visitors.go
mitjat, line 1: The changes in this file are purely boilerplate.

analyzer/runtime/extract.go
```diff
@@ -364,15 +364,40 @@ func ExtractRound(blockHeader nodeapi.RuntimeBlockHeader, txrs []nodeapi.Runtime
...
+				ConsensusAccountsDelegate: func(body *consensusaccounts.Delegate) error {
+					blockTransactionData.Body = body
+					amount = body.Amount.Amount
+					// This is the address of an account in the consensus
+					// layer, not an account in the runtime that generated this tx.
+					// We do not register it as a preimage.
+					if to, err = uncategorized.StringifySdkAddress(&body.To); err != nil {
+						return fmt.Errorf("to: %w", err)
+					}
+					return nil
```
mitjat, line 385: This is analogous to (and copy pasted from) our existing handling of `consensusaccounts.Withdraw` calls

analyzer/runtime/extract.go
```diff
@@ -364,15 +364,40 @@ func ExtractRound(blockHeader nodeapi.RuntimeBlockHeader, txrs []nodeapi.Runtime
...
+					// NOTE: The `from` and `to` addresses have swapped semantics compared to most other txs:
+					// Assume R is a runtime address and C is a consensus address. The inverse of Delegate(from=R, to=C) is Undelegate(from=C, to=R).
+					if _, err := registerRelatedSdkAddress(blockTransactionData.RelatedAccountAddresses, &body.From); err != nil {
+						return fmt.Errorf("from: %w", err)
+					}
+					// The `to` is implicitly the account that created this tx. Ref: https://github.com/oasisprotocol/oasis-sdk/blob/eb97a8162f84ae81d11d805e6dceeeb016841c27/runtime-sdk/src/modules/consensus_accounts/mod.rs#L465-L465
+					to = blockTransactionData.SignerData[0].Address
+					// The `amount` (of tokens) is not contained in the body, only `shares` is. We do not have sufficient information
+					// to convert `shares` to `amount`; we'd need the total amount of shares and tokens in the `from` account.
+					// Assign nothing to `amount`.
```
mitjat, line 398: Can anybody confirm this?

Also, if true, should we store the number of shares in the `amount` variable even though the semantics are different? Both `to` and `amount` are loosely defined and introduced by the indexer; they're a kind of best-effort intentionally-over-generalized fields to help navigate the txs quickly almost regardless of their call method.

The `amount` variable contents end up in the `amount` column in our DB, and eventually in the `amount` field of each tx in the API, but the API docs describe the above looseness pretty well:
> A reasonable "amount" associated with this transaction, if applicable. The meaning varies based on the transaction method.

Still, might be too much of a stretch? And potentially particularly confusing because shares and token amounts are so closely related?

analyzer/runtime/extract.go
```diff
@@ -364,15 +364,40 @@ func ExtractRound(blockHeader nodeapi.RuntimeBlockHeader, txrs []nodeapi.Runtime
...
+				},
+				ConsensusAccountsUndelegate: func(body *consensusaccounts.Undelegate) error {
+					blockTransactionData.Body = body
+					// NOTE: The `from` and `to` addresses have swapped semantics compared to most other txs:
+					// Assume R is a runtime address and C is a consensus address. The inverse of Delegate(from=R, to=C) is Undelegate(from=C, to=R).
+					if _, err := registerRelatedSdkAddress(blockTransactionData.RelatedAccountAddresses, &body.From); err != nil {
+						return fmt.Errorf("from: %w", err)
+					}
+					// The `to` is implicitly the account that created this tx. Ref: https://github.com/oasisprotocol/oasis-sdk/blob/eb97a8162f84ae81d11d805e6dceeeb016841c27/runtime-sdk/src/modules/consensus_accounts/mod.rs#L465-L465
+					to = blockTransactionData.SignerData[0].Address
```
mitjat, line 395: While the `to` address according to tx-internal semantics is indeed the tx sender (see NOTE comment above), it might make more sense to store `body.From` in our `to` field? The `to` field is internally defined by the indexer (though exposed through the API), with [varying meanings](https://github.com/oasisprotocol/nexus/blob/2ea104969cc4242ca896a7d50bd313024c849432/api/spec/v1.yaml#L2368-L2375) but generally being "the address that is most relevant to this tx, other than the sender".

analyzer/runtime/extract.go
```diff
@@ -548,15 +577,11 @@ func extractEvents(blockData *BlockData, relatedAccountAddresses map[apiTypes.Ad
...
-				eventRelatedAddresses := map[apiTypes.Address]bool{}
-				for _, addr := range []apiTypes.Address{fromAddr, toAddr} {
-					eventRelatedAddresses[addr] = true
-				}
 				eventData := EventData{
 					Type:             apiTypes.RuntimeEventTypeAccountsTransfer,
 					Body:             event.Transfer,
 					WithScope:        ScopedSdkEvent{Accounts: event},
-					RelatedAddresses: eventRelatedAddresses,
+					RelatedAddresses: map[apiTypes.Address]bool{fromAddr: true, toAddr: true},
```
mitjat, line 584: Unrelated simplification

analyzer/runtime/extract.go
```diff
@@ -617,6 +642,57 @@ func extractEvents(blockData *BlockData, relatedAccountAddresses map[apiTypes.Ad
 				extractedEvents = append(extractedEvents, &eventData)
 				// .To is from another chain, so exclude?
```
mitjat, line 643: I did not follow this pattern with (un)delegate events, and I'm tempted to undo it for these "old" events to, i.e. I'm tempted to always extract `.To` as a related address. There's not really a downside to overindexing, other than bloating our DB a bit, as far as I know. And it might come handy to be able to look up `consensusaccounts` events related to a certain consensus account.

analyzer/runtime/extract.go
```diff
@@ -364,15 +364,40 @@ func ExtractRound(blockHeader nodeapi.RuntimeBlockHeader, txrs []nodeapi.Runtime
...
 						if to, err = uncategorized.StringifySdkAddress(body.To); err != nil {
 							return fmt.Errorf("to: %w", err)
 						}
 					}
 					return nil
 				},
+				ConsensusAccountsDelegate: func(body *consensusaccounts.Delegate) error {
+					blockTransactionData.Body = body
+					amount = body.Amount.Amount
+					// This is the address of an account in the consensus
```
mitjat, line 379: Can someone confirm this please? `to` is typically a validator account, right? And if one delegated to a non-validator account, nothing bad would happen, it just wouldn't result in any rewards over time.

kostko: Correct for both.

analyzer/runtime/extract.go
kostko: Do not conflate shares and amount as this can lead to incorrect values and user confusion.

You can only know the final amount once the undelegation completes as nodes may be slashed while debonding. However you can compute the amount at any point in time by inspecting the relevant share pool (see the CLI).

analyzer/runtime/extract.go
pro-wh: :+1:

analyzer/runtime/extract.go
pro-wh: ya ok

analyzer/runtime/extract.go
pro-wh: :skull: is this what we do for consensus undelegate?

pro-wh (commented):
lg but let's approve when we're sure there will be no regrets about what "to" in an undelegate means

analyzer/runtime/extract.go
mitjat: By "consensus undelegate", you mean the `ReclaimEscrow` method from [here](https://github.com/oasisprotocol/oasis-core/blob/6dcc5f29ff908f8c6c5bd3481f20969db009c3be/go/staking/api/api.go#L77-L81), right?
When [inserting consensus txs into the DB](https://github.com/oasisprotocol/nexus/blob/2ea104969cc4242ca896a7d50bd313024c849432/analyzer/consensus/consensus.go#L332-L345), we make no attempt to construct a reasonably generalizable `to` field. So I guess there's no prior art :/

I'm actually inclined to set `to = body.From`. Because otherwise there's no added value in setting the `to` (we're storing `body.To` in the tx sender field already), and so it will be hard to surface undelegations relating to a specific validator.

Edit: Doing so after a slack discussion.

analyzer/runtime/extract.go
mitjat: Thanks. I kept the shares in the `amount` field, as was coded initially; I agree we better be careful here.
I updated the comment with your explanation about slashing.

analyzer/runtime/extract.go
mitjat: After more discussion on Slack, we can't think of reasons why this would be problematic. So I'm adding `.To` (for `Withdraw` event) and `.From` (for `Deposit`) to the list of event-related addrs.

mitjat:
After discussions, I slightly tweaked what addresses we index. Reviewers, please take another look. Thanks!

analyzer/runtime/extract.go
```diff
@@ -364,15 +364,42 @@ func ExtractRound(blockHeader nodeapi.RuntimeBlockHeader, txrs []nodeapi.Runtime
...
+						// This is the address of an account in the consensus
+						// layer, not an account in the runtime that generated this tx.
 						// We do not register it as a preimage.
 						if to, err = uncategorized.StringifySdkAddress(body.To); err != nil {
 							return fmt.Errorf("to: %w", err)
 						}
 					}
 					return nil
 				},
+				ConsensusAccountsDelegate: func(body *consensusaccounts.Delegate) error {
```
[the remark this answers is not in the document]
ptrus: I tnink no need, since the appropriate `account` burn/mint events will be emitted by the SDK on the delegation and once the funds are reclaimed. And those are already handled in https://github.com/oasisprotocol/nexus/blob/fb5d2fb1ed3836f0bf15d2edd076fd8b255ba9e6/analyzer/runtime/accounts.go#L18-L31

analyzer/runtime/extract.go
```diff
@@ -364,15 +364,42 @@ func ExtractRound(blockHeader nodeapi.RuntimeBlockHeader, txrs []nodeapi.Runtime
...
+					// NOTE: The `from` and `to` addresses have swapped semantics compared to most other txs:
+					// Assume R is a runtime address and C is a consensus address (likely a validator). The inverse of Delegate(from=R, to=C) is Undelegate(from=C, to=R).
+					// In Undelegate semantics, the inexistent `body.To` is implicitly the account that created this tx, i.e. the delegator R.
+					// Ref: https://github.com/oasisprotocol/oasis-sdk/blob/eb97a8162f84ae81d11d805e6dceeeb016841c27/runtime-sdk/src/modules/consensus_accounts/mod.rs#L465-L465
+					// However, we instead expose `body.From` as the DB/API `to` for consistency with `Delegate`, and because it is more useful: the delegator R is already indexed in the tx sender field.
+					if to, err = registerRelatedSdkAddress(blockTransactionData.RelatedAccountAddresses, &body.From); err != nil {
+						return fmt.Errorf("from: %w", err)
+					}
+					// The `amount` (of tokens) is not contained in the body, only `shares` is. There isn't sufficient information
+					// to convert `shares` to `amount` until the undelegation actually happens (= UndelegateDone event); in the meantime,
```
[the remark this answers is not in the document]
ptrus: I think that no need, since SDK will also emit a `Mint` event on the undelegation: https://github.com/oasisprotocol/oasis-sdk/blob/fac9e70f9444e0f20e62ca56cc597abfd3f54e38/runtime-sdk/src/modules/consensus_accounts/mod.rs#L788

analyzer/runtime/extract.go
ptrus, line 376: This should probably check if the event contains an error https://github.com/oasisprotocol/oasis-sdk/blob/fac9e70f9444e0f20e62ca56cc597abfd3f54e38/client-sdk/go/modules/consensusaccounts/types.go#L201


Actually all `consensus_account` events can contain a potential consensus error in which indicates a failed event, and should probably be skipped here?

See: https://github.com/oasisprotocol/oasis-sdk/blob/fac9e70f9444e0f20e62ca56cc597abfd3f54e38/client-sdk/go/modules/consensusaccounts/types.go#L233-L239

and: https://github.com/oasisprotocol/oasis-sdk/blob/fac9e70f9444e0f20e62ca56cc597abfd3f54e38/runtime-sdk/src/modules/consensus_accounts/mod.rs#L609-L625

for examples.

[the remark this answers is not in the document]
mitjat: Correct. (With a small caveat -- there's an additional Transfer involved.) I was also unsure of this when I started this PR, and it took me some spelunking to find the explanation before I was comfortable writing the code. Good question, Andy, and worth spelling out the answer for future maintainers. See the new "LESSON" comment in the code for an explanation of events.

mitjat: Hmm ... it does feel weird to basically ignore errors, but I'm not sure the indexer needs to do anything special about `consensusaccounts` events with a non-nil `Error`:
- There is no dead reckoning to adjust (or suppress); that all happens via other, co-emitted events
- We [store](https://github.com/oasisprotocol/nexus/blob/f9aae40e235f3dca08bb2845f8fb124603b8f310/analyzer/runtime/extract.go#L397-L397) (and then expose via API) the entire body of the event as JSON; [example](https://index-staging.oasislabs.com/v1/emerald/events?type=consensus_accounts.deposit). If the `Error` field were set, we'd include it there.

Maybe I'm forgetting something?

ptrus: You're right

ptrus (approved):

analyzer/runtime/extract.go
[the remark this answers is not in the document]
Andrew7234: thanks for adding the comnent! v helpful

Andrew7234 (approved):

PR #518, opened 2023-09-14
nodeapi: cache deterministic failure responses

Andrew7234:
https://app.clickup.com/t/8692qau0v

Updates `EvmSimulateCall` to return a wrapped result that can also include a known error.

Testing: Ran the e2e_regression tests with non-block analyzers locally; no issues with running in offline mode.

random thoughts:
- Decided against using FallibleResponse for all node queries since it introduces a lot of unnecessary bloat/boilerplate.
- Think more about how/when we actually want to cache node errors. Are there any cases of `EvmSimulateCall` that might error 'deterministically' but later succeed? ---- No, since `EvmSimulateCall` is queried at a given `round`
- For evm.Code, we currently ignore any errors and retry later, which means that we don't want to cache the error (except maaybe in testing environs). --- Yes, the only errors evm.Code will return are transient ones. If a contract does not exist at a given address, evm.Code will successfully return an empty byte array, not a deterministic error type.

### DIFF
--- a/storage/oasis/nodeapi/api.go
+++ b/storage/oasis/nodeapi/api.go
@@ -209,7 +209,7 @@ type Proposal governance.Proposal
 // Like ConsensusApiLite, but for the runtime API.
 type RuntimeApiLite interface {
 	GetEventsRaw(ctx context.Context, round uint64) ([]RuntimeEvent, error)
-	EVMSimulateCall(ctx context.Context, round uint64, gasPrice []byte, gasLimit uint64, caller []byte, address []byte, value []byte, data []byte) ([]byte, error)
+	EVMSimulateCall(ctx context.Context, round uint64, gasPrice []byte, gasLimit uint64, caller []byte, address []byte, value []byte, data []byte) (*FallibleResponse, error)
 	EVMGetCode(ctx context.Context, round uint64, address []byte) ([]byte, error)
 	GetBlockHeader(ctx context.Context, round uint64) (*RuntimeBlockHeader, error)
 	GetTransactionsWithResults(ctx context.Context, round uint64) ([]RuntimeTransactionWithResults, error)
--- a/storage/oasis/nodeapi/file/kvstore.go
+++ b/storage/oasis/nodeapi/file/kvstore.go
@@ -157,7 +157,7 @@ func fetchTypedValue[Value any](cache KVStore, key CacheKey, value *Value) error
 	}
 	err = cbor.Unmarshal(raw, value)
 	if err != nil {
-		return fmt.Errorf("failed to unmarshal key %s from cache into %T: %v", key.Pretty(), value, err)
+		return fmt.Errorf("failed to unmarshal the value for key %s from cache into %T: %v; raw value was %x", key.Pretty(), value, err, raw)
 	}
 
 	return nil
--- a/storage/oasis/nodeapi/universal_runtime.go
+++ b/storage/oasis/nodeapi/universal_runtime.go
@@ -10,6 +10,7 @@ import (
 	coreCommon "github.com/oasisprotocol/oasis-core/go/common"
 	"github.com/oasisprotocol/oasis-core/go/common/cbor"
 	"github.com/oasisprotocol/oasis-core/go/common/crypto/hash"
+	"github.com/oasisprotocol/oasis-core/go/common/errors"
 	roothash "github.com/oasisprotocol/oasis-core/go/roothash/api/block"
 	coreRuntimeClient "github.com/oasisprotocol/oasis-core/go/runtime/client/api"
 	connection "github.com/oasisprotocol/oasis-sdk/client-sdk/go/connection"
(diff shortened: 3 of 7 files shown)


storage/oasis/nodeapi/api.go
```diff
@@ -206,10 +206,54 @@ type Proposal governance.Proposal
 // ....................  Runtimes  ....................
 // ....................................................
 
+// FallibleResponse stores the response to a node query. If the call was successful,
+// the response is stored in `result`. If the call failed due to a deterministic
+// error, eg call reverted, the error is stored in `deterministicErr`. This struct
+// should _not_ store transient errors or any other errors not considered to be a
+// valid node response.
+type FallibleResponse[Response any] struct {
+	Result           Response
```
mitjat, line 215:
```suggestion
	Ok           *Response
```

Suggestions:
- I find `Result` confusing because it's what Rust uses for `FallibleResponse`.
- Add pointer to allow clearly expressing a missing ok-response.

Also, nit: I find it slightly odd to talk about deterministic errors in the docstring here; I'd probably put it in the docstring of `EVMSimulateCall`, because 1) that's where the semantics of this object are ultimately determined and 2) it can be confusing to consumers to have a return value `(FallibleResponse, error)` that can express errors in two different ways, without an explanation. OTOH the explanation is here, if they only dig a tiny bit. Your call.

storage/oasis/nodeapi/api.go
```diff
@@ -206,10 +206,54 @@ type Proposal governance.Proposal
 // ....................  Runtimes  ....................
 // ....................................................
 
+// FallibleResponse stores the response to a node query. If the call was successful,
```
mitjat, line 209: This file is already kinda long and might need to eventually be split. But for now, I've been trying to at least keep it limited to the top-level types of the node API to keep it easily navigable. Can you please move the newly added lines into a new file, e.g. `evm.go`?

storage/oasis/nodeapi/universal_runtime.go
```diff
@@ -135,8 +136,22 @@ func (rc *UniversalRuntimeApiLite) GetEventsRaw(ctx context.Context, round uint6
 	return evs, nil
 }
 
-func (rc *UniversalRuntimeApiLite) EVMSimulateCall(ctx context.Context, round uint64, gasPrice []byte, gasLimit uint64, caller []byte, address []byte, value []byte, data []byte) ([]byte, error) {
-	return rc.sdkClient.Evm.SimulateCall(ctx, round, gasPrice, gasLimit, caller, address, value, data)
+func (rc *UniversalRuntimeApiLite) EVMSimulateCall(ctx context.Context, round uint64, gasPrice []byte, gasLimit uint64, caller []byte, address []byte, value []byte, data []byte) (*FallibleResponse[[]byte], error) {
+	res, err := rc.sdkClient.Evm.SimulateCall(ctx, round, gasPrice, gasLimit, caller, address, value, data)
+	if errors.Is(err, ErrSdkEVMExecutionFailed) || errors.Is(err, ErrSdkEVMReverted) {
+		localErr := sdkToLocalErr[err]
```
mitjat, line 142: What if the key lookup fails? (It especially might fail because of the whole `==` vs `Is`)

I suggest we simplify this: We don't particularly need the "local" error types because nobody is looking at these errors downstream (and we're not writing an SDK where we'd need to worry about 3rd party devs). We can just assign `DeterministicErr: &DeterministicErr{ msg: err.Error() }` and get rid of the local types and the map.

storage/oasis/nodeapi/api.go
```diff
@@ -206,10 +206,54 @@ type Proposal governance.Proposal
...
+// the response is stored in `result`. If the call failed due to a deterministic
+// error, eg call reverted, the error is stored in `deterministicErr`. This struct
+// should _not_ store transient errors or any other errors not considered to be a
+// valid node response.
+type FallibleResponse[Response any] struct {
```
mitjat, line 214: I suggest skipping the generics until we need them. It will keep things simpler. We can also easily add them back in if ever needed.

storage/oasis/nodeapi/file/kvstore.go
```diff
@@ -157,7 +157,7 @@ func fetchTypedValue[Value any](cache KVStore, key CacheKey, value *Value) error
 	}
 	err = cbor.Unmarshal(raw, value)
 	if err != nil {
-		return fmt.Errorf("failed to unmarshal key %s from cache into %T: %v", key.Pretty(), value, err)
+		return fmt.Errorf("failed to unmarshal key %s and value %x from cache into %T: %v", key.Pretty(), raw, value, err)
```
mitjat, line 160:
```suggestion
		return fmt.Errorf("failed to unmarshal the value for key %s from cache into %T: %v; raw value was %x", key.Pretty(), value, err, raw)
```

For large values, this keeps the more important part of the err msg together and readable.

storage/oasis/nodeapi/universal_runtime.go
```diff
@@ -135,8 +136,22 @@ func (rc *UniversalRuntimeApiLite) GetEventsRaw(ctx context.Context, round uint6
...
+		return &FallibleResponse[[]byte]{
+			Result:           nil,
```
mitjat, line 144: Nit here and below: I think it's more idiomatic to omit the initialization to `nil`. Feel free to leave as-is if you find this more readable.

mitjat (approved):
> Testing: Ran the e2e_regression tests with non-block analyzers locally; no issues with running in offline mode.

:tada: 

Thank you for the feature and the thoughts in the PR description!

storage/oasis/nodeapi/api.go
```diff
@@ -206,10 +206,54 @@ type Proposal governance.Proposal
...
+	return e.msg
+}
+
+// TODO: can we move this to oasis-sdk/client-sdk/go/modules/evm?
+const EVMModuleName = "evm"
+
+var (
+	// Known deterministic errors that may be cached
+	// https://github.com/oasisprotocol/oasis-sdk/blob/runtime-sdk/v0.2.0/runtime-sdk/modules/evm/src/lib.rs#L123
+	ErrSdkEVMExecutionFailed = errors.New(EVMModuleName, 2, "execution failed")
```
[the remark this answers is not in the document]
pro-wh: there's magic for it, but it's here instead: https://github.com/oasisprotocol/oasis-core/blob/v22.2.11/go/common/grpc/errors.go#L90. i.e., when "deserializing" from grpc and various other rpc things, we don't construct a coded error object and populate the fields from data. we look up the actual error object from a registry. thus they'll be equal, which satisfies errors.Is

storage/oasis/nodeapi/api.go
Andrew7234: good point - removed

storage/oasis/nodeapi/api.go
Andrew7234: Updated both, and moved the comment as suggested

storage/oasis/nodeapi/universal_runtime.go
Andrew7234: oo good idea - updated

PR #519, opened 2023-09-14
testing: extend regression tests to item analyzers

Andrew7234:
https://app.clickup.com/t/866aak668

Tests functionality of the `evm_tokens`, `evm_token_balances`, and `evm_contracts` analyzers


tests/e2e_regression/e2e_config_2.yml
```diff
@@ -0,0 +1,48 @@
...
+    # DANGER__WIPE_STORAGE_ON_STARTUP: true
+    migrations: file://storage/migrations
+
+server:
+  chain_name: mainnet
+  endpoint: localhost:8008
+  storage:
+    endpoint: postgresql://api:password@localhost:5432/indexer?sslmode=disable
+    backend: postgres
+
```
mitjat, line 42: Can this be commented out / removed? (Not sure how our config parser works)

tests/e2e_regression/e2e_config_1.yml
```diff
@@ -0,0 +1,42 @@
+analysis:
```
mitjat, line 1: I'm worried about the _1 and _2 config files drifting apart, leading to hard-to-understand failures in e2e tests.

One way to avoid it would be to derive them programmatically from some common template. But I think it would be relatively verbose to do it, and potentially awkward to maintain.

A more light-weight option is to add a step to the test itself that makes sure the two configs are both talking to the same cache, both expecting mainnet (vs testnet), etc. Something like this:

```bash
#!/bin/bash

set -euo pipefail

# Our e2e regression test runs the analyzer twice in close succession. This is a slightly hacky but
# simple way to ensure block analyzers run first, and non-block analyzers perform EVM queries always
# at the same height, thereby hitting the offline response cache.
#
# This script compares the key parameters of the two config files used in the two runs. If any of
# those parameters differ, it shows the diff and exits with an error.

# Element of the config files that we'll compare
important_attrs='{"cache": .analysis.source.cache.cache_dir, "chain_name": .analysis.source.chain_name, "db": .analysis.storage.endpoint}'

# A converter whose only dependency is stock python, which is likely to be preinstalled
alias yaml2json="python -c 'import sys,yaml,json; print(json.dumps(yaml.safe_load(str(sys.stdin.read()))))'" 

# Compare
cat tests/e2e_regression/e2e_config_1.yml | yaml2json | jq "$important_attrs" > /tmp/e2e_config_1.summary
cat tests/e2e_regression/e2e_config_2.yml | yaml2json | jq "$important_attrs" > /tmp/e2e_config_2.summary
diff /tmp/e2e_config_1.summary /tmp/e2e_config_2.summary || { echo "The two config files for e2e tests differ in key parameters! See diff above."; exit 1; }
```

Then in the Makefile, at the beginning of the regression test target, we just need to call it:
```Makefile
@ensure_consistent_config.sh
```

The script is untested apart from the jq part. Please rethink `important_config` with a critical eye.

mitjat (approved):
Ooh I'm really looking forward to this :)

tests/e2e_regression/run.sh
```diff
@@ -64,13 +64,16 @@ testCases=(
...
   'emerald_txs              /v1/emerald/transactions'
   'emerald_tx               /v1/emerald/transactions/a6471a9c6f3307087586da9156f3c9876fbbaf4b23910cd9a2ac524a54d0aefe'
   'emerald_failed_tx        /v1/emerald/transactions/a7e76442c52a3cb81f719bde26c9a6179bd3415f96740d91a93ee8f205b45150'
   'emerald_events           /v1/emerald/events'
   'emerald_events_by_type   /v1/emerald/events?type=sdf'
+  'emerald_tokens           /v1/emerald/evm_tokens'
+  'emerald_token            /v1/emerald/evm_tokens/oasis1qpgcp5jzlgk4hcenaj2x82rqk8rrve2keyuc8aaf'
+  'emerald_token_holders    /v1/emerald/evm_tokens/oasis1qpgcp5jzlgk4hcenaj2x82rqk8rrve2keyuc8aaf/holders'
   'emerald_account          /v1/emerald/accounts/oasis1qrt0sv2s2x2lkt9e7kmr2mzxgme8m0pzauwztprl'
+  'emerald_account_2        /v1/emerald/accounts/oasis1qpwx3ptmvcceqkd4syjmqf9jmdlf90xmuuy0f6y9'
```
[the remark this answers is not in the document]
Andrew7234: Updated! The first account has a native token balance, the second has evm_token balances

tests/e2e_regression/e2e_config_2.yml
Andrew7234: yup - removed

mitjat:
I pushed ~two~ three small commits:
- The first one regenerates the RPC cache. The value found in the old cache was `{"Result": null, "DeterministicErr": {}}`, which is incompatible with the new type now that we renamed `Result` to `Ok`. A little wild that the subsequent attempt to actually call EVMSimulateCall caused a segfault :raised_eyebrow:, but I guess that's a worry for another day?
- The second one is a tiny bugfix that I have in a branch of mine, but fits into this PR much better. When running e2e tests on this PR, I noticed that it affects it also.
- (Ooops, another commit to update the expected e2e outputs)

tests/e2e_regression/e2e_config_1.yml
Andrew7234: got it, thanks for the example! Added with a couple minor tweaks
- `shopt -s expand_aliases` to make the aliases work in script
- changed `python` -> `python3` to be clearer

PR #520, opened 2023-09-20
always update last_download_round for evm_tokens

Andrew7234:
https://app.clickup.com/t/8692rqgz6

The root cause was that the `Oasis Swag EthCC 6` evm token on sapphire testnet would return a `nil` totalSupply. The resulting code path would not return an error or update `last_download_round`. This PR ensures that `last_download_round` is always updated unless an error occured.

### DIFF
--- a/analyzer/evmtokens/evm_tokens.go
+++ b/analyzer/evmtokens/evm_tokens.go
@@ -157,15 +157,23 @@ func (p *processor) ProcessItem(ctx context.Context, batch *storage.QueryBatch, 
 		if err != nil {
 			return fmt.Errorf("downloading mutated token %s: %w", staleToken.Addr, err)
 		}
+		// We may be unable to download the totalSupply because
+		// of a deterministic error or because the contract may
+		// not provide the data.
 		if mutable != nil && mutable.TotalSupply != nil {
 			batch.Queue(queries.RuntimeEVMTokenDownloadedTotalSupplyUpdate,
 				p.runtime,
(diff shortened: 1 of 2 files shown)


analyzer/evmtokens/evm_tokens.go
```diff
@@ -157,15 +157,23 @@ func (p *processor) ProcessItem(ctx context.Context, batch *storage.QueryBatch,
...
 		}
+		// We may be unable to download the totalSupply because
+		// of a deterministic error or because the contract may
+		// not provide the data.
 		if mutable != nil && mutable.TotalSupply != nil {
 			batch.Queue(queries.RuntimeEVMTokenDownloadedTotalSupplyUpdate,
 				p.runtime,
 				staleToken.Addr,
 				mutable.TotalSupply.String(),
-				staleToken.DownloadRound,
```
Andrew7234, line 165: note: was thinking of updating the `RuntimeEVMTokenDownloadedUpsert` query above, but I'm hesitating bc
- the query is already nicely atomic, and it'd be weird to insert a downloaded token without providing the round it was downloaded at immediately
- the subsequent update is idempotent.

ptrus (approved):

pro-wh (approved):
lg, thanks. we may be able to go on to simplify this, I think the nilability of token data / mutable token data is a historical relic that we don't need anymore. I can look into that when we have more tech debt fixing energy

PR #525, opened 2023-09-23
e2e_regression: Increase coverage of tests

mitjat:
Extends e2e regression tests in two ways:
- Increases the number of indexed blocks (from 100 to 1000) to capture more types of scenarios. Wall time is still only 1.5 minutes.
- In addition to API responses, it also tests the contents of certain DB tables for regressions. This captures a lot more detail of the indexer's internal workings.

Resolves https://app.clickup.com/t/866abx7ue


tests/e2e_regression/e2e_config_2.yml
```diff
@@ -0,0 +1,34 @@
+analysis:
```
ptrus, line 1: A readme in `e2e_regression` folder or a doc comment in e2e_config_1.yml and e2e_config_2.yml might be useful explaining why two are needed/used. I assume first _1 runs the block analyzers and once it's done the _2 is used to run the non-block analyzers.

mitjat: Ah that's actually part of Andy's #525. My PR was based on top of it, but GitHub lost track of it when Andy rebased. Sorry this has made for a bulkier and more confusing review.
Your point stands though; I added comments directly in the .yml, thank you.

ptrus: Thanks! :+1:

tests/e2e_regression/run.sh
```diff
@@ -110,16 +140,23 @@ for ((i = 0; i < nCases; i++)); do
...
-  curl --silent --show-error --dump-header "$outDir/$name.headers" "$url" >"$outDir/$name.body"
-  # Try to pretty-print and normalize (for stable diffs) the output.
-  # If `jq` fails, output was probably not JSON; leave it as-is.
-  jq 'if .latest_update_age_ms? then .latest_update_age_ms="UNINTERESTING" else . end' \
-    <"$outDir/$name.body" \
-    >/tmp/pretty 2>/dev/null &&
-    cp /tmp/pretty "$outDir/$name.body" || true
-  # Sanitize the current timestamp out of the response header so that diffs are stable
-  sed -i -E 's/^(Date|Content-Length|Last-Modified): .*/\1: UNINTERESTING/g' "$outDir/$name.headers"
+  if [[ "$param" = /v1* ]]; then
```
ptrus, line 143: Maybe checking if `$name` starts with `db:` would be still hacky but a tiny bit nicer check to decide between this being an API or a DB a check.

tests/e2e_regression/run.sh
```diff
@@ -17,12 +20,39 @@ SCRIPT_DIR="$(cd "$(dirname "${BASH_SOURCE[0]}")" >/dev/null 2>&1 && pwd)"
...
 outDir="$SCRIPT_DIR/actual"
 mkdir -p "$outDir"
 rm "$outDir"/* || true
 
+# Each test case is a pair of (name, SQL query or URL).
+# For SQL queries, the regression-tested output is the result of the query against the indexer DB.
+# For URLs, the regression-tested output is the full HTTP response.
 testCases=(
+  ## Consensus.
+  'db:allowances               select * from chain.allowances order by beneficiary, owner'
```
ptrus, line 36: Seems that Github artifacts doesn't like colons in the path name. 

> Error: Artifact path is not valid: /db:allowances.csv. Contains the following character:  Colon :

https://github.com/oasisprotocol/nexus/actions/runs/6305740450/job/17119705569?pr=525

ptrus (approved):
Nice, this is a great addition to the CI tests :+1:

Andrew7234 (approved):
nice this should help flag bugs before they crawl into main

tests/e2e_regression/run.sh
mitjat: Hehe, my preference for "tiny bit nicer" leans the other way. I'll leave as-is as it seems it's not a strong preference on your end either. Anyway, we can easily change this later too, or introduce a more properly structured approach if needed.

PR #527, opened 2023-09-25
NFT transfer tracking

pro-wh:
stores an 'owner' field with the destination of the last transfer. it's anticipated that "burns" will transfer to 0x00...00 and leave it there. our record of the nft will still exist

### DIFF
--- a/analyzer/queries/queries.go
+++ b/analyzer/queries/queries.go
@@ -624,12 +624,15 @@ var (
       token_address = $2 AND
       nft_id = $3`
 
-	RuntimeEVMNFTInsert = `
-    INSERT INTO chain.evm_nfts
-      (runtime, token_address, nft_id, last_want_download_round)
+	RuntimeEVMNFTUpsert = `
+    INSERT INTO chain.evm_nfts AS old
+      (runtime, token_address, nft_id, owner, num_transfers, last_want_download_round)
     VALUES
-      ($1, $2, $3, $4)
-    ON CONFLICT (runtime, token_address, nft_id) DO NOTHING`
+      ($1, $2, $3, $4, $5, $6)
+    ON CONFLICT (runtime, token_address, nft_id) DO UPDATE
+    SET
+      owner = COALESCE(excluded.owner, old.owner),
+      num_transfers = old.num_transfers + excluded.num_transfers`
 
 	RuntimeEVMNFTAnalysisStale = `
     SELECT
--- a/analyzer/queries/queries.go
+++ b/analyzer/queries/queries.go
@@ -226,7 +226,7 @@ var (
     INSERT INTO chain.claimed_nodes (entity_id, node_id) VALUES ($1, $2)
       ON CONFLICT (entity_id, node_id) DO NOTHING`
 
-	ConsensusEntityInsert = `
+	ConsensusEntityUpsert = `
     INSERT INTO chain.entities (id, address) VALUES ($1, $2)
       ON CONFLICT (id) DO
       UPDATE SET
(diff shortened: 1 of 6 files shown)


analyzer/runtime/extract.go
```diff
@@ -897,6 +917,7 @@ func extractEvents(blockData *BlockData, relatedAccountAddresses map[apiTypes.Ad
 						pt.Mutated = true
 					}
 					registerNFTExist(blockData.PossibleNFTs, eventAddr, tokenID)
+					registerNFTTransfer(blockData.PossibleNFTs, eventAddr, tokenID, toAddr)
```
ptrus, line 920: Afaik ERC-721 allows transferring nft to yourself (e.g. from address=to address). Is it ok that num-transfers is increased in that case as well?

ptrus (approved):

analyzer/runtime/extract.go
pro-wh: yeah sure. it doesn't seem like a high obstacle if for example people have to transfer it between two of their own accounts instead

analyzer/queries/queries.go
```diff
@@ -601,10 +601,13 @@ var (
...
-      (runtime, token_address, nft_id, last_want_download_round)
+      (runtime, token_address, nft_id, owner, num_transfers, last_want_download_round)
     VALUES
-      ($1, $2, $3, $4)
-    ON CONFLICT (runtime, token_address, nft_id) DO NOTHING`
+      ($1, $2, $3, $4, $5, $6)
+    ON CONFLICT (runtime, token_address, nft_id) DO UPDATE
+    SET
+      owner = COALESCE(excluded.owner, chain.evm_nfts.owner),
+      num_transfers = chain.evm_nfts.num_transfers + excluded.num_transfers`
```
mitjat, line 610: Optional: For upserts, I've recently been using an `old` alias like so, and liking it:
```sql
INSERT INTO chain.my_long_named_table AS old
...
ON CONFLICT DO UPDATE SET
  num_transfers = old.num_transfers + excluded.num_transfers
```

analyzer/queries/queries.go
```diff
@@ -601,10 +601,13 @@ var (
 
 	RuntimeEVMNFTInsert = `
```
mitjat, line 602:
```suggestion
	RuntimeEVMNFTUpsert = `
```

mitjat (approved):
:metal:

analyzer/queries/queries.go
pro-wh: k let me add that too

analyzer/runtime/extract.go
pro-wh: ok tentatively on the mint and burn not counting in num_transfers

analyzer/queries/queries.go
pro-wh: ):

pro-wh: no on this. there's a lot of `...Insert` queries that define an ON CONFLICT. really we need better terminology for "the thing the block analyzer does" and "the thing the other analyzer does"

analyzer/runtime/extract.go
pro-wh: ugh I just saw this other comment from the fungible tokens side

> Mints, burns, and zero-value transfers all count as transfers.

so I'm going to preserve consistency with that instead

pro-wh: and I'm opposed to fussing with the abstraction in the other suggestions about inlining registerNFTTransfer and eliminating registerNFTExist

analyzer/queries/queries.go
pro-wh: actually nah changing it is fine. most of the insert on conflict's are DO NOTHING

analyzer/runtime/extract.go
mitjat: Thanks for finding the prior "art" on countin mints/burns. Might be worth mentioning in a comment in our codebase.

Tiny refactor/abstraction: sure, sg.

analyzer/queries/queries.go
mitjat: :+1: if they're not `DO NOTHING`, I'm in favor of renaming them too. Consistency is always nice. But regardless, I find the "upsert" in the name helpful when reading analyzer code; sometimes that clue is enough to guess/remember what the SQL query does, without jumping to the definition.

pro-wh:
image

PR #529, opened 2023-09-26
NFT query by owner

pro-wh:
> /{runtime}/accounts/{address}/evm_nfts

### DIFF
--- a/api/spec/v1.yaml
+++ b/api/spec/v1.yaml
@@ -2621,6 +2621,14 @@ components:
         id:
           <<: *BigIntType
           description: The instance ID of this NFT within the collection represented by `token`.
+        owner:
+          <<: *AddressType
+          description: The Oasis address of this NFT instance's owner.
+          example: "oasis1qpclnnm0wu44pn43mt6vv3me59kl8zk9ty7qyj03"
+        owner_eth:
+          type: string
+          description: The Ethereum address of this NFT instance's owner.
+          example: "0xDEF1009df2d6872C214cd9148c6883893B7c4D91"
         metadata_uri:
           type: string
         metadata_accessed:
--- a/api/spec/v1.yaml
+++ b/api/spec/v1.yaml
@@ -1080,6 +1080,29 @@ paths:
                 $ref: '#/components/schemas/RuntimeAccount'
         <<: *common_error_responses
 
+  /{runtime}/accounts/{address}/nfts:
+    get:
+      summary: |
+        Returns the list of non-fungible token (NFT) instances owned by an account.
+      parameters:
+        - *limit
+        - *offset
+        - *runtime
+        - in: path
+          name: address
+          required: true
+          schema:
+            <<: *StakingAddressType
+          description: The staking address of the owner of the NFT instances.
+      responses:
+        '200':
+          description: The requested instances.
+          content:
+            application/json:
+              schema:
+                $ref: '#/components/schemas/EvmNftList'
+        <<: *common_error_responses
+
   /{runtime}/status:
     get:
       summary: Returns the runtime status.
--- a/storage/client/queries/queries.go
+++ b/storage/client/queries/queries.go
@@ -504,9 +504,9 @@ const (
 			)
 		SELECT
 			chain.evm_nfts.token_address,
-			chain.address_preimages.context_identifier,
-			chain.address_preimages.context_version,
-			chain.address_preimages.address_data,
+			token_preimage.context_identifier,
+			token_preimage.context_version,
+			token_preimage.address_data,
 			chain.evm_tokens.token_name,
 			chain.evm_tokens.symbol,
 			chain.evm_tokens.decimals,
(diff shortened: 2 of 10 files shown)


api/spec/v1.yaml
```diff
@@ -1080,6 +1080,29 @@ paths:
                 $ref: '#/components/schemas/RuntimeAccount'
         <<: *common_error_responses
 
+  /{runtime}/accounts/{address}/evm_nfts:
+    get:
+      summary: |
+        Returns the list of non-fungible token (NFT) instances owned by an account.
+      parameters:
```
csillag, line 1087: Would it be possible to add an optional parameter for filtering by token contract address? (According to the current UI design, when browsing the inventory belonging to an account, we are going to display the different collections separately.)

pro-wh: how does filtering by token contract address fit into this? will it also need a query to get the list of token contracts that the account owns at least one of?

csillag: > will it also need a query to get the list of token contracts that the account owns at least one of?

I think we already have that info. 
https://nexus.oasis.io/v1/sapphire/accounts/oasis1qzxpkka97g4pvl9ghc5da2pharwvzth56uymlsgv
returns
``` JSON
{
  "address": "oasis1qzxpkka97g4pvl9ghc5da2pharwvzth56uymlsgv",
  "address_preimage": {
    "address_data": "7QPqnJbsOQl1SCVuQooWPl9STkc=",
    "context": "oasis-runtime-sdk/address: secp256k1eth",
    "context_version": 0
  },
  "balances": [
    {
      "balance": "1126883292480000000000",
      "token_decimals": 18,
      "token_symbol": "ROSE"
    }
  ],
  "evm_balances": [
    {
      "balance": "42",
      "token_contract_addr": "oasis1qpwe6h0hw6s0x4skujhznld065300p83wgxnvu0n",
      "token_contract_addr_eth": "0x998633BDF6eE32A9CcA6c9A247F428596e8e65d8",
      "token_decimals": 0,
      "token_name": "NFTrout",
      "token_symbol": "TROUT",
      "token_type": "ERC721"
    },
    {
      "balance": "23",
      "token_contract_addr": "oasis1qqkcynmpm5tj9thyfq0fg5k2slvdap9teu02n77d",
      "token_contract_addr_eth": "0xFcfed3be2d333F24854cA8d3A351E772272D5842",
      "token_decimals": 0,
      "token_name": "NFTrout Gen 1",
      "token_symbol": "TROUT1",
      "token_type": "ERC721"
    }
  ],
  "stats": {
    "num_txns": 55,
    "total_received": "4000000000000000000000",
    "total_sent": "0"
  }
}
```

Here, the `evm_balances` field lists the tokens, and our balance. If the balance is not 0, we have some of it.

csillag: > how does filtering by token contract address fit into this?

When looking at tokens, first we list all the collections we have anything from. We can already do that based on the data we get as part of the account info.

Then as a next step, if the user clicks into a collection, then we need to list all the instances the user owns from that collection.

Now, we could do the filtering for collection (based on contract address) locally on the client, but it would be cumbersome, because of pagination. So it would be best if we could query directly the instances that a given account owns, from a given collection.

pro-wh: ah ok

mitjat (commented):
Sweet. Can you add an e2e regression test please?
I see we have some relevant tokens in the test's range of blocks:
```
indexer=# select token_type, count(*) from chain.evm_tokens group by 1;
 token_type | count                                                    
------------+-------                                                   
        721 |     8                                                    
         20 |   101                                                    
```

~I might have to merge https://github.com/oasisprotocol/nexus/pull/525 first.~ Done.

pro-wh:
moved to /{runtime}/accounts/{address}/nfts

api/spec/v1.yaml
pro-wh: adding in #539

pro-wh: resolving here, let's discuss there

storage/client/queries/queries.go
```diff
@@ -516,25 +516,78 @@ const (
...
 			chain.evm_contracts.contract_address = chain.evm_tokens.token_address
+		LEFT JOIN chain.address_preimages AS owner_preimage ON
+			owner_preimage.address = chain.evm_nfts.owner
 		WHERE
 			chain.evm_nfts.runtime = $1::runtime AND
 			chain.evm_nfts.token_address = $2::oasis_addr
 		LIMIT $3::bigint
 		OFFSET $4::bigint`
 
+	AccountEvmNfts = `
```
mitjat, line 544: How different is this from the `EvmNfts` query? Can we reuse/generalize that one instead so it says something like
```
WHERE
  runtime = $1 AND 
  ($2 IS NULL OR token_address = $2)  AND
  ($3 IS NULL OR owner = $3)  AND
```
?

pro-wh: ah yeah let's do that

pro-wh: moreso with #538

pro-wh: combined

storage/client/queries/queries.go
```diff
@@ -516,25 +516,78 @@ const (
...
+		LEFT JOIN token_holders USING (token_address)
+		LEFT JOIN chain.evm_contracts ON
+			chain.evm_contracts.runtime = chain.evm_tokens.runtime AND
+			chain.evm_contracts.contract_address = chain.evm_tokens.token_address
+		LEFT JOIN chain.address_preimages AS owner_preimage ON
+			owner_preimage.address = chain.evm_nfts.owner
+		WHERE
+			chain.evm_nfts.runtime = $1::runtime AND
+			chain.evm_nfts.owner = $2::oasis_addr
+		LIMIT $3::bigint
```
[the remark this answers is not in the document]
pro-wh: unified with that other query

pro-wh:
image

note: migrations are combined with the permission change migration that hasn't appeared in a release yet

pro-wh:
oh yeah, e2e test added

Andrew7234 (approved):

PR #530, opened 2023-09-27
Andrew7234/tx revert encoding

Andrew7234:
Part of https://app.clickup.com/t/8692revhx

Nexus now stores the raw error messages in a dedicated column. It tries to parse the message into a human-readable string, which is stored in the existing `chain.runtime_transactions.error_message` column if successful.

The remaining part of the ticket is to have the abi analyzer use the abi to decode custom errors. That'll go in a follow-up PR.

Todo:
- think about what we want to surface in the api, if anything -- let's hold off for now since users can't decode it without the abi anyway.
- clean up/shorten comments [x]
- add unit tests [x]

```
indexer=# select round, tx_index, error_message, error_message_raw from chain.runtime_transactions;
  round  | tx_index |   error_message    |
              error_message_raw

---------+----------+--------------------+---------------------------------------------------
---------------------------------------------------------------------------------------------
----
 2710850 |        0 |                    |
 2710850 |        1 |                    |
 2710850 |        2 |                    |
 2710851 |        0 |                    |
 2710852 |        0 | too late to submit | reverted: CMN5oAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAA
AAAAAAAgAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAABJ0b28gbGF0ZSB0byBzdWJtaXQAAAAAAAAAAAAAAAAAA
A==
 2710852 |        1 | too late to submit | reverted: CMN5oAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAA
AAAAAAAgAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAABJ0b28gbGF0ZSB0byBzdWJtaXQAAAAAAAAAAAAAAAAAA
A==
 2710852 |        2 |                    | reverted: EYzapwAAAAAAAAAAAAAAAH3y4TP8/2EqpGmV6CpU
M19xa2a+
 2710853 |        0 | too late to submit | reverted: CMN5oAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAA
AAAAAAAgAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAABJ0b28gbGF0ZSB0byBzdWJtaXQAAAAAAAAAAAAAAAAAA
A==
 2710853 |        1 |                    |
 2710854 |        0 |                    | reverted: b+19hQ==
 2710855 |        0 |                    |
```


analyzer/runtime/extract.go
```diff
@@ -578,6 +586,84 @@ func sumGasUsed(events []*EventData) (sum uint64, foundGasUsedEvent bool) {
...
+//
+// Note that the error type can be any data type specified in the abi.
+// We may not have the abi available now, so in those cases the
+// abi analyzer will extract the error message once the contract
+// abi is available.
+func tryParseErrorMessage(msg string) (*string, error) {
+	// Decode revert reason, format is as follows:
+	// "reverted: " || base64(abiEncode(error))
+	ErrRevertedPrefix := "reverted: "
+	if len(msg) < 10 || msg[:10] != ErrRevertedPrefix {
```
ptrus, line 618: Consider using [`strings.HasPrefix`](https://pkg.go.dev/strings#HasPrefix) since it is a bit more idiomatic go code for checking a string prefix.

Also function variables usually always start with lowercase (`errRevertedPrefix`).

analyzer/runtime/extract.go
```diff
@@ -578,6 +586,84 @@ func sumGasUsed(events []*EventData) (sum uint64, foundGasUsedEvent bool) {
...
+// We first check to see if the error message is a plaintext error string,
+// in which case we simply sanitize and return the message.
+// Otherwise, we optimistically try to decode the error as the prevailing
+// Error(string) type and return the error message if successful.
+//
+// Note that the error type can be any data type specified in the abi.
+// We may not have the abi available now, so in those cases the
+// abi analyzer will extract the error message once the contract
+// abi is available.
+func tryParseErrorMessage(msg string) (*string, error) {
```
ptrus, line 614: Would be nice having some unit tests for `tryParseErrorMessage`.

Specially the last part which tries decoding it as `Error(string)`

analyzer/runtime/extract.go
```diff
@@ -578,6 +586,84 @@ func sumGasUsed(events []*EventData) (sum uint64, foundGasUsedEvent bool) {
...
+	//
+	// 08c379a0                                                         <- Function selector
+	// 0000000000000000000000000000000000000000000000000000000000000020 <- Offset of string return value
+	// 0000000000000000000000000000000000000000000000000000000000000047 <- Length of string return value (the revert reason)
+	// 6d7946756e6374696f6e206f6e6c79206163636570747320617267756d656e74 <- First 32 bytes of the revert reason
+	// 7320776869636820617265206772656174686572207468616e206f7220657175 <- Next 32 bytes of the revert reason
+	// 616c20746f203500000000000000000000000000000000000000000000000000 <- Last 7 bytes of the revert reason
+	//
+	ErrorStringSelector, _ := hex.DecodeString("08c379a0") // Keccak256("Error(string)")
+	const FieldOffsetStart int64 = 4
```
ptrus, line 644: Does this parsing code roughly equal to:
```
	typ, _ := abi.NewType("string", "", nil)
	unpacked, err := (abi.Arguments{{Type: typ}}).Pack(msg[10:])
```

This is roughly how web3-gateway used to try optimistically parsing the error message.

analyzer/runtime/extract.go
Andrew7234: updated!

analyzer/runtime/extract.go
Andrew7234: nicee didn't know there was a library for it - updated

analyzer/runtime/extract.go
Andrew7234: Added!

analyzer/runtime/runtime.go
```diff
@@ -220,15 +220,21 @@ func (m *processor) queueDbUpdates(batch *storage.QueryBatch, data *BlockData) {
...
+			error_message_raw = transactionData.Error.RawMessage
 		}
 		if error_message != nil {
 			// Apparently the message does need to be valid UTF-8.
 			// In the rare case it's not, hex encode it.
 			// https://github.com/oasisprotocol/nexus/issues/439
+			if !storage.IsValidText(*error_message_raw) {
+				*error_message_raw = hex.EncodeToString([]byte(*error_message_raw))
+			}
+			// this is likely unnecessary. todo: remove
```
ptrus, line 237: Agreed that it can be removed below since the error_message string is sanitized via `sanitizeString`?

Actually, raw message is sanitized as well, so the above check is not needed either, not sure why it was here before even.

---

I did a minor investigation and I see what happened: 
- Mitja fixed this by sanitizing error messages in: https://github.com/oasisprotocol/nexus/pull/373 (deployed in v0.0.15).
- I fixed this by adding the `storage.IsValidText` check in: https://github.com/oasisprotocol/nexus/pull/440 - because staging at the time was running an older version that did not contain the Mitja's fix and I did not notice that this was already fixed.

So the checks here are redundant, feel free to remove the checks.

analyzer/runtime/extract.go
```diff
@@ -67,7 +69,8 @@ type TxError struct {
 	Module string
 	// `Module` should be present but `Message` may be null
 	// https://github.com/oasisprotocol/oasis-sdk/blob/fb741678585c04fdb413441f2bfba18aafbf98f3/client-sdk/go/types/transaction.go#L488-L492
-	Message *string
+	Message    *string
```
ptrus, line 72: Update the doc comments here to explain the two message fields.

analyzer/runtime/extract.go
ptrus: Thanks, looks great! :+1:

ptrus (approved):
LGTM!

analyzer/runtime/extract_test.go
```diff
@@ -0,0 +1,48 @@
...
+	for _, input := range inputs {
+		msg := tryParseErrorMessage(input)
+		require.NotNil(t, msg, "failed to extract error message")
+		require.Equal(t, input, *msg, "parsed message should match original input message")
+	}
+}
+
+func TestKnownStringError(t *testing.T) {
+	inputs := map[string]string{
+		"too late to submit": "reverted: CMN5oAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAgAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAABJ0b28gbGF0ZSB0byBzdWJtaXQAAAAAAAAAAAAAAAAAAA==",
```
mitjat, line 24: Because the tx can fail for reasons other than being reverted, we should keep the "reverted: " prefix when decoding the base64 stuff. In other words, the correct decoding here would be "reverted: too late to submit".

storage/migrations/19_runtime_tx_errors.up.sql
```diff
@@ -0,0 +1,7 @@
+BEGIN;
+
+ALTER TABLE chain.runtime_transactions
+    -- The unparsed transaction error message.
```
mitjat, line 4:
```suggestion
    -- The unparsed transaction error message. The "parsed" version will be identical in the majority of cases.
    -- One notable exception are txs that were reverted inside the EVM; for those, the raw msg is abi-encoded.
```

mitjat (commented):
Thank you! Looks good overall, my comments are mostly nits, or have to do with a slight misunderstanding around the possible values of the error msg. Should all be easy fixes :+1:

analyzer/runtime/extract.go
```diff
@@ -578,6 +582,65 @@ func sumGasUsed(events []*EventData) (sum uint64, foundGasUsedEvent bool) {
...
+// abi is available.
+func tryParseErrorMessage(msg string) *string {
+	// Decode revert reason, format is as follows:
+	// "reverted: " || base64(abiEncode(error))
+	if !strings.HasPrefix(msg, "reverted: ") {
+		// Likely a plaintext error message.
+		sanitizedMsg := sanitizeString(msg)
+		return &sanitizedMsg
+	}
+	abiEncodedErr, err := base64.StdEncoding.DecodeString(msg[10:])
```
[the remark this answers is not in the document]
kostko: Even better, use `strings.TrimPrefix` instead of this custom code.

analyzer/runtime/extract.go
```diff
@@ -578,6 +582,65 @@ func sumGasUsed(events []*EventData) (sum uint64, foundGasUsedEvent bool) {
...
+// Error(string) type and return the error message if successful.
+//
+// Note that the error type can be any data type specified in the abi.
+// We may not have the abi available now, so in those cases the
+// abi analyzer will extract the error message once the contract
+// abi is available.
+func tryParseErrorMessage(msg string) *string {
+	// Decode revert reason, format is as follows:
+	// "reverted: " || base64(abiEncode(error))
+	if !strings.HasPrefix(msg, "reverted: ") {
```
kostko, line 610: Note that one could also check the error code. The reversion error has a specific error code and anything else should not be treated as a reversion:
```rust
    #[error("reverted: {0}")]
    #[sdk_error(code = 8)]
    Reverted(String),
```
So if module is `evm` and code is `8` the error is a reversion. Otherwise it shouldn't be parsed as one.

analyzer/runtime/extract.go
```diff
@@ -578,6 +582,65 @@ func sumGasUsed(events []*EventData) (sum uint64, foundGasUsedEvent bool) {
 	return
 }
 
+// Attempts to extract the human-readable error message from
+// the raw error returned by the node.
+//
+// In Emerald and older Sapphire versions, transaction revert
```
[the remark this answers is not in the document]
Andrew7234: TIL, thanks

analyzer/runtime/extract.go
```diff
@@ -578,6 +582,65 @@ func sumGasUsed(events []*EventData) (sum uint64, foundGasUsedEvent bool) {
...
+// reasons were decoded by the runtime and returned as
+// human-readable strings, e.g.
+// - "reverted: Incorrect premium amount."
+// - "execution failed: out of gas"
```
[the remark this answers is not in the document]
Andrew7234: Ah I see, I misunderstood errors vs tx revert errors. Updated the docstring!

analyzer/runtime/extract.go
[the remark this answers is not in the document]
Andrew7234: updated!

analyzer/runtime/extract.go
```diff
@@ -578,6 +582,65 @@ func sumGasUsed(events []*EventData) (sum uint64, foundGasUsedEvent bool) {
...
+	// Decode revert reason, format is as follows:
+	// "reverted: " || base64(abiEncode(error))
+	if !strings.HasPrefix(msg, "reverted: ") {
+		// Likely a plaintext error message.
+		sanitizedMsg := sanitizeString(msg)
+		return &sanitizedMsg
+	}
+	abiEncodedErr, err := base64.StdEncoding.DecodeString(msg[10:])
+	if err != nil {
+		// Likely a plaintext error message.
```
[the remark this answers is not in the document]
Andrew7234: Good point - updated the comment accordingly. Thanks for the detail

analyzer/runtime/extract_test.go
Andrew7234: got it, updated the fn to include the `reverted: `

analyzer/runtime/runtime.go
Andrew7234: removed! I updated the tests for `IsValidText` to now test `SanitizeString` as well

analyzer/runtime/extract.go
Andrew7234: got it, updated!

mitjat (approved):
Thank you!

PR #533, opened 2023-09-28
storage: move token type translation to go

pro-wh:
did we want to move this?

### DIFF
--- a/storage/client/client.go
+++ b/storage/client/client.go
@@ -2,6 +2,7 @@ package client
 
 import (
 	"context"
+	"database/sql"
 	"fmt"
 	"math/big"
 	"time"
@@ -44,6 +45,17 @@ type StorageClient struct {
 	txCache    *ristretto.Cache
 
 	logger *log.Logger
+}
+
+func translateTokenType(tokenType common.TokenType) apiTypes.EvmTokenType {
+	switch tokenType {
+	case common.TokenTypeERC20:
+		return apiTypes.EvmTokenTypeERC20
+	case common.TokenTypeERC721:
+		return apiTypes.EvmTokenTypeERC721
+	default:
+		return "unexpected_other_type"
+	}
 }
 
 // runtimeNameToID returns the runtime ID for the given network and runtime name.
@@ -1279,6 +1291,7 @@ func (c *StorageClient) RuntimeEvents(ctx context.Context, p apiTypes.GetRuntime
 	for res.rows.Next() {
 		var e RuntimeEvent
 		var et apiTypes.EvmEventToken
+		var tokenType sql.NullInt32
 		if err := res.rows.Scan(
 			&e.Round,
 			&e.TxIndex,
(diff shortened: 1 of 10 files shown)


storage/client/client.go
```diff
@@ -18,6 +18,7 @@ import (
 	oasisConfig "github.com/oasisprotocol/oasis-sdk/client-sdk/go/config"
 	sdkTypes "github.com/oasisprotocol/oasis-sdk/client-sdk/go/types"
 
+	"github.com/oasisprotocol/nexus/analyzer/runtime/evm"
```
pro-wh, line 21: dang ol storage depending on analyzer

storage/client/client.go
```diff
@@ -46,6 +47,17 @@ type StorageClient struct {
 	logger *log.Logger
 }
 
+func translateTokenType(tokenType evm.EVMTokenType) apiTypes.EvmTokenType {
+	switch tokenType {
+	case evm.EVMTokenTypeERC20:
+		return apiTypes.EvmTokenTypeERC20
```
pro-wh, line 53: dang ol storage depending on api

storage/client/client.go
```diff
@@ -1481,14 +1500,15 @@ func (c *StorageClient) RuntimeTokens(ctx context.Context, p apiTypes.GetRuntime
...
-			&t.Type,
+			&tt,
 			&t.NumHolders,
 			&t.IsVerified,
 		); err2 != nil {
 			return nil, wrapError(err2)
 		}
 
 		t.EthContractAddr = ethCommon.BytesToAddress(addrPreimage).String()
+		t.Type = translateTokenType(tt)
```
pro-wh, line 1511: t.T = tTT(tt)

storage/client/client.go
mitjat: :shrug: you could move it to `common/types.go`? Although it looks OK enough to me as-is. The true storage type is `int`, really ... But it's nice to do the automatic cast into analyzer types, because the compiler will warn us(*) if we change the db input (= analyzer) but not the handling of the db output (= storage/client).

(*) in limited cases, Go's approach to enums is a joke, call me at 1 800 VITRIOL for more pointless Go rants

mitjat (approved):
Hallelujah. Thank you, I've wanted to scratch this itch for a while. Sorry for allowing it to spawn in the first place.

storage/client/client.go
pro-wh: ah yeah, would make sense to move it to common, now that it shows up in both storage and api

go has a lot of progressive ideas about preventing authors from locking down their own stuff, including via enums

pro-wh: moved. also it's now TokenType instead of EVMTokenType

storage/client/client.go
pro-wh: ugh now I think this is fine. it lets us have the list of enum values in the api spec for the price of one copy of the values

storage/client/client.go
pro-wh: correction: doesn't show up in api

PR #535, opened 2023-09-29
kvstore: Add prometheus metric for cache hits/misses

mitjat:
See title. This will be nice to have: We can make better sense of speed-of-indexing grafana charts if we know whether node I/O is involved.

This PR also includes a cleanup commit that enforces strong typing for labels when accessing metrics, and renames local variable that represent metrics so that the variable name matches the metric name. **I recommend reviewing that first commit separately** -- lots of noise, not much effect.

### DIFF
--- a/metrics/storage.go
+++ b/metrics/storage.go
@@ -8,6 +8,9 @@ import (
 
 // Default service metrics for database operations.
 type StorageMetrics struct {
+	// Name of the runtime (or "consensus") that is being analyzed.
+	runtime string
+
 	// Counts of database operations
 	databaseOperations *prometheus.CounterVec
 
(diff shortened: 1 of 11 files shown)


metrics/storage.go
mitjat, line 1: Most of this file existed before under `metrics/database.go`; I renamed the file to accommodate both DB and KVStore metrics, and added the LocalCacheReads counter.

metrics/storage.go
```diff
@@ -0,0 +1,83 @@
...
+		databaseLatencies: prometheus.NewHistogramVec(
+			prometheus.HistogramOpts{
+				Name: fmt.Sprintf("%s_db_latencies", runtime),
+				Help: "How long database operations take, partitioned by operation.",
+			},
+			[]string{"database", "operation"}, // Labels.
+		),
+		localCacheReads: prometheus.NewCounterVec(
+			prometheus.CounterOpts{
+				Name: "local_cache_reads",
```
mitjat, line 54: Other metrics use names in the pattern of "emerald_local_cache_reads". This is nonidiomatic and makes it hard to aggregate metrics across runtimes. For this new metric, I instead added the runtime as a label (= attribute) of this metric. I'd love to later refactor the other metrics too, bu tthat's more work because we'll also need to update grafana.

ptrus (approved):

PR #538, opened 2023-10-06
NFT query by ID

pro-wh:
https://github.com/oasisprotocol/nexus/pull/514#discussion_r1341689174

> for getting data about a specific NFT instance

### DIFF
--- a/api/v1/strict_server.go
+++ b/api/v1/strict_server.go
@@ -298,6 +298,11 @@ func (srv *StrictServerImpl) GetRuntimeEvmTokensAddressNfts(ctx context.Context,
 	return apiTypes.GetRuntimeEvmTokensAddressNfts200JSONResponse(*nfts), nil
 }
 
+func (srv *StrictServerImpl) GetRuntimeEvmTokensAddressNftsId(ctx context.Context, request apiTypes.GetRuntimeEvmTokensAddressNftsIdRequestObject) (apiTypes.GetRuntimeEvmTokensAddressNftsIdResponseObject, error) {
+	// TODO implement me
+	panic("implement me")
+}
+
 func (srv *StrictServerImpl) GetRuntimeTransactions(ctx context.Context, request apiTypes.GetRuntimeTransactionsRequestObject) (apiTypes.GetRuntimeTransactionsResponseObject, error) {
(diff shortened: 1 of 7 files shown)


api/v1/strict_server.go
```diff
@@ -298,6 +298,14 @@ func (srv *StrictServerImpl) GetRuntimeEvmTokensAddressNfts(ctx context.Context,
 	return apiTypes.GetRuntimeEvmTokensAddressNfts200JSONResponse(*nfts), nil
 }
 
+func (srv *StrictServerImpl) GetRuntimeEvmTokensAddressNftsId(ctx context.Context, request apiTypes.GetRuntimeEvmTokensAddressNftsIdRequestObject) (apiTypes.GetRuntimeEvmTokensAddressNftsIdResponseObject, error) {
+	nft, err := srv.dbClient.RuntimeEVMNFT(ctx, request.Address, request.Id)
```
mitjat, line 302: To reduce boilerplate and copy-paste, I'd prefer if we reuse `srv.dbClient.RuntimeEVMNFTs()` here (note the plural in the name). It will need to start taking a new optional param: The NFT ID. 

See `GetRuntimeEvmTokensAddress()` and `GetConsensusValidatorsEntityId()` for two places with a very similar problem, and a solution along the lines above.

pro-wh: k converted

pro-wh:
```diff
 diff --git a/home/runner/work/nexus/nexus/tests/e2e_regression/expected/window_size.body b/home/runner/work/nexus/nexus/tests/e2e_regression/actual/window_size.body
index bed16d7..431442b 100644
--- a/home/runner/work/nexus/nexus/tests/e2e_regression/expected/window_size.body
+++ b/home/runner/work/nexus/nexus/tests/e2e_regression/actual/window_size.body
@@ -1,4 +1 @@
-{
-  "window_size_seconds": 300,
-  "windows": []
-}
+{"window_size_seconds":300,"windows":[]}
```

oh shoot let me try with explicit `else .`

pro-wh:
ok yeah. I must have a different jq version locally

Andrew7234 (commented):
looks good, thanks for the adding the regression tests

Andrew7234 (approved):

PR #540, opened 2023-10-06
abi analyzer

Andrew7234:
Part of https://app.clickup.com/t/866ad5hg1

Adds an analyzer to re-parse event body, tx data, and tx revert reason using known abis. 

Note: does not require a reindex

### DIFF
--- a/analyzer/evmabibackfill/evm_abi_backfill.go
+++ b/analyzer/evmabibackfill/evm_abi_backfill.go
@@ -1,0 +1,324 @@
+package evmabibackfill
+
+import (
+	"bytes"
+	"context"
+	"encoding/base64"
+	"encoding/json"
+	"fmt"
+	"strings"
+
+	"github.com/ethereum/go-ethereum/accounts/abi"
+	sdkEVM "github.com/oasisprotocol/oasis-sdk/client-sdk/go/modules/evm"
+
+	"github.com/oasisprotocol/nexus/analyzer"
+	"github.com/oasisprotocol/nexus/analyzer/item"
+	"github.com/oasisprotocol/nexus/analyzer/queries"
+	"github.com/oasisprotocol/nexus/analyzer/runtime/abiparse"
+	"github.com/oasisprotocol/nexus/common"
+	"github.com/oasisprotocol/nexus/config"
+	"github.com/oasisprotocol/nexus/log"
+	"github.com/oasisprotocol/nexus/storage"
+)
+
+const (
+	evmAbiAnalyzerPrefix = "evm_abi_"
+)
+
+type processor struct {
+	runtime common.Runtime
+	target  storage.TargetStorage
+	logger  *log.Logger
+}
+
+var _ item.ItemProcessor[*abiEncodedItem] = (*processor)(nil)
+
+type abiEncodedTx struct {
+	TxHash         string
+	TxData         []byte
+	TxRevertReason *string
+}
+
+type abiEncodedEvent struct {
+	Round     uint64
+	TxIndex   *int
+	EventBody sdkEVM.Event
+}
+
+type abiEncodedItem struct {
+	Tx           *abiEncodedTx
+	Event        *abiEncodedEvent
+	ContractAddr string
+	Abi          json.RawMessage
+}
+
+type abiEncodedArg struct {
+	Name    string      `json:"name"`
+	EvmType string      `json:"evm_type"`
+	Value   interface{} `json:"value"`
+}
+
+func NewAnalyzer(
+	runtime common.Runtime,
+	cfg config.ItemBasedAnalyzerConfig,
+	target storage.TargetStorage,
+	logger *log.Logger,
+) (analyzer.Analyzer, error) {
+	logger = logger.With("analyzer", evmAbiAnalyzerPrefix+runtime)
+	p := &processor{
+		runtime,
+		target,
+		logger,
+	}
+	return item.NewAnalyzer[*abiEncodedItem](
+		evmAbiAnalyzerPrefix+string(runtime),
+		cfg,
+		p,
+		target,
+		logger,
+	)
+}
+
+// Transaction data is canonically represented as a byte array. However,
+// the transaction body is stored as a JSONB column in postgres, which
+// causes the tx body->>data to be returned as a base64-encoded string
+// enclosed by escaped double quote characters.
+func cleanTxData(raw string) ([]byte, error) {
+	s := strings.TrimPrefix(strings.TrimSuffix(raw, "\""), "\"")
+	return base64.StdEncoding.DecodeString(s)
+}
+
+func (p *processor) GetItems(ctx context.Context, limit uint64) ([]*abiEncodedItem, error) {
+	// There are two types of data we process using a contract abi: transactions and events.
+	// Within a transaction, we process the call data and the revert reason. Since they are
+	// colocated in the same table we can fetch them using a single query.
+	var items []*abiEncodedItem
+	txRows, err := p.target.Query(ctx, queries.RuntimeEvmVerifiedContractTxs, p.runtime, limit)
+	if err != nil {
+		return nil, fmt.Errorf("querying verified contract txs: %w", err)
+	}
+	defer txRows.Close()
+	for txRows.Next() {
+		var rawTxData string
+		var tx abiEncodedTx
+		var item abiEncodedItem
+		item.Tx = &tx
+		if err = txRows.Scan(
+			&item.ContractAddr,
+			&item.Abi,
+			&tx.TxHash,
+			&rawTxData,
+			&tx.TxRevertReason,
+		); err != nil {
+			return nil, fmt.Errorf("scanning verified contract tx: %w", err)
+		}
+		if tx.TxData, err = cleanTxData(rawTxData); err != nil {
+			return nil, fmt.Errorf("error decoding tx data from db: %w", err)
+		}
+		items = append(items, &item)
+	}
+	// Short circuit.
+	if len(items) == int(limit) {
+		return items, nil
+	}
+	eventRows, err := p.target.Query(ctx, queries.RuntimeEvmVerifiedContractEvents, p.runtime, int(limit)-len(items))
+	if err != nil {
+		return nil, fmt.Errorf("querying verified contract evs: %w", err)
+	}
+	defer eventRows.Close()
+	for eventRows.Next() {
+		var ev abiEncodedEvent
+		var item abiEncodedItem
+		item.Event = &ev
+		if err = eventRows.Scan(
+			&item.ContractAddr,
+			&item.Abi,
+			&ev.Round,
+			&ev.TxIndex,
+			&ev.EventBody,
+		); err != nil {
+			return nil, fmt.Errorf("scanning verified contract event: %w", err)
+		}
+		items = append(items, &item)
+	}
+	return items, nil
+}
+
+// Transaction revert reasons for failed evm transactions have been encoded
+// differently over the course of Oasis history. Older transaction revert
+// reasons were returned as one of
+// - "reverted: Incorrect premium amount"
+// - "reverted: base64(up to 1024 bytes of revert data)"
+//
+// Note that if the revert reason was longer than 1024 bytes it was truncated.
+// Newer transaction revert reasons are returned as
+// - "reverted: base64(revert data)"
+//
+// In all cases the revert reason has the "reverted: " prefix, which we first
+// strip. We then attempt to base64-decode the remaining string to recover
+// the error message. It should be noted that if the b64 decoding fails, it's
+// likely an older error message.
+//
+// See the docstring of tryParseErrorMessage in analyzer/runtime/extract.go
+// for more info.
+func cleanTxRevertReason(raw string) ([]byte, error) {
+	s := strings.TrimPrefix(raw, "reverted: ")
+	return base64.StdEncoding.DecodeString(s)
+}
+
+func (p *processor) ProcessItem(ctx context.Context, batch *storage.QueryBatch, item *abiEncodedItem) error {
+	// Unmarshal abi
+	contractAbi, err := abi.JSON(bytes.NewReader(item.Abi))
+	if err != nil {
+		return fmt.Errorf("error unmarshalling abi: %w", err)
+	}
+	// Parse data
+	if item.Event != nil { //nolint:nestif // has complex nested blocks (complexity: 12) (nestif)
+		abiEvent, abiEventArgs, err := abiparse.ParseEvent(item.Event.EventBody.Topics, item.Event.EventBody.Data, &contractAbi)
+		if err != nil {
+			queueIncompatibleEventUpdate(batch, p.runtime, item.Event.Round, item.Event.TxIndex)
+			p.logger.Warn("error processing event using abi", "contract address", item.ContractAddr, "err", err)
+			return nil
+		}
+		eventArgs, err := marshalArgs(abiEvent.Inputs, abiEventArgs)
+		if err != nil {
+			queueIncompatibleEventUpdate(batch, p.runtime, item.Event.Round, item.Event.TxIndex)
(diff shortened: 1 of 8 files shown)


pro-wh:
:+1:

analyzer/queries/queries.go
```diff
@@ -808,4 +834,104 @@ var (
...
+
+	RuntimeEvmVerifiedContractTxs = `
+    WITH abi_contracts AS (
+      SELECT 
+        runtime,
+        contract_address AS addr,
+        abi
+      FROM chain.evm_contracts
+      WHERE
+        runtime = $1 AND verification_info_downloaded_at IS NOT NULL
```
mitjat, line 846:
```suggestion
        runtime = $1 AND abi IS NOT NULL
```
Nit. This is a little more explicit and a little safer (because we're not relying on the fact that other code will only ever set `downloaded_at` when they also set `abi`)

analyzer/queries/queries.go
```diff
@@ -808,4 +834,104 @@ var (
...
+      abi_contracts.addr,
+      abi_contracts.abi,
+      txs.tx_hash,
+      txs.body->'data',
+      txs.error_message_raw,
+    FROM abi_contracts 
+    JOIN chain.runtime_transactions as txs ON
+      txs.runtime = abi_contracts.runtime AND
+      txs.to = abi_contracts.addr AND
+      txs.method = 'evm.Call' -- note: does not include evm.Create txs
```
mitjat, line 858:
```suggestion
      txs.method = 'evm.Call' -- note: does not include evm.Create txs; their payload is never encrypted
```

analyzer/queries/queries.go
```diff
@@ -808,4 +834,104 @@ var (
...
+    WHERE
+      runtime = $1 AND
```
mitjat, line 860: Should/can we further narrow the selection here by checking for empty `body` and `error_message_raw`?

analyzer/queries/queries.go
```diff
@@ -808,4 +834,104 @@ var (
...
+        runtime = $1 AND verification_info_downloaded_at IS NOT NULL
+    )
+    SELECT COUNT(*)
+    FROM abi_contracts 
+    JOIN chain.runtime_transactions as txs ON
+      txs.runtime = abi_contracts.runtime AND
+      txs.to = abi_contracts.addr AND
+      txs.method = 'evm.Call' -- note: does not include evm.Create txs
+    WHERE
+      runtime = $1 AND -- todo: redundant; confirm that it does not improve performance and remove
```
mitjat, line 881: This query (as a whole) feels like it could be very expensive. At the same time, it is not very important; it's just for our internal grafana dashboard.
Can you try running it manually on the prod db? If it takes more than a second or maybe two, I'd say that's undue DB load. In that case, I suggest we use the same approach we use for pagination, and be ok with clipping the true count. So, to run something like
```go
db.Query("SELECT count(*) FROM (" + RuntimeEvmVerifiedContractTxs + ") foo"`, 10000)
// ^ 10000 is the limit for the inner query
```

analyzer/queries/queries.go
```diff
@@ -808,4 +834,104 @@ var (
...
+    JOIN chain.runtime_events as evs ON
+      evs.runtime = contracts.runtime AND
+      evs.round = txs.round AND
+      evs.tx_index = txs.tx_index
+    WHERE
+      evs.processed_at IS NULL
+    ORDER BY addr
+    LIMIT $2`
+
+	RuntimeEvmVerifiedContractEventsCount = `
```
mitjat, line 915: Same performance concern and suggestion as for tx count

analyzer/evmabibackfill/evm_abi_backfill.go
```diff
@@ -0,0 +1,214 @@
...
+			item.Event.Round,
+			item.Event.TxIndex,
+			abiEvent.RawName,
+			eventArgs,
+			abiEvent.Sig,
+		)
+	} else if item.Tx != nil {
+		method, args, err := abiparse.ParseData(item.Tx.TxData, &contractAbi)
+		if err != nil {
+			return fmt.Errorf("error processing tx using abi for contract %s: %w", item.ContractAddr, err)
```
mitjat, line 160: This and other error handling in this func: If ABI processing fails, it's almost certainly because of uncompatible data+ABI, which is the tx's fault. We should not fail the item (which in turn will fail the batch), we should mark the item as processed in the DB and just not populate the parsed fields.

storage/migrations/02_runtimes.up.sql
```diff
@@ -54,6 +54,10 @@ CREATE TABLE chain.runtime_transactions
   "to"        oasis_addr,   -- Exact semantics depend on method. Extracted from body; for convenience only.
   amount      UINT_NUMERIC, -- Exact semantics depend on method. Extracted from body; for convenience only.
 
+  -- Added in 22_runtime_abi.up.sql
+  -- evm_fn_name TEXT,
+  -- evm_fn_params JSONB,
+
```
mitjat, line 60: Nice, thank you!
Edit: The filename is outdated by now :)

mitjat (commented):
Thank you! Looks good, only one substantial comment around error handling.

Sorry for the late reply; I did do an early pass of the PR, but I see now that it remained unsubmitted. It doesn't change anything though.

analyzer/queries/queries.go
Andrew7234: good idea, added!

analyzer/queries/queries.go
Andrew7234: good eye - turns out even the original queries used in `GetItems` are too slow for prod emerald bc of the join on the events table... will experiment further

analyzer/evmabibackfill/evm_abi_backfill.go
Andrew7234: oh good catch; fixed!

storage/migrations/22_runtime_abi.up.sql
```diff
@@ -0,0 +1,24 @@
+BEGIN;
+
+ALTER TABLE chain.runtime_transactions
+    -- For evm.Call transactions, we store both the name of the function and
+    -- the function paramters.
+    ADD COLUMN evm_fn_name TEXT,
+    ADD COLUMN evm_fn_params JSONB,
```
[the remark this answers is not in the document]
Andrew7234: Hm actually the JSONB will have a list of (k,v) tuples, where k is the param name and v is the param value. It shouldn't need to be JOINed with the abi since we do that during event parsing already.

From my understanding the mapping of event param names to values should be 1:1. Warren's [ParseEvent](https://github.com/oasisprotocol/nexus/blob/main/analyzer/runtime/abiparse/abiparse.go#L109) extracts the input values whether they're stored in the topics or in the event data, which should be exhaustive. We fetch param names from the [Event](https://pkg.go.dev/github.com/ethereum/go-ethereum/accounts/abi#Event) Inputs, which should also be exhaustive since go eth will [autogenerate names for unnamed args](https://github.com/ethereum/go-ethereum/blob/v1.13.5/accounts/abi/event.go#L70).

analyzer/queries/queries.go
Andrew7234: Added the clipped query for abi events; transactions query was fast enough

analyzer/queries/queries.go
```diff
@@ -850,4 +876,108 @@ var (
...
+      txs.tx_index,
+      evs.body
+    FROM abi_contracts
+    JOIN chain.runtime_transactions as txs ON
+      txs.runtime = abi_contracts.runtime AND
+      txs.to = abi_contracts.addr AND
+      txs.method = 'evm.Call'
+    JOIN chain.runtime_events as evs ON
+      evs.runtime = abi_contracts.runtime AND
+      evs.tx_hash = txs.tx_hash
```
mitjat, line 950: Based on today's slack conversation with Warren, we actually want to avoid joining on the txs because the contract that was the target of tx is not necessarily also the contract that emitted the event. The contract that emitted the event is in the event's `body.address`. So something like

```suggestion
    FROM abi_contracts
    JOIN chain.runtime_events as evs ON
      evs.type = 'evm.log' AND
      evs.runtime = abi_contracts.runtime AND
      decode(evs.body->>'address', 'base64') = eth_preimage(abi_contracts.contract_address)
```

That might require an index on body.address (IDK if it's needed, and/or if it would help), or maybe even a [generated column](https://www.postgresql.org/docs/current/ddl-generated-columns.html) for `decode(evs.body->>'address', 'base64')` and then an index on top of _that_. See how it goes with the simplest approach first :)

analyzer/queries/queries.go
```diff
@@ -453,10 +453,34 @@ var (
...
+      processed_at = CURRENT_TIMESTAMP
+    WHERE
+      runtime = $1 AND
+      tx_hash = $2`
+
 	RuntimeEventInsert = `
     INSERT INTO chain.runtime_events (runtime, round, tx_index, tx_hash, tx_eth_hash, timestamp, type, body, related_accounts, evm_log_name, evm_log_params, evm_log_signature)
       VALUES ($1, $2, $3, $4, $5, $6, $7, $8, $9, $10, $11, $12)`
 
+	RuntimeEventUpdate = `
```
mitjat, line 472:
```suggestion
	RuntimeEventEvmParsedFieldsUpdate = `
```

or similar, so it's clearer from the name what we're doing. Same for `RuntimeTransactionUpdate`.

mitjat (commented):
Thanks! I realized there's another gotcha around events that might in require a little more performance testing/tweaking. Sorry we didn't think of it before. But in general, getting there :)

analyzer/queries/queries.go
```diff
@@ -850,4 +876,108 @@ var (
...
+      JOIN chain.runtime_events as evs ON
+        evs.runtime = abi_contracts.runtime AND
+        evs.tx_hash = txs.tx_hash
+      WHERE
+        evs.abi_parsed_at IS NULL
+      LIMIT $2 -- limit count for performance reasons
+    )
+    SELECT 
+      COUNT(*)
+    FROM abi_events;`
```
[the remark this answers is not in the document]
Andrew7234: Got it, updated!

analyzer/queries/queries.go
Andrew7234: Hrm it's slow, even with a generated column. Seems like the `eth_preimage` fn is the bottleneck - will try to just join over address_preimages instead
```
oasisindexer=*> explain analyze
WITH abi_contracts AS (
      SELECT
        runtime,
        contract_address AS addr,
        abi
      FROM chain.evm_contracts
      WHERE
        runtime = 'emerald' AND
        abi IS NOT NULL
    )
    SELECT
      abi_contracts.addr,
      abi_contracts.abi,
      evs.round,
      evs.tx_index,
      evs.body
    FROM abi_contracts
    JOIN chain.runtime_events as evs ON
      evs.type = 'evm.log' AND
      evs.runtime = abi_contracts.runtime AND
      evm_contract_address_eth = eth_preimage(abi_contracts.addr)
    LIMIT 1000;
                                                                  QUERY PLAN
----------------------------------------------------------------------------------------------------------------------------------------------
 Limit  (cost=0.00..181689.03 rows=1000 width=982) (actual time=42717.383..61984.504 rows=197 loops=1)
   ->  Nested Loop  (cost=0.00..2232049.74 rows=12285 width=982) (actual time=42717.381..61984.447 rows=197 loops=1)
         Join Filter: (evs.evm_contract_address_eth = eth_preimage(evm_contracts.contract_address))
         Rows Removed by Join Filter: 326507
         ->  Seq Scan on runtime_events evs  (cost=0.00..1580270.46 rows=614246 width=215) (actual time=15.804..55033.975 rows=46672 loops=1)
               Filter: ((runtime = 'emerald'::runtime) AND (type = 'evm.log'::text))
               Rows Removed by Filter: 20543893
         ->  Materialize  (cost=0.00..678.53 rows=4 width=807) (actual time=0.000..0.001 rows=7 loops=46672)
               ->  Seq Scan on evm_contracts  (cost=0.00..678.51 rows=4 width=807) (actual time=1.088..2.911 rows=7 loops=1)
                     Filter: ((abi IS NOT NULL) AND (runtime = 'emerald'::runtime))
                     Rows Removed by Filter: 7988
 Planning Time: 3.994 ms
 Execution Time: 61984.703 ms
(13 rows)
```

mitjat: :+1: is simple to use but not very efficient. That said, I thought it wouldn't matter here because we have so few verified contracts. But the query plan doesn't choose to go contract by contract, trying to find matching events for each contract :(

Andrew7234: Updated to join on `chain.address_preimages`. Also added an index on `chain.runtime_events (runtime, type)`, which makes the query performant enough.
```
oasisindexer=*> explain analyze
WITH abi_contracts AS (
      SELECT
        runtime,
        contract_address AS addr,
        abi
      FROM chain.evm_contracts
      WHERE
        runtime = 'emerald' AND
        abi IS NOT NULL
    )
    SELECT
      abi_contracts.addr,
      abi_contracts.abi,
      evs.round,
      evs.tx_index,
      evs.body
    FROM abi_contracts
    join chain.address_preimages as preimages on abi_contracts.addr = preimages.address JOIN chain.runtime_events as evs ON
      evs.type = 'evm.log' AND
      evs.runtime = abi_contracts.runtime AND
      decode(body->>'address', 'base64') = preimages.address_data
    LIMIT 1000;
                                                                                     QUERY PLAN
------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
 Limit  (cost=10154.37..1170949.61 rows=114 width=982) (actual time=257.164..2266.097 rows=197 loops=1)
   ->  Gather  (cost=10154.37..1170949.61 rows=114 width=982) (actual time=257.162..2266.052 rows=197 loops=1)
         Workers Planned: 2
         Workers Launched: 2
         ->  Hash Join  (cost=9154.37..1169938.21 rows=48 width=982) (actual time=1015.169..2243.789 rows=66 loops=3)
               Hash Cond: (decode((evs.body ->> 'address'::text), 'base64'::text) = preimages.address_data)
               ->  Parallel Bitmap Heap Scan on runtime_events evs  (cost=8442.09..1167623.43 rows=256322 width=183) (actual time=5.136..2204.427 rows=15564 loops=3)
                     Recheck Cond: ((runtime = 'emerald'::runtime) AND (type = 'evm.log'::text))
                     Heap Blocks: exact=7413
                     ->  Bitmap Index Scan on ix_runtime_events_type  (cost=0.00..8288.29 rows=615173 width=0) (actual time=9.769..9.769 rows=46692 loops=1)
                           Index Cond: ((runtime = 'emerald'::runtime) AND (type = 'evm.log'::text))
               ->  Hash  (cost=712.23..712.23 rows=4 width=828) (actual time=5.272..5.274 rows=7 loops=3)
                     Buckets: 1024  Batches: 1  Memory Usage: 14kB
                     ->  Nested Loop  (cost=0.41..712.23 rows=4 width=828) (actual time=1.554..5.236 rows=7 loops=3)
                           ->  Seq Scan on evm_contracts  (cost=0.00..678.51 rows=4 width=807) (actual time=1.501..5.043 rows=7 loops=3)
                                 Filter: ((abi IS NOT NULL) AND (runtime = 'emerald'::runtime))
                                 Rows Removed by Filter: 7988
                           ->  Index Scan using address_preimages_pkey on address_preimages preimages  (cost=0.41..8.43 rows=1 width=68) (actual time=0.024..0.024 rows=1 loops=21)
                                 Index Cond: ((address)::text = (evm_contracts.contract_address)::text)
 Planning Time: 1.216 ms
 Execution Time: 2266.211 ms
(21 rows)
```

Side note: The generated column for the event address ended up being not needed, but will keep in mind for future performance testing.

mitjat (approved):
:tada: 

It's unfortunate that our e2e (or other) tests don't currently cover sourcify, and so cannot test this. I assume you tested locally. Luck favors the brave :roll_eyes:

PR #544, opened 2023-10-12
parse unencrypted results for confidential txs

Andrew7234:
https://app.clickup.com/t/8692tq704

```
oasisindexer=> select count(*) from chain.runtime_transactions where evm_encrypted_format='encrypted/x25519-deoxysii' and method='evm.Call' and success=true and evm_encrypted_result_data is not NULL;
 count
-------
 15318 <-- encrypted results
(1 row)

oasisindexer=> select count(*) from chain.runtime_transactions where evm_encrypted_format='encrypted/x25519-deoxysii' and method='evm.Call' and success=true and evm_encrypted_result_data is NULL;
 count
--------
 212680 <-- unencrypted results
```

Before:
```
indexer=# select runtime, round, tx_index, method, success, evm_encrypted_result_data from chain.runtime_transactions
;
 runtime  |  round  | tx_index |  method  | success | evm_encrypted_result_data
----------+---------+----------+----------+---------+---------------------------
 sapphire | 2993605 |        0 | evm.Call | t       |
 sapphire | 2993605 |        1 | evm.Call | t       |
 sapphire | 2993606 |        0 | evm.Call | t       |
 sapphire | 2993606 |        1 | evm.Call | t       |
 sapphire | 2993607 |        0 | evm.Call | t       |
 sapphire | 2993607 |        1 | evm.Call | t       |
 sapphire | 2993608 |        0 | evm.Call | f       |
(7 rows)
```

After
```
indexer=# select runtime, round, tx_index, method, success, evm_encrypted_result_data from chain.runtime_transactions
;
 runtime  |  round  | tx_index |  method  | success |          evm_encrypted_result_data
----------+---------+----------+----------+---------+----------------------------------------------
 sapphire | 2993605 |        0 | evm.Call | t       | \x76115a6703f475bdc9eee0747ba441cedc4ed9ebfe
 sapphire | 2993605 |        1 | evm.Call | t       | \x7c495d2602e702b7b490779a5e029f436fcfb02c7b
 sapphire | 2993606 |        0 | evm.Call | t       |
 sapphire | 2993606 |        1 | evm.Call | t       | \x40d51310f0ab71197168e70d45b3bff595c10840d5
 sapphire | 2993607 |        0 | evm.Call | t       | \xa8eafc2f40b431323d3017142d8c87208f8e5d8d28
 sapphire | 2993607 |        1 | evm.Call | t       | \x74c6ed9bd74cdbfeec89c4c55e3ada66dc94f2b8b5
 sapphire | 2993608 |        0 | evm.Call | f       |
(7 rows)
```

On testnet, after block 542000 we stopped parsing the tx results correctly... will need a full reindex.

Future Todo: Make sure this is covered in the e2e_regression tests when we add sapphire to them.

### DIFF
--- a/analyzer/runtime/evm/client.go
+++ b/analyzer/runtime/evm/client.go
@@ -74,6 +74,11 @@ type EVMEncryptedData struct {
 	DataData    []byte
 	ResultNonce []byte
 	ResultData  []byte
+	// Unlike `ResultData` and `DataData`, this field is not encrypted data. Rather,
+	// it is extracted during encrypted tx parsing and processed in extract.go.
+	// If non-null, the tx is an older Sapphire tx and it failed, and `ResultNonce`
+	// and `ResultData` will be empty.
+	FailedCallResult *sdkTypes.FailedCallResult
 }
 
 type EVMContractData struct {
@@ -307,6 +312,111 @@ func EVMDownloadTokenBalance(ctx context.Context, logger *log.Logger, source nod
 // EVMMaybeUnmarshalEncryptedData breaks down a possibly encrypted data +
 // result into their encryption envelope fields. If the data is not encrypted,
 // it returns nil with no error.
+//
+// LESSON: Oasis paratime transaction structure.
+//
+// Oasis paratime transactions conform to the oasis format transaction payload:
+// https://github.com/oasisprotocol/oasis-sdk/blob/04e8918a9b3eb90caffe115ea2d46b5742438714/client-sdk/go/types/transaction.go#L212
+//
+// Transaction payloads may be encrypted in oasis. To express that, the
+// Transaction.Call field of the transaction would have a call format set, e.g.
+//     Call.Format: 'encrypted/x25519-deoxysii'
+//
+// This gives rise to the concept of outer/inner calls for oasis-format
+// transactions, where the inner call would be encrypted and stored in the
+// outer Call.Body. However, when the runtime evm module converts ethereum
+// format transactions to oasis format transaction payloads, they are never
+// in this encrypted mode because there is no native ethereum way to express
+// an encrypted transaction.
+//
+// Thus, for ethereum format transactions, the outer oasis format Call.Format
+// is always 0/plain (or equivalently, unset). This outer Call directly has
+// the method and body of the transaction. For example, the method may be
+// "evm.Call" and the body would be the corresponding struct:
+// https://github.com/oasisprotocol/oasis-sdk/blob/04e8918a9b3eb90caffe115ea2d46b5742438714/client-sdk/go/modules/evm/types.go#L13
+//
+// Here is where we may encounter sapphire-style encryption. In the first and
+// most basic case, there is no sapphire-style encryption at all and the fields
+// of the evm.Call struct are the exact bytes that would go into the evm.
+// Note that Sapphire supports this mode.
+//
+// The result of this transaction is stored in the associated CallResult struct
+// that is returned alongside the transaction in TransactionWithResult:
+// https://github.com/oasisprotocol/oasis-sdk/blob/main/client-sdk/go/client/client.go#L146
+//
+// A transaction may also be in sapphire encrypted mode. In this case, the
+// evm.Call.Data or evm.Create.Initcode field is itself a cbor-encoded oasis
+// format runtime-sdk Call struct that we term the "sapphire outer call". This
+// struct is the same oasis format Transaction.Call mentioned above:
+// https://github.com/oasisprotocol/oasis-sdk/blob/04e8918a9b3eb90caffe115ea2d46b5742438714/client-sdk/go/types/transaction.go#L363
+//
+// In encrypted mode, the sapphire outer call has a Call.Format of
+// 'encrypted/x25519-deoxysii', and the call.Body is a cbor-encoded encryption
+// envelope, which can be decrypted and parsed into another oasis format Call
+// struct that we term the "sapphire inner call". Note that Nexus never has
+// the keys the decrypt this envelope to get the inner struct since the
+// Sapphire transaction is encrypted. The sapphire inner call.Body has the evm
+// input data, and the inner call.Format is always unset.
+//
+// The results of sapphire encrypted transactions are stored similarly. In the
+// first case, we should note that sapphire encrypted transactions may have
+// unencrypted results if the inputs were not encrypted. Unencrypted transaction
+// results are stored in the same associated CallResult as the results of
+// unencrypted transactions. We term this CallResult the "outer call result."
+//
+// Encrypted results can be recovered by cbor-unmarshalling the outer
+// CallResult.Ok field into another CallResult struct that we term the
+// "sapphire outer call result." Currently, the sapphire outer CallResult.Ok
+// stores the result envelope of successful sapphire encrypted transactions.
+// However, in older versions of Sapphire and for non-evm transactions, this
+// result envelope is stored in the sapphire outer CallResult.Unknown.
+//
+// The result envelope contains the encrypted "sapphire inner call result",
+// where the sapphire inner CallResult.Ok field contains the output data
+// from the evm. Note that Nexus does not have the keys to decrypt and access
+// this sapphire inner call result.
+//
+// Side note: The CallResult struct does not specify its own encryption format.
+// Rather, it follows the format of its corresponding oasis format Call. The
+// "outer call" Call.Format corresponds to the outer call result, and the
+// "sapphire outer call" Call.Format corresponds to the sapphire outer call
+// result.
+//
+// Currently, if an encrypted transaction fails, the error is bubbled up to
+// the outer CallResult.Failed. However, in older versions of Sapphire the
+// error was stored in the sapphire outer CallResult.Failed.
+//
+// Lastly, there's also a "not-encryption-but-sapphire-something" case where
+// the sapphire outer Call.Format is set to "plain". In this case, there's no
+// sapphire inner call and the sapphire outer Call.Body has the evm input data.
+// Nexus does not extract the evm input data directly here like the way it does
+// for unencrypted evm.Call transactions.
+//
+// The results of these "plain" transactions are returned in the outer
+// CallResult.
+//
+// Outer oasis Call.Body
+//   -> inner oasis Call
+//   -> evm.Call
+//        -> sapphire outer Call.Body
+//             -> sapphire inner Call
+//
+// To summarize:
+// * Non-Sapphire
+//   * Tx Body: Stored in oasis outer Call.Body
+//   * Success Result: Stored in outer oasis CallResult.Ok
+//   * Failure Result: Stored in outer oasis CallResult.Failed
+// * Sapphire Encrypted
+//   * Tx Body: Encryption envelope stored in sapphire outer Call.Body
+//   * Success Result: Encryption envelope stored in either sapphire outer
+//                     CallResult.Ok (new) or CallResult.Unknown (old)
+//   * Failure Result: Stored in either the sapphire outer CallResult.Failed (old)
+//                     or the outer CallResult.Failed (new)
+// * Sapphire Plain(0)
+//   * Tx Body: Stored in the sapphire outer Call.Body
+//   * Success Result: Stored in outer oasis CallResult.Ok
+//   * Failure Result: Stored in outer oasis CallResult.Failed
+
 func EVMMaybeUnmarshalEncryptedData(data []byte, result *[]byte) (*EVMEncryptedData, error) {
 	var encryptedData EVMEncryptedData
 	var call sdkTypes.Call
@@ -325,6 +435,9 @@ func EVMMaybeUnmarshalEncryptedData(data []byte, result *[]byte) (*EVMEncryptedD
 		encryptedData.PublicKey = callEnvelope.Pk[:]
 		encryptedData.DataNonce = callEnvelope.Nonce[:]
 		encryptedData.DataData = callEnvelope.Data
+	// Plain txs have no encrypted fields to extract.
+	case sdkTypes.CallFormatPlain:
+		return nil, nil
 	// If you are adding new call formats, remember to add them to the
 	// database call_format enum too.
 	default:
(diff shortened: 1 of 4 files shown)


analyzer/runtime/evm/client.go
```diff
@@ -313,21 +313,30 @@ func EVMMaybeUnmarshalEncryptedData(data []byte, result *[]byte) (*EVMEncryptedD
...
-				return nil, fmt.Errorf("outer call format %s (%d) not supported", call.Format, call.Format)
+				return nil, fmt.Errorf("inner callResult failed")
 			}
+			encryptedData.ResultNonce = resultEnvelope.Nonce[:]
+			encryptedData.ResultData = resultEnvelope.Data
+		default:
+			// We have already checked when decoding the call envelope,
+			// but I'm keeping this default case here so we don't forget
+			// if this code gets restructured.
+			return nil, fmt.Errorf("outer call format %s (%d) not supported", call.Format, call.Format)
```
mitjat, line 339: Heads up: I've noticed this error in logs quite a bit:
```
{
analyzer: "sapphire"
caller: "extract.go:499"
err: "outer call format plain (0) not supported"
level: "error"
module: "analysis_service"
msg: "error unmarshalling encrypted data and result, omitting encrypted fields"
}
```

The `caller` above is a line in `extract.go` that calls the method from this PR (`EVMMaybeUnmarshalEncryptedData`). So basically, we've got txs with `call.Format == sdkTypes.CallFormatPlain`, and we're not handling them. (This function generates the same error message twice, once here and once in line 309.)
I haven't created a ticket for this IIRC, but I'm wondering if it's on your radar already since you've been looking at encryption a lot lately? It's probably worth handling in the next days as you're working in this area.

analyzer/runtime/evm/client.go
```diff
@@ -313,21 +313,30 @@ func EVMMaybeUnmarshalEncryptedData(data []byte, result *[]byte) (*EVMEncryptedD
...
+			var resultEnvelope sdkTypes.ResultEnvelopeX25519DeoxysII
+			switch {
+			// The result is encrypted.
+			case callResult.IsUnknown():
 				if err := cbor.Unmarshal(callResult.Unknown, &resultEnvelope); err != nil {
-					return nil, fmt.Errorf("outer call result unmarshal unknown: %w", err)
+					return nil, fmt.Errorf("outer call result unmarshal ok: %w", err)
+				}
+			// The result can be unencrypted if the input was unencrypted.
+			case callResult.IsSuccess() && !callResult.IsUnknown():
```
mitjat, line 326: Which of `callResult.IsSuccess()` and `!callResult.IsUnknown()` indicates unencrypted input? (I'm referring to your comment.)

Can a `callResult` have `IsSuccess()` and `IsUnknown()` at the same time? If not, let's just check for `IsSuccess()` here. Edit: Urgh I see how those two methods are defined. Not nice. Worth commenting on here, I think.

Why do we handle the case of tx status=failed (as opposed to status={ok, unknown}) in such a different place, [here](https://github.com/oasisprotocol/nexus/blob/5881904c1f9a6de691dabda76152f33c0d98d7f5/analyzer/runtime/extract.go#L344-L344)? Are error messages never encrypted? Also worth commenting on here..

mitjat (commented):
Thank yo ufor making sense of this.
I only have comments/questions that will hopefully make it make sense to myself and future maintainers.

mitjat:
> On testnet, after block 542000 we stopped parsing the tx results correctly... will need a full reindex.

Huh. How come we were parsing them correctly before? / What caused us to start indexing incorrectly?
Also, this is about the production deployment, presumably?

analyzer/runtime/evm/client.go
Andrew7234: Thanks for raising this! I think in this case, the analyzer is working as expected since we aren't able to extract `evm_encrypted_*` fields from a plain tx. I'll add an explicit switch case to reduce the log spam here though.

The one equivalent field we _could_ extract is the `data_data`, eg the evm call data. However, it likely requires its own dedicated column, and also we currently store that as part of the `body -> data`.

analyzer/runtime/evm/client.go
```diff
@@ -313,21 +313,30 @@ func EVMMaybeUnmarshalEncryptedData(data []byte, result *[]byte) (*EVMEncryptedD
...
-				var resultEnvelope sdkTypes.ResultEnvelopeX25519DeoxysII
+		switch call.Format {
+		case sdkTypes.CallFormatEncryptedX25519DeoxysII:
+			var resultEnvelope sdkTypes.ResultEnvelopeX25519DeoxysII
+			switch {
+			// The result is encrypted.
+			case callResult.IsUnknown():
 				if err := cbor.Unmarshal(callResult.Unknown, &resultEnvelope); err != nil {
-					return nil, fmt.Errorf("outer call result unmarshal unknown: %w", err)
+					return nil, fmt.Errorf("outer call result unmarshal ok: %w", err)
```
pro-wh, line 323: why is this changed to 'ok'? we're still unmarshalling the .Unknown

analyzer/runtime/evm/client.go
pro-wh: `!callResult.IsUnknown()` should be covered by the above case not matching

analyzer/runtime/evm/client.go
pro-wh: wait what do you mean by "twice" and "log spam"? it shows up in the code lexically twice, but they return when either one happens

Andrew7234:
> Huh. How come we were parsing them correctly before? / What caused us to start indexing incorrectly?

Hmm iirc there was some talk of how the data in `callResult.Unknown` was moved to `callResult.Ok` for encrypted txs, but I'll need to follow up on that

> Also, this is about the production deployment, presumably?

Ah it's actually staging testnet

analyzer/runtime/evm/client.go
pro-wh: I need to read through this thread more tho

analyzer/runtime/visitors.go
```diff
@@ -111,7 +111,7 @@ func VisitCall(call *sdkTypes.Call, result *sdkTypes.CallResult, handler *CallHa
 			if !result.IsUnknown() && result.IsSuccess() {
 				var ok []byte
 				if err := cbor.Unmarshal(result.Ok, &ok); err != nil {
-					return fmt.Errorf("unmarshal evm create result: %w", err)
+					return fmt.Errorf("unmarshal evm call result: %w", err)
```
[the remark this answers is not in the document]
pro-wh: oops thanks

analyzer/runtime/evm/client.go
Andrew7234: fixed!

analyzer/runtime/evm/client.go
pro-wh: ok I don't get the code comment here either.

I thought the .Unknown, .Ok, and .Failed are mutually exclusive. Unknown is allowed by the SDK, but I think I heard that Sapphire moved to making the successfulness of transactions public, meaning it chooses to use .Ok and .Failed instead of .Unknown.

Additionally, I think the implication in this PR is that .Failed is always in cleartext in Sapphire? Is that the case? Is there a systematic way to tell whether .Failed is encrypted or not?

pro-wh: https://github.com/oasisprotocol/oasis-sdk/blob/runtime-sdk/v0.7.4/runtime-sdk/src/callformat.rs#L253 yes it look like failure gets passed through unencrypted under this `expose_failure` setting

pro-wh: ok correction: it's not Sapphire specifically doing this, newer versions of the SDK do this for the whole EVM module https://github.com/oasisprotocol/oasis-sdk/blob/runtime-sdk/v0.7.4/runtime-sdk/modules/evm/src/lib.rs#L651

analyzer/runtime/evm/client.go
Andrew7234: Hm so when the outer call format is `plain (0)`, do we still want to log an error? My original idea was to simply return no error in this case, since the docstring says that `If the data is not encrypted,it returns nil with no error.`. I might be missing something though. If the callformat is 0, does it guarantee that the transaction is not encrypted?

pro-wh: "if the data is not encrypted" here refers to a certain thing about how Sapphire style encryption is. the data field of an otherwise unencrypted Ethereum format transaction is a valid CBOR-encoded Oasis `Call` structure (the hope is that no contract will incidentally expect anything compatible as the actual data, initcode for "create" as well).

so the phrase in the description (as well as the "Maybe" in the name) there refers to handling this ambiguity. it's only to say you can pass in whatever the data/initcode you get and it'll do the right thing.

as for how plain (0) should fit in, imo return nil with no error. the goal of this function is to populate the encrypted envelope fields for insertion into the database. and the change in this PR to look for the encrypted envelope in callResult.Ok is correct too.

pro-wh: addendum: now that I think about it, maybe it's more suitable to handle invalid `result` CBOR with similar logic inside this function instead of special-casing the code around it to pass `nil` for "create."

analyzer/runtime/evm/client.go
aefhm: Agree with @pro-wh

aefhm: Oof, I suppose the history is tough because we need to support pre-SDK change where errors and successes were both in `Unknown`.

pro-wh: pre-change will be fine, we continue to look in Unknown if it's Unknown

mitjat: > My current understanding is that if the inner sdk callResult fails, that error is bubbled up to the outer evm callResult.Failed and so we can handle failed txs in one place at the higher level

This is a good comment, can you add it as a code comment to the `default:` branch on line 338 (which is where we handle errors)?

By `callResult.Failed`, do you mean `txr.Failed` as checked [here](https://github.com/oasisprotocol/nexus/blob/5881904c1f9a6de691dabda76152f33c0d98d7f5/analyzer/runtime/extract.go#L344-L344)?

Regardless, I'm still a little confused by the flow: If the transaction was encrypted and failed, we'll detect the failure in the `default:` branch of `case` (line 338), we'll return an error, but our caller is here and it just logs a parsing failure.
https://github.com/oasisprotocol/nexus/blob/9e33faed3fc86d90edd16892b51826a38cc569fc/analyzer/runtime/extract.go#L514-L517
Should we set `chain.runtime_transactions.success`? At the very least, we probably shouldn't be logging a parsing error.

mitjat (approved):
Thank you. Approving so it can be merged when convenient. The first comment thread still needs more work though, or I'm misunderstanding something.

analyzer/runtime/evm/client.go
Andrew7234: yup!

analyzer/runtime/evm/client.go
Andrew7234: confirmed that when the tx fails, the parsing error isn't reached since `VisitCall` sets result to nil. I'll add a coment

pro-wh: it could be unknown outer + fail inner for non-evm transactions and for old transactions

mitjat:
IIRC you mentioned in Split that the logic likely needs to fixed in some way still?

I don't remember the conclusion in terms of actual code, but I do remember we said it would be useful to document the expected layout (i.e. contents of outer and inner result) for all the combos:
{old format, new format} x {success, failure} x {encrypted, "encrypted" with Plain format, unencrypted}

It's a lot of them, but that's also the reason we're having such long conversations, and lower-than-ideal confidence in correctness. If we can make meaningful unit tests for them, even better. (That's kinda scope creep for the PR, so feel free to push back. But it's only _kinda_ scope creep, and it covers very real existing tech/documentation debt.)

If you don't know the answer for some of the combos, we can produce that combo on testnet, or find it in the wild, or ask around.

As I've felt several times with this PR: Thank you for your service 🫡

analyzer/runtime/evm/client.go
```diff
@@ -313,21 +317,40 @@ func EVMMaybeUnmarshalEncryptedData(data []byte, result *[]byte) (*EVMEncryptedD
...
+			// to run iff the call succeeded.
+			case callResult.IsSuccess() && !callResult.IsUnknown():
+				if err := cbor.Unmarshal(callResult.Ok, &resultEnvelope); err != nil {
+					return nil, fmt.Errorf("outer call result unmarshal ok: %w", err)
+				}
+			// For non-evm txs as well as older Sapphire txs, the outer callResult may
+			// be Unknown and the inner callResult Failed. In this case, we extract the
+			// failed callResult.
+			case callResult.Failed != nil:
+				encryptedData.FailedCallResult = callResult.Failed
```
Andrew7234, line 343: per pro-wh's comment:
> it could be unknown outer + fail inner for non-evm transactions and for old transactions

Adjusted to extract the `FailedCallResult` and bubble it up to the calling fn in `extract.go`, which in turn uses the failed call result to populate the error fields.

Andrew7234:
Ok added tests and confirmed that they work as expected. I still need to locate some older sapphire encrypted txs that have the tx/results wrapped in the sapphire outer call within the oasis outer call, which requires archive node access

mitjat (approved):
We were discussing this ticket with Tadej during sprint planning, and the one thing he immediately said when he learned about your [writeup in Google Docs](https://docs.google.com/document/d/1LHb94T0i9IpT_zI6WU80kiErtT7T1asvdOsKrcPoNE0/edit) was how nice it would be to have it close to the codebase so it doesn't get lost.

You did add nice inline code comments but they serve more so the function of "now that we know how the data is laid out, the comments guide us through how the nested ifs correspond to that layout".
I too would like to encourage you to include the contents of the Google Doc (with links converted to Markdown(?), and the table rendered as nested bullet points) in a place where a) it's discoverable when browsing the code and b) it's accessible to the public. It would be a shame if the results of days of digging and summarizing end up rotting away in some forgotten doc.

Some placement options:
- Huge docstring for `EVMMaybeUnmarshalEncryptedData()`
- A separate markdown file (in this directory?), referenced prominently from the docstring of `EVMMaybeUnmarshalEncryptedData()`.

I'm re-LGTMing -- this feels well researched now, and even if it still fails in some corner case, we can deal with that later. I bet it will feel good to move on :). Thank you!

analyzer/runtime/extract_test.go
```diff
@@ -12,6 +19,26 @@ type mockError struct {
...
+}
+
+type mockEvmEncrypted struct {
+	Format      common.CallFormat
+	PublicKey   string
+	DataNonce   string
+	DataData    string
+	ResultNonce string
+	ResultData  string
+}
```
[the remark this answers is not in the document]
Andrew7234: Good eye - I still need to find some old txs that store the failed result in the `outer sapphire call result`. Will add shortly

analyzer/runtime/evm/client.go
```diff
@@ -68,12 +68,13 @@ type EVMTokenBalanceData struct {
...
-	DataData    []byte
-	ResultNonce []byte
-	ResultData  []byte
+	Format           common.CallFormat
+	PublicKey        []byte
+	DataNonce        []byte
+	DataData         []byte
+	ResultNonce      []byte
+	ResultData       []byte
+	FailedCallResult *sdkTypes.FailedCallResult
```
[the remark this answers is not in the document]
Andrew7234: Added! Updated to 
```
        // Unlike `ResultData` and `DataData`, this field is not encrypted data. Rather,
	// it is extracted during encrypted tx parsing and processed in extract.go.
	// If non-null, the tx is an older Sapphire tx and it failed, and `ResultNonce`
	// and `ResultData` will be empty.
```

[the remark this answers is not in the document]
mitjat: > If non-null, the tx is an older Sapphire tx

Oh. So where do we store failure details for all other transactions? (notably, new Sapphire)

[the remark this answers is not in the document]
Andrew7234: It's stored in the "outer call result" one level up

analyzer/runtime/extract_test.go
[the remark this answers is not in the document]
Andrew7234: Ok I couldn't find any - will keep searching and add in a follow on PR

PR #548, opened 2023-10-17
refactor: Vendor oasis-core v22.2, Use mostly vendored oasis-core

mitjat:
This PR is in preparation for supporting the E version (v23.x) of consensus. It is **functionally a no-op**, a refactor that will make v23 support easier/smoother.

Before this PR, the indexer-internal types were a mix of vendored v21.x (=Cobalt) oasis-core types, and unvendored oasis-core types.

**Changes at a high level:**
After this PR, nexus mostly only depends on _vendored_ oasis-core; the remaining uses of  "live" oasis-core are limited to a handful of very common and stable types like `staking.Address`, and to modules `/go/common/{cbor,quantity,crypto/address,entity}`
- NOTE: We used to also depend on the `/go/common/node` module of `oasis-core`, but the module types changed with v23, so in preparation for that, this PR also vendors `/go/common/node` and starts using it in Cobalt and Damask compatibility layers.

**Changes at a low level:**
The largest diffs are due to the vendored copy of oasis-core. Therefore, it's **probably better to review commit by commit**. Notable commits:
- `vendor-oasis-core: Support v22.2.11 (Damask)` - Tweaks in the vendoring script. The script was previously only used for vendoring v21.x, and v22.x required some small changes. Notably, it required the inclusion of a few more enum constants because we (ab)use the v22.x types for two purposes:
      1) For serde of gRPC requests/responses (as we also do for Cobalt, and later E)
      2) As nexus's internal data representation (this needs a few more enums to interpret/manipulate; in an ideal world, we'd create our own types here)
- `Add vendored oasis-core v22.2.11 (Damask)` - Huge diff, but it's just the result of running `vendor_oasis_core.sh`. Not worth reviewing IMO.
- `Replace usage of oasis-core with vendored 22.2.11; BREAKS THE BUILD` and `nodeapi/damask: Use raw gRPC instead of oasis-core wrappers` - What the first commit message says. The first commit is just a mechanic find-replace, the second fixes the build by removing the use of oasis-core's convenience gRPC client wrappers. (The vendored oasis-core doesn't have those, it only has types.)
- `vendor-oasis-core: Support to vendor "node" module too` and `Re-vendor oasis-core v22.2.11 (incl node module); BREAKS THE BUILD` and `Use the damask vendored `node` module for cobalt too`: As described above, this PR adds the `go/common/node` module from oasis-core into the vendored copy. These commits achieve that. I could interleave them with the above commits (which vendor all the other parts of oasis-core), but since it affects both Damask and Cobalt, I left it separate.

**Testing:**
- To double-check compatibility with old types: Indexed blocks 8_048_956 (Damask genesis) to 8_049_964 using a cache that predates this migration.
- To test e2e: Indexed blocks 8_048_956 (Damask genesis) to 8_065_753, verified there were no errors/warnings.


analyzer/consensus/consensus.go
```diff
@@ -9,10 +9,10 @@ import (
 	"os"
 	"strings"
 
+	genesis "github.com/oasisprotocol/nexus/coreapi/v22.2.11/genesis/api"
```
pro-wh: needs to go in later commit?

pro-wh (approved):
ya

pro-wh:
oh one other thing: is there any extra setup and logic in the oasis-core grpc client?

scripts/vendor-oasis-core/patches/v22.2.11/assorted.patch
```diff
@@ -0,0 +1,240 @@
...
+diff --git a/coreapi/v22.2.11/staking/api/address.go b/coreapi/v22.2.11/staking/api/address.go
+index 1f706bbe..ea721dce 100644
+--- a/coreapi/v22.2.11/staking/api/address.go
++++ b/coreapi/v22.2.11/staking/api/address.go
+@@ -1,7 +1,9 @@
+ package api
+ 
+ import (
+-        original "github.com/oasisprotocol/oasis-core/go/staking/api"
++	original "github.com/oasisprotocol/oasis-core/go/staking/api"
```
Andrew7234, line 169: just curious - do we not vendor these types because they're unlikely to change in future oasis-core versions?

storage/oasis/nodeapi/damask/node.go
```diff
@@ -27,48 +32,48 @@ import (
 // compatible with Damask gRPC API, this struct just trivially wraps the
 // convenience methods provided by oasis-core.
 type DamaskConsensusApiLite struct {
-	client consensus.ClientBackend
+	grpcConn *grpc.ClientConn
```
Andrew7234, line 35: will we use raw grpc for E* too?

Andrew7234 (approved):
thanks for going through all this, and the commit descriptions

scripts/vendor-oasis-core/patches/v22.2.11/assorted.patch
mitjat: Yeah, that's the hope anyway. If they do change, we can always go back and vendor them fully. (Just like I did in this PR with `node` -- exact same situation.)

Main motivation:
Because these types are so commonly used, it's very convenient to have them aliased to the exact same type across all our vendored copies of oasis-core. This way, Go is happier to perform automatic casts from e.g. the cobalt version of MyTypeThatContainsAddress to the damask version of MyTypeThatContainsAddress.

storage/oasis/nodeapi/damask/node.go
mitjat: Yes, I decided to go that way in order to lose the asymmetry between versions. The use of oasis-core helper functions (as opposed to raw grpc) in Damask was to a large extent just inertia from back when we only supported a single version of oasis-core. In the end, those helper funcs don't benefit us all that much IMO: Code is a tiny bit more readable and shorter, but not much.

mitjat:
> oh one other thing: is there any extra setup and logic in the oasis-core grpc client?

Shouldn't be, our connection setup is basically copy-pasted from oasis-core: https://github.com/oasisprotocol/nexus/blob/9e33faed3fc86d90edd16892b51826a38cc569fc/storage/oasis/connections/raw_grpc.go#L19-L19

pro-wh:
aw nice now I can hack this to use no TLS for a TCP connection

analyzer/consensus/consensus.go
mitjat: :+1: fixed

PR #549, opened 2023-10-18
Support node v23.0

mitjat:
This PR adds support for querying and indexing the E version (v23.x) of consensus.

Similarly to #548 , most of the diff lines are a result of vendoring oasis-core 23.x, so it's best to review commit by commit. Here are the notable commits:
- `vendor-oasis-core: Support v23.0` - tweaks/patches that make the vendored v23 types compile. Most of these patches were already used for v22 and v21.
- `Add vendored coreapi v23.0` - Just a direct result of running `vendor-oasis-core.sh v23.0`. Large diff but not worth reviewing.
- `bump default oasis-core to 23.0` - We hardly rely on oasis-core any more, but we do use some of its `Address` types (among other things). The vendored v23 oasis-core uses uses some new `Address` methods, so we just bring in the more-featureful (but backwards-compatible) v23 Address for everyone.
- `nodeapi: Add enigma (v23.0) client` - The meat of the PR. Most of it is copied from Cobalt (!), then tweaked for E.
- `nodeapi: Add RuntimeSuspendedEvent; Rename RuntimeRegisteredEvent -> RuntimeStartedEvent` and `nodeapi/enigma: Parse runtime (un)suspension events` - Before E, the only runtime-related event was `Runtime`, which meant "runtime registered". There were no events for a runtime getting suspened or (re)starting. E introduces those events, which we now track.
- `update statecheck to use oasis-core v23` - statecheck depends directly on "live" oasis-core, without the the `nodeapi` abstraction/generalization. It only needs to run against the newest node (so there's no _need_ for `nodeapi`), and bypassing `nodeapi` should also help catch any bugs in it (because statecheck will obtain the "truth" without using that abstraction).

Note: Some parts of our code refer to "generations"/versions/archives by name: `damask`, `cobalt`, etc. For those uses, I nicknamed v23 `enigma`, with the full expectation that this will change once v23 is deployed on mainnet. At that point I'll rename it in the nexus codebase too.

**Testing**:
Indexed ~60k blocks (from 17_751_681 (E genesis) to 17_811_730) with no issues.


storage/oasis/nodeapi/api.go
```diff
@@ -6,12 +6,13 @@ import (
...
 
 	coreCommon "github.com/oasisprotocol/oasis-core/go/common"
+
 	hash "github.com/oasisprotocol/oasis-core/go/common/crypto/hash"
 	"github.com/oasisprotocol/oasis-core/go/common/crypto/signature"
 	"github.com/oasisprotocol/oasis-core/go/common/errors"
-	node "github.com/oasisprotocol/oasis-core/go/common/node"
 
 	beacon "github.com/oasisprotocol/nexus/coreapi/v22.2.11/beacon/api"
+	node "github.com/oasisprotocol/nexus/coreapi/v22.2.11/common/node"
```
pro-wh, line 15: was this from #548?

go.mod
```diff
@@ -2,132 +2,148 @@ module github.com/oasisprotocol/nexus
...
 	github.com/ethereum/go-ethereum v1.12.1
 	github.com/go-chi/chi/v5 v5.0.7
 	github.com/go-kit/log v0.2.1
 	github.com/golang-migrate/migrate/v4 v4.15.2
 	github.com/google/uuid v1.3.0
 	github.com/jackc/pgx/v5 v5.3.1
 	github.com/knadh/koanf v1.4.1
-	github.com/oasisprotocol/oasis-core/go v0.2202.11
-	github.com/oasisprotocol/oasis-sdk/client-sdk/go v0.6.0
+	github.com/oasisprotocol/oasis-core/go v0.2300.0
```
pro-wh, line 26: tag exists now, could you update "bump default oasis-core to 23.0" commit message

coreapi/v23.0/beacon/api/vrf.go
```diff
@@ -1,8 +1,9 @@
 package api
 
 import (
-	"github.com/oasisprotocol/nexus/coreapi/v23.0/consensus/api/transaction"
 	"github.com/oasisprotocol/oasis-core/go/common/crypto/signature"
+
+	"github.com/oasisprotocol/nexus/coreapi/v23.0/consensus/api/transaction"
```
pro-wh, line 6: aw no more separate goimports commit

storage/oasis/nodeapi/enigma/convert.go
```diff
@@ -0,0 +1,480 @@
...
+				}
+			case e.Staking.Escrow.DebondingStart != nil: // Note: Event started appearing mid-Enigma.
+				ret = nodeapi.Event{
+					StakingDebondingStart: &nodeapi.DebondingStartEscrowEvent{
+						Owner:           e.Staking.Escrow.DebondingStart.Owner,
+						Escrow:          e.Staking.Escrow.DebondingStart.Escrow,
+						Amount:          e.Staking.Escrow.DebondingStart.Amount,
+						ActiveShares:    e.Staking.Escrow.DebondingStart.ActiveShares,
+						DebondingShares: e.Staking.Escrow.DebondingStart.DebondingShares,
+						DebondEndTime:   0, // Not provided in Enigma; added in core v22.0, i.e. the testnet-only Damask precursor.
```
pro-wh, line 306: woa is it really not?

coreapi/v21.1.1/registry/api/sanity_check.go
```diff
@@ -1,58 +0,0 @@
...
-// and SanityCheckRuntimes for cross referencing purposes.
-// removed func
-
-// SanityCheckStake ensures entities' stake accumulator claims are consistent
-// with general state and entities have enough stake for themselves and all
-// their registered nodes and runtimes.
-// removed func
-
-// Runtimes lookup used in sanity checks.
-type sanityCheckRuntimeLookup struct {
```
pro-wh: should we maybe exclude these lowercase things by default? are there any that we keep?

storage/oasis/nodeapi/enigma/convert.go
```diff
@@ -0,0 +1,480 @@
...
+	nodeEnigma "github.com/oasisprotocol/nexus/coreapi/v23.0/common/node"
+	txResultsEnigma "github.com/oasisprotocol/nexus/coreapi/v23.0/consensus/api/transaction/results"
+	genesisEnigma "github.com/oasisprotocol/nexus/coreapi/v23.0/genesis/api"
+	governanceEnigma "github.com/oasisprotocol/nexus/coreapi/v23.0/governance/api"
+	registryEnigma "github.com/oasisprotocol/nexus/coreapi/v23.0/registry/api"
+	schedulerEnigma "github.com/oasisprotocol/nexus/coreapi/v23.0/scheduler/api"
+	stakingEnigma "github.com/oasisprotocol/nexus/coreapi/v23.0/staking/api"
+)
+
+func convertProposal(p *governanceEnigma.Proposal) *governance.Proposal {
```
pro-wh, line 36: ok so how come we're converting from e* to damask instead of always forward?

storage/oasis/nodeapi/enigma/convert.go
```diff
@@ -0,0 +1,480 @@
+package enigma
```
pro-wh, line 1: I'm just gonna throw this in here: `estar` would be a less creative option before the name somehow gets chosen

Andrew7234 (approved):

storage/oasis/nodeapi/enigma/convert.go
mitjat: Hehe
:shrug: 
`estoy` indiferente
Not a bad suggestion though.

storage/oasis/nodeapi/enigma/convert.go
mitjat: The PR description of #548 alludes to it:
> we (ab)use the v22.x types for two purposes:
> 1) For serde of gRPC requests/responses (as we also do for Cobalt, and later E)
> 2) As nexus's internal data representation (this needs a few more enums to interpret/manipulate; in an ideal world, we'd create our own types here)

So philosophically in this file, we're converting to `nodeapi` types, i.e. types that are internal to nexus. They are currently almost entirely a subset of damask types, but they can and will gain fields to be able to represent relevant data from all versions of oasis-core past and future.

coreapi/v21.1.1/registry/api/sanity_check.go
mitjat: IIRC we do use some, transitively. I'm not motivated enough to try harder to prune the vendored set; what we have is "close enough" IMO, though it could certainly be a more precise subset. In an ideal world, we'd run some sort of tree shaking on the source code.

pro-wh: k

storage/oasis/nodeapi/enigma/convert.go
pro-wh: ah

storage/oasis/nodeapi/api.go
mitjat: Yep. I put it in the wrong commit apparently; apologies. #548 is merged now, so leaving this one as-is.

go.mod
mitjat: updated

coreapi/v23.0/beacon/api/vrf.go
mitjat: I tried hard to not have it! Buncha rebasing. Still ended up with some weird conflicts and I think one small post-fix commit :/

storage/oasis/nodeapi/enigma/convert.go
mitjat: The comment is False. Thank you, fixed. It was a remnant of the Cobalt code (from which this file is derived), combined with a little overly-automatic find-replace (Cobalt -> Enigma). I went through all the comments in this file and found another field that we can populate in Enigma but couldn't populate in Damask.

go.mod
```diff
@@ -2,132 +2,148 @@ module github.com/oasisprotocol/nexus
...
 go 1.19
 
-replace github.com/tendermint/tendermint => github.com/oasisprotocol/tendermint v0.34.21-oasis1
+replace ( // All from oasis-core
+	github.com/cometbft/cometbft => github.com/oasisprotocol/cometbft v0.37.2-oasis1
+
+	// v1.5.0 has broken uint parsing, use my commit with fixes instead until
+	// the maintainers merge my PR: https://github.com/spf13/cast/pull/144
+	github.com/spf13/cast => github.com/oasisprotocol/cast v0.0.0-20220606122631-eba453e69641
+	github.com/tendermint/tendermint => github.com/oasisprotocol/tendermint v0.34.21-oasis1
```
ptrus, line 11: This one is not needed anymore (tendermint is now cometbft).

go.mod
ptrus: There is v0.2300.2 now already :-)

go.mod
```diff
@@ -2,132 +2,148 @@ module github.com/oasisprotocol/nexus
 
 go 1.19
```
ptrus, line 3: Hm i thought oasis-core/go v 23 requires go 1.21, apparently not? Probably good idea to bump in a followup PR.

storage/oasis/nodeapi/enigma/node.go
```diff
@@ -0,0 +1,217 @@
...
+	registryEnigma "github.com/oasisprotocol/nexus/coreapi/v23.0/registry/api"
+	roothashEnigma "github.com/oasisprotocol/nexus/coreapi/v23.0/roothash/api"
+	schedulerEnigma "github.com/oasisprotocol/nexus/coreapi/v23.0/scheduler/api"
+	stakingEnigma "github.com/oasisprotocol/nexus/coreapi/v23.0/staking/api"
+)
+
+// EnigmaConsensusApiLite provides low-level access to the consensus API of a
+// Enigma node. To be able to use the old gRPC API, this struct uses gRPC
+// directly, skipping the convenience wrappers provided by oasis-core.
+type EnigmaConsensusApiLite struct {
```
ptrus, line 39: Wouldn't it make more sense for this to just be `ConsensusApiLight`. Since enigma is already the package name, it's likely always used as `enigma.ConsensusApiLite`?

tests/statecheck/consensus_test.go
```diff
@@ -317,15 +316,12 @@ func validateNodes(t *testing.T, genesis *registryAPI.Genesis, source consensusA
 		}
 
 		vrfPubkey := ""
-		if n.VRF != nil {
-			vrfPubkey = n.VRF.ID.String()
-		}
+		vrfPubkey = n.VRF.ID.String()
```
ptrus, line 319: Can now remove the above line and do
```suggestion
		vrfPubkey := n.VRF.ID.String()
```

docker/oasis-net-runner/Dockerfile
```diff
@@ -1,6 +1,6 @@
 FROM golang:1.19-buster AS oasis-core
```
ptrus, line 1: I think buster doesn't have recent enough glibc to work with 23.x oasis-core (GLIBC>=2.32)

ptrus: I guess that also fixture and genesis.json used in tests will require updates (no sure how these were generated, by hand?)

go.mod
mitjat: :shrug:  apparently.

Upgrade: Not opposed to it but I don't see a benefit to the busywork unless it's needed?

storage/oasis/nodeapi/enigma/node.go
mitjat: I think the linter with its default agrees with you :|
I'm not a fan of the rule, but open to renaming this at some point in the way you're suggesting. Right now though, we've already got `CobaltConsensusApiLite` and `DamaskConsensusApiLite` so IMO the higher priority is to keep local consistency.
I'll try to remember when I'm renaming Enigma to whatever E name we come up with.

tests/statecheck/consensus_test.go
mitjat: :+1:  We can even inline `vrfPubkey` now. Fixed.

ptrus (approved):

docker/oasis-net-runner/Dockerfile
mitjat: buster: I was almost done debugging this when I saw your comment, and sure enough, the final run confirmed it :). Thank you.
As predicted, I then also had to update the fixture in a tiny way.

mitjat:
I implemented a few more changes:
- Fixes to the synthetic e2e test (all related to test config/setup), and some extra logging/conveniences that I needed when debugging the failing e2e test
- Resolved two outstanding TODOs around handling new-to-v23.0 fields:
    - `Validator.EntityID` we now ignore, but knowingly; we don't need it
    - `TakeEscrowEvent.DebondingAmount` is now used for dead-reckoning

e2e passes locally, branch should be good to merge now :crossed_fingers:

PR #552, opened 2023-10-31
NFT original metadata

pro-wh:
this leads to some data duplication. but eh, in the hypothetical situation where there's more than one token standard, `name`, `description`, and `image` will have different routines for how to populate them, and `metadata` will be whatever appropriate metadata that standard provides, e.g. it might be arranged differently, or name etc. will be from an on-chain call instead, or whatever.


analyzer/evmnfts/evm_nfts.go
```diff
@@ -118,6 +118,7 @@ func (p *processor) ProcessItem(ctx context.Context, batch *storage.QueryBatch,
 		staleNFT.DownloadRound,
 		nftData.MetadataURI,
 		nftData.MetadataAccessed,
+		nftData.Metadata,
```
pro-wh, line 121: apparently this makes it so that the string is the JSON serialized value. I have no idea how, hypothetically, to tell the db  that "the JSON value was a string, here's that string."

storage/client/client.go
```diff
@@ -1603,6 +1603,7 @@ func (c *StorageClient) RuntimeEVMNFTs(ctx context.Context, limit *uint64, offse
 			&ownerAddrData,
 			&nft.MetadataUri,
 			&metadataAccessedN,
+			&nft.Metadata,
```
pro-wh, line 1606: and then this does unmarshal the JSON from the db into Metadata interface{} as like maps and stuff :shrug: and our API layer later re-marshals it :repeat:

api/spec/v1.yaml
```diff
@@ -2688,6 +2688,10 @@ components:
           type: string
         metadata_accessed:
           type: string
+        metadata:
+          description: |
+            A metadata document for this NFT instance.
+            Currently only ERC-721 is supported, where the document is an Asset Metadata from the ERC721 Metadata JSON Schema.
```
pro-wh, line 2694: https://eips.ethereum.org/EIPS/eip-721 their schema is here. if higher effort, we could import that and say it could be that schema

storage/migrations/23_evm_nfts.up.sql
```diff
@@ -0,0 +1,6 @@
+BEGIN;
+
+ALTER TABLE chain.evm_nfts
+    ADD COLUMN metadata JSONB;
```
pro-wh, line 4: using a json type to be nice to people who want to do things in the db. but beware there are no indices. downside is that both go and postgresql will parse the JSON code. if they ever have any difference in opinion on what's valid JSON, then someone could break our indexer.

pro-wh:
omg regression tests are gonna need to be updated huh

pro-wh:
> {"args":["emerald","oasis1qrtn6tdgevwyzl0cjulwe8zs285zv60kpq8vks2v","862",1004297,"https://airose.mypinata.cloud/ipfs/QmSEfuX5f33Pxet1HHq536aDwAE6e (truncated 21 bytes)","2023-10-31T00:18:03.01781408Z","","","",""],"caller":"client.go:60","db":"indexer","err":"ERROR: invalid input syntax for type json (SQLSTATE 22P02)","level":"error","module":"postgres","msg":"Query","pid":87,"sql":"\n    UPDATE chain.evm_nfts SET\n      last_download_round = $4,\n      metadata_uri = $5,\n      metadata_accessed = $6,\n      metadata = $7,\n      name = $8,\n      description = $9,\n      image = $10\n    WHERE\n      runtime = $1 AND\n      token_address = $2 AND\n      nft_id = $3","time":"280.201µs","ts":"2023-10-31T00:18:03.658721433Z"}

oh shoot what

pro-wh:
oh it's trying to give uninitialized empty string when there's no metadata available, which isn't a valid json value

pro-wh:
ok I'm gonna make the EVMNFTData struct fields nil-able. sql NULL in the metadata column will mean we didn't get the document for whatever reason. json null will theoretically mean the metadata document was null, although that doesn't fit the schema and would probably fail to demarshal first. empty strings in the NFT table will be from legacy code, although we haven't enabled the analyzer in our deployments yet, so we shouldn't see any (other than from contracts that actually give empty string)

tests/e2e_regression/expected/emerald_account_nfts.body
```diff
@@ -4,6 +4,21 @@
       "description": "Swarm Teddies are the first Teddy Bear-like collection in Oasis Network, coming from space, these awesome creatures are unique, exclusive and hand designed with dedication and love.Born from an original concept, Swarm Teddies consists of 1000 cute bear NFTs that quickly adapted to their favorite themes, these creatures are not here to take part, they are here to take over.",
       "id": "177",
       "image": "ipfs://QmctsL3PKCbu3QGNmVfu9opWC3ujFecKR9u2DjuAhWSqkD/Casual-102.png",
+      "metadata": {
```
pro-wh, line 7: new

tests/e2e_regression/expected/emerald_account_nfts.body
```diff
@@ -221,12 +755,9 @@
       }
     },
     {
-      "description": "",
```
pro-wh, line 224: these fields are removed by following the logic posted in the PR:
- we can't get metadata from this URL (unresolvable hostname)
- so the name+description+image are left as nil in the EVMNFTData struct during analysis
- which go in as NULL in the database
- which I suppose we are leaving them as field-absent in our API output :shrug:

ptrus (approved):

PR #553, opened 2023-11-02
common: Log gRPC library messages

mitjat:
Can be informative for low-level debugging. The library is not too chatty; `info` level mostly contains info about connecting to nodes, nothing about individual requests.

### DIFF
--- a/cmd/common/common.go
+++ b/cmd/common/common.go
@@ -6,6 +6,8 @@ import (
 	"io"
 	"os"
 
+	"google.golang.org/grpc/grpclog"
+
 	coreLogging "github.com/oasisprotocol/oasis-core/go/common/logging"
 
 	"github.com/oasisprotocol/nexus/config"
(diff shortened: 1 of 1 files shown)


pro-wh:
this is so custom, does cause us to mix different log formats? should we not have a logger impl that sends things through our existing logger and have the latter take care of what json format and what output stream and what verbosity?

mitjat:
> this is so custom, does cause us to mix different log formats? should we not have a logger impl that sends things through our existing logger and have the latter take care of what json format and what output stream and what verbosity?

You're right, that would be a lot cleaner. But then I'd also need to worry about the format that `grpc` uses to write to our new glue writers (is it JSON, or just a string msg?). Overall, not worth the hassle. I'll just drop this PR. The logging was something that I thought would help me solve a bug, and is vaguely potentially nice-to-have, but it hasn't been obviously needed so far.

PR #555, opened 2023-11-02
Connect to oasis-node lazily

mitjat:
Before this PR, nexus connected to all the oasis nodes listed in the config immediately on startup, and failed to start if any of the nodes were unavailable.
This PR makes nexus connect to the nodes lazily, on-demand.

This has multiple benefits:
- **More robust startup**; e.g. a downed Sapphire node does not stop us from indexing consensus. (Before, indexing consensus in this scenario required temporarily disabling Sapphire in the config.)
   - Of short-term interest: We should even be able to add a v23.0 node to the config before it's even deployed, and nexus should smoothly switch to indexing it once it reached the right height. Untested.
- **Less special-cased e2e regression test exdcution path** across config, Go code, and scripts: e2e regression tests do not have node connectivity in CI, so we had to mainain a special "CI mode"
  (`query_on_cache_miss` flag) that allowed nexus to start without attempting to connect to any nodes, but still pretend the nodes were there so that it was able to query their cache.

### DIFF
--- a/storage/oasis/connections/raw_grpc.go
+++ b/storage/oasis/connections/raw_grpc.go
@@ -1,7 +1,9 @@
 package connections
 
 import (
+	"context"
 	"crypto/tls"
+	"sync"
 
 	cmnGrpc "github.com/oasisprotocol/oasis-core/go/common/grpc"
 	sdkConfig "github.com/oasisprotocol/oasis-sdk/client-sdk/go/config"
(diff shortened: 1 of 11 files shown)


storage/oasis/connections/raw_grpc.go
```diff
@@ -33,3 +35,56 @@ func RawConnect(nodeConfig *config.NodeConfig) (*grpc.ClientConn, error) {
...
+
+type LazyGrpcConn struct {
+	inner *grpc.ClientConn
+	lock  sync.Mutex // For lazy initialization.
+
+	nodeConfig config.NodeConfig // The node to connect to.
+}
+
+func (c *LazyGrpcConn) ensureConn() error {
+	if c.inner != nil {
```
ptrus, line 55: I think this is technically not safe to do in go, but tbh I am also not aware how problematic can it be in practice. 

To avoid locking on every `Invoke` you could use an additional atomic signalling that the connection was established and use that for the fast-path (this is also how [`sync.Do`](https://github.com/golang/go/blob/642b1311111ef949f54606317ec447838c29024c/src/sync/once.go#L48-L67)  is implemented, which solves a similar problem).

mitjat: With the caveat that I'm no expert in concurrency, this still looks ok to me even with your additional input.

The problem that `sync.Once(f func(...))` (nice link!) is solving is that if two callers call `f` simultaneously, we want `Once()` to return to both only after `f` has finished. In our example, we analogously do not want to do an early return from `ensureConn` unless we're sure that our `f` (which is implicit, and equal to `f() { c.inner, err = RawConnect(&c.nodeConfig) }`) has completed. In other words, the current code can be dangerous if `c.inner` can be in a state where it is `!= nil`, but also not fully initialized yet.
I think such a state is impossible because we first construct the connection with `RawConnect()`, and only then do we assign it (in a single, atomic operation of copying a single pointer) to `c.inner`.
The code _could_ be risky if `c.inner` (or the value it points to) potentially took multiple operations to build out, e.g. if the connection was established with something like `err = RawConnect(ctx, c.inner)`.

ptrus: > I think such a state is impossible because we first construct the connection with RawConnect(), and only then do we assign it (in a single, atomic operation of copying a single pointer)

I'm not sure this part is true in the go memory model. The assigning of the pointer is indeed most likely atomic, but _i think_ that there is no guarantee that **from the perspective of other go-routines** the write to `c.inner` happens after RawConnect is completely done.
But like i stated before, not sure if this is actually problematic in practice, but the afaik the go documentation does mention this being an undefined behaviour. 


FWIW the below test fails with a "WARNING: DATA RACE" if run with `go test -race .`
```
type test struct {
	inner *string
	lock  sync.Mutex
}

func initString() *string {
	a := "foo"
	return &a
}

func (t *test) init() *string {
	if t.inner != nil {
		return nil
	}
	t.lock.Lock()
	defer t.lock.Unlock()

	if t.inner != nil {
		return nil
	}

	// Initialize `inner`.
	t.inner = initString()

	return t.inner
}

func TestTest(t *testing.T) {
	fmt.Println("im here")
	test := &test{}
	for i := 0; i < 100; i++ {
		go test.init()
	}
}
```

pro-wh: lock it, why wouldn't we

mitjat: > from the perspective of other go-routines

Oh. Nasty filthy compilerses and their optimizations, they tricked us!
Makes sense. Thank you for this patient explanation and the in-depth research.

Instead of going with the additional cheap lock-like atomic int (like in `sync.Once`), I just opted to keep the code simpler and more maintainable, and just use the mutex for every `inner != nil` check. With our usage pattern, I think that will still be plenty performant. If not ... we now know what to do next :+1:

ptrus (approved):

PR #556, opened 2023-11-02
pgx: Transfer entire DB batch in one request

mitjat:
pgx offers two ways of sending query batches to the DB: A faster one with poor error reporting, and a slower one with better errors.

Before this PR, we toggled between the two based on an env variable. This PR refactors the code (read the diff bottom to top) so that we always use the fast method, but if it fails, we retry with the slow method to obtain a better error msg. Since a failed tx is automatically reverted and since the retry is not async, there should be no harm in attempting the batch twice.

**Testing:** Deployed in staging mainnet. Sapphire block analyzer went from spending ~300ms per block on writing to ~40ms. :tada:

### DIFF
--- a/storage/postgres/client.go
+++ b/storage/postgres/client.go
@@ -122,7 +122,7 @@ func (c *Client) sendBatchWithOptionsFast(ctx context.Context, batch *storage.Qu
 	}
 	defer common.CloseOrLog(batchResults, c.logger)
 
-	// Exec indiviual queries in the batch.
+	// Read the results of indiviual queries in the batch.
 	for i := 0; i < pgxBatch.Len(); i++ {
 		if _, err := batchResults.Exec(); err != nil {
 			rollbackErr := ""
@@ -170,21 +170,26 @@ func (c *Client) sendBatchWithOptionsSlow(ctx context.Context, batch *storage.Qu
 	// Commit the transaction.
 	err = tx.Commit(ctx)
 	if err != nil {
-		c.logger.Error("failed to submit tx",
-			"error", err,
-			"batch", batch.Queries(),
-		)
 		return err
 	}
 	return nil
 }
 
 func (c *Client) SendBatchWithOptions(ctx context.Context, batch *storage.QueryBatch, opts pgx.TxOptions) error {
-	if err := c.sendBatchWithOptionsFast(ctx, batch, opts); err == nil {
+	var err error
+	if err = c.sendBatchWithOptionsFast(ctx, batch, opts); err == nil {
 		// The fast path succeeded. This should happen most of the time.
 		return nil
 	}
-	// There was an error. The tx was reverted, so we can resubmit. This time, use the slow method for better error msgs.
+
+	// There was an error. The tx was reverted(*), so we can resubmit.
+	//   (*) Theoretically it's possible that the tx was committed but the confirmation of commit
+	//       didn't reach us due to networking issues. We don't handle this case.
+	// This time, use the slow method for better error msgs.
+	c.logger.Warn("failed to submit tx using the fast path; falling back to slow path",
+		"error", err,
+		"batch", batch.Queries(),
+	)
 	return c.sendBatchWithOptionsSlow(ctx, batch, opts)
 }
 
--- a/storage/postgres/client.go
+++ b/storage/postgres/client.go
@@ -100,6 +100,15 @@ func (c *Client) SendBatch(ctx context.Context, batch *storage.QueryBatch) error
 // However, it reports errors poorly: If _any_ query is syntactically
 // malformed, called with the wrong number of args, or has a type conversion problem,
 // pgx will report the _first_ query as failing.
+//
+// For efficiency and simplicity, the method does not use explicit transactions
+// with BEGIN/COMMIT unless required by the tx options. Even so, the batch is processed
+// atomically, because:
+//  1) We use pgx in its default QueryExecMode.
+//  2) This in turn makes pgx use postgresql in pipeline mode.
+//  3) Postgresql pipeline mode implies transactions-like behavior:
+//     https://www.postgresql.org/docs/15/libpq-pipeline-mode.html#LIBPQ-PIPELINE-ERRORS
+
 func (c *Client) sendBatchWithOptionsFast(ctx context.Context, batch *storage.QueryBatch, opts pgx.TxOptions) error {
 	pgxBatch := batch.AsPgxBatch()
 	var batchResults pgx.BatchResults
@@ -108,7 +117,7 @@ func (c *Client) sendBatchWithOptionsFast(ctx context.Context, batch *storage.Qu
 	var err error
 
 	// Begin a transaction.
-	useExplicitTx := opts != emptyTxOptions
+	useExplicitTx := opts != emptyTxOptions // see function docstring for more info
 	if useExplicitTx {
 		// set up our own tx with the specified options
 		tx, err = c.pool.BeginTx(ctx, opts)
--- a/storage/postgres/client.go
+++ b/storage/postgres/client.go
@@ -5,7 +5,6 @@ package postgres
 import (
 	"context"
 	"fmt"
-	"os"
 
 	"github.com/jackc/pgx/v5"
 	"github.com/jackc/pgx/v5/pgxpool"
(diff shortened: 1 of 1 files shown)


ptrus:
Nice, i'm interested in hearing if this has any impact.

storage/postgres/client.go
```diff
@@ -96,89 +95,99 @@ func (c *Client) SendBatch(ctx context.Context, batch *storage.QueryBatch) error
...
+// pgx will report the _first_ query as failing.
+func (c *Client) sendBatchWithOptionsFast(ctx context.Context, batch *storage.QueryBatch, opts pgx.TxOptions) error {
+	pgxBatch := batch.AsPgxBatch()
+	var batchResults pgx.BatchResults
+	var emptyTxOptions pgx.TxOptions
+	var tx pgx.Tx
+	var err error
+
+	// Begin a transaction.
+	useExplicitTx := opts != emptyTxOptions
```
pro-wh, line 111: should always be a transaction, right? we don't want the batch to be partially applied

storage/postgres/client.go
```diff
@@ -96,89 +95,99 @@ func (c *Client) SendBatch(ctx context.Context, batch *storage.QueryBatch) error
...
-						rollbackErr = fmt.Sprintf("; also failed to rollback tx: %s", err2.Error())
-					}
+		batchResults = c.pool.SendBatch(ctx, &pgxBatch)
+	} else {
+		// use implicit tx provided by SendBatch; see https://github.com/jackc/pgx/issues/879
+		batchResults = c.pool.SendBatch(ctx, &pgxBatch)
+	}
+	defer common.CloseOrLog(batchResults, c.logger)
+
+	// Exec indiviual queries in the batch.
```
pro-wh, line 125: "Exec" here is a little awkward, in the batch mode we only tell pgx to read each result as the result of an exec. it's already been sent out to exec prior to this though

storage/postgres/client.go
```diff
@@ -96,89 +95,99 @@ func (c *Client) SendBatch(ctx context.Context, batch *storage.QueryBatch) error
...
+		if _, err := batchResults.Exec(); err != nil {
+			rollbackErr := ""
+			if useExplicitTx {
+				err2 := tx.Rollback(ctx)
+				if err2 != nil {
+					rollbackErr = fmt.Sprintf("; also failed to rollback tx: %s", err2.Error())
 				}
-				return fmt.Errorf("query %d %v: %w%s", i, batch.Queries()[i], err, rollbackErr)
 			}
+			return fmt.Errorf("query %d %v: %w%s", i, batch.Queries()[i], err, rollbackErr)
```
pro-wh, line 135: actually is this even worse than the diagnostics we get from one-query-at-a-time mode? it's all the same info

storage/postgres/client.go
```diff
@@ -96,89 +95,99 @@ func (c *Client) SendBatch(ctx context.Context, batch *storage.QueryBatch) error
...
 	}
 	return nil
 }
 
+func (c *Client) SendBatchWithOptions(ctx context.Context, batch *storage.QueryBatch, opts pgx.TxOptions) error {
+	if err := c.sendBatchWithOptionsFast(ctx, batch, opts); err == nil {
+		// The fast path succeeded. This should happen most of the time.
+		return nil
+	}
+	// There was an error. The tx was reverted, so we can resubmit. This time, use the slow method for better error msgs.
```
pro-wh, line 187: > was reverted

or there was an error in `tx.Commit()`, and we don't know :grimacing:

storage/postgres/client.go
```diff
@@ -96,89 +95,99 @@ func (c *Client) SendBatch(ctx context.Context, batch *storage.QueryBatch) error
...
+			"error", err,
+			"batch", batch.Queries(),
+		)
+		return err
 	}
 	return nil
 }
 
+func (c *Client) SendBatchWithOptions(ctx context.Context, batch *storage.QueryBatch, opts pgx.TxOptions) error {
+	if err := c.sendBatchWithOptionsFast(ctx, batch, opts); err == nil {
```
pro-wh, line 183: shouldn't throw away `err` here imo, we might not get the same error or any error from -Slow

pro-wh (commented):
this feels a little weird because something different might happen in the retry. do we actually get better debugging in the "slow" mode?

some comments unrelated to the change :skull:

mitjat:
> do we actually get better debugging in the "slow" mode?

Yes, see the docstring of `sendBatchWithOptionsFast`:
```
// However, it reports errors poorly: If _any_ query is syntactically
// malformed, called with the wrong number of args, or has a type conversion problem,
// pgx will report the _first_ query as failing.
```

storage/postgres/client.go
mitjat: It is always a tx, but we don't usually need to make it explicit with separate `BEGIN` and `COMMIT` calls. See the comment in line 120 for more context.

storage/postgres/client.go
mitjat: Good catch. Changed comment to "Read the results".

storage/postgres/client.go
mitjat: IIRC it always reports failure on query 0, even when the actual error appears lower down. It's been a while since I tried this though. Let me see again, and update the comment.

:exclamation: TODO

storage/postgres/client.go
mitjat: Hm ... that is an interesting corner case.

If there was a problem _sending_ the `tx.Commit()`, or the commit operation somehow errored out, we're fine because the tx indeed never committed.
If there was a problem _receiving the results_ the `tx.Commit()`, we're in trouble. However the only case where I can think of this happening is with connectivity problems. If the db was unreachable _just for the commit_ (and not when sending the batch), it seems extremely unlikely for it to be suddenly reachable again in the next split second when resending the tx.

However I'm inclined to risk it, because **no aspect of this problem is specific to the retry logic in this PR**. If pgx reports an error when sending the tx, we (and presumably most other clients) always assume the tx had no side effects. I don't see a way around it other than some sort of two-phase commits.

Also, specific to nexus: If a tx succeeds but we don't learn about it so we then erroneously re-send the same tx, in many cases the doubled tx will cause PK violations, so that's another failsafe.

storage/postgres/client.go
pro-wh: the comment:

> 		// use implicit tx provided by SendBatch; see https://github.com/jackc/pgx/issues/879

pro-wh: ok why do they link to the extended query flow? the section explains,

> In the extended protocol, the frontend first sends a Parse message ...
>
> Note: The query string contained in a Parse message cannot include more than one SQL statement; else a syntax error is reported. This restriction does not exist in the simple-query protocol, but it does exist in the extended protocol, because allowing prepared statements or portals to contain multiple commands would complicate the protocol unduly.

which seems to be rather that there's no grouping of multiple statements into an implicit transaction

by contrast, the simple query section explains:

> When a simple Query message contains more than one SQL statement (separated by semicolons), those statements are executed as a single transaction, unless explicit transaction control commands are included to force a different behavior.

consistent with the summary in the issue comment

which flow does pgx use?

storage/postgres/client.go
pro-wh: thx

storage/postgres/client.go
pro-wh: pk violations :pray: 

although we're also moving more to upsert style in various places

pro-wh:
ok I'm looking through the pgx batch implementation. errors in preparing+binding the queries are managed as a so-called "early" error. 
there is indeed some poor bookkeeping in pgx that loses information about the source of the error
- handling error in prepare (i.e. syntax error) does not record which query it came from https://github.com/jackc/pgx/blob/v5.5.3/conn.go#L1125
- actually that's it
- client-side problems in binding include mismatched arg count and error encoding the args. these get reported with the query formatted into the error message, so we at least stand a chance

that early gets returned on the first `.Exec()` read. we can also get it through a batch tracer `TraceBatchEnd`, but that doesn't seem like it gives much more utility

storage/postgres/client.go
mitjat: pgx uses the extended mode. See [docs](https://github.com/jackc/pgx/blob/7ec6ee7b0ad9ecb4cf885b6630fd36b022e89755/conn.go#L36). I also used wireshark to snoop on the connection, and all PREPARE statements (triggered by both fast and slow tx-sending mode from this PR) contain only a single statement each, which implies extended mode.

In extended mode, pgx always also enables "pipeline mode" ([here](https://github.com/jackc/pgx/blob/7ec6ee7b0ad9ecb4cf885b6630fd36b022e89755/conn.go#L1062)), which "allows applications to send a query without having to read the result of the previously sent query", [per postgres docs](https://www.postgresql.org/docs/current/libpq-pipeline-mode.html). Those same docs also say that "If the pipeline used an implicit transaction(✱), then operations that have already executed are rolled back and operations that were queued to follow the failed operation are skipped entirely," which matches the description by jackc from https://github.com/jackc/pgx/issues/879

✱ Those are statements not inside a BEGIN/COMMIT block. I cannot find an explicit (heh) definition of that exact term, but a [google search](https://www.google.com/search?q=postgres+implicit+transaction) makes the meaning pretty clear, also via official postgres docs.

storage/postgres/client.go
mitjat: It wasn't thrown away, it was logged inside the Fast() method. But that was not a logical place for it :+1:.
Moved the logging here now.

storage/postgres/client.go
pro-wh: thanks, good find on the pipeline docs saying that

storage/postgres/client.go
pro-wh: ok so from other threads, there's no application layer database transaction, so we can resubmit

storage/postgres/client.go
pro-wh: philosophically resolved. could you add a code comment about:
- we use pgx cache execution mode
- pgx cache execution mode uses postgresql pipeline
- postgresql pipeline makes an implicit transaction https://www.postgresql.org/docs/current/libpq-pipeline-mode.html#LIBPQ-PIPELINE-SENDING

storage/postgres/client.go
mitjat: I think [your new research](https://github.com/oasisprotocol/nexus/pull/556#issuecomment-1928184551) agrees with what I wrote. PREPARE also fails if the privileges are off (e.g. on a table), or there are issues with types. These are only expected to happen during development or experimental deploys, but it's very helpful to have a good error message in those rare cases.

Also, I just found out that when we send the results of genesis analysis to the db (= a huge tx), the fast path fails with `unexpected pipeline result: *pgconn.PipelineSync` but the slow path succeeds. I think that's a bug in pgx; postgres docs do mention that "writing clients using the pipeline mode is more complex". But I don't even want to follow up on it for such a corner case for which we even have a workaround already, inadvertently.

storage/postgres/client.go
mitjat: :+1: added

pro-wh (approved):
ya

PR #557, opened 2023-11-02
Bump Go (1.21.3), golangci-lint (1.55.1)

mitjat:
This PR:
- Bumps the Go version used to build Nexus
- Bumps the version of the linter, and updates code and linting rules to make the new linter happy
- Updates CI so it can detect `go mod tidy` issues. Those were previously only checked-for in the release process, thus allowing us to check in code that would later fail to release.

The backstory is a little convoluted:
1) Release v0.1.18 [failed to produce a docker image](https://github.com/oasisprotocol/nexus/actions/runs/6739050114/job/18319820031) because goreleaser first runs `go mod tidy` [as a hook](https://github.com/oasisprotocol/nexus/blob/e4f3b280395459566a74dc1316b63237d81a5410/.goreleaser.yaml#L5-L5), which failed with
```
github.com/oasisprotocol/nexus/analyzer/uncategorized imports
        github.com/ethereum/go-ethereum/crypto imports
        github.com/btcsuite/btcd/btcec/v2/ecdsa tested by
        github.com/btcsuite/btcd/btcec/v2/ecdsa.test imports                                                       
        github.com/btcsuite/btcd/chaincfg/chainhash: ambiguous import: found package github.com/btcsuite/btcd/chaincfg/chainhash in multiple modules:
        github.com/btcsuite/btcd v0.20.1-beta (/home/mitjat/go/pkg/mod/github.com/btcsuite/btcd@v0.20.1-beta/chaincfg/chainhash)
        github.com/btcsuite/btcd/chaincfg/chainhash v1.0.1 (/home/mitjat/go/pkg/mod/github.com/btcsuite/btcd/chaincfg/chainhash@v1.0.1)
```

2) A sort-of fix was to add an explicit dependency in an attempt to enforce a specific version: `go get github.com/btcsuite/btcd/chaincfg/chainhash@v1.0.1`. After this, nexus builds and `go mod tidy` is happy, _but_ the latter also removes the chainhash entry from go.mod, and a second invocation of `go mod tidy` fails. This smells buggy.

3) It turns out that bumping the Go version resolves the buggy `go mod tidy` behavior. Go v1.21 is also [used by oasis-core](https://github.com/oasisprotocol/oasis-core/blob/v23.0/go/go.mod#L196) and @ptrus recently [suggested](https://github.com/oasisprotocol/nexus/pull/549#discussion_r1374237671) we use it in nexus too, so it's a good idea to upgrade anyway. However, this also requires us to bump the linter version (the old linter does not support Go v1.21), which in turn requires us to resolve some new linter warnings. This PR does that.

**Developers:** After this goes in, you might need to update your local env accordingly:
```bash
# Download go v1.21
go install golang.org/dl/go1.21.3 && go1.21.3 download

# Make it the default
sudo ln -s -f $GOPATH/bin/go1.21.3 /usr/bin/go

# Make it the "Oasis default" (used by some Makefile targets)
(edit ~/.bashrc or similar, update OASIS_GO variable)

# Install new golangci-lint
curl -sSfL https://raw.githubusercontent.com/golangci/golangci-lint/master/install.sh | sh -s -- -b $(go env GOPATH)/bin v1.55.1
```


Andrew7234 (approved):

Andrew7234 (commented):
oh ok I see the slack messages/CI now

ptrus (approved):

mitjat:
Additional data point: @peternose points out that oasis-core had the same issue in https://github.com/oasisprotocol/oasis-core/pull/5268, and also resolved it by switching to a newer Go.

PR #558, opened 2023-11-04
analyzer: Slow-sync analyzers wait only for _relevant_ fast-sync analyzers to complete

mitjat:
This PR changes the sequencing of analyzers to what the title describes.

Previously, _all_ slow-sync analyzers were started only once _all_ fast-sync analyzers completed. But it doesn't make sense for e.eg. Sapphire to wait for fast-sync consensus to complete. These arbitrary delays were expecially noticeable when doing a full reindex (where fast-sync can take a week or more), or if one of the runtimes/consensus encountered an error during fast-sync and got stuck.

Resolves https://app.clickup.com/t/8692uz0a5

### DIFF
--- a/cmd/analyzer/analyzer.go
+++ b/cmd/analyzer/analyzer.go
@@ -586,14 +586,3 @@ func Register(parentCmd *cobra.Command) {
 	analyzeCmd.Flags().StringVar(&configFile, "config", "./config/local.yml", "path to the config.yml file")
 	parentCmd.AddCommand(analyzeCmd)
 }
-
-// For testing purposes only.
-type ServiceTester struct {
-	Service
-}
-
-func (a *ServiceTester) SetAnalyzers(fastSyncAnalyzers []SyncedAnalyzer, analyzers []SyncedAnalyzer) {
-	a.fastSyncAnalyzers = fastSyncAnalyzers
-	a.analyzers = analyzers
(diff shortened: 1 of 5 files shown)


cmd/analyzer/analyzer.go
```diff
@@ -550,3 +586,14 @@ func Register(parentCmd *cobra.Command) {
 	analyzeCmd.Flags().StringVar(&configFile, "config", "./config/local.yml", "path to the config.yml file")
 	parentCmd.AddCommand(analyzeCmd)
 }
+
+// For testing purposes only.
+type ServiceTester struct {
```
ptrus, line 591: It's also possible to move the `TestSequencing` test into the `package analyzer`  (e.g. in sequencing_test.go file). Then the test will have access to package private fields directly and this workaround is not needed.

ptrus (approved):

cmd/analyzer/analyzer.go
mitjat: Ah nice, thanks. Converted.

I had looked up a random existing test and figured that the separate `_test` package is a strong Go convention for tests. I see now that both types of testing (from `mypackage_test` and from `mypackage`) are endorsed. And sure, `mypackage_test` is a little cleaner, but ... not much, with my `ServiceTester` hack. So, not worth the hassle in this case, IMO.

PR #559, opened 2023-11-07
analyzer/block: more efficient db query on startup

mitjat:
This PR speeds up a query that runs on every startup. Tested on current staging testnet (~20M rows in table): The old query takes 77s, the new query takes ~32s. Still slow, but at least better.

The PR also adds a unit test for the query, and emits a log line before running the expensive query so that logs make more sense at nexus startup time when the program apparently freezes for a bit.

Resolves https://app.clickup.com/t/8692yjjtq

### DIFF
--- a/analyzer/queries/queries.go
+++ b/analyzer/queries/queries.go
@@ -150,11 +150,17 @@ var (
       SELECT COALESCE(max(height), -1) as height
       FROM analysis.processed_blocks
       WHERE analyzer = $1
+    ),
+    gaps AS (
+      SELECT h
+      FROM highest_encountered_block, generate_series(GREATEST(1, $2::bigint), LEAST(highest_encountered_block.height, $3::bigint)) AS h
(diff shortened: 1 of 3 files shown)


analyzer/queries/queries.go
```diff
@@ -150,11 +150,17 @@ var (
       SELECT COALESCE(max(height), -1) as height
       FROM analysis.processed_blocks
       WHERE analyzer = $1
+    ),
+    gaps AS (
+      SELECT h
+      FROM highest_encountered_block, generate_series(GREATEST(1, $2::bigint), LEAST(highest_encountered_block.height, $3::bigint)) AS h
+      EXCEPT
```
pro-wh, line 157: :information_source: EXCEPT them, whereas previously it fully relied on ON CONFLICT DO NOTHING

pro-wh (approved):
sql wiz

PR #561, opened 2023-11-07
runtime/evm: prevent saving invaild NFT metadata JSON

pro-wh:
previous code was designed to save whatever the server said, mismatched with the db schema which expected JSON. this new routine round-trips the result through Go json decode+encode (+decode :roll_eyes:) so that we only keep valid JSON.

### DIFF
--- a/analyzer/runtime/evm/erc721.go
+++ b/analyzer/runtime/evm/erc721.go
@@ -7,7 +7,6 @@ import (
 	"fmt"
 	"io"
 	"math/big"
-	"strings"
 	"time"
 
 	ethCommon "github.com/ethereum/go-ethereum/common"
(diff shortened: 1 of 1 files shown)


analyzer/runtime/evm/erc721.go
```diff
@@ -108,20 +107,46 @@ func evmDownloadNFTERC721Metadata(ctx context.Context, logger *log.Logger, sourc
...
+		}
+		return nil
 	}
 	if err = rc.Close(); err != nil {
 		return fmt.Errorf("closing metadata reader: %w", err)
 	}
-	nftData.Metadata = common.Ptr(metadataBuilder.String())
+	// 2. Re-serialize into JSON for database. This is to normalize the
+	//    syntax, in case Go and our database have slightly different opinions
+	//    on what's valid JSON.
```
mitjat, line 129: For my understanding: 

This is more hypothetical and was not the source of the errors we were observing, right? Go's json doesn't allow unquoted keys. ([example](https://go.dev/play/p/fl8D2dGIp1a))
So probably the most important change is the new `return nil` in line 122 that demands that metadata be (Go-style) well-formed.

analyzer/runtime/evm/erc721.go
```diff
@@ -108,20 +107,46 @@ func evmDownloadNFTERC721Metadata(ctx context.Context, logger *log.Logger, sourc
...
-		logger.Info("error decoding token metadata",
+	if err = json.NewDecoder(limitedReader).Decode(&metadataAny); err != nil {
+		logger.Info("error decoding token metadata as any",
 			"uri", nftData.MetadataURI,
 			"err", err,
 		)
+		if err = rc.Close(); err != nil {
+			return fmt.Errorf("closing metadata reader: %w", err)
+		}
+		return nil
```
mitjat, line 122: Will this result in one of those deterministic errors, meaning we won't retry? Or will we keep parsing the same JSON and failing every 2 minutes or whatever the analyzer's interval is?

mitjat (approved):
Thank you!

analyzer/runtime/evm/erc721.go
pro-wh: right, hypothetical

analyzer/runtime/evm/erc721.go
pro-wh: this is outside of the "collect deterministic errors" zone, it's just logged for our information and entirely treated as success. "we successfully got the metadata and extracted what we could from it. which was nothing because it was unparseable, but still"

PR #568, opened 2023-11-16
evmnfts, evmtokens: utf-8-clean strings from chain

pro-wh:
values that come from chain: use our SanitizeString, as they may contain postgresql-displeasing nul characters

edit: as well as values that come from servers :roll_eyes:


analyzer/evmnfts/evm_nfts.go
```diff
@@ -116,12 +116,13 @@ func (p *processor) ProcessItem(ctx context.Context, batch *storage.QueryBatch,
...
 		staleNFT.ID,
 		staleNFT.DownloadRound,
-		nftData.MetadataURI,
+		storage.SanitizeStringP(nftData.MetadataURI),
 		nftData.MetadataAccessed,
-		nftData.Metadata,
-		nftData.Name,
-		nftData.Description,
-		nftData.Image,
+		// If SanitizeString can affect JSON validity, we're doomed.
```
mitjat, line 121: :(
It does seem low risk.
We could have another wrapper, `ensureValidJSON()`, that returns the input as-is if it's valid JSON, otherwise logs a warning and returns `{}` or `null`. Your call on whether that's warranted. I think you already made the call with this comment :)

mitjat (approved):
Thank you!

Worth mentioning why we're implementing this PR:
We found invalid metadata JSON in the wild (in particular, it was invalid UTF-8 in that it contained `\x00`, but the Go JSON parser didn't balk at that, but Postgres does).
Hope I got that right :)

pro-wh:
e2e regression flaked, http 500 from ipfs gateway

pro-wh:
oh clarification about @mitjat 's explanation: \x00 NUL bytes are valid UTF-8, but they're not allowed in PostgreSQL due to a technical limitation.

incidentally, they're also not allowed in JSON, as strings aren't allowed to contain control characters (\uxxxx escapes of them are allowed), and the grammar doesn't allow them anywhere else either.

PR #570, opened 2023-11-22
storage: handle NULL NFT owner

pro-wh:
we only set owner on transfers, so when we see e.g. Approval, we can add an NFT to the db without setting an owner (hm, but anyway). prevent the storage glue from breaking (attempts to scan NULL into owner address context identifier string) when viewing these entries

### DIFF
--- a/storage/client/client.go
+++ b/storage/client/client.go
@@ -1579,8 +1579,9 @@ func (c *StorageClient) RuntimeEVMNFTs(ctx context.Context, limit *uint64, offse
 		var contractAddrContextVersion int
 		var contractAddrData []byte
 		var tokenType sql.NullInt32
-		var ownerAddrContextIdentifier string
-		var ownerAddrContextVersion int
+		// Owner might not be known, so these preimage fields are also nilable.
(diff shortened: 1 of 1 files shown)


storage/client/client.go
```diff
@@ -1579,8 +1579,8 @@ func (c *StorageClient) RuntimeEVMNFTs(ctx context.Context, limit *uint64, offse
 		var contractAddrContextVersion int
 		var contractAddrData []byte
 		var tokenType sql.NullInt32
-		var ownerAddrContextIdentifier string
-		var ownerAddrContextVersion int
+		var ownerAddrContextIdentifier *string
```
mitjat, line 1582: Let's add a short inline comment here explaining why this is nullable.

mitjat (approved):
:+1:

Andrew7234 (approved):

storage/client/client.go
pro-wh: ya

oo come to think of it we should use `if nft.Owner != nil` instead

PR #571, opened 2023-11-22
common.BigInt.String() and some unprovoked changes

### LINKED ISSUE #566: reexamine how common.BigInt goes out
how does common.BigInt leave our system? let's make sure that's convenient. let me write down some cases

- as a parameter in a db query
- as a field in an API result

and some minor internal ways

- formatted into a string, e.g. an error message via. I believe we usually use `%v`
- as a value in a log

---

> **TL;DR and suggested action: `big.Int` API sucks again. Revert this line and implement non-pointer-receiver String for our BigInt:**
```go
// big.Int already defines `String()`, but only for pointer receivers
func (b BigInt) String() string {
	return (&b.Int).String()
}
```
(Note: As the playground link shows, calling `b.Int.String()` works too, but I don't know why, so I prefer to be explicit with the `&`.

_Originally posted by @mitjat in https://github.com/oasisprotocol/nexus/pull/563#discussion_r1393166206_

pro-wh:
fixes #566

### DIFF
--- a/common/types.go
+++ b/common/types.go
@@ -5,7 +5,6 @@ import (
 	"errors"
 	"fmt"
 	"math/big"
-	"strings"
 
 	"github.com/jackc/pgx/v5/pgtype"
 	"github.com/oasisprotocol/oasis-core/go/common/quantity"
@@ -21,21 +20,21 @@ func NewBigInt(v int64) BigInt {
 	return BigInt{*big.NewInt(v)}
 }
 
-func (b BigInt) MarshalText() ([]byte, error) {
-	return []byte(b.String()), nil
-}
-
-func (b *BigInt) UnmarshalText(text []byte) error {
-	return b.Int.UnmarshalText(text)
-}
-
 func (b BigInt) MarshalJSON() ([]byte, error) {
-	return []byte(fmt.Sprintf(`"%s"`, b.String())), nil
+	t, err := b.MarshalText()
+	if err != nil {
+		return nil, err
+	}
+	return json.Marshal(string(t))
 }
 
 func (b *BigInt) UnmarshalJSON(text []byte) error {
-	v := strings.Trim(string(text), "\"")
-	return b.Int.UnmarshalJSON([]byte(v))
+	var s string
+	err := json.Unmarshal(text, &s)
+	if err != nil {
+		return err
+	}
+	return b.UnmarshalText([]byte(s))
 }
 
 func (b BigInt) String() string {
--- a/common/types.go
+++ b/common/types.go
@@ -38,6 +38,20 @@ func (b *BigInt) UnmarshalJSON(text []byte) error {
 	return b.Int.UnmarshalJSON([]byte(v))
 }
 
+func (b BigInt) String() string {
+	// *big.Int does have a String() method. But the way the Go language
+	// works, that method on a pointer receiver doesn't get included in
+	// non-pointer BigInt's method set. In some places this doesn't matter,
+	// because *big.Int's methods are included in pointer *BigInt's method
+	// set, and a completely different part of the language set says that
+	// writing b.String() is fine; it's shorthand for (&b).String(). But
+	// reflection-driven code like fmt.Printf only looks at method sets and
+	// not shorthand trickery, so we need this method to make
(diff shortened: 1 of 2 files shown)


common/types.go
```diff
@@ -22,20 +21,32 @@ func NewBigInt(v int64) BigInt {
...
 	return b.Int.UnmarshalText(text)
 }
 
 func (b BigInt) MarshalJSON() ([]byte, error) {
-	return []byte(fmt.Sprintf(`"%s"`, b.String())), nil
+	t, err := b.MarshalText()
+	if err != nil {
+		return nil, err
+	}
+	return json.Marshal(string(t))
```
pro-wh, line 36: philosophically, we said "we want these to be strings in JSON." the issue with the previous code is that it further defines how strings are encoded in JSON. this updated, 5-times-as-many-lines code rather does less, allowing the json module to define how strings are encoded in json

common/types_test.go
```diff
@@ -0,0 +1,31 @@
...
+	"testing"
+
+	"github.com/stretchr/testify/require"
+)
+
+func TestBigInt(t *testing.T) {
+	var v BigInt
+
+	textRef := []byte("11111111111111111111")
+	err := v.UnmarshalText(textRef)
```
pro-wh, line 15: is there a slightly more end-to-end way to exercise MarshalText/UnmarshalText? would better to do that so we know the pointerness of the receiver is right

common/types.go
```diff
@@ -22,20 +21,32 @@ func NewBigInt(v int64) BigInt {
...
+	var s string
+	err := json.Unmarshal(text, &s)
+	if err != nil {
+		return err
+	}
+	return b.UnmarshalText([]byte(s))
+}
+
+func (b BigInt) String() string {
+	return b.Int.String()
```
mitjat, line 49:
```suggestion
	// For some reason, big.Int only defines String() for a pointer receiver, so we need to define BigInt.String() manually.
	return b.Int.String()
```

common/types.go
```diff
@@ -22,20 +21,32 @@ func NewBigInt(v int64) BigInt {
...
 
 func (b *BigInt) UnmarshalJSON(text []byte) error {
-	v := strings.Trim(string(text), "\"")
-	return b.Int.UnmarshalJSON([]byte(v))
+	var s string
+	err := json.Unmarshal(text, &s)
+	if err != nil {
+		return err
+	}
+	return b.UnmarshalText([]byte(s))
```
mitjat, line 45: Much more proper :+1:

mitjat (approved):
Thank you!

common/types.go
mitjat, line 49: Nit/curiosity: In https://github.com/oasisprotocol/nexus/issues/566, I suggested `(&b.Int).String()`, because we need a `*big.Int` to call `String()` on it, whereas `b.Int` is of type `big.Int`. In my understanding, the code as written shouldn't work, yet it does. Do you know why?

common/types_test.go
mitjat: I searched around and it looks like the only prominent use of MarshalText is by `encoding/json` for objects that don't define `MarhsalJSON`. So:
- Maybe technically we could omit (Un)marshalJSON and retain the same behavior
- Or we could keep MarshalJSON and remove MarshalText because the latter doesn't get used much in isolation

Anyway. I'd call the current state of affairs Good Enough. Thank you for the tests.

common/types.go
pro-wh: doesn't it usually automatically use pointer receivers? let me see if I can find a reference for this

pro-wh: https://go.dev/ref/spec#Calls

> If x is [addressable](https://go.dev/ref/spec#Address_operators) and &x's method set contains m, x.m() is shorthand for (&x).m()

thanks go :skull:

common/types_test.go
pro-wh: ok I've found that we can't omit marshal/unmarshal json. we need these to override big.Int's marshal/unmarshal json, which use json numbers instead of strings

proceeding with removing our marhsal/unmarshal text. we're still able to call the embeded big.Int's implementation. I'm leaving the test in.

PR #572, opened 2023-11-22
metrics: Add timings for data fetch, data analysis

mitjat:
This PR adds histogram prometheus metrics for measuring how long it takes to fetch data for a block from the node, and how long to process it.

It also slightly refactors the analyzers (extracts data analysis into a separate method) and the metrics setup (expands the scope of the previously-db-focused "storage.go" to include all analysis metrics).

It also increases the granularity of the (previously very coarse) histograms. We now have 50 buckets per histogram. Hopefully not too much because we don't have a ton of histograms.

Resolves part of https://app.clickup.com/t/8692uy0j9

TODO: Use these in Grafana.


pro-wh (approved):

metrics/analysis.go
```diff
@@ -0,0 +1,135 @@
...
+	// Latencies of fetching a block's data from the node.
+	blockFetchLatencies *prometheus.HistogramVec
+}
+
+type CacheReadStatus string
+
+const (
+	CacheReadStatusHit      CacheReadStatus = "hit"
+	CacheReadStatusMiss     CacheReadStatus = "miss"
+	CacheReadStatusBadValue CacheReadStatus = "bad_value" // Value in cache was not valid (likely because of mismatched types / CBOR encoding).
```
Andrew7234, line 36: oh nice do we not have that panic anymore when the value doesn't exist in the cache

analyzer/consensus/consensus.go
```diff
@@ -300,6 +284,36 @@ func (m *processor) ProcessBlock(ctx context.Context, uheight uint64) error {
...
+	fetchTimer.ObserveDuration() // We make no observation in case of a data fetch error; those timings are misleading.
+
+	// Process data, prepare updates.
+	analysisTimer := m.metrics.BlockAnalysisLatencies()
+	batch := &storage.QueryBatch{}
+	err = m.queueDbUpdates(batch, *data)
+	analysisTimer.ObserveDuration()
+	if err != nil {
+		return err
+	}
```
Andrew7234, line 315: Hm what do you think of only observing when the analysis succeeds?

Andrew7234 (approved):

metrics/analysis.go
mitjat: Discussed offline. If our memory serves, the panic was only triggered in very specific circumstances, which were made impossible in #555, with the removal of `query_on_cache_miss`.

analyzer/consensus/consensus.go
mitjat: No very strong opinion, but I'm leaning towards marking it always. Just like we do for db txs, for example; we time it even if they need to be rolled back because of a deadlock. I think that makes sense; we're still spending time on analysis, even if it eventually fails, and the goal of the metrics is to help us understand where time is being spent.

mitjat (commented):
Thanks. I was going to test this by deploying from a branch, but I never got around to it and now this PR is getting to be dangerously old; also #577 is semi-dependent on it. So I'll merge now, and follow up with tweaks in a separate PR if they're needed.

csillag:
> Resolves part of the task.
>
> TODO: Use these in Grafana.

Commented by @mitjat elsewhere:

> Added to grafana for staging mainnet.

PR #573, opened 2023-11-25
analyzer/consensus: Support ParametersChange proposal

ptrus:


analyzer/consensus/consensus.go
```diff
@@ -907,6 +908,19 @@ func (m *processor) queueSubmissions(batch *storage.QueryBatch, data *governance
...
 				submission.ClosesAt,
 			)
+		case submission.Content.ChangeParameters != nil:
+			batch.Queue(queries.ConsensusProposalSubmissionChangeParametersInsert,
+				submission.ID,
+				submission.Submitter.String(),
+				submission.State.String(),
+				submission.Deposit.String(),
+				submission.Content.ChangeParameters.Module,
+				submission.Content.ChangeParameters.Changes,
```
ptrus, line 918: Changes is `cbor.RawMessage` of one of the various [`ConsensusParameterChanges`](https://github.com/search?q=repo%3Aoasisprotocol%2Foasis-core+ConsensusParameterChanges&type=code) types.

We could probably try unmarshaling this into `var changes map[string]interface{}` and then marshaling it into JSON, and storing a JSONB. I think that should work in all cases that we are likely to ever hit, but not absolutely certain if there are any potential edge cases, so for now opted to just storing the raw value. The frontend or the API controller could still do this.

analyzer/consensus/genesis.go
```diff
@@ -489,24 +491,43 @@ VALUES
...
+				query += fmt.Sprintf(
+					"\t(%d, '%s', '%s', %d, NULL, NULL, NULL, NULL, NULL, NULL, '%s', decode('%s', 'hex'), %d, %d, %d)",
+					proposal.ID,
+					proposal.Submitter.String(),
+					proposal.State.String(),
+					proposal.Deposit.ToBigInt(),
+					// 5 hardcoded NULLs for the proposal.Content.Upgrade fields.
+					// 1 hardocded NULL for the proposal.Content.CancelUpgrade.ProposalID field.
+					proposal.Content.ChangeParameters.Module,
+					hex.EncodeToString(proposal.Content.ChangeParameters.Changes),
```
mitjat, line 524: :facepalm: one day we'll use a sql library in genesis processing too ... it's just not justifiable to spend a day doing that busy work.

api/spec/v1.yaml
```diff
@@ -2040,6 +2040,16 @@ components:
           description: |
             The proposal to cancel, if this proposal proposes
             cancelling an existing proposal.
+        parameters_change_module:
+          type: string
+          description: |
+            The name of the module whose parameters are to be changed
+            by this 'parameters_change' proposal.
+        parameters_change:
+          type: string
```
mitjat, line 2049: It would be preferable to also have `format: byte` here.
Examples:
https://github.com/oasisprotocol/nexus/blob/6b2fe2e4d580e7f9be194081684c19dcf7f1a6bf/api/spec/v1.yaml#L152-L152
https://github.com/oasisprotocol/nexus/blob/6b2fe2e4d580e7f9be194081684c19dcf7f1a6bf/api/spec/v1.yaml#L2295-L2303

Then our openapi generator will represent it as a simple `[]byte` on the Go side, which is more truthful. And when serializing to JSON, it will automatically base64 encode it because that's what the openapi standard prescribes. So you'll be able to drop the ugly manual transformation (`p.ParametersChange = common.Ptr(base64.StdEncoding.EncodeToString(parametersChange))` and related lines).

mitjat (approved):
Getting to be one fat table, this :)

Thank you! LGTM assuming you agree with the type change (see comments).

FYI keep an eye out on #574 in case it goes in before yours.

api/spec/v1.yaml
mitjat: thank you! oh alternatively, I like the JSON solution (that you mention in the PR description) also

Andrew7234:
> keep an eye out on https://github.com/oasisprotocol/nexus/pull/574 in case it goes in before yours.

I'll hold off on merging 574 until this and #540 go in in an effort to minimize the number of migrations 😄

api/spec/v1.yaml
ptrus: Used the raw cbor message (bytes) for now.

ptrus:
> > keep an eye out on #574 in case it goes in before yours.
> 
> I'll hold off on merging 574 until this and #540 go in in an effort to minimize the number of migrations 😄

Thanks, merging this now!

PR #574, opened 2023-11-29
consolidate migrations

Andrew7234:
Resolves https://app.clickup.com/t/8692wz78k

Functionally a no-op. It would be nice to do a full reindex once this goes in, however, future deployments will continue to work since the current deployed databases have much higher 'current migration' number. However, any future migrations applied on top of this will also not apply until a reindex is done.

Validation:
Comparison of the dumped database schema (using `pg_dump -s`) for main branch / feature branch:
```
~/oasis/oasis-block-indexer(andrew7234/consolidate-migrations*) » diff main.sql new.sql                                                andrewlow@Beleriand

Explanation: function was removed since it was no longer used
218,230d217
< --
< -- Name: floor_5min(timestamp with time zone); Type: FUNCTION; Schema: public; Owner: rwuser
< --
<
< CREATE FUNCTION public.floor_5min(ts timestamp with time zone) RETURNS timestamp with time zone
<     LANGUAGE sql IMMUTABLE
<     AS $_$
<     SELECT date_trunc('hour', $1) + date_part('minute', $1)::int / 5 * '5 minutes'::interval;
< $_$;
<
<
< ALTER FUNCTION public.floor_5min(ts timestamp with time zone) OWNER TO rwuser;
<

Explanation: different ordering of columns in evm_nfts table
515a503,505
>     owner public.oasis_addr,
>     num_transfers integer DEFAULT 0 NOT NULL,
>     metadata jsonb,
520,523c510
<     image text,
<     owner public.oasis_addr,
<     num_transfers integer DEFAULT 0 NOT NULL,
<     metadata jsonb
---
>     image text

Explanation: different ordering of columns within a table
669a657
>     "timestamp" timestamp with time zone NOT NULL,
676d663
<     "timestamp" timestamp with time zone NOT NULL,

Explanation: pkey constraint on main is named differently since the table originally had a different name. No effect
962c949
< -- Name: evm_token_balances evm_token_balance_analysis_pkey; Type: CONSTRAINT; Schema: analysis; Owner: rwuser
---
> -- Name: evm_token_balances evm_token_balances_pkey; Type: CONSTRAINT; Schema: analysis; Owner: rwuser
966c953
<     ADD CONSTRAINT evm_token_balance_analysis_pkey PRIMARY KEY (runtime, token_address, account_address);
---
>     ADD CONSTRAINT evm_token_balances_pkey PRIMARY KEY (runtime, token_address, account_address);
```


ptrus:
> However, any future migrations applied on top of this will also not apply until a reindex is done.

Cant we just manually update the schema version (to 0) on live deployments and avoid reindexing?

mitjat (approved):
Thank you for the validation and the listing+explanation of diffs! Makes for an easy LGTM :)

Also, God's work :)

mitjat:
Oh when cutting a release, we should call special attention to this. We should also bump the minor version because this is a backwards-incompatible DB schema change.

PR #575, opened 2023-11-29
analyzer/runtime: Support WROSE events Deposit, Withdrawal

mitjat:
analyzer/runtime: Support WROSE events Deposit, Withdrawal

This PR adds support for two events specific to the WROSE contract (which is a copy of WETH): `Deposit` and `Withdrawal`.

The contract is a valid ERC-20 token, but it also generates these special events when converting between ROSE and WROSE.

**Background**
A user reported that nexus was reporting an incorrect balance of WROSE for them. Also, after they wrapped and subsequently unwrapped 1 ROSE as an experiment, the nexus balance never updated.
We traced this back to the fact that native token (un)wrapping requires special-casing: we need to recognize those events and at least query the node, but ideally also dead reckon the balances.
For details, see the larger comment blocks in the new code.

<details><summary>Addresses etc from the bug report</summary>

User wrapped 1 ROSE in mainnet Emerald block 7847770 ([txs](https://index.oasislabs.com/v1/emerald/transactions?block=7847770), [events](https://index.oasislabs.com/v1/emerald/events?block=7847770)), then unwrapped in 7847845 ([txs](https://index.oasislabs.com/v1/emerald/transactions?block=7847845), [events](https://index.oasislabs.com/v1/emerald/events?block=7847845))

```
User addr:
  0x2435ff763095d7c8ABfc1F05d1C4031B44013914
  oasis1qrrc6m8ulfcd00fct2jkedptkpc3gp7yzvhs30n5
  AAAAAAAAAAAAAAAAJDX/djCV18ir/B8F0cQDG0QBORQ= (as EVM log topic)
WROSE addr:
  0x21C718C22D52d0F3a789b752D4c2fD5908a8A733
  oasis1qpgcp5jzlgk4hcenaj2x82rqk8rrve2keyuc8aaf
Deposit signature:
  0xe1fffcc4923d04b559f4d29a8bfc6cda04eb5b0d3c460751c2402c5c5cc9109c
Withdrawal signature:
  0x7fcf532c15f0a6db0bd6d0e038bea71d30d808c7d98cb3bf7268a95bf5081b65
  f89TLBXwptsL1tDgOL6nHTDYCMfZjLO/cmipW/UIG2U= (as EVM log topic)
```

</details> 

**Testing**
1) e2e regressions. I verified the diff manually: The new WROSE balances are correct at round 1_004_297 (where regression test stops).
2) I also ran the analyzer over the range of blocks around the above bug report, and confirmed that both ROSE and WROSE get dead-reckoned as expected.


analyzer/runtime/extract.go
```diff
@@ -1106,6 +1107,87 @@ func extractEvents(blockData *BlockData, relatedAccountAddresses map[apiTypes.Ad
...
+					registerTokenDecrease(blockData.TokenBalanceChanges, evm.NativeRuntimeTokenAddress, ownerAddr, amount)
+
+					// ^ The above includes dead-reckoning for the native token because no events are emitted for native token transfers.
+					//
+					// Example: in mainnet Emerald block 7847845, account A withdrew 1 WROSE from the WROSE contract,
+					//   i.e. unwrapped 1 WROSE into ROSE. The effect is that A's WROSE balance decreases by 1,
+					//   and the WROSE contract transfers 1 ROSE to A.
+					//   However, that block shows a single event: evm.log Withdrawal(from=A, value=1000000000000000000)
+					//   Similarly, in block 7847770, A deposited 1 ROSE into the WROSE contract.
+					//   No ROSE events were emitted, only evm.log Deposit(to=A, value=1000000000000000000)
```
ptrus, line 1129: Don't we already dead-reckon the caller and callee for every evm.Call? e.g. via: 

https://github.com/oasisprotocol/nexus/blob/a1ba62c6028ae18e4777a83fe39a5ac57f0ee27b/analyzer/runtime/extract.go#L531-L548 

or am I missing something/

mitjat: Well spotted. However, there's a subtlety: We dead-reckone native ROSE for every evm.Call _that has an empty body_ (line 539). The call to the `withdraw()` method of WROSE, for example, is not like that. See it [here](https://index.oasislabs.com/v1/emerald/transactions?block=7847845). The real value (1 ROSE) is hidden in the `data` field (which contains the signature of the  `withdraw()` method (`2e1a7d4d`), and `1e18`, which is how many wei we're withdrawing).

We _could_ skip dead reckoning of ROSE in the new code, because the dead reckoning (by a delta=0) in the code you linked would cause the evm_token_balances analyzer to later fetch the true balance from the runtime. But with explicit dead reckoning (by the correct amount), the UX should be a little better.

pro-wh: problem here is deposit can be empty or non-empty body. anyone want to disable the old reckoning? in my recollection it was a hack we added to support wrose

mitjat: @pro-wh the "hack" is not just for WROSE; any contract can (and multiple do) transfer ROSE under the hood.

@ptrus We do always enqueue both the tx signer and the smart contract account to be re-queried for ROSE balance; see lines 544-547.

The only difference between empty vs nonempty `Call::data` is just this: If `data` is nonempty empty, we assume it's too hard to guess know what the contract will do, so we just enqueue the accounts to be re-queried, but we do not dead-reckon any changes. OTOH if `data` is empty, we assume that `Call::value` attribute is the amount of ROSE that the call will transfer, and we optimistically use that value for dead-reckoning (but still enqueue the accounts to be re-queried).

So the ROSE balance was never off for long in the `WROSE::deposit()` case that this PR addresses; we did already enqueue the user's account to be re-queried. But the _WROSE balance_ we did not query; there was no reason to, because the WROSE contract was not (from the perspective of observing just the `Deposit` or `Withdrawal` event or tx) an ERC-20.

This gives me the idea to try and enqueue every `(contract, account)` pair (where `account` calls `contract`) to be queried for `account`'s balance of the `contract` EVM token. I've been playing with that and it enqueues a whole bunch, but IDK yet if it's meaningful. Querying the node for EVM balances gives me HTTP 502 a lot of the time.

ptrus: > So the ROSE balance was never off for long in the WROSE::deposit() case that this PR addresses; we did already enqueue the user's account to be re-queried

Ok, yeah that's what I was wondering. So only the token WROSE was problematic. I think I actually misread the ticked description initially, I understood that the ROSE balance was wrong as well, which didn't make sense to me.

ptrus (approved):

analyzer/runtime/extract.go
pro-wh: > @pro-wh the "hack" is not just for WROSE; any contract can (and multiple do) transfer ROSE under the hood.

contracts that transfer ROSE are the cases that the hack doesn't support. the hack assumes the ROSE goes from the tx sender to the contract and stays there, which supports the wROSE deposit case

pro-wh: > No ERC20 balances were corrected in the e2e regression test block range(*). So I'll not include that enqueuing for now.

wait I thought we were talking about native balances

PR #577, opened 2023-11-30
[metrics] add block analyzer queuelength metrics

Andrew7234:
Resolves https://app.clickup.com/t/8692qbwf4

To enable the queuelength metric for runtime block analyzers, nexus needs the current chain heights of the runtimes. The `node_stats` analyzer currently fetches/stores this data only for consensus, so this PR also adds runtime support for the node_stats analyzer.

Note: We _could_ avoid this by fetching the chain heights directly in the block analyzers. Although the height would be fresher, it would also require a second round-trip in the main block processing loop. Since the metrics are internal-only I opted to use the chain heights we already store in `chain.latest_node_heights`. Open to suggestions though.

Alternatively alternatively, we could put the metric updating thing in a concurrent loop separate from the main block processing loop. Not sure if it's worth the extra complexity since slow-sync is, by name, expected to be slower.

```
~/oasis/oasis-block-indexer(andrew7234/block-analyzer-queuelength*) » curl localhost:8009/metrics | grep queue_length                 
100 43060    0 43060    0     0  7934k      0 --:--:-- --:--:-- --:--:-- 41.0M
# HELP consensus_queue_length number of blocks left to process for the consensus analyzer
# TYPE consensus_queue_length gauge
consensus_queue_length 0 
^ failed bc couldn't run consensus analyzer locally. Default value of a prometheus gauge is 0, which is what we see here

# HELP emerald_queue_length number of blocks left to process for the emerald analyzer
# TYPE emerald_queue_length gauge
emerald_queue_length 2.392834e+06
# HELP sapphire_queue_length number of blocks left to process for the sapphire analyzer
# TYPE sapphire_queue_length gauge
sapphire_queue_length 3.726054e+06
```

### DIFF
--- a/analyzer/block/block.go
+++ b/analyzer/block/block.go
@@ -17,6 +17,7 @@ import (
 	"github.com/oasisprotocol/nexus/analyzer/util"
 	"github.com/oasisprotocol/nexus/config"
 	"github.com/oasisprotocol/nexus/log"
+	"github.com/oasisprotocol/nexus/metrics"
 	"github.com/oasisprotocol/nexus/storage"
 )
 
@@ -61,8 +62,9 @@ type blockBasedAnalyzer struct {
 
 	processor BlockProcessor
 
-	target storage.TargetStorage
-	logger *log.Logger
+	target  storage.TargetStorage
+	logger  *log.Logger
+	metrics metrics.AnalysisMetrics
 
 	slowSync bool
 }
@@ -259,6 +261,31 @@ func (b *blockBasedAnalyzer) ensureSlowSyncPrerequisites(ctx context.Context) (o
 	return true
 }
 
+// sendQueueLengthMetric updates the relevant Prometheus metric with an approximation
+// of the number of known blocks that require processing, calculated as the difference
+// between the analyzer's highest processed block and the current block height
+// on-chain (fetched by the node-stats analyzer).
+// Note that the true count may be higher during fast-sync.
+func (b *blockBasedAnalyzer) sendQueueLengthMetric(queueLength uint64) {
+	// For block-based analyzers, the analyzer name is identical to the layer name.
+	b.metrics.QueueLength(b.analyzerName).Set(float64(queueLength))
+}
+
+// Returns the chain height of the layer this analyzer is processing.
+// The heights are fetched and added to the database by the node-stats analyzer.
+func (b *blockBasedAnalyzer) nodeHeight(ctx context.Context) (int, error) {
+	var nodeHeight int
+	err := b.target.QueryRow(ctx, queries.NodeHeight, b.analyzerName).Scan(&nodeHeight)
+	switch err {
+	case nil:
+		return nodeHeight, nil
+	case pgx.ErrNoRows:
+		return -1, nil
+	default:
+		return -1, fmt.Errorf("error fetching chain height for consensus: %w", err)
+	}
+}
+
 // Start starts the block analyzer.
 func (b *blockBasedAnalyzer) Start(ctx context.Context) {
 	// Run prework.
@@ -325,6 +352,10 @@ func (b *blockBasedAnalyzer) Start(ctx context.Context) {
 			backoff.Failure()
 			continue
 		}
+		nodeHeight, err := b.nodeHeight(ctx)
+		if err != nil {
+			b.logger.Warn("error fetching current node height: %w", err)
+		}
 
 		// Process blocks.
 		b.logger.Debug("picked blocks for processing", "heights", heights)
@@ -337,6 +368,12 @@ func (b *blockBasedAnalyzer) Start(ctx context.Context) {
 			// in the batch will exceed the current lock expiry of 5min. The analyzer will terminate
 			// the batch early and attempt to refresh the locks for a new batch.
 			if b.slowSync {
+				// In slow-sync mode, we update the node-height at each block for a more
+				// precise measurement.
+				nodeHeight, err = b.nodeHeight(ctx)
+				if err != nil {
+					b.logger.Warn("error fetching current node height: %w", err)
+				}
 				select {
 				case <-time.After(backoff.Timeout()):
 					// Process the next block
@@ -378,6 +415,10 @@ func (b *blockBasedAnalyzer) Start(ctx context.Context) {
 				// Unlock a failed block, so it can be retried sooner.
 				b.unlockBlocks(ctx, []uint64{height})
 				continue
+			}
+			// If we successfully fetched the node height earlier, update the estimated queue length.
+			if nodeHeight != -1 {
+				b.sendQueueLengthMetric(uint64(nodeHeight) - height)
 			}
 			cancel()
 			backoff.Success()
@@ -425,13 +466,16 @@ func NewAnalyzer(
 	if batchSize == 0 {
 		batchSize = defaultBatchSize
 	}
-	return &blockBasedAnalyzer{
+	a := &blockBasedAnalyzer{
 		blockRange:   blockRange,
 		batchSize:    batchSize,
 		analyzerName: name,
 		processor:    processor,
 		target:       target,
 		logger:       logger.With("analyzer", name, "mode", mode),
+		metrics:      metrics.NewDefaultAnalysisMetrics(name),
 		slowSync:     mode == analyzer.SlowSyncMode,
-	}, nil
+	}
+
+	return a, nil
 }
--- a/analyzer/consensus/consensus.go
+++ b/analyzer/consensus/consensus.go
@@ -297,13 +297,13 @@ func (m *processor) ProcessBlock(ctx context.Context, uheight uint64) error {
 	// Fetch all data.
 	fetchTimer := m.metrics.BlockFetchLatencies()
 	data, err := fetchAllData(ctx, m.source, m.network, height, m.mode == analyzer.FastSyncMode)
+	fetchTimer.ObserveDuration()
 	if err != nil {
 		if strings.Contains(err.Error(), fmt.Sprintf("%d must be less than or equal to the current blockchain height", height)) {
 			return analyzer.ErrOutOfRange
 		}
 		return err
 	}
-	fetchTimer.ObserveDuration() // We make no observation in case of a data fetch error; those timings are misleading.
 
 	// Process data, prepare updates.
 	analysisTimer := m.metrics.BlockAnalysisLatencies()
--- a/analyzer/node_stats/node_stats.go
+++ b/analyzer/node_stats/node_stats.go
@@ -15,6 +15,7 @@ import (
 	"github.com/oasisprotocol/nexus/storage/oasis/nodeapi"
 
 	consensusAPI "github.com/oasisprotocol/nexus/coreapi/v22.2.11/consensus/api"
+	runtimeAPI "github.com/oasisprotocol/nexus/coreapi/v22.2.11/runtime/client/api"
 )
 
 const (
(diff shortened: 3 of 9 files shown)


analyzer/block/block.go
```diff
@@ -337,6 +354,10 @@ func (b *blockBasedAnalyzer) Start(ctx context.Context) {
 			// in the batch will exceed the current lock expiry of 5min. The analyzer will terminate
 			// the batch early and attempt to refresh the locks for a new batch.
 			if b.slowSync {
+				// Only update this metric on slow-sync mode.
+				if err := b.sendQueueLengthMetric(ctx, height); err != nil {
+					b.logger.Warn("error updating queue length metric: %w", err)
+				}
```
Andrew7234, line 360: could put this in a separate goroutine and fetch both the latest processed block and latest chain block from the db

analyzer/node_stats/node_stats.go
```diff
@@ -54,22 +61,44 @@ func NewAnalyzer(
...
+	var latestBlock *nodeapi.RuntimeBlockHeader
+	var err error
+	switch layer {
+	case common.LayerEmerald:
+		latestBlock, err = p.emeraldSource.GetBlockHeader(ctx, runtimeAPI.RoundLatest)
+	case common.LayerSapphire:
+		latestBlock, err = p.sapphireSource.GetBlockHeader(ctx, runtimeAPI.RoundLatest)
+	default:
 		return fmt.Errorf("unsupported layer %s", layer)
 	}
```
mitjat, line 95: Stylistic only: Would it make sense to unify the consensus flow and the runtime flow more? Something like this:

```suggestion
	latestHeight := -1  // will be fetched from the node
	switch layer {
	case common.LayerConsensus:
		latestBlock, err = p.source.GetBlock(ctx, consensusAPI.HeightLatest)
		if err != nil {
			return fmt.Errorf("error fetching latest block height for layer %s: %w", layer, err)
		}
		latestHeight = latestBlock.Height
	case common.LayerEmerald:
		latestBlock, err = p.emeraldSource.GetBlockHeader(ctx, runtimeAPI.RoundLatest)
		if err != nil {
			return fmt.Errorf("error fetching latest block height for layer %s: %w", layer, err)
		}
		latestHeight = latestBlock.Round
	case common.LayerSapphire:
		latestBlock, err = p.sapphireSource.GetBlockHeader(ctx, runtimeAPI.RoundLatest)
		if err != nil {
			return fmt.Errorf("error fetching latest block height for layer %s: %w", layer, err)
		}
		latestHeight = latestBlock.Round		
	default:
		return fmt.Errorf("unsupported layer %s", layer)
	}
```

analyzer/block/block.go
```diff
@@ -259,6 +266,16 @@ func (b *blockBasedAnalyzer) ensureSlowSyncPrerequisites(ctx context.Context) (o
 	return true
 }
 
+func (b *blockBasedAnalyzer) sendQueueLengthMetric(ctx context.Context, height uint64) error {
+	queueLength, err := b.processor.QueueLength(ctx, height)
+	if err != nil {
+		b.logger.Warn("error fetching queue length for %s: %w", b.analyzerName, err)
```
mitjat, line 272: Nit: This method (`sendQueueLengthMetric()`) is only called once, and the caller already logs any errors. So let's log either here or there. I'd say it's more canonical to log outside, and return a `fmt.Errorf()` here.

analyzer/block/block.go
```diff
@@ -50,6 +51,11 @@ type BlockProcessor interface {
 	// The method should be a no-op if all the blocks up to `height` have been analyzed
 	// in slow-sync mode, assuming no bugs in dead reckoning code.
 	FinalizeFastSync(ctx context.Context, lastFastSyncHeight int64) error
+	// QueueLength returns the number of known blocks that still need to be processed,
+	// calculated as the difference between the analyzer's highest processed block
+	// and the current block height on-chain (fetched by the node-stats analyzer).
+	// Note that the true count may be higher during fast-sync.
+	QueueLength(ctx context.Context, height uint64) (int, error)
```
mitjat, line 58: Does this really need to be implemented by the `processor`? The `blockBasedAnalyzer` knows the height that's being processed currently, and the node height is stored in the DB anyway.

analyzer/block/block.go
mitjat: This metric will actually be especially valuable during fast-sync, which is when the queue is of a non-trivial size. We can report an approximate value during fast-sync, that's fine. (Your docstring for QueueLength already already foresees this.)

We could fetch the latest node block at the beginning of the batch (around where we call fetchBatchForProcessing()), store it in a local variable, and update the metric at the end of `if err := b.processor.ProcessBlock() { ... }`. This ties into my comment about `processor::QueueLength()` probably not being needed.

cmd/analyzer/analyzer.go
```diff
@@ -460,7 +460,15 @@ func NewService(cfg *config.AnalysisConfig) (*Service, error) { //nolint:gocyclo
...
-			return nodestats.NewAnalyzer(cfg.Analyzers.NodeStats.ItemBasedAnalyzerConfig, sourceClient, dbClient, logger)
+			emeraldClient, err1 := sources.Runtime(ctx, common.RuntimeEmerald)
+			if err1 != nil {
+				return nil, err1
+			}
+			sapphireClient, err1 := sources.Runtime(ctx, common.RuntimeSapphire)
+			if err1 != nil {
+				return nil, err1
+			}
+			return nodestats.NewAnalyzer(cfg.Analyzers.NodeStats.ItemBasedAnalyzerConfig, sourceClient, emeraldClient, sapphireClient, dbClient, logger)
```
mitjat, line 471: Thank you for expanding the node stats analyzer!

It's not great that we're hardcoding the dependency on all the runtimes here. Imagine Sapphire becomes unavailable and we still want stats for consensus; or that we're wanting to test `nodestats` locally or in (nonexistent :/) tests, but don't go through the hassle of setting up access to all runtimes.

Three options I can think of:
- Add a config flag (list) for specifying which layers to include; this is probably the most "proper" solution, and quite usable too if the default is `[consensus, sapphire, emerald]`.
- Decide which layers to include based on the presence/absence of their block analyzers. That's very convenient but leads to implicit dependencies between analyzers / sections of config, which is super ugly.
- Leave as-is. If e.g. sapphire becomes unavailable, things will continue to work (thanks to lazy initialization of node clients in #555), we'll just see a good amount of error spam in logs.

Your call on whether to go with 1 or 3. I think both are justifiable. I'm partially writing them out to check if option 3 holds the way I've described it.

mitjat (commented):
Thanks Andy!

analyzer/block/block.go
Andrew7234: Yeah I was on the fence about this - I initially aimed to mirror item analyzers but I like your suggestion, it's cleaner. Will update

analyzer/block/block.go
Andrew7234: Hmm good point. Updated to your suggestion - with an additional tweak of also fetching the `nodeHeight` at each loop during slow-sync

analyzer/block/block.go
```diff
@@ -425,13 +466,20 @@ func NewAnalyzer(
...
 		logger:       logger.With("analyzer", name, "mode", mode),
-		slowSync:     mode == analyzer.SlowSyncMode,
-	}, nil
+		queueLengthMetric: prometheus.NewGauge(prometheus.GaugeOpts{
+			Name: fmt.Sprintf("%s_queue_length", name),
+			Help: fmt.Sprintf("number of blocks left to process for the %s analyzer", name),
+		}),
+		slowSync: mode == analyzer.SlowSyncMode,
+	}
+	prometheus.MustRegister(a.queueLengthMetric)
```
mitjat, line 482: I think this might fail when you have parallelism, because prometheus only lets you register a metric once. This won't be a problem if you go with the suggested mini refactor.

analyzer/block/block.go
```diff
@@ -425,13 +466,20 @@ func NewAnalyzer(
...
 		analyzerName: name,
 		processor:    processor,
 		target:       target,
 		logger:       logger.With("analyzer", name, "mode", mode),
-		slowSync:     mode == analyzer.SlowSyncMode,
-	}, nil
+		queueLengthMetric: prometheus.NewGauge(prometheus.GaugeOpts{
+			Name: fmt.Sprintf("%s_queue_length", name),
+			Help: fmt.Sprintf("number of blocks left to process for the %s analyzer", name),
+		}),
```
mitjat, line 479: Request for small refactor: Can you predefine this metric in `metrics/analysis.go`? Having it there is nice because
- it keeps metric definitions in one place, making it easier to understand what our prometheus endpoint exposes
- it defines helper funcs that let you create metrics with _type-checked_ labels; see e.g. `DatabaseLatencies(...)`
- (more theoretical) it makes it easier to mock out the entirety of prometheus, should we ever want to

So the idea is to
1. define a new gauge-typed `queueLength` field inside `AnalysisMetrics`, with `analyzer` being one of the labels, possibly the only one
2. use this predefined metric from block.go but also item.go (so neither will need a `queueLengthMetric` field any more)

analyzer/consensus/consensus.go
```diff
@@ -300,6 +284,36 @@ func (m *processor) ProcessBlock(ctx context.Context, uheight uint64) error {
...
+	// Fetch all data.
+	fetchTimer := m.metrics.BlockFetchLatencies()
+	data, err := fetchAllData(ctx, m.source, m.network, height, m.mode == analyzer.FastSyncMode)
+	if err != nil {
+		if strings.Contains(err.Error(), fmt.Sprintf("%d must be less than or equal to the current blockchain height", height)) {
+			return analyzer.ErrOutOfRange
+		}
+		return err
+	}
+	fetchTimer.ObserveDuration() // We make no observation in case of a data fetch error; those timings are misleading.
```
mitjat, line 306: I was just answering a similar thread in #572 :)
IDK that they're misleading. If, for example, the data fetch fails because of a timeout in 90% of the cases, it's very informative to learn that we're spending a minute per block in data fetch, rather than having the chart show the average fetch time for the other 10% of cases.
It's maybe a tiny bit easier to argue that failure cases are misleading when the data fetch fails quickly because of e.g. a malformed request. But even then ... that's objectively the time spent on data fetch, right? The anomaly in the graph will alert us to something.

mitjat (approved):
Thank you! The are one or two not-entirely trivial comment threads still unresolved, but I expect smooth sails from here on out. I'm LGTMing now as holidays will make it harder to sync.

It looks like you might also have to be careful with the rebase now that #572 is merged? Hopefully not :crossed_fingers:

cmd/analyzer/analyzer.go
Andrew7234: Went with option 1!

Some notes for future reference:
The code here creates the consensus/runtime clients for all layers even if they weren't specified in the config file. The LazyGrpcConnect function mentioned above allows us to connect to oasis-node only when needed. However, when we instantiate a new runtime client we also make a connections.SDKConnect call a few lines above [here,](https://github.com/oasisprotocol/nexus/blob/main/storage/oasis/nodeapi/history/runtime.go#L27) which seems like it would error. However, the underlying connection code only establishes the connection and checks the chainContext; it does not fetch any runtime-specific info. In almost all cases, the default rpc node specified in the config file will pass this check successfully. Thus the only failure case should be when a runtime node is explicitly specified and is down. In this case, the failure is immediate and obvious and can be circumvented by either a) restoring the node or b) removing the problematic layer from the node-stats config list.

analyzer/block/block.go
Andrew7234: Done!

analyzer/consensus/consensus.go
Andrew7234: Hmm true, updated!

PR #578, opened 2023-12-04
(unfinished) load partial verifications too

pro-wh:
cc @mitjat 

here's whatever I had in an unfinished branch. might have been off on a wrong track though


pro-wh:
#648

PR #579, opened 2023-12-05
api: add NFT-oriented event filtering

pro-wh:
/{runtime}/events:
- contract_address
- nft_id

aaa it's so divisive

### DIFF
--- a/api/spec/v1.yaml
+++ b/api/spec/v1.yaml
@@ -973,6 +973,31 @@ paths:
             A filter on the evm log signatures.
             Note: The filter will only match on parsed (verified) EVM events.
           example: '0xddf252ad1be2c89b69c2b068fc378daa952ba7f163c4a11628f55a4df523b3ef'
+        - in: query
+          name: contract_address
+          schema:
+            type: string
+          description: |
+            A filter on a smart contract. Every returned event will have been
+            emitted by the contract at this Oasis address.
+          example: *staking_address_1
+        - in: query
+          name: nft_id
+          schema:
+            type: string
+          example: "999"
+          description: |
+            A filter on NFT events. Every returned event will be specifically
+            about this NFT instance ID. You must specify the contract_address
+            filter with this filter.
+            Currently this only supports ERC-721 Transfer events.
+            This may expand to support other event types in the future.
+            If you want only ERC-721 Transfer events, specify
+            evm_log_signature=ddf252ad1be2c89b69c2b068fc378daa952ba7f163c4a11628f55a4df523b3ef
+            to avoid inadvertently getting other event types if they are
+            supported later.
+            Using an evm_log_signature filter with this set to any other value
+            will match no events.
       responses:
         '200':
           description: |
--- a/api/spec/v1.yaml
+++ b/api/spec/v1.yaml
@@ -140,13 +140,13 @@ x-type-annotations:
     example: "oasis1qpg2xuz46g53737343r20yxeddhlvc2ldqsjh70p"
     x-go-type: staking.Address
     x-go-type-import: { name: staking, path: "github.com/oasisprotocol/nexus/coreapi/v22.2.11/staking/api" }
-    description: An oasis-style (bech32) address.
+    description: An Oasis-style (bech32) address.
   address: &AddressType
     type: string
     pattern: '^oasis1[a-z0-9]{40}$'
     example: "oasis1qpg2xuz46g53737343r20yxeddhlvc2ldqsjh70p"
     x-go-type: Address
-    description: An oasis-style (bech32) address.
+    description: An Oasis-style (bech32) address.
   pubkey: &Ed25519PubKeyType
     type: string
     format: byte  # means base64-encoded raw bytes
@@ -875,8 +875,8 @@ paths:
             the sender or the recipient of tokens.
             Nexus detects related accounts inside EVM transactions and events on a
             best-effort basis. For example, it inspects ERC20 methods inside `evm.Call` txs.
-            However, you must provide the oasis-style derived address here, not the Eth address.
-            See `AddressPreimage` for more info on oasis-style vs Eth addresses.
+            However, you must provide the Oasis-style derived address here, not the Eth address.
+            See `AddressPreimage` for more info on Oasis-style vs Eth addresses.
       responses:
         '200':
           description: |
@@ -1786,18 +1786,18 @@ components:
       description: |
         The data from which a consensus-style address (`oasis1...`)
         was derived. Notably, for EVM runtimes like Sapphire,
-        this links the oasis address and the Ethereum address.
+        this links the Oasis address and the Ethereum address.
 
         Oasis addresses are derived from a piece of data, such as an ed25519
         public key or an Ethereum address. For example, [this](https://github.com/oasisprotocol/oasis-sdk/blob/b37e6da699df331f5a2ac62793f8be099c68469c/client-sdk/go/helpers/address.go#L90-L91)
-        is how an Ethereum is converted to an oasis address. The type of underlying data usually also
+        is how an Ethereum is converted to an Oasis address. The type of underlying data usually also
         determines how the signatuers for this address are verified.
 
         Consensus supports only "staking addresses" (`context="oasis-core/address: staking"`
         below; always ed25519-backed).
         Runtimes support all types. This means that every consensus address is also
         valid in every runtime. For example, in EVM runtimes, you can use staking
-        addresses, but only with oasis tools (e.g. a wallet); EVM contracts such as
+        addresses, but only with Oasis tools (e.g. a wallet); EVM contracts such as
         ERC20 tokens or tools such as Metamask cannot interact with staking addresses.
       type: object
       required: [context, address_data]
@@ -1805,7 +1805,7 @@ components:
         context:
           $ref: "#/components/schemas/AddressDerivationContext"
           description: |
-            The method by which the oasis address was derived from `address_data`.
+            The method by which the Oasis address was derived from `address_data`.
           example: "oasis-runtime-sdk/address: secp256k1eth"
         context_version:
           type: integer
@@ -1816,7 +1816,7 @@ components:
           type: string
           format: byte
           description: |
-            The base64-encoded data from which the oasis address was derived.
+            The base64-encoded data from which the Oasis address was derived.
             When `context = "oasis-runtime-sdk/address: secp256k1eth"`, this
             is the Ethereum address (in base64, not hex!).
           example: 'INLp2Ih3YIdcA+zFNhM+SIGyFgKsYYc9SKQeKRKe2uI='
@@ -1848,7 +1848,7 @@ components:
           description: Number of tokens held, in base units.
         token_contract_addr:
           type: string
-          description: The oasis address of this token's contract.
+          description: The Oasis address of this token's contract.
           example: 'oasis1qzk5pr2x8ah04lgjee3lv06fmyvqvz45egjv4ps0'
         token_contract_addr_eth:
           type: string
@@ -1888,7 +1888,7 @@ components:
       properties:
         holder_address:
           type: string
-          description: The oasis address of the account holder.
+          description: The Oasis address of the account holder.
           example: *staking_address_1
         eth_holder_address:
           type: string
--- a/storage/client/client.go
+++ b/storage/client/client.go
@@ -3,12 +3,14 @@ package client
 import (
 	"context"
 	"database/sql"
+	"encoding/base64"
 	"fmt"
 	"math/big"
 	"strings"
 	"time"
 
 	"github.com/dgraph-io/ristretto"
+	"github.com/ethereum/go-ethereum/accounts/abi"
 	ethCommon "github.com/ethereum/go-ethereum/common"
 	"github.com/jackc/pgx/v5"
 	"github.com/jackc/pgx/v5/pgtype"
@@ -19,6 +21,7 @@ import (
 	oasisConfig "github.com/oasisprotocol/oasis-sdk/client-sdk/go/config"
 	sdkTypes "github.com/oasisprotocol/oasis-sdk/client-sdk/go/types"
 
+	"github.com/oasisprotocol/nexus/analyzer/evmabi"
 	beacon "github.com/oasisprotocol/nexus/coreapi/v22.2.11/beacon/api"
 	staking "github.com/oasisprotocol/nexus/coreapi/v22.2.11/staking/api"
 
@@ -1294,6 +1297,18 @@ func (c *StorageClient) RuntimeEvents(ctx context.Context, p apiTypes.GetRuntime
 		h := ethCommon.HexToHash(*p.EvmLogSignature)
 		evmLogSignature = &h
 	}
+	var NFTIdB64 *string
+	if p.NftId != nil {
+		erc721TransferTokenIdBI := &big.Int{}
+		if err := erc721TransferTokenIdBI.UnmarshalText([]byte(*p.NftId)); err != nil {
+			return nil, fmt.Errorf("unmarshalling erc721_transfer_token_id: %w", err)
+		}
+		erc721TransferTokenIdBuf, err := abi.Arguments{evmabi.ERC721.Events["Transfer"].Inputs[2]}.Pack(erc721TransferTokenIdBI)
+		if err != nil {
+			return nil, fmt.Errorf("ABI-packing erc721_transfer_token_id: %w", err)
+		}
+		NFTIdB64 = common.Ptr(base64.StdEncoding.EncodeToString(erc721TransferTokenIdBuf))
+	}
 
 	res, err := c.withTotalCount(
 		ctx,
@@ -1305,6 +1320,8 @@ func (c *StorageClient) RuntimeEvents(ctx context.Context, p apiTypes.GetRuntime
 		p.Type,
 		evmLogSignature,
 		p.Rel,
+		p.ContractAddress,
+		NFTIdB64,
 		p.Limit,
 		p.Offset,
 	)
--- a/storage/client/client.go
+++ b/storage/client/client.go
@@ -1297,6 +1297,9 @@ func (c *StorageClient) RuntimeEvents(ctx context.Context, p apiTypes.GetRuntime
 		h := ethCommon.HexToHash(*p.EvmLogSignature)
 		evmLogSignature = &h
 	}
+	if p.NftId != nil && p.ContractAddress == nil {
+		return nil, fmt.Errorf("must specify contract_address with nft_id")
+	}
 	var NFTIdB64 *string
 	if p.NftId != nil {
 		erc721TransferTokenIdBI := &big.Int{}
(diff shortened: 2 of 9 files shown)


storage/migrations/26_runtime_events_nft_transfers.up.sql
```diff
@@ -0,0 +1,10 @@
+BEGIN;
+
+-- CREATE INDEX ix_runtime_events_nft_transfers
```
pro-wh, line 3: :smiling_face_with_tear:

api/spec/v1.yaml
```diff
@@ -961,6 +961,17 @@ paths:
...
+          name: evm_log_contract_addr_eth
+          schema:
+            type: string
+          description: Get EVM log events from the contract at this Ethereum address.
+        - in: query
+          name: erc721_transfer_token_id
+          schema:
+            type: string
+          example: "999"
+          description: Get ERC-721 Transfer events about this token ID.
```
mitjat, line 974: It feels wrong to couple the type of the event (ERC-721 transfer) and the token ID so tightly in a single filter. Especially since we already have the `evm_log_signature` filter. 
I suggest we rename this simply `nft_token_id`, and adjust the description accordingly: "A filter on the token (instance) ID of an NFT. All returned events will involve this token." This leaves the API more amenable to future changes.

I'm guessing you bundled these two together due to performance concerns? I share those :). We can add to the description: "If this parameter is provided, `evm_log_contract_addr_eth` has to be provided too, and `evm_log_signature` has to equal 0x... (the ERC-721 Transfer event)". We can then enforce this at the param validation layer, before we call `RuntimeEvents()`. This puts a tiny bit more onus on the user, but makes the query more explicit, and keeps it possible for us to relax the constraint in the future, should we want to (ERC-1155 support, Cipher support, ...), without having to deprecate this query param and invent new ones.

api/spec/v1.yaml
```diff
@@ -961,6 +961,17 @@ paths:
             A filter on the evm log signatures.
             Note: The filter will only match on parsed (verified) EVM events.
           example: '0xddf252ad1be2c89b69c2b068fc378daa952ba7f163c4a11628f55a4df523b3ef'
+        - in: query
+          name: evm_log_contract_addr_eth
+          schema:
+            type: string
```
mitjat, line 967: Do we tolerate the leading 0x? Add to the description and/or add an example value.

api/spec/v1.yaml
```diff
@@ -961,6 +961,17 @@ paths:
             A filter on the evm log signatures.
             Note: The filter will only match on parsed (verified) EVM events.
           example: '0xddf252ad1be2c89b69c2b068fc378daa952ba7f163c4a11628f55a4df523b3ef'
+        - in: query
+          name: evm_log_contract_addr_eth
```
mitjat, line 965: I'd drop the `_eth` suffix. I checked and we never use it (or the `eth_` prefix) for _query_ params. I think it's explicit enough if the docstring says we want the hex-encoded eth address. Also, if we ever want to support oasis-style addresses, I expect we'll follow the lead of a few other [similar params](https://github.com/oasisprotocol/nexus/blob/f495dac742ed91fd6e32336dfbc99d00955c9abd/api/spec/v1.yaml#L883-L895) so far, where we accept either format.

I'd _also_ drop `evm_log`, because that's true now but might not be in the future (Cipher or similar).

How do you like `emitting_contract_addr` to make it clearer how this contract is related? Or just `contract_add`.

api/spec/v1.yaml
mitjat, line 1: Can you please also add an entry in `e2e_regression/run.sh` to test the new param?

mitjat (commented):
Thank you, Warren! This gives more tangible context to your Slack questions. I'm not so worried about query performance (as long as it's reasonable), but I do have some thoughts on keeping this feature a little less hardcoded to EVM, ERC-721 NFTs.

storage/migrations/26_runtime_events_nft_transfers.up.sql
pro-wh: hadn't tried with this actually

storage/client/client.go
```diff
@@ -1269,6 +1272,22 @@ func (c *StorageClient) RuntimeEvents(ctx context.Context, p apiTypes.GetRuntime
...
+		erc721TransferTokenIdBI := &big.Int{}
+		if err := erc721TransferTokenIdBI.UnmarshalText([]byte(*p.Erc721TransferTokenId)); err != nil {
+			return nil, fmt.Errorf("unmarshalling erc721_transfer_token_id: %w", err)
+		}
+		erc721TransferTokenIdBuf, err := abi.Arguments{evmabi.ERC721.Events["Transfer"].Inputs[2]}.Pack(erc721TransferTokenIdBI)
+		if err != nil {
+			return nil, fmt.Errorf("ABI-packing erc721_transfer_token_id: %w", err)
+		}
+		erc721TransferTokenIdB64 = common.Ptr(base64.StdEncoding.EncodeToString(erc721TransferTokenIdBuf))
+	}
```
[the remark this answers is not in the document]
pro-wh: yeah, it is more consistent with the project's style to read that into new columns. let's not do that in this PR though

api/spec/v1.yaml
```diff
@@ -961,6 +961,17 @@ paths:
...
+          schema:
+            type: string
+          description: Get EVM log events from the contract at this Ethereum address.
```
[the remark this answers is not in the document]
pro-wh: why's language gotta be hard lol

api/spec/v1.yaml
pro-wh: correctness:

- the implementation behind this literally only looks at ERC-721 transfers. and although ERC-721 is the only supported standard which would waive us having to specify ERC-721 here, it doesn't look at other event types from ERC-721, including Approval, and debatably, ApprovalForAll.

maintainability:

- if we add an NFT standard other than ERC-721 without broadening this, this name will become wrong

api/spec/v1.yaml
pro-wh: up to ethCommon.HexToAddress and I didn't look :smiling_face_with_tear:

api/spec/v1.yaml
pro-wh: ya

api/spec/v1.yaml
pro-wh: > we never use [_eth suffix] (or the eth_ prefix)

image

pro-wh: I think what I'll do is take oasis format address like everything else and go through address_preimages in the db :skull:

api/spec/v1.yaml
mitjat: Welp this thread was intended as the nudge to look. Anyway. I looked it up now, the 0x is optional. Can you please add that to the description and add an example value.

storage/migrations/26_runtime_events_nft_transfers.up.sql
mitjat: Not sure which part the "hadn't tried" refers to. But really the part I'm most curious about is:
How fast/slow is the query without the index, i.e. as-is?

storage/client/queries/queries.go
```diff
@@ -404,10 +404,29 @@ const (
...
+						context_identifier = 'oasis-runtime-sdk/address: secp256k1eth' AND
+						context_version = 0
+				)
+			)) AND
+			($9::text IS NULL OR (
+			    -- Currently this only supports ERC-721 Transfer events.
+			    evs.type = 'evm.log' AND
+			    evs.evm_log_signature = '\xddf252ad1be2c89b69c2b068fc378daa952ba7f163c4a11628f55a4df523b3ef' AND
+			    jsonb_array_length(body -> 'topics') = 4 AND
+			    body -> 'topics' ->> 3 = $9::text
```
mitjat, line 425: Nit: Mix of spaces and tabs for indentation.

storage/client/queries/queries.go
```diff
@@ -404,10 +404,29 @@ const (
...
+			    -- Currently this only supports EVM smart contracts.
+			    evs.type = 'evm.log' AND
+			    evs.body ->> 'address' = (
+			        SELECT encode(address_data, 'base64')
+			        FROM chain.address_preimages
+			        WHERE
+			            address = $8::text AND
+						context_identifier = 'oasis-runtime-sdk/address: secp256k1eth' AND
+						context_version = 0
+				)
```
mitjat, line 418: An equivalent expression is `encode(eth_preimage($8::oasis_addr), 'base64')`, via [this func](https://github.com/oasisprotocol/nexus/blob/f495dac742ed91fd6e32336dfbc99d00955c9abd/storage/migrations/18_refactor_dead_reckoning.up.sql#L8-L25)

storage/client/client.go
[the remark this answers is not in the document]
mitjat: OK. It will be a fair bit more work to refactor this later than to do it right now I think, especially because of the reindex. But fair enough to push to get this PR (and feature) out.

api/spec/v1.yaml
mitjat: I thought about that option too. It's fairly contrived because now the frontend will have to convert from eth to bech32, only for us to convert (somewhat expensively) back. If you feel like supporting both formats (bech32 and eth-hex), I'm in favor of it; it's the most consistent with analogous params like `tx_hash`.

api/spec/v1.yaml
mitjat: I don't think I get your points above. They sound like summaries of my comment to me :)
- correctness: Agreed. Hence my suggestion to "add to the description" (of the param).
- maintainability: Also agreed, hence my suggestion to rename the param.

Regardless, thank you for the changes. I still think we should enforce at the API level that if `nft_id` is given, then `contract_address` has to be given, and (as you point out) `evm_log_signature` has to be `0xddf252ad1be2c89b69c2b068fc378daa952ba7f163c4a11628f55a4df523b3ef`, for two reasons:
- Explicitness/Correctness. If users provide a different `evm_log_signature` (e.g. for Approval), we won't correctly return those events with the current code; the SQL query hardcodes the Transfer signature.
- **Preventing expensive queries. If people request all events with `ntf_id==0` (with no `contract_address` specified), that will result in a very expensive linear scan of the events.** We can fight that by adding an index (and then continuing to ensure that this wacko query is performant, yuck), or by preventing this type of queries.

storage/client/queries/queries.go
```diff
@@ -404,10 +404,29 @@ const (
...
+			)) AND
+			($9::text IS NULL OR (
+			    -- Currently this only supports ERC-721 Transfer events.
+			    evs.type = 'evm.log' AND
+			    evs.evm_log_signature = '\xddf252ad1be2c89b69c2b068fc378daa952ba7f163c4a11628f55a4df523b3ef' AND
```
mitjat, line 423: If you go with my push to require an explicit evm_log_signature param, this can go away because it will be enforced via $6.

mitjat (commented):
Thanks! I'm gunning for one more feature, see below. The rest are tiny suggestions.

storage/client/queries/queries.go
pro-wh: o thanks

api/spec/v1.yaml
pro-wh: it's `contract_address` now, only taking oasis format, with an example provided

storage/client/queries/queries.go
pro-wh: switched to that

api/spec/v1.yaml
mitjat: OK! Yeah if it's not technically more difficult to support arbitrary event signatures, great, let's do that. :+1: 

I was suggesting we could demand the specific ERC721 Transfer signature _for now_ because that's all that our Explorer needs (it has a dedicated tab or showing past token transfers) and because limiting the scope of what we're able to retrieve efficiently can make the task easier. In particular, I thought we might need to use the commented-out `CREATE INDEX ...WHERE evm_log_signature=0xddf...`.

pro-wh: ok I'll test that on staging and uncomment it if it works

storage/migrations/26_runtime_events_nft_transfers.up.sql
pro-wh: aha. I have been able to observe it being used on staging mainnet

without index:
```
                                                                                                                                                                     QUERY PLAN                                                                                                                                                                      
-----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
 Limit  (cost=7715455.36..7715455.37 rows=1 width=353)
   ->  Sort  (cost=7715455.36..7715455.37 rows=1 width=353)
         Sort Key: round DESC, tx_index, ((body)::text)
         ->  Index Scan Backward using ix_runtime_events_evm_log_signature on runtime_events  (cost=0.57..7715455.35 rows=1 width=353)
               Index Cond: ((runtime = 'sapphire'::runtime) AND (evm_log_signature = '\xddf252ad1be2c89b69c2b068fc378daa952ba7f163c4a11628f55a4df523b3ef'::bytea))
               Filter: ((type = 'evm.log'::text) AND (jsonb_array_length((body -> 'topics'::text)) = 4) AND (((body -> 'topics'::text) ->> 3) = 'AAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAOQ='::text) AND ((body ->> 'address'::text) = encode(eth_preimage(('oasis1qpwe6h0hw6s0x4skujhznld065300p83wgxnvu0n'::text)::oasis_addr), 'base64'::text)))

```

with index:
```
                                                                           QUERY PLAN                                                                            
-----------------------------------------------------------------------------------------------------------------------------------------------------------------
 Limit  (cost=690.69..690.70 rows=1 width=353)
   ->  Sort  (cost=690.69..690.70 rows=1 width=353)
         Sort Key: round DESC, tx_index, ((body)::text)
         ->  Index Scan using ix_runtime_events_nft_transfers on runtime_events  (cost=0.42..690.68 rows=1 width=353)
               Index Cond: ((runtime = 'sapphire'::runtime) AND (((body -> 'topics'::text) ->> 3) = 'AAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAOQ='::text))
               Filter: ((body ->> 'address'::text) = encode(eth_preimage(('oasis1qpwe6h0hw6s0x4skujhznld065300p83wgxnvu0n'::text)::oasis_addr), 'base64'::text))
```

pro-wh:
ok there, tested with the index on a simplified query. fixed up the commented-out query and un-commented it.

it does look like postgresql tries to sort the result though. it's like it doesn't get any benefit from the index being ordered by round. so this may slow down a lot later for NFTs with long transfer histories

api/spec/v1.yaml
pro-wh: wait did I reply to the wrong thread aaaaa

pro-wh: deleted replies to wrong thread

the index is un-commented-out. we're currently not forcing the user to set $6 = (that event signature hash). we can later add `Allow` events similarly if desired

storage/client/client.go
```diff
@@ -1294,6 +1297,21 @@ func (c *StorageClient) RuntimeEvents(ctx context.Context, p apiTypes.GetRuntime
 		h := ethCommon.HexToHash(*p.EvmLogSignature)
 		evmLogSignature = &h
 	}
+	if p.NftId != nil && p.ContractAddress == nil {
+		return nil, fmt.Errorf("must specify contract_address with nft_id")
+	}
```
Andrew7234, line 1302: Can we add a check to make sure that `type`/`evm_log_signature` weren't specified to be anything other than `evm.log`/`erc721Transfer_signature` in the query if `NftId` was provided? Otherwise such a query would return no results because the filters clash, but it wouldn't be obvious to the user why this happened.

```
	if p.NftId != nil && (p.Type != 'evm.log' || evmLogSignature != 'erc 721 signature here') {
	        return nil, fmt.Errorf("nft_id field currently only evm erc-721 transfer events")
	 }
```

mitjat: Because they would want to list e.g. the `Allow` calls for a certain `NftId`.
Or, here's a related not-great scenario: User will not specify any `evm_log_signature` (which seems to be the expected usage right now) and then will be surprised that they only get the Transfer events, not also Allow.

In this [other thread](https://github.com/oasisprotocol/nexus/pull/579#discussion_r1418351008), what I thought I was agreeing to was to "[...] support arbitrary event signatures", which in my understanding was to remove this from SQL: `evs.evm_log_signature = '\xddf252ad1be2c89b69c2b068fc378daa952ba7f163c4a11628f55a4df523b3ef'`.

I still think we should either:
- Remove that thing from SQL (but that might be prohibitively slow, and also semantics are sketchy because we can't be sure that `(body -> 'topics' ->> 3)` is the ntf id), OR
- Require the query to also specify `evm_log_signature`, as Andy is suggesting. We can say "unsupported value" to any value other than `erc721Transfer_signature` for now, and later expand to allowing other values.

If the above+Andy convinces you, great. If not, I've lost the wind in my sails to belabor this relatively minor point further, convinced as I might be. Therefore LGTMing.

mitjat (approved):
Alright this was a long one. Hoping I can win you over on that one last comment thread, but if not, I'm still not going to veto the PR. Thanks for the work+discussions!

storage/client/client.go
pro-wh: let's try adding support for Allow events. you meant events, not calls right? this wasn't meant to limit the API to transfers. we just hadn't implemented it

pro-wh: ugh though wasn't this nft_id field documented as something specifically for ERC-721 transfer events? that's going to be on them for filtering for ERC-721 transfer events that aren't ERC-721 transfer events

pro-wh: omg that's been removed since the comment on it was posted. did anyone ask for that? should I put it back :skull:

storage/client/client.go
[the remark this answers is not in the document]
pro-wh: gonna file an issue and resolve this thread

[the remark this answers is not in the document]
pro-wh: #601

storage/migrations/26_runtime_events_nft_transfers.up.sql
pro-wh: gonna resolve, it seemed to work

storage/client/client.go
pro-wh: ok in conclusion:

- don't wanna
- adding a note to the description though

storage/client/queries/queries.go
pro-wh: choosing not to require evm_log_signature param

api/spec/v1.yaml
pro-wh: reinstating specifying the _current_ limitation to ERC-721 Transfer events

pro-wh: adding a note that users should specify evm_log_signature if they want to get only Transfer, in case we add support for Approval later

PR #582, opened 2023-12-06
Improve documentation of the block hash argument

csillag:
Follow-up to #581

The updated wording of the docs is more in sync with how we describe the other filters.

### DIFF
--- a/api/spec/v1.yaml
+++ b/api/spec/v1.yaml
@@ -212,7 +212,7 @@ paths:
           name: hash
           schema:
             type: string
-          description: The block hash of the block to return.
+          description: A filter on the block hash.
           example: *block_hash_1
       responses:
         '200':
@@ -825,7 +825,7 @@ paths:
           name: hash
           schema:
             type: string
-          description: The block hash of the block to return.
+          description: A filter on the block hash.
           example: *block_hash_1
       responses:
         '200':


api/spec/v1.yaml
```diff
@@ -212,7 +212,11 @@ paths:
           name: hash
           schema:
             type: string
-          description: The block hash of the block to return.
+          description: |
+            The block hash of the block to return.
+            (Please note that if you set this arg, then whatever you set for limit,
+            the response will always contain exactly zero or one block,
+            never any subsequent blocks, unlike the use case when you search using the "from" arg.)
```
mitjat, line 219: I disagree with calling this out here. All filters work that way, not just `hash`: they limit the output to only those objects (txs, events, blocks, whatever) that match the filter. So if anything, the `from` filter is unusual, though I find the wording there reasonably clear too: "A filter on minimum block height, inclusive." It doesn't return "subsequent" blocks; rather all blocks `[N, N+1, N+2, ...]` match the filter `from=N`.

@csillag Do you find either of the following two would increase readability?
- Extend the description of `from` (and analogously `to`, `before`, `after`) to say "All returned blocks will have height that is greater than or equal to this value."
- Replace the description of `hash` with "A filter on the block hash. All returned blocks will have this hash. Because hashes are globally unique, this means the response will contain at most one block."

csillag: > So if anything, the `from` filter is unusual,

I think the `offset` option is the most outstanding one, because it's not a filter at all.

> @csillag Do you find either of the following two would increase readability?

Yes, I think option 2 would help, even if we only use the first sentence. ("A filter on the block hash. ")

csillag: @mitjat I have updated the wording of this PR accordingly. Is it OK now?

mitjat: Thank you! LGTM. It also makes the wording more consistent with the other params/filters, compared to the old docstring :+1:

PR #586, opened 2023-12-14
Andrew7234/abi api

Andrew7234:
After [540](https://github.com/oasisprotocol/nexus/pull/540), we now parse tx data and event data using the abi we have available. The parsed fields are stored in several new columns which we now expose in the api. 

Note: The abi-parsed event data is stored in the existing `evm_log_params` column, which requires no changes.

Note**: This PR includes a breaking change of renaming the `EvmEventParam` type to `EvmAbiParam`, which is more accurate and also necessary because this type is also used by the runtime transaction `evm_fn_params` and `txErr.params`.

```
» curl "http://localhost:8008/v1/sapphire/transactions/92d09287efdc88fff15c6cf63f3405529441ff719d2ea5f98ff52283753eccb4" | jq                andrewlow@Beleriand
100  1780  100  1780    0     0   111k      0 --:--:-- --:--:-- --:--:--  158k
{
  "total_count": 1,
  "transactions": [
    {
      "amount": "0",
      "body": {
        "address": "mYYzvfbuMqnMpsmiR/QoWW6OZdg=",
        "data": "nEW2bgAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAABgAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAKAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAwAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAABAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAIwAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAADAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAACAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAQAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAgAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAADtiYWZ5cmVpZmtybmJoZWhwdXNzY215bW5sdDd5dml0aGppdjRrZnF5ZjRjN3BqcTJ2em9lbnNrNGwzaQAAAAAA",
        "value": ""
      },
      "charged_fee": "11351000000000000",
      "eth_hash": "b087daa98fc02bf7c9357d44305cfc57a11a8900a70bb44f83ffef71fa872b21",
      "evm_fn_name": "acceptTaskResults",
      "evm_fn_params": [
        {
          "evm_type": "uint256[]",
          "name": "_taskIds",
          "value": [
            140
          ]
        },
        {
          "evm_type": "bytes",
          "name": "_proof",
          "value": ""
        },
        {
          "evm_type": "bytes",
          "name": "_report",
          "value": "AAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAACAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAQAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAgAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAADtiYWZ5cmVpZmtybmJoZWhwdXNzY215bW5sdDd5dml0aGppdjRrZnF5ZjRjN3BqcTJ2em9lbnNrNGwzaQAAAAAA"
        }
      ],
      "fee": "1500000000000000000",
      "gas_limit": 15000000,
      "gas_used": 113510,
      "hash": "92d09287efdc88fff15c6cf63f3405529441ff719d2ea5f98ff52283753eccb4",
      "index": 0,
      "method": "evm.Call",
      "nonce_0": 146,
      "round": 410922,
      "sender_0": "oasis1qr5tj6pk8rm447u3k588a7hgkxrzus5vxqq3wsyq",
      "sender_0_eth": "0x7Df2e133FCFF612aA46995E82A54335F716B66be",
      "size": 558,
      "success": true,
      "timestamp": "2023-05-04T15:53:11-07:00",
      "to": "oasis1qpwe6h0hw6s0x4skujhznld065300p83wgxnvu0n",
      "to_eth": "0x998633BDF6eE32A9CcA6c9A247F428596e8e65d8"
    }
  ]
}
```


api/spec/v1.yaml
```diff
@@ -1500,6 +1500,15 @@ components:
...
           description: The message of a failed transaction.
+        params:
+          type: array
+          items:
+            $ref: '#/components/schemas/EvmAbiParam'
+          description: |
+            The error parameters, as decoded using the contract abi.
+            Only present for errors from verified contracts. Note that
+            users should be cautious when evaluating error data since the
+            data origin is not tracked and the error can be faked.
```
mitjat, line 1511: IIRC technically the error is never (except at old heights) just a plain string; `message` is our stringification of a name-and-anonymous-param structure that's something like `Error("this is my error message")`. So in my mind, it doesn't make sense to expose `message` _and_ `params`; `message` is just a human-friendly rendition of `params` (if there are any; or just the plain text instead; and also possibly a non-evm prefix stirng like "reverted:").
Or maybe we want to argue that rendering the message is tricky (especially with the "reverted: " prefix and whatnot) even when structured params are given, and that pre-rendering into `message` is helpful, but it's also helpful for people to be able to access the structured part? I'm doubtful about the latter, but not strongly opposed to it.

The prior behavior on parsing errors, see RawMessage vs Message (= parsed RawMessage) here:
https://github.com/oasisprotocol/nexus/blob/6d23919d10ce55b8446ed3edddbb7ce4c7219545/analyzer/runtime/extract.go#L73-L77
You can see how the latter is derived from the former, and there are also unit tests.

Would it make sense for the backfilling parser to reuse more of the non-backfilling parser code? E.g. this function:
https://github.com/oasisprotocol/nexus/blob/6d23919d10ce55b8446ed3edddbb7ce4c7219545/analyzer/runtime/extract.go#L676-L676
It currently only attempts to parse with a single ABI, `Error(string)`, see line 698. The func would have to made a little more generic so the backfilling parser can pass in a custom ABI. Then we could use it in the backfilling analyzer.

At this point, I think I might be missing something. https://github.com/oasisprotocol/nexus/blob/6d23919d10ce55b8446ed3edddbb7ce4c7219545/analyzer/evmabibackfill/evm_abi_backfill.go#L162-L165 even mentions `tryParseErrorMessage()` (which I'm trying to get us to reuse here) but then does not call it. Maybe we can clarify in a chat or call?

mitjat (commented):
Hey. Looking at this makes me feel like I might have missed some stuff in the precursor PR with the analyzer, #540 . If so, apologies :(. Hopefully it's just been a long day for me and I'm just misunderstanding/misreading stuff in this PR.

storage/migrations/01_runtimes.up.sql
```diff
@@ -56,8 +56,6 @@ CREATE TABLE chain.runtime_transactions
   -- For evm.Call transactions, we store both the name of the function and
   -- the function parameters.
   evm_fn_name TEXT,
-  -- The function parameter values. Refer to the abi to see the parameter
-  -- names. Note that the parameters may be unnamed.
```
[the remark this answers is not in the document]
Andrew7234: Yep, this and the error args are stored as an ordered json array of abiEncodedArg structs that json-serialize to `{name: ..., evm_type: ..., value: ...}`. For unnamed arguments, the name is just an empty string. The ordering of the args is preserved.

Added the `CHECK` and a comment/example!

api/spec/v1.yaml
Andrew7234: > expose message and params; message

Hm the `txErr.Message` field stores the error message for all types of tx errors. Non evm revert errors, such as `execution failed: out of gas`, are still returned as plaintext even in the most recent upgrade, so we can't remove the `message` field and replace it with `params`.

Agreed that the repetition between `message` and `params` for newer evm reverted txs is somewhat messy. We _could_ avoid this by not populating `params` when we specifically see that the abi error has the format `Error(string)`, but imo it's not worth special-casing this.

Edit**: Actually it looks like the `Error(string)` type is not included by default in the abi [due to quirks with evm](https://docs.soliditylang.org/en/latest/control-structures.html#revert). `reverted("my_error")` is automatically encoded as the `Error(string)` type but isn't included in the abi. The backfilling analyzer shouldn't be able to parse such errors since they're not explicitly defined in the abi. Therefore we should continue optimistically trying to parse tx evm reverted errors as `Error(string)`. One caveat: some contracts may explicitly define the type, which would lead to the error message being replicated in `message` and `params`, but we have no control over this and it should be rare.

> Would it make sense for the backfilling parser to reuse more of the non-backfilling parser code?

Yeah the explicit abi logic is a remnant from the days before Warren added the `abiparse` package. We should define a bare-bones `abi` struct with an `Error(string)` error and update `tryParseErrorMessage` to use the new `abiparse.ParseError(err, abi)` function. I'd prefer to do this in a follow-up PR; filed a ticket to track: https://app.clickup.com/t/8693epjgf

api/spec/v1.yaml
```diff
@@ -1500,6 +1500,15 @@ components:
         message:
           type: string
           description: The message of a failed transaction.
+        params:
```
mitjat, line 1503: We probably need another field for the _name_ (type) of the error? Analogous to `evm_fn_name` (which we're missing) and `evm_fn_params`.
It's unfortunate that we're creating so much boileplate and fields for something so small ... but I don't think it's worth taking shortcuts here. By contrast, in the DB, we _could_ (ab)use the first field of `error_params` to store just the plain-string name of the error type. If you go with that (instead of adding yet another column to the table), please be sure to update the docstring.

mitjat (commented):
Getting there!
I only left 1 long comment (that will hopefully just require to read+copy+paste) and 1 short but meaty comment about not tracking the error type. Other than that, LGTM!

api/spec/v1.yaml
Andrew7234: Hm we currently we return the error name in the `TxError.message` field, which is otherwise empty. So a custom error like `InsufficientBalance(threshold uint64)` would be returned as
```
TxError.message          | TxError.params
__________________________________
InsufficentBalance       | [{name: "threshold", type: uint64, value: 100}]
```

However, one quirk is that default revert errors with a format`Error(string)`, e.g., `revert("insufficient balance")` are returned as
```
TxError.message          | TxError.params
__________________________________
insufficient balance     |  
```

whereas if we were strictly consistent with formatting as the other error types, it 'should' look like
```
TxError.message          | TxError.params
__________________________________
Error                             | [{name: "", type: string, value: "insufficient balance"}]
```

Imo the former case is easier to understand, so I'd be okay with keeping it as-is. Also, the `revert("")`/`Error(string)` error types are implicit, which makes me a bit more willing to special case this type. I'd imagine that actual user-defined error types in the abi would have more descriptive error names/types, which would be more appropriate to return in the error_message field than just `Error`.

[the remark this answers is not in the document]
Andrew7234: sounds good, updated!

api/spec/v1.yaml
Andrew7234: Updated, thanks!

> `message` will contain the revert reason, either in its raw ABI-encoded form

We currently do not return the raw abi-encoded message; we only store it in `runtime_transactions.error_message_raw`. We could easily change the api to return the raw abi-encoded message if no parsed message is found, though I'm not sure how useful that would be for most users. Perhaps the main use case would be for contract devs who have the abi and are able to decode it themselves?

api/spec/v1.yaml
Andrew7234: Addendum to the above^:

We should include the `reverted: ` prefix for all abi error names as well. Updated the descriptions accordingly

mitjat: Thank you for fixing the `reverted: ` prefix!

I still think that in an ideal world, the error type (e.g. `InsufficientBalance`) should be stored and reported in a separate field like `revert_error_type`. `revert_error_type` and `revert_params` are the machine-readable version of the revert reason, whereas `message` is the human-readable version.

That said, let's just wrap this up for now and be pragmatic. It's reasonably easy and reliable to extract the type from `message`, and I don't see users needing this precise breakdown very often.

**The one thing I still would like to change** now that I understand the behavior is to make `message` indeed the human-readable version of the full error. Which IMO would be something like `reverted: <error_type>(<arg1_value>, <arg2_value>, ...)`. For the example above, it would be `reverted: InsufficientBalance(100)`. This will be very handy for the Explorer too, because presumably this is the value they'll want to show by default.

LMK if you disagree, otherwise #shipit! :+1:

mitjat (approved):

api/spec/v1.yaml
Andrew7234: Good call, added!

PR #590, opened 2023-12-19
runtime: add explicit PossibleNFT.Burned field

pro-wh:
less use of special empty-string fake Address to signal weird conditions.

fields reordered too, so that the more important NumTransfers, which affects the validity of other fields, comes first


mitjat (approved):
Thx

Andrew7234 (approved):

PR #592, opened 2023-12-21
common/types: BigInt: CBOR-serialize correctly

mitjat:
This PR adds custom (de)serialization for BigInt because it was broken.

The current version of our code CBOR-serializes every `BigInt` instance as `{}`, the empty struct (CBOR: `0xa0`). It was then deserialized as `0`.

BigInt is defined in our code as
```go
type BigInt struct {
	big.Int
}
```

https://github.com/fxamacker/cbor/blob/master/encode.go explains that structs are serialized as maps. The inner `big.Int` is stored in an implicit `BigInt.Int` field, and is public (as evidenced by some sloppy usage in our own code). So IDK why it's not being serialized by cbor, and I'm fairly sure it was serialized correctly in the past.

Regardless, we did not have tests, which this PR also adds.

**Note:** Because the new serialization format is no longer map-based (if it ever was), my hope was that old cached values won't be decoded successfully, and affected cache entries will be automatically invalidated. That's not the case; the `cbor` library decodes `0xa0` (empty map) into the default object automatically. I found a workaround but it's verbose and would require us to special-case 0 when encoding.

The only affected cached value is the return value of `GetNativeBalance()`. I suggest we rename its cache key to "forget" old cache values in the near future. (Thanks @pro-wh for the idea.) I can't do so in this PR because the node is having issues ("Bad Gateway").

### DIFF
--- a/common/types.go
+++ b/common/types.go
@@ -20,6 +20,55 @@ func NewBigInt(v int64) BigInt {
 	return BigInt{*big.NewInt(v)}
 }
 
+func (b BigInt) Plus(other BigInt) BigInt {
+	result := big.Int{}
+	result.Set(&b.Int) // creates a copy of b.Int
+	result.Add(&result, &other.Int)
+	return BigInt{result}
+}
+
+func (b BigInt) Times(other BigInt) BigInt {
+	result := big.Int{}
+	result.Set(&b.Int) // creates a copy of b.Int
+	result.Mul(&result, &other.Int)
+	return BigInt{result}
+}
+
+func (b BigInt) MarshalText() ([]byte, error) {
+	return b.Int.MarshalText()
+}
+
+func (b *BigInt) UnmarshalText(text []byte) error {
+	return b.Int.UnmarshalText(text)
+}
+
+// Used by the `cbor` library.
+func (b BigInt) MarshalBinary() ([]byte, error) {
+	var sign byte = 1
+	if b.Int.Sign() == -1 {
+		sign = 255
+	}
+	return append([]byte{sign}, b.Bytes()...), nil // big.Int only supports serializing the abs value
+}
+
+func (b *BigInt) UnmarshalBinary(data []byte) error {
+	if len(data) == 0 {
+		return errors.New("empty data")
+	}
+	var inner big.Int
+	inner.SetBytes(data[1:])
+	switch data[0] {
+	case 1:
+		// do nothing
+	case 255:
+		inner.Neg(&inner)
+	default:
+		return fmt.Errorf("invalid sign byte: %d", data[0])
+	}
+	b.Int = inner
+	return nil
+}
+
 func (b BigInt) MarshalJSON() ([]byte, error) {
 	t, err := b.MarshalText()
 	if err != nil {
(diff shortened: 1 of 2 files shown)


common/types.go
```diff
@@ -20,6 +20,55 @@ func NewBigInt(v int64) BigInt {
...
+func (b BigInt) MarshalText() ([]byte, error) {
+	return b.Int.MarshalText()
+}
+
+func (b *BigInt) UnmarshalText(text []byte) error {
+	return b.Int.UnmarshalText(text)
+}
+
+// Used by the `cbor` library.
+func (b BigInt) MarshalBinary() ([]byte, error) {
```
ptrus, line 46: Wow indeed weird. It looks like this is due to cbor's internal handling of BigInt marsahaling and not detecting it whenever it's embedded in another struct. 

Would it be safer to use the cbor's library internal marshal/unmarshal methods for big ints? Seem a bit more complex (but didn't look in depth) e.g.:
https://github.com/fxamacker/cbor/blob/a1d4c29efb5eac2ad571787a350fc6b9c46995b1/encode.go#L1179-L1212


One way to achieve this would be to do:
```
type BigInt struct {
	big.Int `cbor:"int"`
}
```
with the added cbor tag it seems to work.

mitjat: Interesting. I don't know that it's safer, but it would be more interoperable/standard :/. Though it'd better to go with a more efficient encoding:
```go
type BigInt struct {
	_   struct{} `cbor:",toarray"`
	Int big.Int
}
```
so we're not encoding the map structure over and over.

FWIW, nothing special is needed for `cbor` to encode the `big.Int` using the special encoder. It's enough even just to make the field explicit:
```go
type BigInt struct {
	Int big.Int
}
```
In this case, it encodes a large number as
```
A1                                      # map(1)
   63                                   # text(3)
      496E74                            # "Int"
   C2                                   # tag(2)
      50                                # bytes(16)
         3FFFFFFFFFFFFFFF0000000000000001 # "?\xFF\xFF\xFF\xFF\xFF\xFF\xFF\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0001"
```
Values smaller than MaxInt64 were roundtripped correctly, but the large numbers in the test errored out during decoding with "cbor: CBOR tag isn't allowed".
This is because nexus (re)uses the oasis-core `cbor` wrapper for max compatibility, and that wrapper disallows tags by default: https://github.com/oasisprotocol/oasis-core/blob/799668903b5a993d0f8af0110c99f2916fb6eb11/go/common/cbor/cbor.go#L43

So our options are:
- Leave PR as-is. Downside: Not a universally understood encoding; you need our BigInt definition to decode.
- Create our own CBOR utility funcs that allow BigInt tags, do not use oasis-core's utilities. Downside: Boilerplate, potential (though unlikely) tiny discrepancies with oasis-core behavior.

I prefer the first option largely because it's less work at this point, and because this will be used internally only. But open to stronger opinions.

kostko: I would leave the PR as-is, the encoding seems similar to `Quantity` (except for the extra sign) and we do not really use CBOR tags elsewhere.

kostko:
Also, this issue should be reported upstream.

ptrus (approved):

mitjat:
Reported https://github.com/fxamacker/cbor/issues/454 upstream

PR #593, opened 2023-12-21
analyzer/runtime: fast-sync: do not update common pool, fee accumulator

mitjat:
During fast-sync, updates to the `runtime_sdk_balances` table often causes deadlocks.

I analyzed the last 10k errors. Of those, almost all are due to only two addresses: 7582x  fee accumulator and 2322x common pool.

**This PR:**
- disables updates to the balance of those two accounts during fast-sync
- expands `FinalizeFastSync()` so it queries the balance of those two accounts

**Testing:**
I cannot test on Damask mainnet (the regular e2e-regression range) because of the "Bad Gateway" errors when attempting to query the balance of the two accounts :/
However I tested manually on Eden, range 7_875_130 to 7_876_129 (1000 blocks); rounds up to and including 7_876_033 were fast-sync. Compared to slow-sync, at "final" round 7_876_129:
 - The fee accumulator has identical balance (2213600000000000)
 - The common pool has a different balance! Slow-sync: 118953099302760004, Fast-sync: 121166699302760033. However I verified with a manual query that the fast-sync derived balance is correct. So somewhere in the 7_875_130 to 7_876_033 range we **perform incorrect/incomplete dead reckoning on the common pool**. Which is noteworthy and we need to follow up on it, but does not invalidate this PR.


storage/client/queries/queries.go
```diff
@@ -421,10 +421,10 @@ const (
 			) AS eth_creation_tx,
 			creation_bytecode,
 			runtime_bytecode,
-			(
+			COALESCE((
 				SELECT gas_for_calling FROM chain.runtime_accounts ra
 				WHERE (ra.runtime = $1) AND (ra.address = $2::text)
-			) AS gas_for_calling,
+			), 0) AS gas_for_calling,
```
mitjat, line 427: unrelated tiny bugfix; it's possible to have a known contract with no associated known account, IIRC because we learned about the contract from sourcify

analyzer/runtime/accounts.go
```diff
@@ -32,6 +34,13 @@ func (m *processor) queueAccountsEvents(batch *storage.QueryBatch, blockData *Bl
 	}
 }
 
+// Accounts whose balance changes in most blocks.
+// We don't update their balance during fast-sync, or else the db sees lots of deadlocks.
+var veryHighTrafficAccounts = []sdkTypes.Address{
+	sdkTypes.NewAddressFromBech32("oasis1qz78phkdan64g040cvqvqpwkplfqf6tj6uwcsh30"), // common pool
```
ptrus, line 40: This is duplicated with https://github.com/oasisprotocol/nexus/blob/a52b646c72e723adab383cac81f16b91ae7e6782/analyzer/runtime/runtime.go#L99-L100

Andrew7234 (approved):
looks good

analyzer/runtime/accounts.go
mitjat: Went ahead and defined them in a new central location. I don't think it's a win for readability though :/, and the "true" definition is in the runtimes anyway. Oh well. Bit of a bikeshed.

tests/e2e_regression/expected/db__sdk_balances.csv
```diff
@@ -80,7 +80,7 @@ emerald,oasis1qz629557xs0j6vk9ulwk4k3mjkdpvfncrs5h0zg8,ROSE,8591305825000000000
 emerald,oasis1qz670t637yyxshnlxhjj5074wgwl94d0x5x69zqd,ROSE,34225830891942957224
 emerald,oasis1qz6m48jljzlpdz52prgfhrppnz0ffshlt5j7kjaz,ROSE,2906029534421611553
 emerald,oasis1qz72lvk2jchk0fjrz7u2swpazj3t5p0edsdv7sf8,ROSE,411567692021454489
-emerald,oasis1qz78phkdan64g040cvqvqpwkplfqf6tj6uwcsh30,ROSE,131
+emerald,oasis1qz78phkdan64g040cvqvqpwkplfqf6tj6uwcsh30,ROSE,121166699302776158
```
mitjat, line 83: **NOTE:** This is the fee accumulator addr. **The diff is expected** because the previously, slow-sync (under which regime the whole e2e regression runs) started with the (known incorrect) assumption that runtime balances are all 0. Now, the `finalizeFastSync()` method runs first (even though fast-sync itself doesnt; that method is really more of a `preparePreconditionsForSlowSync()`) and pulls the initial balance from the node.

(*) Although in

PR #594, opened 2023-12-22
Revert "analyzer/block: more efficient db query on startup"

pro-wh:
this query for enqueuing gaps seems to be crashing the database.

shall we revert to the slow and steady one?


mitjat (approved):
Thank you, SGTM, it's not even that much slower.
But since we don't know if it really is the `EXCEPT` that's killing the DB, I suggest deploying this fix from your branch first to verify that it actually helps.

pro-wh:
let me try running this actual query and seeing if the staging db crashes :hand_over_mouth:

pro-wh:
tested on staging. db survived

PR #595, opened 2023-12-22
runtime: upsert evm contract creation info

pro-wh:
in fast sync, a later call to the contract can get it marked as a contract candidate, and the bytecode analyzer may upsert it into the EVM contracts db before we get to analyzing its creation.


mitjat (approved):
Nice, thank you!

PR #596, opened 2023-12-23
e2e-regression: Use block range in Eden

mitjat:
This PR moves the target range of the e2e regression test: We now query a subrange in Eden.

I have manually changed many of the target URLs in `run.sh` so that they pull data that is actually indexed in this limited set of blocks.

Benefits of moving to Eden:
- We are more likely to notice any mishandling of the data as presented by the current (= most relevant) version of the node
- Internal deployment of Damask node is currently semi-broken, preventing any changes to the e2e-regression outputs. Moving away from Damask lets us make more changes to nexus without disabling the e2e-regression tests.

Downsides of moving to Eden:
- Emerald blocks (at least the ones I tried) have less activity overall, so outputs are sparser and less likely to catch bugs.


pro-wh (approved):
interesting that some changes between chunked and content-length-known response headers

PR #598, opened 2024-01-04
tests: require.Nil -> require.NoError

ptrus:
Use `{require,assert}.NoError` instead of `.Nil` in all test cases where error existence is checked. This makes the output more readable, which is especially useful in the statecheck tests:

```
    consensus_test.go:168:
                Error Trace:    /nexus/tests/statecheck/consensus_test.go:168
                Error:          Expected nil, but got: &status.Error{s:(*status.Status)(0xc000307448)}
                Test:           TestGenesisFull
```
vs
```
    consensus_test.go:164:
                Error Trace:    /nexus/tests/statecheck/consensus_test.go:164
                Error:          Received unexpected error:
                                rpc error: code = Unimplemented desc = Only some methods are allowed on this proxy.
                Test:           TestGenesisFull
```


pro-wh (approved):
I'm in favor of this

PR #600, opened 2024-01-11
tests/statecheck/runtime: Report the number of balance discrepancies

ptrus:
Useful for quickly seeing the number of account discrepancies in a statecheck run.


pro-wh (approved):
didn't even realize assert returns normally when it fails

mitjat (approved):
nice, thank you!

PR #602, opened 2024-01-12
runtime-transactions: fix calculated `charged_fee` for non-evm txs

ptrus:
For non-evm transactions there are no fee refunds as there are for EVM transactions (where this is done to ensure EVM compatibility).

### DIFF
--- a/api/spec/v1.yaml
+++ b/api/spec/v1.yaml
@@ -2434,7 +2434,8 @@ components:
           description: |
             The fee that was charged for the transaction execution (total, native denomination,
             ParaTime base units, as a string).
-            Calculated as `gas_price * gas_used`, where `gas_price = fee / gas_limit`.
+            For EVM transactions this is calculated as `gas_price * gas_used`, where `gas_price = fee / gas_limit`, for compatibility with Ethereum.
+            For other transactions this equals to `fee`.
         size:
           type: integer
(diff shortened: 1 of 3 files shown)


api/spec/v1.yaml
```diff
@@ -2434,7 +2434,8 @@ components:
           description: |
             The fee that was charged for the transaction execution (total, native denomination,
             ParaTime base units, as a string).
-            Calculated as `gas_price * gas_used`, where `gas_price = fee / gas_limit`.
+            For EVM transactions this is calculated as `gas_price * gas_used`, where `gas_price = fee / gas_limit`.
```
mitjat, line 2437:
```suggestion
            For EVM transactions this is calculated as `gas_price * gas_used`, where `gas_price = fee / gas_limit`, for compatibility with Ethereum.
```

mitjat (approved):
It's a beaut. Thank you for digging into this!

pro-wh (approved):
woa do we really do this in the client query area

mitjat:
> woa do we really do this in the client query area

Haha I was similarly taken aback. We have similarly misplaced (= not abstracted into a special layer) transformations in a few other places too, but I think it's tolerable (if not great) since these kinds of transformations are rare and lightweight in nexus for now.

PR #603, opened 2024-01-15
statecheck: Add support for sapphire statecheck

ptrus:


mitjat (approved):
Thanks! 
For my understanding: Did this require any additional changes, other than on the ops side (special node URL)?

ptrus:
Just the charts change for the spcial node URL. 

(i am still running the modified version that avoids the big queries though, although the non modified version would likely work here due to the much lower number of accounts).

PR #607, opened 2024-01-19
active_accounts query optimization

Andrew7234:
@lukaw3d reported that the latest transaction volume stats weren't up to date:

https://nexus.oasis.io/v1/emerald/blocks
```
"timestamp":"2024-01-19T17:30:48Z"
```
https://nexus.oasis.io/v1/emerald/stats/tx_volume?window_size_seconds=86400&window_step_seconds=86400&limit=30&offset=0
```
"window_end":"2023-11-26T00:00:00Z"
```

Upon closer inspection, the root cause was that the aggregate stats analyzer query to calculate `active_accounts` was taking a long time and blocking the calculation of other stats, including the transaction volume. This PR makes the queries more performant by 1) adding an index for consensus, and 2) rewriting the runtime active accounts query per @mitjat 's suggestion so that postgres now uses the existing index on the `runtime_related_transactions` table.

The index has been added to {mainnet, testnet} x {staging, production} manually, and the existing migrations were edited directly.

Minor change: renames an index to `ix_accounts_related_transactions_address` for accuracy.

Todo: Consider adding a generous timeout to the query context so that 1) queries will not block each other, 2) timeout errors will show up in the logs. Edit*: done

### DIFF
--- a/analyzer/aggregate_stats/aggregate_stats.go
+++ b/analyzer/aggregate_stats/aggregate_stats.go
@@ -23,10 +23,13 @@ const (
 	// This limits database batch sizes in cases if the stats analyzer would be far behind
 	// latest indexed block (e.g. if it would not be enabled during initial sync).
 	// Likely never the case in normal operation.
-	statsBatchLimit = 1_000
+	statsBatchLimit = 100
 
 	// Name of the consensus layer.
 	layerConsensus = "consensus"
+
+	// Timeout when processing a single batch of a single statistic.
+	statsComputationTimeout = 3 * time.Minute
 
 	// Interval between stat computation iterations.
 	// The workers only computes new data and iteration with no updates are cheap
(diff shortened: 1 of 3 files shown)


pro-wh (approved):
as discussed, thx

mitjat (approved):
Thank you, looks good!

As discussed on Slack, IMO it's worth expanding this PR with your suggested TODO (= adding a timeout for the db query). I wouldn't even be particularly generous. Maybe 3s, since the current queries take 1.5s. Maybe even just 2s. In general we want them to be pretty fast because we need to recompute all stats after a reindex. For 3 years of data and a granularity of 5min, that's `3*365*24*12` data points (times 3 statistics, for now). Which is 1e6 queries, or 260h if the average query takes 1s.

Andrew7234:
Added the timeout with a slight tweak. Instead of having a per-query timeout, we have a timeout over a batch of a single computation. This gives us some more flexibility to have expensive stat queries over longer time ranges, e.g. monthly active accounts. 

This timeout is set to 3 minutes, and the batch size has been reduced to 100. This means that for each stat computation, the analyzer will run the query as many times as it can until 1) there are no more windows to compute, 2) the 3 minute timeout is hit, 3) the batch has been filled. The successfully computed queries are then applied to the db.

Note: the current parameters allow for an average query time of 180 sec / 100 = 1.8sec per query. If any queries are slower, we will see timeout errors in the logs (assuming we are in catch-up mode).

PR #610, opened 2024-01-23
bugfix: abi analyzer no longer overwrites error_message on failure

Andrew7234:
The `evm_abi_backfill` analyzer incorrectly overwrites the `runtime_transactions.error_message` field to the empty string if the analyzer is unable to parse the error message. This PR simplifies/fixes the logic in the analyzer.

### DIFF
--- a/analyzer/evmabibackfill/evm_abi_backfill.go
+++ b/analyzer/evmabibackfill/evm_abi_backfill.go
@@ -198,6 +198,7 @@ func (p *processor) ProcessItem(ctx context.Context, batch *storage.QueryBatch, 
 			abiEvent.ID,
 		)
 	} else if item.Tx != nil {
+		// Parse fn call.
 		method, abiTxArgs, err := abiparse.ParseData(item.Tx.TxData, &contractAbi)
 		if err != nil {
 			queueIncompatibleTxUpdate(batch, p.runtime, item.Tx.TxHash)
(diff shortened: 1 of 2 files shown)


analyzer/evmabibackfill/evm_abi_backfill.go
```diff
@@ -210,52 +211,46 @@ func (p *processor) ProcessItem(ctx context.Context, batch *storage.QueryBatch,
...
-					method.RawName,
-					txArgs,
-					nil, // error name
-					nil, // error args
-				)
+				// not parse the error with the abi.
+				//
+				// Note: Here and below, we do not need to call `queueIncompatibleTxUpdate`
+				// because the RuntimeTransactionEvmParsedFnUpdate query above will
+				// mark the transaction as processed.
```
mitjat, line 231: (This whole comment is informational only, now that I've wrapped my mind around the diff. I think the code is fine as-is; this analyzer is not that large or complex.)

FWIW, I find the new flow harder to follow because there's global-ish state one has to keep in mind (= whether the tx is already marked as processed). This comment is useful (and thank you for it!), but it's even better if we don't need it in the first place.

I'd probably structure it as follows:
1) During the parsing step, only convert `item` into a set of variables (basically lines 213-216 in the red diff). However, make all those variables nilable. If a field cannot be parsed into, set it to nil. Do not generate DB queries (to avoid global state), do not return (to allow the fn parser and err parser to both run).
2) Then update the DB, but not when the variable is nil. Sth like `UPDATE ... SET error_message = COALESCE($6, error_message), ...`.

This does have the downside that we cannot actively override any fields with a `NULL`. But that's not ever needed IIUC.

mitjat (approved):
Thank you!

A side effect of the fact that we don't have a separate db column for the error name (type) and we instead abuse the error_message column :grimacing: . Wish I had pushed for it harder in the original PR :)

I also feel quite sheepish about the fact that we don't have a way to enshrine this fixed behavior in a test. Life was simpler when we had fewer item analyzers; all these code-verification-related analyzers interact with each other in so many ways :/

analyzer/evmabibackfill/evm_abi_backfill.go
Andrew7234: Oh I kind of like that, filed https://github.com/oasisprotocol/nexus/pull/613 as an alternative. LMK what you prefer!

mitjat: :star_struck: it looks nicer than I thought :), I'd say let's go with that!

Andrew7234:
Closing in favor of https://github.com/oasisprotocol/nexus/pull/613

PR #612, opened 2024-01-23
api: enum proposal states

lukaw3d:
Based on https://github.com/oasisprotocol/nexus/blob/dc95e6bcf9e75dee797c7bf17b91f67439ff940a/coreapi/v23.0/governance/api/proposal.go#L22-L25


ptrus (approved):

mitjat (approved):
Thank you. I extracted the enum into an explicitly named type and also used it on the query side (where you can specify which proposals you want to retrieve).

PR #613, opened 2024-01-23
use coalesce for backfill analyzer queries

Andrew7234:
Alternative to #610

### DIFF
--- a/analyzer/evmabibackfill/evm_abi_backfill.go
+++ b/analyzer/evmabibackfill/evm_abi_backfill.go
@@ -9,6 +9,7 @@ import (
 	"strings"
 
 	"github.com/ethereum/go-ethereum/accounts/abi"
+	ethCommon "github.com/ethereum/go-ethereum/common"
 	sdkEVM "github.com/oasisprotocol/oasis-sdk/client-sdk/go/modules/evm"
 
 	"github.com/oasisprotocol/nexus/analyzer"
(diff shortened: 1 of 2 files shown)


analyzer/evmabibackfill/evm_abi_backfill.go
```diff
@@ -167,134 +168,102 @@ func cleanTxRevertReason(raw string) ([]byte, error) {
...
+
 func (p *processor) ProcessItem(ctx context.Context, batch *storage.QueryBatch, item *abiEncodedItem) error {
 	// Unmarshal abi
 	contractAbi, err := abi.JSON(bytes.NewReader(item.Abi))
 	if err != nil {
 		return fmt.Errorf("error unmarshalling abi: %w", err)
 	}
 	// Parse data
+	p.logger.Debug("processing item using abi", "contract_address", item.ContractAddr)
 	if item.Event != nil { //nolint:nestif // has complex nested blocks (complexity: 12) (nestif)
```
mitjat, line 239: ohhh you know, as I was writing the comment in #610 I was thinking "and I wonder if the parsing decomposes into helper funcs really nicely, or if there's shared state I'm missing". Glad it works out and you factored these out, looks sleek!

And the linter says we don't need the `nolint` any more ;)

analyzer/evmabibackfill/evm_abi_backfill.go
```diff
@@ -167,134 +168,102 @@ func cleanTxRevertReason(raw string) ([]byte, error) {
...
+	txArgs, err := marshalArgs(method.Inputs, abiTxArgs)
+	if err != nil {
+		p.logger.Warn("error processing tx args using abi", "err", err)
+		return nil, nil
+	}
+
+	return &method.RawName, txArgs
+}
+
+func (p *processor) ProcessTxErr(tx *abiEncodedTx, contractAbi abi.ABI) (*string, []*abiEncodedArg) {
```
mitjat, line 201: nit: can be private (= lowercase), right?

analyzer/evmabibackfill/evm_abi_backfill.go
```diff
@@ -167,134 +168,102 @@ func cleanTxRevertReason(raw string) ([]byte, error) {
 	return base64.StdEncoding.DecodeString(s)
 }
 
+func (p *processor) processEvent(ev *abiEncodedEvent, contractAbi abi.ABI) (*string, []*abiEncodedArg, *ethCommon.Hash) {
```
mitjat, line 171: For this and the next two funcs: Can you document the return values (the types are not very descriptive), or use named return params?

Optional: My instinct is to name these `parseX` rather than `processX`, as I think we normally use `processX()` for "parse + enqueue db statements". But either works; your call.

mitjat (approved):
Nice

analyzer/evmabibackfill/evm_abi_backfill.go
Andrew7234: Thanks! Added docstrings, and also renamed the methods

PR #616, opened 2024-01-26
 api: add roothash messages

pro-wh:
/consensus/roothash_messages

### DIFF
--- a/api/spec/v1.yaml
+++ b/api/spec/v1.yaml
@@ -88,6 +88,8 @@ x-examples:
     - &epoch_2 8048966
   event-type:
     - &event_type_1 'staking.escrow.take'
+  roothash-message-type:
+    - &roothash_message_type_1 'staking.transfer'
   entity-id:
     - &entity_id_1 'gb8SHLeDc69Elk7OTfqhtVgE2sqxrBCDQI84xKR+Bjg='
   node-id:
@@ -372,7 +374,7 @@ paths:
         - in: query
           name: type
           schema:
-            type: string
+            allOf: [$ref: '#/components/schemas/RoothashMessageType']
         - in: query
           name: rel
           schema:
@@ -1670,6 +1672,18 @@ components:
       description: |
         An event emitted by the consensus layer.
 
+    RoothashMessageType:
+      type: string
+      enum:
+        - staking.transfer
+        - staking.withdraw
+        - staking.add_escrow
+        - staking.reclaim_escrow
+        - registry.update_runtime
+        - governance.cast_vote
+        - governance.submit_proposal
+      example: *roothash_message_type_1
+
     RoothashMessageList:
       allOf:
         - $ref: '#/components/schemas/List'
@@ -1694,7 +1708,7 @@ components:
           type: integer
           format: int32
         type:
-          type: string
+          allOf: [$ref: '#/components/schemas/RoothashMessageType']
         body:
           type: object
         error_module:
--- a/api/spec/v1.yaml
+++ b/api/spec/v1.yaml
@@ -364,6 +364,10 @@ paths:
         - *offset
         - in: query
           name: runtime
+          # There's only an index on (runtime, round) for now. Feel free to
+          # index on consensus height and make this parameter optional if we
+          # want to get recent messages from any indexed runtime.
+          required: true
           schema:
             allOf: [$ref: '#/components/schemas/Runtime']
         - in: query
--- a/api/spec/v1.yaml
+++ b/api/spec/v1.yaml
@@ -1701,22 +1701,67 @@ components:
       properties:
         runtime:
           type: string
+          description: |
+            The runtime that sent this message.
+          example: sapphire
         round:
           type: integer
           format: int64
+          description: |
+            The block round when the runtime sent this message.
+          example: 1357490
         index:
           type: integer
           format: int32
+          description: |
+            The 0-based index of this message in the block.
+          example: 0
         type:
           allOf: [$ref: '#/components/schemas/RoothashMessageType']
+          description: |
+            The type of thies message.
+          example: staking.withdraw
         body:
           type: object
+          description: |
+            The "body" of a message is a structure within the
+            `github.com/oasisprotocol/oasis-core/go/roothash/api/message`
+            `Message` structure
+            (https://pkg.go.dev/github.com/oasisprotocol/oasis-core/go/roothash/api/message#Message).
+            For example, if the type is `staking.withdraw`, the body is the Go
+            `Message` structure's `.Staking.Withdraw` field, which is a
+            `github.com/oasisprotocol/oasis-core/go/staking/api` `Withdraw`
+            structure
+            (https://pkg.go.dev/github.com/oasisprotocol/oasis-core/go/staking/api#Withdraw),
+            with `from` and `amount` fields in JSON.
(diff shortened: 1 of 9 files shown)


storage/client/queries/queries.go
```diff
@@ -92,6 +92,26 @@ const (
...
+			body,
+			error_module,
+			error_code,
+			result
+		FROM chain.roothash_messages
+		WHERE
+			($1::runtime IS NULL OR runtime = $1) AND
+			($2::bigint IS NULL OR round = $2) AND
+			($3::text IS NULL OR type = $3) AND
+			($4::oasis_addr IS NULL OR related_accounts @> ARRAY[$4])
```
pro-wh, line 110: needs index

storage/client/client.go
```diff
@@ -477,6 +477,57 @@ func (c *StorageClient) Events(ctx context.Context, p apiTypes.GetConsensusEvent
 	return &es, nil
 }
 
+func (c *StorageClient) RoothashMessages(ctx context.Context, p apiTypes.GetConsensusRoothashMessagesParams) (*apiTypes.RoothashMessageList, error) {
+	res, err := c.withTotalCount(
+		ctx,
+		queries.RoothashMessages,
+		p.Runtime,
+		p.Round,
```
pro-wh, line 485: should check that runtime filter is set

api/spec/v1.yaml
```diff
@@ -392,6 +392,36 @@ paths:
...
+            $ref: '#/components/schemas/Runtime'
+        - in: query
+          name: round
+          schema:
+            type: integer
+            format: int64
+        - in: query
+          name: type
+          schema:
+            type: string
```
mitjat, line 412: would be good to have an enum

storage/client/queries/queries.go
pro-wh: added

storage/client/client.go
pro-wh: added a check

api/spec/v1.yaml
pro-wh: k added

storage/client/client.go
```diff
@@ -491,6 +491,59 @@ func (c *StorageClient) Events(ctx context.Context, p apiTypes.GetConsensusEvent
...
+		var resultCBOR *[]byte
+		if err := res.rows.Scan(
+			&m.Runtime,
+			&m.Round,
+			&m.Index,
+			&m.Type,
+			&m.Body,
+			&m.ErrorModule,
+			&m.ErrorCode,
+			&resultCBOR,
```
Andrew7234, line 530: would it be better to store the result as a json object in the db during analysis time?

pro-wh: we need the message type to do the conversion, and we get the type and result at different times during analysis, potentially even in reverse order with fast sync. it would be nice, but it's kinda weird to implement.

api/spec/v1.yaml
```diff
@@ -355,6 +357,36 @@ paths:
                 $ref: '#/components/schemas/ConsensusEventList'
         <<: *common_error_responses
 
+  /consensus/roothash_messages:
+    get:
+      parameters:
+        - *limit
+        - *offset
+        - in: query
+          name: runtime
```
mitjat, line 366: This is presumably a required parameter? (Edit: I see now that it is, except if `round` is omitted. We should mention this in arg descriptions.)

api/spec/v1.yaml
```diff
@@ -1638,6 +1670,52 @@ components:
...
+        round:
+          type: integer
+          format: int64
+        index:
+          type: integer
+          format: int32
+        type:
+          allOf: [$ref: '#/components/schemas/RoothashMessageType']
+        body:
+          type: object
```
mitjat, line 1711: Can we add a pointer to the relevant structures in oasis-core here? So people have a semi-formal way of anticipating the expected data structures, even if we give no strict guarantees about the shape of this `object`. Ditto for `result`.

api/spec/v1.yaml
```diff
@@ -1638,6 +1670,52 @@ components:
...
+          properties:
+            roothash_messages:
+              type: array
+              items:
+                allOf: [$ref: '#/components/schemas/RoothashMessage']
+
+    RoothashMessage:
+      type: object
+      required: [runtime, round, index]
+      properties:
```
mitjat, line 1699: Each of the properties needs a `description` and ideally an `example`.

api/spec/v1.yaml
```diff
@@ -355,6 +357,36 @@ paths:
                 $ref: '#/components/schemas/ConsensusEventList'
         <<: *common_error_responses
 
+  /consensus/roothash_messages:
```
mitjat, line 360: Would it make sense to move this endpoint under `/{runtime}/roothash_messages`? It would better reflect the fact that `runtime` is pretty much a required arg (see also thread below), and it would probably align better with how it will be used in Explorer.

I know that the runtime is not where messages live architecturally in oasis-core, but I think the implied expectation is that we'll be showing these on the block page for each _runtime_ block? If we'll be showing these somewhere on the consensus part of the Explorer, do we need the ability to filter this by consensus height? (Specifying consensus_height=H would probably mean "give me roothash messages of runtime rounds that were finalized at H")

mitjat (commented):
Thank you! My only real question is about the external API.

api/spec/v1.yaml
pro-wh: huh maybe we should just make it required. in practice you can't get recent messages from all runtimes because there's no index for that

api/spec/v1.yaml
pro-wh: not sure. maybe, if we go down the route of requiring the 'runtime' filter. and agreed on the weirdness of having a consensus layer thing not in /consensus/.

it would be good to see it on the consensus side because these are so much like consensus transactions. but it'll take some thought on where to put them and how we'll want to query them

mitjat: > these are so much like consensus transactions

Mmm that's a strong point. And the runtime is like the tx sender. So this dilemma is analogous to having `/transactions?sender=X` rather than `/accounts/X/transactions`, and we do do the former. Let's keep it as-is then 👍 .

api/spec/v1.yaml
mitjat: yea I think postgres is not smart enough to scan n_runtimes indexes and merge the results. Let's make this required for now, and we can relax it later if we find a use case?

pro-wh: ok requiring

api/spec/v1.yaml
pro-wh: k keeping it here

api/spec/v1.yaml
pro-wh: leme see if we have a place that documents these

pro-wh: https://pkg.go.dev/github.com/oasisprotocol/oasis-core/go/roothash/api/message#Message ah yeah

api/spec/v1.yaml
pro-wh: adding :bicyclist:

Andrew7234 (approved):

pro-wh:
thanks for reviewing, gonna merge

PR #618, opened 2024-01-27
Andrew7234/account stats

Andrew7234:
@lukaw3d noticed that a query to the runtime accounts endpoint for the fee accumulator address (https://nexus.oasis.io/v1/sapphire/accounts/oasis1qp3r8hgsnphajmfzfuaa8fhjag7e0yt35cjxq0u4) was timing out. Further investigation showed that the db operation to calculate the account statistics for `total_sent` and `total_received` were timing out because they were calculated on the fly and the fee accumulator account was involved in ~30% of all transfers. The query optimizer was therefore doing a full table scan on `chain.runtime_transfers` instead of using the existing (flawed) index.

This PR includes changes to dead-reckon the `total_sent` and `total_received` amounts in the `chain.runtime_accounts` table. It also fixes the existing indexes on the `chain.runtime_transfers` table.


storage/migrations/07_runtime_accounts_stats.up.sql
```diff
@@ -0,0 +1,13 @@
+BEGIN;
+
+DROP INDEX IF EXISTS chain.ix_runtime_transfers_sender;
+DROP INDEX IF EXISTS chain.ix_runtime_transfers_receiver;
+
+CREATE INDEX ix_runtime_transfers_sender ON chain.runtime_transfers(runtime, sender);
```
ptrus, line 6: If we are adding the columns to runtime accounts are these indices needed?

Do the `*Recompute` methods end up using these indices?

analyzer/runtime/runtime.go
```diff
@@ -135,6 +135,12 @@ func (m *processor) FinalizeFastSync(ctx context.Context, lastFastSyncHeight int
 	m.logger.Info("recomputing number of txs for every account")
 	batch.Queue(queries.RuntimeAccountNumTxsRecompute, m.runtime, lastFastSyncHeight)
 
+	m.logger.Info("recomputing total_sent for every account")
+	batch.Queue(queries.RuntimeAccountTotalSentRecompute, m.runtime, lastFastSyncHeight)
```
ptrus, line 139: Did you maybe test how long do these two added queries take on a live db? 

Since these are run only once after fast-sync completes it doesn't really matter all that much, but would be interesting to know the rough running time.

ptrus (commented):
Looks good to me, couple of minor questions

analyzer/runtime/runtime.go
Andrew7234: Each takes ~10-15 seconds on sapphire staging mainnet, though I'd expect them to take longer on emerald.

storage/migrations/07_runtime_accounts_stats.up.sql
Andrew7234: Good catch - I thought the recompute queries would use the indices but turns out they don't. Removed them entirely.

storage/client/client.go
```diff
@@ -1504,15 +1504,24 @@ func (c *StorageClient) RuntimeAccount(ctx context.Context, address staking.Addr
...
-	a.Stats.TotalReceived, err = common.NumericToBigInt(totalReceived)
-	if err != nil {
+	)
+
+	switch err {
+	case nil:
+		a.Stats.TotalSent, err = common.NumericToBigInt(totalSent)
+		if err != nil {
+			return nil, wrapError(err)
+		}
```
mitjat, line 1514: Is this needed? The idea is that we should be able to read directly into `a.Stats.TotalSent`, and the conversion will happen automatically because we implement the `NumericDecoder` pgx interface: https://github.com/oasisprotocol/nexus/blob/f6af89d1101fd6c983ade51755a44dee0b6d0597/common/types.go#L112-L121

storage/client/client.go
```diff
@@ -1504,15 +1504,24 @@ func (c *StorageClient) RuntimeAccount(ctx context.Context, address staking.Addr
...
+		a.Stats.TotalReceived, err = common.NumericToBigInt(totalReceived)
+		if err != nil {
+			return nil, wrapError(err)
+		}
+	case pgx.ErrNoRows:
+		// If an account address has no activity, default to 0.
+		a.Stats.TotalSent = common.NewBigInt(0)
+		a.Stats.TotalReceived = common.NewBigInt(0)
+		a.Stats.NumTxns = 0
```
mitjat, line 1523: Your call on whether it's actually better, but you _could_ apply defaults in SQL:
```sql
SELECT
  COALESCE(computed.total_sent, 0),
  COALESCE(computed.total_received, 0),
  COALESCE(computed.num_txs, 0)
FROM
(VALUES(NULL)) AS dummy -- even if SELECT below matches no accounts, ensures that there will be a row to return
LEFT JOIN (
	SELECT total_sent, total_received, num_txs
	FROM chain.runtime_accounts
	WHERE
		(runtime = $1) AND
		(address = $2::text)
) AS computed ON true;
```

mitjat (approved):
Nice, thank you! Nits only from my side.

storage/client/client.go
Andrew7234: Ah you are right, it's unnecessary. (The comment on line 679 points out that we only need to do this conversion when reading into a pointer). Updated!

storage/client/client.go
Andrew7234: Hm I'd prefer to do it here and keep all the logic in one file.

storage/client/client.go
mitjat: Thanks. FWIW, that comment is probably outdated. It refers to `big.Int`, whereas we now use `BigInt`, and one of the reasons for introducing a custom type was that we can support pgx serde.

PR #619, opened 2024-01-29
storage/oasis: Cache init failure is not longer fatal

mitjat:
When the node RPC cache (backed by a `pogreb` db) fails to initialize, our current code prevents the entire analyzer from starting. This PR makes it so that a failure to initialize the cache merely reports a warning, and continues without the cache.

Testing: Manual.
- Ran the analyzer with a functioning cache; works.
- Ran the analyzer with `cache_dir: /does-not-exist`, produced a warning and continued as expected.

### DIFF
--- a/storage/oasis/client.go
+++ b/storage/oasis/client.go
@@ -9,6 +9,7 @@ import (
 
 	"github.com/oasisprotocol/nexus/common"
 	"github.com/oasisprotocol/nexus/config"
+	"github.com/oasisprotocol/nexus/log"
 	"github.com/oasisprotocol/nexus/storage/oasis/nodeapi"
 	"github.com/oasisprotocol/nexus/storage/oasis/nodeapi/file"
 	"github.com/oasisprotocol/nexus/storage/oasis/nodeapi/history"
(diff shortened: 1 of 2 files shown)


mitjat:
Aborting this PR. As discussed internally, I'd prefer to keep cache failure handling explicit (= one has to disable cache in the config) so we are more aware of caching issues. The team agrees.

Mitigations to the root cause will follow in future PRs:
The problematic situation is when
a) the pogreb store needs reindexing and
b) the sapphire or emerald nodes are unavailable.

Because of (b), nexus initialization keeps failing in a fatal way. (Nexus is lazy about connecting to the node, but not for the runtimes because it uses the SDK in addition to the raw RPC connection.) But by the time that fatal error occurs, pogreb has already started the recovery process, including .bac creation. Then k8s keeps restarting nexus. At this point, if the runtime node became available, all would be good.But it typically stays unavailable or a while, so the .bac files pile up, until eventually the pogreb recovery fails (because it cannot create backups) and nexus terminates before it even tries to connect to the runtime node. At that point, every restart will fail, so  when the runtime node eventually comes back up, nexus never learns about it.

We should ideally stop using the SDK, or change it so it can connect lazily. We should also pre-clean-up the .bac files to prevent pogreb from failing just due to filename buildup.

PR #620, opened 2024-02-01
pogreb: Remove .bac.bac files manually

mitjat:
This PR adds manual management of pogreb backups to work around a problematic situation when
a) the pogreb store needs reindexing and
b) the sapphire or emerald nodes are unavailable.

Because of (b), nexus initialization keeps failing in a fatal way. But by the time that fatal error occurs, pogreb has already started the recovery process, including .bac creation. Then k8s keeps restarting nexus, so the .bac files pile up, until eventually the pogreb recovery fails (because it cannot create backups) and nexus terminates before it even tries to connect to the runtime node. At that point, every restart will fail, so when the runtime node eventually comes back up, nexus never learns about it.

With this PR, nexus cleans up the backups itself using the following cleanup procedure: If a backup is indicated, per pogreb logic(*), then nexus first:
- creates a backup directory
- if the backup directory is empty, moves all files except .psg and lock into it, otherwise leaves them to be deleted (see below). The assumption is that the "original" backups are in the backup dir, and should remain there until manually cleared.
- hands over to pogreb. pogreb will create 0 backup files (since there is nothing to back up) and reindex.

The logic is: If the `lock` file exists, OR if checksum etc checks fail. The first one is the only one we've seen in practice, and we can check for it easily. The second one, let's ignore, because it's hard to reimplement. It does mean that the .bac.bac.bac.bac.... situation is still possible. So nexus also deletes all *.bac.bac files.

**Testing:** Manual only. Created a `lock` file in the cache dir, started nexus twice, observed that the dir gets backed up as expected. It's low impact if this cleanup feature fails, so I don't think it's worth investing in proper tests.

### DIFF
--- a/storage/oasis/nodeapi/file/kvstore.go
+++ b/storage/oasis/nodeapi/file/kvstore.go
@@ -2,6 +2,8 @@ package file
 
 import (
 	"fmt"
+	"os"
+	"path/filepath"
 	"sync/atomic"
 	"time"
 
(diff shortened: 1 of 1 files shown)


storage/oasis/nodeapi/file/kvstore.go
```diff
@@ -87,7 +89,115 @@ func (s pogrebKVStore) Close() error {
...
+		return fmt.Errorf("unable to glob for files to move: %w", err)
+	}
+
+	// Create the destination directory.
+	if err := os.MkdirAll(dst, 0o700); err != nil {
+		return fmt.Errorf("unable to create destination directory %s: %w", dst, err)
+	}
+
+	for _, srcFile := range files {
+		dstFile := filepath.Join(dst, filepath.Base(srcFile))
```
ptrus, line 140: Maybe worth adding a comment that this is safe to do because the glob pattern that is used in `preBackup` will only return objects from a single folder. Otherwise this could override objects with same name from different folders.

ptrus (approved):

pro-wh:
ok on the logic in the description

storage/oasis/nodeapi/file/kvstore.go
mitjat: Technically, I think `os.Rename()` will fail for the 2nd and all subsequent files with the same filename, and this func will return with an error. I added a warning in the func docstring though, it's a good caveat to call out.

storage/oasis/nodeapi/file/kvstore.go
```diff
@@ -87,7 +89,115 @@ func (s pogrebKVStore) Close() error {
...
+	for _, f := range files {
+		if err := os.Remove(f); err != nil {
+			lastErr = fmt.Errorf("unable to delete file %s: %w", f, err)
+		}
+	}
+	return lastErr
+}
+
+// Gets rid of excessively backed-up pogreb index files.
+// If we know pogreb will reindex, delete or possibl .
```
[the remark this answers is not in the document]
mitjat: Thanks, I completed the sentence now.

PR #622, opened 2024-02-01
optimize daily active accounts query

Andrew7234:
Resolves https://app.clickup.com/t/8693pa35b

Query changed as suggested by @pro-wh

### DIFF
--- a/analyzer/aggregate_stats/queries.go
+++ b/analyzer/aggregate_stats/queries.go
@@ -88,15 +88,16 @@ const (
 	// chain.runtime_related_transactions table. Without this conversion,
 	// postgres does not realize that the timestamps correspond to a sequential
 	// set of rounds and does a full table scan.
+	//
+	// Note 2: The subselects push postgres to use the `timestamp` index
+	// on `runtime_blocks`.
 	QueryRuntimeActiveAccounts = `
-		WITH relevant_rounds AS (
-			SELECT min(round) as min_round, max(round) as max_round
-			FROM chain.runtime_blocks AS b
-			WHERE (b.runtime = $1 AND b.timestamp >= $2::timestamptz AND b.timestamp < $3::timestamptz)
-		)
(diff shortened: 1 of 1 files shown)


pro-wh (approved):
as discussed.

:+1: on semi-open interval tweak

analyzer/aggregate_stats/queries.go
```diff
@@ -89,14 +89,12 @@ const (
...
-			WHERE (b.runtime = $1 AND b.timestamp >= $2::timestamptz AND b.timestamp < $3::timestamptz)
-		)
-		SELECT COUNT(DISTINCT account_address)
-		FROM chain.runtime_related_transactions AS rt, relevant_rounds
-		WHERE (rt.runtime = $1 AND tx_round >= relevant_rounds.min_round AND tx_round <= relevant_rounds.max_round)
+	SELECT COUNT(DISTINCT account_address)
+  	FROM chain.runtime_related_transactions
+  	WHERE
+		runtime = $1 AND
+		tx_round >= (SELECT round FROM chain.runtime_blocks WHERE runtime = $1 AND timestamp >= $2::timestamptz ORDER BY timestamp ASC LIMIT 1) AND
```
mitjat, line 96: Let's document why we're doing these shenanigans
```suggestion
		-- The subselect pushes postgres to use the `timestamp` index on `runtime_blocks`.
		tx_round >= (SELECT round FROM chain.runtime_blocks WHERE runtime = $1 AND timestamp >= $2::timestamptz ORDER BY timestamp ASC LIMIT 1) AND
```

mitjat (approved):
:raised_hands: 

Thank you both for figuring this one out!

The biggest surprise to me (from conversation in Slack) is that with the old query, postgres was not wanting to walk the `timestamp` index in order to find all the blocks that belong to a specific 24h hour period. It apparently (wrongly) estimated that there will be too many such blocks. `LIMIT 1` is really nice and explicit.

PR #623, opened 2024-02-01
PR 620 follow-up: tweaks in comments

mitjat:
This addresses comments from #620; I forgot to `git push` before merging the PR :facepalm:

### DIFF
--- a/storage/oasis/nodeapi/file/kvstore.go
+++ b/storage/oasis/nodeapi/file/kvstore.go
@@ -125,6 +125,7 @@ func glob(patterns []string, antipatterns []string) ([]string, error) {
 }
 
 // Moves all files that match the src glob patterns to the destination directory.
+// NOTE: If multiple source files have the same filename, one will clobber the others when moved!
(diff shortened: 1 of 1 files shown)


storage/oasis/nodeapi/file/kvstore.go
```diff
@@ -125,6 +125,7 @@ func glob(patterns []string, antipatterns []string) ([]string, error) {
 }
 
 // Moves all files that match the src glob patterns to the destination directory.
+// NOTE: If multiple source files have the same filename, one will clobber the others when moved!
```
pro-wh, line 128: what's the thing about multiple source files with the same filename?

pro-wh:
> auto-merge enabled

storage/oasis/nodeapi/file/kvstore.go
pro-wh: update: it was if the input patterns would cover multiple dirs that contained files that had the same filename

pro-wh (approved):
alright hit it

PR #625, opened 2024-02-05
Sync gitlint config with other repos

buberdds:


mitjat (approved):
Sure. Note that this is currently not applied in CI though.

buberdds:
yes I have noticed it when I was creating PR. I think a plan is to enable it.

PR #626, opened 2024-02-06
analyzer/item: GetItems: Use stronger isolation level

mitjat:
This PR addresses a possible source of data inaccuracies. We do not know if this has actually caused problems in the past.

The evm_token_balances analyzer fetches the list of work items using the default `READ COMMITTED` mode, which [allows nonrepeatable reads](https://www.postgresql.org/docs/current/transaction-iso.html). This means that the work item could say something like "we think the balance of token X at height H was N", with the following sequencing:
1. balances analyzer reads H at the beginning of the SELECT as the highest round processed so far, 
2. block analyzer completes block H+1 (async) and commits all changes. It also records that H+1 is now analyzed, and updates the balance of X with a dead-reckoned step
3. balances analyzer reads N, thinking that it represents the state at H when it really represents the state at H+1.

Testing: None yet.

### DIFF
--- a/analyzer/evmtokens/evm_tokens.go
+++ b/analyzer/evmtokens/evm_tokens.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"fmt"
 
+	"github.com/jackc/pgx/v5"
 	"github.com/jackc/pgx/v5/pgtype"
 
 	"github.com/oasisprotocol/nexus/analyzer"
@@ -79,7 +80,13 @@ type StaleToken struct {
 
 func (p *processor) GetItems(ctx context.Context, limit uint64) ([]*StaleToken, error) {
 	var staleTokens []*StaleToken
-	rows, err := p.target.Query(ctx, queries.RuntimeEVMTokenAnalysisStale, p.runtime, limit)
+	// Use strict isolation; cannot afford non-repeatable reads where the total supply and the download round would be read
+	// before and after the block analyzer adds a new block, respectively.
+	rows, err := p.target.QueryWithOptions(
+		ctx,
+		pgx.TxOptions{IsoLevel: pgx.Serializable, AccessMode: pgx.ReadOnly},
+		queries.RuntimeEVMTokenAnalysisStale, p.runtime, limit,
+	)
 	if err != nil {
 		return nil, fmt.Errorf("querying discovered tokens: %w", err)
 	}
--- a/analyzer/item/item.go
+++ b/analyzer/item/item.go
@@ -154,7 +154,14 @@ func (a *itemBasedAnalyzer[Item]) processBatch(ctx context.Context) (int, error)
 	// Commit the changes from all successfully processed items to the database.
 	batch := &storage.QueryBatch{}
 	for _, b := range batches {
+		// Each item's batch is wrapped in a transaction because:
+		//  - This makes txs smaller, reducing the impact of deadlocks.
+		//  - We can specify a high isolation level. Item analyzers run in parallel with block analyzers,
+		//    and might need a consistent view of the database. E.g. the `evm_token_balances` analyzer
+		//    needs to read the latest height and the balance at that height.
+		batch.Queue("BEGIN TRANSACTION ISOLATION LEVEL SERIALIZABLE")
 		batch.Extend(b)
+		batch.Queue("COMMIT")
 	}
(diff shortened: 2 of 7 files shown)


analyzer/item/item.go
```diff
@@ -154,7 +154,14 @@ func (a *itemBasedAnalyzer[Item]) processBatch(ctx context.Context) (int, error)
...
 	for _, b := range batches {
+		// Each item's batch is wrapped in a transaction because:
+		//  - This makes txs smaller, reducing the impact of deadlocks.
+		//  - We can specify a high isolation level. Item analyzers run in parallel with block analyzers,
+		//    and might need a consistent view of the database. E.g. the `evm_token_balances` analyzer
+		//    needs to read the latest height and the balance at that height.
+		batch.Queue("BEGIN TRANSACTION ISOLATION LEVEL SERIALIZABLE")
 		batch.Extend(b)
+		batch.Queue("COMMIT")
 	}
```
mitjat, line 165: This change is somewhat unrelated, and will have a much higher impact on our potential for deadlocks.
I think we don't need this for the general item analyzers, and not even for the evm balances analyzer.
But I wanted to show the option here, and hear comments. For now, I plan to remove this from the PR before merging.

If we do merge this, we'll have to evaluate carefully, as it might tank performance.

pro-wh: this here as an anchor for discussion. we can't literally do this because our send-batch-fast-retry-on-error is hardcoded for complete rollback on error

analyzer/evmtokens/evm_tokens.go
```diff
@@ -79,7 +80,13 @@ type StaleToken struct {
 
 func (p *processor) GetItems(ctx context.Context, limit uint64) ([]*StaleToken, error) {
 	var staleTokens []*StaleToken
-	rows, err := p.target.Query(ctx, queries.RuntimeEVMTokenAnalysisStale, p.runtime, limit)
+	// Use strict isolation; cannot afford non-repeatable reads where the total supply and the download round would be read
```
pro-wh, line 83: serializable good :smiling_face_with_tear: from the explanation. I think the thing that has me all weird is that it doesn't particularly need repeatable reads, only that two reads for different things (height H and balance N) need to be consistent

pro-wh: or is it, docs say

> Read Committed is the default isolation level in PostgreSQL. When a transaction uses this isolation level, a SELECT query (without a FOR UPDATE/SHARE clause) sees only data committed before the query began; it never sees either uncommitted data or changes committed by concurrent transactions during the query's execution. In effect, a SELECT query sees a snapshot of the database as of the instant the query begins to run.

which should prevent the later read of N from giving the data from height H+1

mitjat: I don't know how to consolidate your snippet ([from here](https://www.postgresql.org/docs/current/transaction-iso.html#XACT-READ-COMMITTED)):

> In effect, a SELECT query sees a snapshot of the database as of the instant the query begins to run.

with [this](https://www.postgresql.org/docs/current/transaction-iso.html):

> nonrepeatable read: A transaction re-reads data it has previously read and finds that data has been modified by another transaction (that committed since the initial read).
> (later in the table: In Read Committed mode, nonrepeatable reads are possible.)

Edit: Oh ... I see. Your link says  (emphasis added)

> In effect, a SELECT query sees a snapshot of the database as of the instant the **query** begins to run.

and also

> Also note that two successive SELECT commands can see different data, even though they are within a single transaction, if other transactions commit changes after the first SELECT starts and before the second SELECT starts.

This was all part of the slack discussion. I first thought that Read Committed ensured consistency within a single SELECT, then convinced myself that it does not (and started this PR), but now I see that was wrong. Stupid subtlety :shakefist:.

I'll close this PR, I think it doesn't really make sense any more. Thanks for the pushback.

analyzer/item/item.go
mitjat: I believe it would work if send-batch-fast used explicit txs, rather than the pipeline stuff. With an explicit tx, this change would produce nested txs, whereas send-batch-fast would roll back the top-level tx.
OTOH the large parent tx would also negate the proposed benefits here.

I'm getting annoyed with how the pipeline mode needs special consideration. I'll remove (= always start an explicit tx) it in a separate PR.

PR #628, opened 2024-02-07
bugfix: abi_backfill: key event updates by event body

Andrew7234:
@mitjat noticed that the abi-backfill analyzer was not selective enough when updating runtime events with parsed fields. Instead of only updating the corresponding event, the analyzer would update all events from the transaction with the same parsed fields. 

This PR fixes the update query to additionally filter based on event body. Although this is not a unique key for events within a given transaction, events with the same body will parse to the same evm_log_name/params/signature, which maintains correctness. 

Testing: Manually tested near the nftrout creation range. In particular, there is one block `410435` that emits many events from the same transaction. With the bugfix, we see that different events from the same tx are parsed independently as expected.

Screenshot 2024-02-07 at 5 24 38 PM


mitjat (approved):
Thanks. Also, nice description :)

It would be really good to have some tests with this. That said, I'm adding e2e testing for ABI-related stuff (verification, parsing) in my WIP branch ... and I guess that'll have to do for now. More unit-like testing is messy for this particular analyzer, we'd need pretty large fixtures :/

PR #629, opened 2024-02-08
consensus: remove separator comment queries

pro-wh:
These put pgx's pipeline out of sync when it expects to find an Exec result and instead finds the next query's output.


mitjat (approved):
Thank you, and nice debugging!

What I think the implication is: Before this PR, genesis processing would first always fail with `failed to submit tx using the fast path; falling back to slow path`, then succeed via the slow path. After this PR, the fast path succeeds also.

Correct?

pro-wh:
correct

PR #630, opened 2024-02-08
consensus: use batch for genesis queries

pro-wh:
this kinda sucks

- db slower ~13s -> ~17s on eden genesis file
- can no longer print out queries into a migration-like file

also some of the ON CONFLICT is not exactly preserved, so might be some hidden bugs in network upgrades

measurements, very informal, only one run each

reuse
{"analyzer":"consensus","caller":"genesis.go:52","count":355437,"height":"genesis","level":"info","module":"analysis_service","msg":"generated genesis queries","ts":"2024-02-08T01:42:57.758442953Z"}
{"analyzer":"consensus","caller":"consensus.go:204","level":"info","module":"analysis_service","msg":"genesis document queries generated","process_duration":"32.732929903s","ts":"2024-02-08T01:42:57.758504654Z"}
{"analyzer":"consensus","caller":"consensus.go:228","level":"info","module":"analysis_service","msg":"genesis document processed","send_batch_duration":"17.026809672s","ts":"2024-02-08T01:43:14.785349655Z"}

control
{"analyzer":"consensus","caller":"genesis.go:56","count":369,"height":"genesis","level":"info","module":"analysis_service","msg":"generated genesis queries","ts":"2024-02-08T02:30:31.547038915Z"}
{"analyzer":"consensus","caller":"consensus.go:204","level":"info","module":"analysis_service","msg":"genesis document queries generated","process_duration":"43.796619503s","ts":"2024-02-08T02:30:31.547102908Z"}
{"analyzer":"consensus","caller":"consensus.go:229","level":"info","module":"analysis_service","msg":"genesis document processed","send_batch_duration":"13.060895665s","ts":"2024-02-08T02:30:44.608061096Z"}

### DIFF
--- a/analyzer/consensus/genesis.go
+++ b/analyzer/consensus/genesis.go
@@ -5,23 +5,22 @@ package consensus
 import (
 	"encoding/hex"
 	"encoding/json"
-	"fmt"
 	"sort"
-	"strings"
 
 	"github.com/oasisprotocol/oasis-core/go/common/cbor"
 	"github.com/oasisprotocol/oasis-core/go/common/entity"
 
+	"github.com/oasisprotocol/nexus/analyzer/queries"
+	"github.com/oasisprotocol/nexus/common"
 	beacon "github.com/oasisprotocol/nexus/coreapi/v22.2.11/beacon/api"
 	genesis "github.com/oasisprotocol/nexus/coreapi/v22.2.11/genesis/api"
 	registry "github.com/oasisprotocol/nexus/coreapi/v22.2.11/registry/api"
 	staking "github.com/oasisprotocol/nexus/coreapi/v22.2.11/staking/api"
+	"github.com/oasisprotocol/nexus/storage"
 
 	"github.com/oasisprotocol/nexus/log"
 	"github.com/oasisprotocol/nexus/storage/oasis/nodeapi"
 )
-
-const bulkInsertBatchSize = 1000
 
 // GenesisProcessor generates sql statements for indexing the genesis state.
 type GenesisProcessor struct {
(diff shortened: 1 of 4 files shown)


pro-wh:
oh and hide whitespace on the diff due to some indentation decrease

analyzer/consensus/genesis.go
```diff
@@ -35,124 +36,62 @@ func NewGenesisProcessor(logger *log.Logger) *GenesisProcessor {
...
+		batch.Queue(queries.ConsensusEntityUpsert,
 			entity.ID.String(),
 			staking.NewAddress(entity.ID).String(),
 		)
-		if i != len(document.Registry.Entities)-1 {
-			query += ",\n"
-		}
 	}
-	query += `
-ON CONFLICT (id) DO UPDATE SET address = EXCLUDED.address;`
```
mitjat, line 82: You mention in the description that not all `ON CONFLICT` are exactly preserved. That does worry me a little, because the behavior can be nuanced. A good test would be to run e2e with a tweaked `e2e_config_1.yml` so that it uses fast sync for half (?) of the normal block range. After fast sync, analyzers would run StateToGenesis and test our genesis processing against a prepopulated db. But that's probably out of the question because StateToGenesis is unavailable. Kinda ironic because StateToGenesis being unavailable is what made you pause you previous work and go do this PR instead :). cc @gw0 as a data point towards https://app.clickup.com/t/8693mg8z2

mitjat (approved):
Huh, I see the dilemma.
- The 13->17s does suck for local development.
- The dumpable queries ... I don't care for particularly. They can be post-processed pretty easily to show in a readable format, IMO. Sth like output the query template only if it differs from the previous query, otherwise only output the args. Not that we should implement that now; it's just that whoever needs to debug genesis queries that deeply (I haven't, in recent memory) can hack together some insight easily enough.
- The code is _so_ much cleaner, and more resilient to future DB changes.

I vote to check this in. I can wait the extra 4 seconds (times 30 iterations) next time I'm working on something that requires genesis processing.

And it _can_ be sped up if needed:
- The queries are absolutely dominated by account balances (~280k) and staking delegations (~70); there's <10k other rows. So we only really need to speed the first two, or even just the balances.
- The data types involved are simple, so using the COPY protocol [like so](https://pkg.go.dev/github.com/jackc/pgx#hdr-Copy_Protocol) should be simple too.
  - The caveat is that COPY cannot do upserts. We do need upserts because we apply the genesis after fast sync. So you'd need to COPY into a temp table, and then do a `INSERT FROM mytemptable INTO accounts ON CONFLICT ...`.
  - The exact implementation is a little messy too, but it's not super bad. We'd need to either pass the db connection into genesis.go (inversion of control from what we have now; I'm not a fan), or `Process()` in `genesis.go` would need to return a triple: the `[][]interface{}` for balances, the `[][]interface{}` for delegations, and the batch. The caller would then have to make both CopyFrom() calls plus exec the batch (which would contain commands for upserting from the temp tables).
- Conveniently, those are also the only two queries you needed to newly add into `queries.go`. (That's mostly irrelevant, except emotionally :))

Whether you want to take a stab at that is up to you. Even if not, I think this PR is a step in the right direction. Nice cleanup, thank you.

analyzer/consensus/genesis.go
```diff
@@ -35,124 +36,62 @@ func NewGenesisProcessor(logger *log.Logger) *GenesisProcessor {
...
-	tee_hardware = EXCLUDED.tee_hardware,
-	key_manager = EXCLUDED.key_manager;`
-		queries = append(queries, query)
+	for _, runtime := range document.Registry.Runtimes {
+		batch.Queue(queries.ConsensusRuntimeUpsert,
+			runtime.ID.String(),
+			false,
+			runtime.Kind.String(),
+			runtime.TEEHardware.String(),
+			common.StringOrNil(runtime.KeyManager),
```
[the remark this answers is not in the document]
pro-wh: oh my god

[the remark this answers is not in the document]
pro-wh: #631 :telescope:

analyzer/consensus/genesis.go
```diff
@@ -290,299 +192,134 @@ VALUES
 			account.Escrow.Debonding.Balance.ToBigInt(),
 			account.Escrow.Debonding.TotalShares.ToBigInt(),
 		)
-
-		if (i+1)%bulkInsertBatchSize == 0 {
```
pro-wh, line 294: > Error: const `bulkInsertBatchSize` is unused (unused)

oo nice lint find

analyzer/consensus/genesis.go
pro-wh: added a test in #667, no regressions

pro-wh:
let's go with how this is now

PR #632, opened 2024-02-09
Draft: backfill analyzer to re-parse all runtime events for evm_log_* fields

Andrew7234:
do not merge

Context: A bug in the abi_analyzer accidentally overwrote the `evm_log_*` fields for some `evm.log` runtime events. This PR adapts the abi analyzer to re-parse all runtime `evm.log` events and correctly update the `evm_log_*` fields. 

Also, the existing analyzer did not properly insert null values for `evm_log_name` and `evm_log_signature`. This analyzer inserts null values correctly.

The bulk of ParseEvent is a stripped down version of the existing event parsing code in extract.go

Testing:
I ran the analyzer on sapphire for ~500 blocks. I took a snapshot of the database, and then ran the modified abi analyzer on both sapphire til completion. I took a second snapshot of the database and compared it. The ordering of the rows was off, but most of the other data remained the same. The only different fields were the corrected null fields and `abi_parsed_at` as expected

### DIFF
--- a/analyzer/queries/queries.go
+++ b/analyzer/queries/queries.go
@@ -953,4 +953,14 @@ var (
     WHERE
       evs.abi_parsed_at IS NULL
     LIMIT $2`
+
+	RuntimeEvmEvents = `
+    SELECT
+      round, tx_index, body
+    FROM chain.runtime_events
+    WHERE
+      runtime = $1 AND
+      type = 'evm.log' AND
+      abi_parsed_at IS NULL
+    LIMIT $2`
 )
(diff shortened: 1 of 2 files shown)


analyzer/queries/queries.go
```diff
@@ -477,9 +477,9 @@ var (
 	RuntimeEventEvmParsedFieldsUpdate = `
     UPDATE chain.runtime_events
     SET
-      evm_log_name = COALESCE($5, evm_log_name),
-      evm_log_params = COALESCE($6, evm_log_params),
-      evm_log_signature = COALESCE($7, evm_log_signature),
+      evm_log_name = $5,
+      evm_log_params = $6,
+      evm_log_signature = $7,
```
Andrew7234, line 482: we definitely want to overwrite the old values here

Andrew7234:
Applied to staging; events were reparsed successfully. No longer needed

Andrew7234:
Applied to staging; events were reparsed successfully. No longer needed

PR #633, opened 2024-02-09
e2e regression tests: HTTP caching proxy, new cases

mitjat:
This PR expands our e2e testing capabilities; the motivation is the upcoming support for partial verifications in Sourcify.

The changes are:
- A new caching HTTP proxy is introduced, based on our existing KVStore. It allows e2e tests to cache HTTP responses (e.g. from sourcify) on an initial run, and then replay them deterministically on every subsequent run.
  - Small refactor: KVStore and the new proxy are moved to a dedicated package.
- The `evmverifier` analyzer is now included in e2e tests, using the new proxy.
  - The PR also includes two new `db__*` test cases that aim to capture the "derived" columns of txs and events, notably the ABI-parsing results.
  - Despite that, there are no diffs between before and after `evmverifier`. Our current e2e block range does not engage with any contracts that are fully verified :grimacing: 
  - This will hopefully change when #627 lands and re-introduces a Damask range, and will certainly change when we support partially-verified contracts (though minimally; there's only 1 partially verified contract that this e2e block range interacts with.
- The item analyzer sleep logic has changed a little. The old `stop_on_empty_queue` has been replaced with a duration-based flag ("exit only if the queue has been empty for N seconds"). This is helpful for item analyzers that interact with each other.


cmd/analyzer/analyzer.go
```diff
@@ -512,6 +526,16 @@ func (a *Service) Start() {
 	ctx, cancelAnalyzers := context.WithCancel(context.Background())
 	defer cancelAnalyzers() // Start() only returns when analyzers are done, so this should be a no-op, but it makes the compiler happier.
 
+	// Start helpers.
+	for _, helper := range a.helpers {
```
ptrus, line 530: Should probably also try to gracefully `Shutdown()` the server helpers in `received interrupt, shutting down` part. And call `Close()` on the servers in `cleanup`.

cmd/analyzer/analyzer.go
```diff
@@ -149,6 +151,7 @@ func wipeStorage(cfg *config.StorageConfig) error {
 type Service struct {
 	analyzers         []SyncedAnalyzer
 	fastSyncAnalyzers []SyncedAnalyzer
+	helpers           []*http.Server
```
ptrus, line 154: Should this be just named (caching) proxies, at least while caching proxies are the only type of helpers there is?

pro-wh:
accessing this cache over HTTP, can you talk about this

in existing code that does similar, there's a Go interface and a caching implementation of that

tests/e2e_regression/e2e_config_2.yml
```diff
@@ -9,16 +9,20 @@ analysis:
...
-    evm_tokens_emerald:         { stop_on_empty_queue: true }
-    evm_nfts_emerald:           { stop_on_empty_queue: true }
-    evm_token_balances_emerald: { stop_on_empty_queue: true }
-    evm_contract_code_emerald:  { stop_on_empty_queue: true }
-    evm_abi_emerald:            { stop_on_empty_queue: true }
+    evm_tokens_emerald:            { stop_if_queue_empty_for: 1s }
+    evm_nfts_emerald:              { stop_if_queue_empty_for: 1s }
+    evm_token_balances_emerald:    { stop_if_queue_empty_for: 1s }
+    evm_contract_code_emerald:     { stop_if_queue_empty_for: 1s }
+    evm_abi_emerald:               { stop_if_queue_empty_for: 10s }
```
Andrew7234, line 20: Mind adding a comment here about how we expect to fetch all the abis within 10s, and that we may need to tune this in the future? (Alternatively we could add a third step via an `e2e_config_3.yml` but that seems overkill for now)

tests/e2e_regression/run.sh
```diff
@@ -46,6 +46,8 @@ testCases=(
   'db__account_related_txs      select * from chain.runtime_related_transactions order by runtime, tx_round, tx_index, account_address'
   'db__runtime_accounts         select * from chain.runtime_accounts order by runtime, address'
   'db__runtime_transfers        select * from chain.runtime_transfers order by runtime, round, sender, receiver'
+  'db__runtime_txs              select runtime, round, tx_hash, "to", fee, gas_used, method, evm_fn_name, evm_fn_params, error_message from chain.runtime_transactions order by runtime, round, tx_index'
+  'db__runtime_events           select runtime, round, type, tx_hash, evm_log_name, evm_log_params from chain.runtime_events order by runtime, round, tx_index, type, body'
```
Andrew7234, line 50: maybe add `evm_log_signature` here too?

pro-wh (commented):
can we switch to a mockable http client later

analyzer/item/item.go
```diff
@@ -79,15 +79,15 @@ func NewAnalyzer[Item any](
...
-		metrics.NewDefaultAnalysisMetrics(name),
+		maxBatchSize:        cfg.BatchSize,
+		stopIfQueueEmptyFor: cfg.StopIfQueueEmptyFor,
+		fixedInterval:       cfg.Interval,
+		interItemDelay:      cfg.InterItemDelay,
+		analyzerName:        name,
+		processor:           processor,
+		target:              target,
+		logger:              logger,
+		metrics:             metrics.NewDefaultAnalysisMetrics(name),
```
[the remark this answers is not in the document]
mitjat: If style guidelines recommend something, my guess is they do recommend listing the keys once the number of attributes is large enough, e.g. >3. I'm not going off of anything formal though. I just find it more readable myself, and (which is what prompted it) helpful for finding where variables get set, either via ctrl+F or fancier IDE features.

mitjat:
> can we switch to a mockable http client later

Yeah I'm totally OK with that. Right now the http proxy is only being used for sourcify, and it's a low enough number of requests that we could mock them more explicitly. FWIW, the caching proxy achieves pretty much the same effect, with the debugging downside that it's a lot harder to see the cached/mocked responses.

analyzer/item/item.go
```diff
@@ -195,12 +195,16 @@ func (a *itemBasedAnalyzer[Item]) Start(ctx context.Context) {
...
 
-	for {
+	for firstIter := true; ; firstIter = false {
 		delay := backoff.Timeout()
 		if a.fixedInterval != 0 {
 			delay = a.fixedInterval
 		}
+		if firstIter {
+			delay = 0 // Don't sleep before first iteration.
+		}
```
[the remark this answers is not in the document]
mitjat: :astonished:  I somehow pulled the software equivalent of eating soup with a fork

Fixed, thanks.

analyzer/evmverifier/evmverifier.go
```diff
@@ -194,7 +203,11 @@ func (p *processor) ProcessItem(ctx context.Context, batch *storage.QueryBatch,
 	}
 
 	batch.Queue(
-		queries.RuntimeEVMVerifyContractUpdate,
+		// NOTE: This also updates `verification_info_downloaded_at`, causing the `evm_abi` to re-parse
+		//       the contract's txs and events.
+		// NOTE: We upsert rather than update; if the contract is not in the db yet, UPDATE would ignore the
+		//       contract and this analyzer would keep retrying it on every iteration.
```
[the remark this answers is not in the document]
mitjat: An insert can happen if sourcify has a contract that we haven't encountered yet, e.g. because we're scanning just a subset of blocks. GetItems no longer limits the items to `chain.evm_contracts` addresses as of #634 (which is not merged; I accidentally merged it into this branch and then undid that, but I'll duplicate the PR to merge into `main` once this PR goes in). The reasons for the change are pretty lightweight: we now don't have to worry about what happens as the number of contracts grows, and I found the new flow easier to understand, especially with the added partial/full verification.

tests/e2e_regression/e2e_config_2.yml
mitjat: +1 on both points about `_3.yml`. Comment added!

cmd/analyzer/analyzer.go
ptrus: Today's end-to-end tests are tomorrow's production features :grin:

pro-wh:
what's the item analyzer exit delay for? is there a new test that relies on it?

new config api can't say to do what we did before, you instead have to set a very small delay

mitjat:
> what's the item analyzer exit delay for? is there a new test that relies on it?

Yeah the e2e. We ideally would want to run in 3 stages ...
https://github.com/oasisprotocol/nexus/pull/633#discussion_r1498319871
```
evm_abi_emerald:               { stop_if_queue_empty_for: 10s } # Give evm_contract_verifier time to fetch ABIs first. The 10s has been enough in practice, but might need to be tuned in the future, especially if the caching proxy has an empty cache.
```

> new config api can't say to do what we did before, you instead have to set a very small delay

True. I worried about it a little at first, but then decided it's just an aesthetic problem, not a functional one. And it only applies to testing. So screw it.

mitjat:
After rebasing on top of #627 and debugging, I realized that our internal Damask Emerald node is currently having issues. I therefore disabled the Emerald part of the new Damask e2e test. This version of this PR still makes e2e tests usefully more comprehensive to help with merging #634.

PR #635, opened 2024-02-09
api: Route /pontusx requests to /sapphire (TEMP HACK)

mitjat:
This is to enable frontenders to test against nexus as if it supported pontusx.

Should be reverted once pontusx launches.


mitjat:
@csillag handled the rerouting on the frontend, so this PR is not needed

PR #636, opened 2024-02-10
pgx: fast batches: always create an explicit tx

mitjat:
This PR simplifies the way we send txs to postgres.
For the longest time, we've used implicit txs and the "pipeline mode" (see deleted comments in code). #556 added those comments, and at the time, there was a lot of research and back and forth involved already. Now in #626, we identified additional potential caveats with implicit txs. Time to get rid of them. 

Speed impact: TODO. I expect it to be negligible since everything is tx-like in postgres anyway.


ptrus (approved):

storage/postgres/client.go
```diff
@@ -95,93 +96,75 @@ func (c *Client) SendBatch(ctx context.Context, batch *storage.QueryBatch) error
...
+// Adapted from https://github.com/jackc/pgx/blob/v4.18.1/tx.go#L108
+func (c *Client) WithTx(
+	ctx context.Context,
+	txOptions pgx.TxOptions,
+	fn func(pgx.Tx) error,
+) error {
+	var err error
+	var tx pgx.Tx
+
+	tx, err = c.pool.BeginTx(ctx, txOptions)
```
Andrew7234, line 111: Would this work?

```suggestion
	tx, err := c.pool.BeginTx(ctx, txOptions)
```

Andrew7234 (approved):

storage/postgres/client.go
mitjat: yep, that's better, thanks :+1:

PR #639, opened 2024-02-13
Add Pontus-X support

mitjat:
This mostly involves just copy-paste of all the sapphire initialization code.

I did a tiny refactor in `node_stats`. Further decluttering should be possible in `cmd/analyzer` in a future PR. But in general, the whole process was fairly painless and quick.

Testing: None yet; I don't have access to a functioning node.


go.mod
```diff
@@ -6,6 +6,7 @@ toolchain go1.21.3
 
 replace (
 	github.com/cometbft/cometbft => github.com/oasisprotocol/cometbft v0.37.2-oasis1
+	github.com/oasisprotocol/oasis-sdk/client-sdk/go => /home/mitjat/oasis/oasis-sdk/client-sdk/go
```
mitjat, line 9: PR requires https://github.com/oasisprotocol/oasis-sdk/pull/1638/files and a new SDK release, then this override can go away. In a rush, we can hardcode this part of the SDK config into nexus temporarily.

csillag:
> Also update api spec?

Can we use https://github.com/oasisprotocol/nexus/pull/637 for that?

mitjat:
> > Also update api spec?
> 
> Can we use #637 for that?

Thanks, I prefer to do it here though so all the changes are in one PR.

csillag:
> > > Also update api spec?
> > 
> > Can we use #637 for that?
> 
> Thanks, I prefer to do it here though so all the changes are in one PR. Commit added.

Yeah we don't need a separate PR.
My point with that PR was only to provide a base for code generation for explorer.

As as long as your API changes are compatible, we are good.

PR #640, opened 2024-02-13
make evm_log_* fields null for non evm.log events

Andrew7234:
While cleaning up the `evm_log_*` fields in #632 I noticed that these fields weren't properly set to null. `evm_log_name` was set to the empty string and `evm_log_signature` was set to a zero bytestring. 

Testing: Manually ran the indexer and confirmed that there were no rows where `evm_log_*` was non-null for any non evm.log events. 
```
select * from chain.runtime_events where type != 'evm.log' and evm_log_signature is not null limit 10;
```


tests/e2e_regression/expected/emerald_events.body
```diff
@@ -9,7 +9,7 @@
         "from": "oasis1qp3r8hgsnphajmfzfuaa8fhjag7e0yt35cjxq0u4",
         "to": "oasis1qp334gzlzrap6k2ch6wc9vxxplw9sg3v9cfvvgsy"
       },
-      "evm_log_name": "",
+      "evm_log_name": null,
```
mitjat, line 12: I was surprised by all these explicit nulls, so I did some digging. These fields would be simply missing if they were tagged as `json:"...,omitempty"`. The JSON tags are generated by oapi-codegen. And the reason that the tag does not get generated (it [normally does](https://github.com/deepmap/oapi-codegen/blob/v1.16.2/pkg/codegen/codegen_test.go#L126-L128)) is that the `evm_log_name` field is marked `nullable` in our openapi spec. Which, [per openapi](https://swagger.io/docs/specification/data-models/data-types/#null), means "this variable can contain an explicit `null` value".

Looking at our openapi spec, we have 5 instances of `nullable: true`, and I don't think any of them are justified. What we want as the behavior is to simply omit those fields when they are null.

Can I ask you to fix that also, probably in a separate PR, or let me know when this PR goes in so I send one out myself? Thanks!

mitjat (approved):
nice :+1:

tests/e2e_regression/expected/emerald_events.body
Andrew7234: Thanks for digging into that! I'll fix it in a follow-up PR.

Andrew7234: https://github.com/oasisprotocol/nexus/pull/642

PR #641, opened 2024-02-13
openapi: Use named types everywhere

mitjat:
This PR removes YAML references in favor of openapi types and `$ref`-style references for all reusable types, notably `Address` and `BigInt`. It was prompted by https://github.com/oasisprotocol/explorer/issues/711

It also wraps all `$ref` references into `allOf`; this is a [workaround](https://github.com/OAI/OpenAPI-Specification/issues/1514#issuecomment-1386613645) (thank you @lukaw3d) for the fact `$ref` on its own [does not allow sibling attributes](https://swagger.io/docs/specification/using-ref/) like a custom description.

---

**OUTDATED** description follows, from before the workaround was known.

I'm on the fence whether to actually merge this PR:
- GOOD: Nexus still compiles with the new openapi, requiring no changes.
- BAD: When I compared the autogenerated Go code, I noticed all code comments for variables of type `Address` and `BigInt` were gone.
    - This is because `$ref` does not allow a custom description. That's part of the spec: https://swagger.io/docs/specification/using-ref/
    - GOOD: On the other hand, our HTML documentation generator ignores that part of openapi spec and helpfully still includes the docstrings in `v1.html`.

In my mind, losing per-field descriptions is a deal breaker. But since we're not losing them in the human-oriented HTML, maybe that's enough.

@lukaw3d please evaluate the pros and cons on the frontend side.

### DIFF
--- a/api/spec/v1.yaml
+++ b/api/spec/v1.yaml
@@ -242,13 +242,13 @@ paths:
         - in: query
           name: minFee
           schema:
-            allOf: [$ref: '#/components/schemas/BigInt']
+            allOf: [$ref: '#/components/schemas/TextBigInt']
           description: A filter on minimum transaction fee, inclusive.
           example: "1000"
         - in: query
           name: maxFee
           schema:
-            allOf: [$ref: '#/components/schemas/BigInt']
+            allOf: [$ref: '#/components/schemas/TextBigInt']
           description: A filter on maximum transaction fee, inclusive.
           example: "100000000000000000000000"
         - in: query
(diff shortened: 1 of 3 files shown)


lukaw3d:
If we upgraded to OpenAPI 3.1 it should support it, but https://github.com/deepmap/oapi-codegen/issues/373 doesn't support OpenAPI 3.1 yet (I tried installing the latest version, bumping `openapi: 3.0.3`, and regenerating, just in case).

Until then, we can use https://github.com/OAI/OpenAPI-Specification/issues/1514#issuecomment-1386613645 workaround:
```diff
        deposit:
-           $ref: '#/components/schemas/BigInt'
+          allOf:
+            - $ref: '#/components/schemas/BigInt'
            description: The deposit attached to this proposal.
```
and that fixes generated Go, and still works in generated TypeScript

mitjat:
Thanks for the research and for finding the workaround, @lukaw3d ! I implemented it, it works, so let's go with that.

api/spec/v1.yaml
```diff
@@ -276,13 +242,13 @@ paths:
         - in: query
           name: minFee
           schema:
-            <<: *BigIntType
+            allOf: [$ref: '#/components/schemas/BigInt']
```
lukaw3d, line 245: Can this be named anything else, e.g. `StringifiedBigInt`
It is misleading in typescript, where BigInt is already a different global type

mitjat: Hm doesn't typescript do namespacing quite well?
Anyway, renamed to `TextBigInt` to keep the name a bit shorter.

ptrus (approved):

api/spec/v1.yaml
lukaw3d: Our code generator handles it correctly as long as we don't have any native BigInts in that file; so it would only be misleading when looking at the code and intellisense

PR #642, opened 2024-02-14
[api] remove nullable tags

Andrew7234:
From https://github.com/oasisprotocol/nexus/pull/640#discussion_r1488734482

per @mitjat 's comment:
> I was surprised by all these explicit nulls, so I did some digging. These fields would be simply missing if they were tagged as json:"...,omitempty". The JSON tags are generated by oapi-codegen. And the reason that the tag does not get generated (it [normally does](https://github.com/deepmap/oapi-codegen/blob/v1.16.2/pkg/codegen/codegen_test.go#L126-L128)) is that the evm_log_name field is marked nullable in our openapi spec. Which, [per openapi](https://swagger.io/docs/specification/data-models/data-types/#null), means "this variable can contain an explicit null value".

> Looking at our openapi spec, we have 5 instances of nullable: true, and I don't think any of them are justified. What we want as the behavior is to simply omit those fields when they are null.

This PR removes the `nullable` tags in our openapi spec. The only ones that might require further confirmation are the `event.{tx_index/tx_hash}`. For non-tx events, these fields will no longer exist and we should double check that the Explorer doesn't make assumptions otherwise.

### DIFF
--- a/api/spec/v1.yaml
+++ b/api/spec/v1.yaml
@@ -1594,14 +1594,12 @@ components:
         tx_index:
           type: integer
           format: int32
-          nullable: true
           description: |
             0-based index of this event's originating transaction within its block.
             Absent if the event did not originate from a transaction.
           example: 5
         tx_hash:
           type: string
-          nullable: true
           description: |
             Hash of this event's originating transaction.
             Absent if the event did not originate from a transaction.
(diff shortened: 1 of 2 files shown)


mitjat (approved):
Thank you!

PR #645, opened 2024-02-20
analyzer/runtime: Deposit/Withdraw txs: make "to" explicit

mitjat:
For txs with method `consensus.Deposit` or `consensus.Withdraw`, the `to` field in the body is optional; when omitted, the implied semantics are that the deposit/withdrawal is happening to the account of the tx sender.

Nexus's `/transactions` endpoint lists txs with their raw body (where we don't want to inject the implicit `to`), but also with an additional `to` [field](https://github.com/oasisprotocol/nexus/blob/fb45f22b971ce8dc82a04d2094a2a8abd32d80fb/api/spec/v1.yaml#L2518-L2518) that we extract from txs regardless of the method. This PR makes it so that this field is always filled, even if `to` in the tx body is missing.

**Testing:** Only manual :(. Our current e2e range does not include suitable blocks. I indexed round 316043 from `pontusx` that is known to contain a tx with an implicit `to`.


pro-wh:
is this started from another branch?

pro-wh:
lg, cool tests

analyzer/runtime/runtime_test.go
mitjat, line 1: I cannot fully decide if this is the right abstraction level. We could narrow the scope of tests and analyze the internal variables that the runtime analyzer populates during analysis, and then uses them to generate DB statements. The upside of that would be a more focused test. The downside would be having to treat the analyzer a lot more as a white box. Another downside is that it would make it harder to port these tests to consensus: The consensus analyzer doesn't have that convenient intermediate in-memory representation, it produces DB statements more directly from node data. 
So I decide to go with this, and tweak/refactor it later if needed. :shrug:

analyzer/runtime/runtime.go
mitjat, line 1: Unrelated changes; I just found the old wording slightly misleading when debugging pontusx.

PR #646, opened 2024-02-22
consensus: save block metadata

pro-wh:
image

proposer and signers


analyzer/consensus/registration.go
```diff
@@ -37,3 +38,11 @@ func registerRelatedOCAddress(relatedAddresses map[apiTypes.Address]struct{}, oc
 	relatedAddresses[addr] = struct{}{}
 	return addr, nil
 }
+
+func stringifyCmtAddress(cmtAddr cmtTypes.Address) (apiTypes.Address, error) {
+	var sdkAddr sdkTypes.Address
+	if err := sdkAddr.UnmarshalBinary(cmtAddr); err != nil {
```
pro-wh, line 44: this is wrong, it's not the same kind of address.

see oasis's cometbft address impl. need to make a lookup

pro-wh: ok after a lot of fuss, the lookup is made

storage/client/queries/queries.go
```diff
@@ -24,13 +24,39 @@ const (
...
+			time,
+			num_txs,
+			metadata,
+			ROW(
+			    proposer_node.id,
+			    proposer_node.entity_id,
+			    proposer_entity.address,
+			    proposer_entity.meta
+			) AS proposer_node_info,
+			ARRAY(
```
mitjat, line 39: :hot_face: swanky. nice job.

Maybe a comment for future readers? `-- Maps signer_node_consensus_pubkey_addresses to an array of structs that each describe the corresponding node` or sth

go.mod
```diff
@@ -66,6 +65,7 @@ require (
 	github.com/go-git/gcfg v1.5.1-0.20230307220236-3a3c6141e376 // indirect
 	github.com/go-git/go-billy/v5 v5.5.0 // indirect
 	github.com/go-git/go-git/v5 v5.11.0 // indirect
+	github.com/go-kit/kit v0.12.0 // indirect
```
mitjat, line 68: Do you understand how this was added? We're not using it directly, and the only other depedendency change is that cometbft went from being an implicit dep to an explicit one. Maybe Go only adds indirect deps when they are actually used, rather than all transitive deps?

api/spec/v1.yaml
```diff
@@ -1357,6 +1357,12 @@ components:
           format: int32
           description: Number of transactions in the block.
           example: 17
+        proposer:
```
mitjat, line 1360: I would expect to see these beauts in e2e diffs. Probably just needs to be re-run?

mitjat (commented):
Thank you thank you for figuring these out!
Generally LGTM. Most of my comments are about explaining the semantics to our API consumers (and to future nexus code maintainers).

storage/migrations/00_consensus.up.sql
```diff
@@ -25,18 +25,22 @@ CREATE DOMAIN public.eth_addr BYTEA CHECK(length(VALUE) = 20);
 CREATE TABLE chain.blocks
 (
   height     UINT63 PRIMARY KEY,
-  block_hash HEX64 NOT NULL,
-  time       TIMESTAMP WITH TIME ZONE NOT NULL,
-  num_txs    UINT31 NOT NULL,
+  block_hash HEX64 NOT NULL,  -- NULL in 10_block_meta.up.sql
```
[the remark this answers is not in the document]
pro-wh: 14 now :skull:

analyzer/consensus/consensus.go
```diff
@@ -345,8 +347,19 @@ func (m *processor) ProcessBlock(ctx context.Context, uheight uint64) error {
 }
 
 func (m *processor) queueBlockInserts(batch *storage.QueryBatch, data *consensusBlockData) error {
+	var cmtMeta cometbft.BlockMeta
+	if err := cbor.Unmarshal(data.BlockHeader.Meta, &cmtMeta); err != nil {
+		m.logger.Warn("could not unmarshal block meta, may be incompatible version",
+			"height", data.BlockHeader.Height,
+			"err", err,
+		)
+		// This is fine, we don't have earlier versions of the struct.
```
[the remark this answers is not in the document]
pro-wh: haven't tested on damask either, back when it was a "tendermint" structure

analyzer/consensus/consensus.go
```diff
@@ -345,8 +347,19 @@ func (m *processor) ProcessBlock(ctx context.Context, uheight uint64) error {
 }
 
 func (m *processor) queueBlockInserts(batch *storage.QueryBatch, data *consensusBlockData) error {
+	var cmtMeta cometbft.BlockMeta
```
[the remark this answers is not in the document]
pro-wh: we/our as in the oasis-core team? probably would be for consistency. different consensus backends could have different information

api/spec/v1.yaml
```diff
@@ -1371,6 +1377,17 @@ components:
       description: |
         A consensus block.
 
+    NodeInfo:
```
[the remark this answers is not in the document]
pro-wh: it's like just a reference to a validator, enough stuff to render a link to their explorer page. not full details. probably does need to be named something other than "info" but imo "signer" is too specific to this use case. maybe node reference

api/spec/v1.yaml
```diff
@@ -1371,6 +1377,17 @@ components:
...
+      type: object
+      properties:
+        node_id:
+          type: string
+        entity_id:
+          type: string
+        entity_address:
+          type: string
+        entity_metadata: {}
```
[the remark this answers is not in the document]
pro-wh: not equivalent, `type: object` is more specific. having an explicit schema would be good, and that including the "type." but I'm against being the first mover with a half measure like adding `type: object` alone. import the metadata schema throughout the codebase, and not in this pr

analyzer/consensus/consensus.go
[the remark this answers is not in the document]
mitjat: OK let's say "pre-Eden" in the comment then; it's more informative than "earlier".

> back when it was a "tendermint" structure

And that's the structure that we "don't have", per the comment? The comment still confuses me. Is it that we cannot find that struct anymore (sounds unlikely), or that we haven't bothered importing/implementing it, or that that structure appears somewhere other than the `data.BlockHeader.Meta` field and so it does not apply here?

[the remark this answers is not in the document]
pro-wh: this BlockMeta structure, we haven't been vendoring it. the comment is meant to say that we try to decode any block metadata into the latest BlockMeta structure as directly imported oasis-core, we don't have earlier version of the struct to help with decoding earlier blocks' metadata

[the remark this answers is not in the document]
pro-wh: oo shouldn't return here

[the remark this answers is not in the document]
pro-wh: comment updated

analyzer/consensus/consensus.go
```diff
@@ -347,8 +349,24 @@ func (m *processor) ProcessBlock(ctx context.Context, uheight uint64) error {
...
+		)
+		// We only try to unmarshal into the current version of the metadata
+		// structure (oasis-core Eden + CometBFT at time of writing). This may
+		// fail on blocks from an incompatible earlier version. Skip indexing
+		// the block metadata in that case.
+	}
+
+	var proposerAddr *string
+	if cmtMeta.Header != nil {
+		proposerAddr = common.Ptr(cmtMeta.Header.ProposerAddress.String())
```
Andrew7234, line 366: Can we convert this to an oasis-address (assuming we just feed the bytes to a bech32 derivation) and store it like that instead? It'd be nice to avoid another address format and to keep things consistent, but it'd also be v helpful for supporting queries on signed blocks/proposed blocks/uptime for a given validator.

Same for the `prevSigners` addresses below

Also, does the `proposerAddr` address refer to the node_id (type `base64_ed25519_pubkey` in the db), but encoded in hex?

storage/migrations/14_block_meta.up.sql
```diff
@@ -0,0 +1,19 @@
...
+ALTER TABLE chain.blocks
+    ALTER COLUMN block_hash DROP NOT NULL,
+    ALTER COLUMN time DROP NOT NULL,
+    ALTER COLUMN num_txs DROP NOT NULL,
+    ALTER COLUMN namespace DROP NOT NULL,
+    ALTER COLUMN version DROP NOT NULL,
+    ALTER COLUMN type DROP NOT NULL,
+    ALTER COLUMN root_hash DROP NOT NULL,
+    ADD COLUMN proposer_node_consensus_pubkey_address TEXT,
+    ADD COLUMN signer_node_consensus_pubkey_addresses TEXT[];
```
Andrew7234, line 17: Heads up - we have a request to support queries for proposed/signed blocks for a given validator (https://app.clickup.com/t/86941rbrt), and I imagine we'll have to add indexes on these two columns to support that (gin index for the latter). Your call on whether you want to add that in this PR though.

storage/migrations/14_block_meta.up.sql
```diff
@@ -0,0 +1,19 @@
+BEGIN;
+
+ALTER TABLE chain.nodes
+    ADD COLUMN consensus_pubkey_address TEXT;
+
+CREATE INDEX ix_nodes_consensus_pubkey_address ON chain.nodes (consensus_pubkey_address);
+
+ALTER TABLE chain.blocks
+    ALTER COLUMN block_hash DROP NOT NULL,
```
ptrus, line 9: Maybe add a comment on why these can now be null? Due to `ConsensusBlockSignersUpsert` inserting metadata for previous block, which doesn't necessarily yet exists (due to no order guarantees when processing blocks in fast-sync) right?

ptrus: Alternatively we could have also split out the metadata into a separate table, I guess that was considered and decided against it?

pro-wh: oh yeah could use a separate table

analyzer/consensus/consensus.go
```diff
@@ -349,8 +351,24 @@ func (m *processor) ProcessBlock(ctx context.Context, uheight uint64) error {
...
 func (m *processor) queueBlockInserts(batch *storage.QueryBatch, data *consensusBlockData) error {
+	var cmtMeta cometbft.BlockMeta
+	if err := cbor.Unmarshal(data.BlockHeader.Meta, &cmtMeta); err != nil {
+		m.logger.Warn("could not unmarshal block meta, may be incompatible version",
+			"height", data.BlockHeader.Height,
+			"err", err,
+		)
+		// We only try to unmarshal into the current version of the metadata
+		// structure (oasis-core Eden + CometBFT at time of writing). This may
+		// fail on blocks from an incompatible earlier version. Skip indexing
```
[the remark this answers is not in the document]
pro-wh: ok yeah

storage/client/client.go
```diff
@@ -309,10 +330,35 @@ func (c *StorageClient) Blocks(ctx context.Context, r apiTypes.GetConsensusBlock
...
-		if err := res.rows.Scan(&b.Height, &b.Hash, &b.Timestamp, &b.NumTransactions); err != nil {
+		var metadata json.RawMessage
+		var proposerRow nodeInfoRow
+		var signerRows []nodeInfoRow
+		if err := res.rows.Scan(
+			&b.Height,
+			&b.Hash,
+			&b.Timestamp,
+			&b.NumTransactions,
+			&metadata,
```
mitjat, line 341: Currently unused; let's not pull it from the DB in the first place. (I'm supportive of having the raw meta field in the db in case we need to extract more info from it later)

api/spec/v1.yaml
mitjat: Ah yes, I see e2e fails with exactly the expected diffs :+1:

storage/client/client.go
pro-wh: woa is it not used?

pro-wh: huh did I intend to add that to the response or something

storage/migrations/14_block_meta.up.sql
mitjat, line 17: Can we add a comment about why this is not `oasis_addr[]`, and what the actual values are here?

Alternatively, let's store the oasis addr here (as suggested in https://github.com/oasisprotocol/nexus/pull/646/files#r1591463407 by Andy). @jberci said (and Warren probably knows well) that we can derive this on the fly, rather than looking it up from the DB. The downside is that we're less robust if very hypothetically oasis-core were to change the derivation scheme; negligible risk, IMO.

go.mod
```diff
@@ -170,6 +170,7 @@ require (
 
 require (
 	github.com/akrylysov/pogreb v0.10.1
+	github.com/cometbft/cometbft v0.0.0-00010101000000-000000000000
```
Andrew7234, line 173: Is there a reason for this version? It seems quite old: https://pkg.go.dev/github.com/cometbft/cometbft?tab=versions

storage/migrations/28_block_meta.up.sql
```diff
@@ -0,0 +1,32 @@
+BEGIN;
+
+CREATE TABLE chain.block_nodes
+(
+    id      TEXT PRIMARY KEY, -- The consensus address of the signer/proposer node.
+    address oasis_addr NOT NULL UNIQUE, -- The oasis address of the node (can't be directly derived from the consensus address).
+    CONSTRAINT address_combo UNIQUE (id, address)
+);
```
Andrew7234, line 8: Neat idea! Yeah fast-sync is annoying to deal with. Although if we use the node consensusAddr -> entityID mapping, we hopefully wouldn't need this or `block_signer_links` since we can store the entityID directly.

api/spec/v1.yaml
```diff
@@ -1356,6 +1362,17 @@ components:
       description: |
         A consensus block.
 
+    NodeInfo:
+      type: object
+      properties:
+        node_id:
```
Andrew7234, line 1368: `node_id` is not needed to satisfy the consensus mocks, let's leave it out since it simplifies things. (Plus, i doubt most users would be able to use it, `entity_id` is more helpful since it corresponds to the validator)

Andrew7234 (commented):
Thanks! Looking better and better, left a suggestion about tracking consensusAddr -> entity ID and some follow-up comments off of that. Let me know what you think!

analyzer/consensus/consensus.go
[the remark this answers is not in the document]
Andrew7234: Damask worked for me, didn't try cobalt though

analyzer/queries/queries.go
```diff
@@ -166,9 +166,30 @@ var (
...
+      proposer_entity_id = excluded.proposer_entity_id`
+
+	ConsensusBlockSignersUpsert = `
+    INSERT INTO chain.blocks
+      (height, signer_entity_ids)
+    VALUES
+      ($1, $2)
+    ON CONFLICT (height) DO UPDATE
+    SET
+      signer_entity_ids = excluded.signer_entity_ids`
```
Andrew7234, line 192: Would this work if we used an `UPDATE` instead? We could then keep the `not null` constraints for the other `chain.blocks` columns. I _think_ they were originally removed to support updating the signers for the `from`-1 block (since the consensus analyzer didn't analyze the block before `from`, it won't have any of the fields populated). As a bonus, the migration becomes shorter/faster.

Andrew7234 (approved):
Looks good! There's a few remaining comments from ptrus/mitja which need to be addressed, but hopefully those are minor changes. Thanks

storage/migrations/32_block_meta.up.sql
```diff
@@ -0,0 +1,10 @@
+BEGIN;
+
+ALTER TABLE chain.blocks
+    ADD COLUMN proposer_entity_id base64_ed25519_pubkey,
+    ADD COLUMN signer_entity_ids base64_ed25519_pubkey[];
```
ptrus, line 5: What was the reasoning for storing Entity IDs instead of Node IDs here? 

A validator entity can have multiple validator nodes registered. In that case, we lose information about which exact node signed the block. 

In practice, only a single node will be active at any given epoch (since the network limits to one at this point), so this shouldn't be that problematic. However, I feel we could easily have used node IDs here to avoid this.

ptrus: Actually, it probably makes sense/is fine, since we mostly don't care about individual validator nodes in other places/in the API.

pro-wh: yeah, it's for reasons of API simplicity like that. the general style of nexus is to have analysis precompute stuff that the API side will need, even sometimes at the cost of lossiness

Andrew7234: > we mostly don't care about individual validator nodes in other places/in the API.

Yeah it was a combination of this, also because we don't track historical validator entity<>node mappings (though it is possible to add now in the `history.validators` table).

storage/migrations/00_consensus.up.sql
```diff
@@ -25,14 +25,14 @@ CREATE DOMAIN public.eth_addr BYTEA CHECK(length(VALUE) = 20);
 CREATE TABLE chain.blocks
 (
   height     UINT63 PRIMARY KEY,
-  block_hash HEX64 NOT NULL,
-  time       TIMESTAMP WITH TIME ZONE NOT NULL,
-  num_txs    UINT31 NOT NULL,
+  block_hash HEX64 NOT NULL,  -- NULL in 32_block_meta.up.sql
```
ptrus, line 28: These comments are wrong  `32_block_meta.up.sql` does not make these fields `NULL`-able.

analyzer/consensus/consensus.go
```diff
@@ -380,8 +422,32 @@ func (m *processor) queueBlockInserts(batch *storage.QueryBatch, data *consensus
...
+				m.logger.Warn("could not convert block signer address to entity id (address not found)",
+					"height", data.BlockHeader.Height,
+					"signer", cs.ValidatorAddress.String(),
+				)
+				continue
+			}
+			prevSigners = append(prevSigners, entity.String())
+		}
+		batch.Queue(
+			queries.ConsensusBlockAddSigners,
```
ptrus, line 445: This doesn't work correctly in the fast-sync case. When in fast-sync, blocks are not processed in order, so the previous block (`cmtMeta.LastCommit.Height`) might not exist, and therefore this query won't insert any singers.

Some quick suggestion on how to fix this:
- doing an Upsert, and making all other fields NULL-able (I think a previous version of this PR was done in that way)
- storing the signers in a separate table 

I noticed this in E2E tests of https://github.com/oasisprotocol/nexus/pull/764 since some blocks are missing signers in the `edenfast` test case.

PR #650, opened 2024-02-28
[analyzer] return default message if no revert reason specified

Andrew7234:
Per discussion on slack, we want to return a more human-friendly default error message if no transaction revert reason is provided. 

Note:
* This PR only handles the case where the error message is `reverted: `. It does not handle the case where the error message is entirely empty, since this should never occur (staging mainnet has no instances of a failed tx with a rawErrMessage of `""`)

Testing:
* Unfortunately our e2e_regression test ranges don't include a suitable failed tx with an empty revert message. Tested manually on the [originally reported transaction](https://explorer.oasis.io/testnet/sapphire/tx/0x872d95e14bd497f35209539341958346a4e20ddf66d1ef2f282251294aa5d3d0):

```
~/oasis/oasis-block-indexer(andrew7234/default-runtime-tx-err*) » curl "http://localhost:8008/v1/sapphire/transactions/872d95e14bd497f35209539341958346a4e20ddf66d1ef2f282251294aa5d3d0" | jq
100  2249    0  2249    0     0   107k      0 --:--:-- --:--:-- --:--:--  244k
{
  "total_count": 1,
  "transactions": [
    {
      "amount": "0",
      "body": {
        "address": "llRjnKXeSEg9muiB+exn7IR/G3o=",
        "data": "OALN1QAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAgAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAGAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAADIAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAOgAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAOAAAAAAAAAAAAAAAACx/yhbXkLNKgq/Z+RVLPWmmG7bpAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAEgAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAYAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAACQAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAY2DMN3FAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAoAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAElByb29mLW9mLUF1dGhvcml0eQAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAC5RbWNvMUI4QWFHVEttSnJrVjNYdEp6c3FWRmFKR3B6UjhhU056RnJ4TTNjZDhzAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAEAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAIAAAAAAAAAAAAAAAALH/KFteQs0qCr9n5FUs9aaYbtukAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAEAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAnt9AAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAdkYW9zaWduAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAACe30AAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAQWwftSr0/smPCfeMrrBKn85WcjfkOPC3e+9HVm6wUnY6SgZ0fyq9gfBtfb5qL61LIeq7lH3lqj6k5Tu24NFaPzQcAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAuUW1aNm41Nkg5THhQcnVDUWRlRVJVazFpdVZvNzRrRWFuZFpxdnk2dGp1cFNQMwAAAAAAAAAAAAAAAAAAAAAAAA==",
        "value": ""
      },
      "charged_fee": "30870585000000000",
      "error": {
        "code": 8,
        "message": "reverted without a message",
        "module": "evm"
      },
      "eth_hash": "872d95e14bd497f35209539341958346a4e20ddf66d1ef2f282251294aa5d3d0",
      "fee": "10100250000000000000",
      "gas_limit": 10050000,
      "gas_used": 30717,
      "hash": "7b8f21f8f233412855758ae5a8a5f5032649b0dffa578a69b4516bc762e82ccb",
      "index": 0,
      "method": "evm.Call",
      "nonce_0": 10,
      "round": 5029734,
      "sender_0": "oasis1qqmkkw69a56lqsrqqhvetknlup3gcl6885g8lqjm",
      "sender_0_eth": "0x1e59ce931B4CFea3fe4B875411e280e173cB7A9C",
      "size": 1197,
      "success": false,
      "timestamp": "2024-02-23T05:46:50-05:00",
      "to": "oasis1qqd94y2hqwl48t4sh7ck2f72y8g06f42fvz82lmp",
      "to_eth": "0x9654639cA5De48483D9Ae881f9ec67eC847f1b7A"
    }
  ]
}
```

### DIFF
--- a/analyzer/evmabibackfill/evm_abi_backfill.go
+++ b/analyzer/evmabibackfill/evm_abi_backfill.go
@@ -164,7 +164,6 @@ func (p *processor) parseEvent(ev *abiEncodedEvent, contractAbi abi.ABI) (*strin
 	}
 	eventArgs, err := marshalArgs(abiEvent.Inputs, abiEventArgs)
 	if err != nil {
-		p.logger.Warn("error processing event args using abi", "err", err)
 		return nil, nil, nil, fmt.Errorf("error processing event args using abi: %w", err)
 	}
 
@@ -186,9 +185,10 @@ func (p *processor) parseTxCall(tx *abiEncodedTx, contractAbi abi.ABI) (*string,
 	return &method.RawName, txArgs, nil
 }
 
-// Attempts to parse the transaction revert reason into the error name and args
-// as defined by the abi of the contract that was called.
-func (p *processor) parseTxErr(tx *abiEncodedTx, contractAbi abi.ABI) (*string, []*abiEncodedArg) {
+// Attempts to parse the transaction revert reason into the error name (or plaintext message;
+// see chain.runtime_transactions.error_message in DB) and args as defined by the abi of the
+// contract that was called.
+func (p *processor) parseTxErr(tx *abiEncodedTx, contractAbi abi.ABI) (*string, []*abiEncodedArg, error) {
 	var abiErrMsg string
 	var abiErr *abi.Error
 	var abiErrArgs []interface{}
@@ -200,22 +200,30 @@ func (p *processor) parseTxErr(tx *abiEncodedTx, contractAbi abi.ABI) (*string, 
 			// as "reverted: Ownable: caller is not the owner". In this case, we do
 			// not parse the error with the abi.
 			p.logger.Info("encountered likely old-style reverted transaction", "revert reason", tx.TxRevertReason, "tx hash", tx.TxHash, "err", err)
-			return nil, nil
+			return nil, nil, nil
+		}
+		// If no revert reason was provided, we return nil here to skip the update
+		// and preserve the existing error_message, which should be the default
+		// tx error message defined in runtime.DefaultTxRevertErrMsg.
+		//
+		// Note: Although we could return the default error message here, conceptually
+		// the abi analyzer should only update fields that were parsed using the abi.
+		if len(txrr) == 0 {
+			p.logger.Info("no revert reason provided, skipping update")
+			return nil, nil, nil
 		}
 		abiErr, abiErrArgs, err = abiparse.ParseError(txrr, &contractAbi)
 		if err != nil || abiErr == nil {
-			p.logger.Warn("error processing tx error using abi", "contract address", "err", err)
-			return nil, nil
+			return nil, nil, fmt.Errorf("error processing error using abi: %w", err)
 		}
 		abiErrMsg = runtime.TxRevertErrPrefix + abiErr.Name + prettyPrintArgs(abiErrArgs)
 		errArgs, err = marshalArgs(abiErr.Inputs, abiErrArgs)
 		if err != nil {
-			p.logger.Warn("error processing tx error args", "err", err)
-			return nil, nil
+			return nil, nil, fmt.Errorf("error marshalling tx err args: %w", err)
 		}
 	}
 
-	return &abiErrMsg, errArgs
+	return &abiErrMsg, errArgs, nil
 }
 
 func (p *processor) ProcessItem(ctx context.Context, batch *storage.QueryBatch, item *abiEncodedItem) error {
@@ -248,7 +256,11 @@ func (p *processor) ProcessItem(ctx context.Context, batch *storage.QueryBatch, 
 			p.logger.Warn("error parsing tx with abi", "err", err, "contract_address", item.ContractAddr, "tx_hash", item.Tx.TxHash)
 			// Write to the DB regardless of error so we don't keep retrying the same item.
 		}
-		errMsg, errArgs := p.parseTxErr(item.Tx, contractAbi)
+		errMsg, errArgs, err := p.parseTxErr(item.Tx, contractAbi)
+		if err != nil {
+			p.logger.Warn("error parsing tx err with abi", "err", err, "contract_address", item.ContractAddr, "tx_hash", item.Tx.TxHash)
+			// Write to the DB regardless of error so we don't keep retrying the same item.
+		}
 		batch.Queue(
 			queries.RuntimeTransactionEvmParsedFieldsUpdate,
 			p.runtime,
--- a/analyzer/evmabibackfill/evm_abi_backfill_test.go
+++ b/analyzer/evmabibackfill/evm_abi_backfill_test.go
@@ -1,0 +1,221 @@
+package evmabibackfill
+
+import (
+	"encoding/base64"
+	"encoding/json"
+	"testing"
+
+	ethCommon "github.com/ethereum/go-ethereum/common"
+
+	"github.com/oasisprotocol/oasis-sdk/client-sdk/go/modules/evm"
+	"github.com/stretchr/testify/require"
+
+	"github.com/oasisprotocol/nexus/analyzer/evmabi"
+	"github.com/oasisprotocol/nexus/common"
+	"github.com/oasisprotocol/nexus/log"
+)
+
+// TODO: Update WROSE source and add tests for more txs, mismatched txs, and custom errors
+
+type mockParsedEvent struct {
+	Name *string
+	Args []*abiEncodedArg
+	Sig  *ethCommon.Hash
+}
+
+type mockParsedTxCall struct {
+	Name *string
+	Args []*abiEncodedArg
+}
+
+func unmarshalEvmEvent(t *testing.T, body []byte) *abiEncodedEvent {
+	var ev evm.Event
+	err := json.Unmarshal(body, &ev)
+	require.Nil(t, err)
+	return &abiEncodedEvent{
+		Round:     0,
+		TxIndex:   nil,
+		EventBody: ev,
+	}
+}
+
+func unmarshalEvmTx(t *testing.T, body string, error_message *string) *abiEncodedTx {
+	txData, err := base64.StdEncoding.DecodeString(body)
+	require.Nil(t, err)
+	return &abiEncodedTx{
+		TxHash:         "", // not relevant for current unit tests
+		TxData:         txData,
+		TxRevertReason: error_message,
+	}
+}
+
+func TestParseEvent(t *testing.T) {
+	abi := evmabi.WROSE
+	p := &processor{
+		runtime: common.RuntimeSapphire,
+		target:  nil,
+		logger:  log.NewDefaultLogger("testing"),
+	}
+	ev := unmarshalEvmEvent(t, []byte(`{"data": "AAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAADCopscuTr4A=", "topics": ["3fJSrRviyJtpwrBo/DeNqpUrp/FjxKEWKPVaTfUjs+8=", "AAAAAAAAAAAAAAAAw3+DQaxuSpRTgwK81NSc8IUtMMA=", "AAAAAAAAAAAAAAAAWZiaj7/4vj0d6v/ytGtmtjrLeX4="], "address": "i8KwMLKZlk7vteHgs2mRNS5W0tM="}`))
+	expected := mockParsedEvent{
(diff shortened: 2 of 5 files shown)


analyzer/evmabibackfill/evm_abi_backfill.go
```diff
@@ -202,6 +202,10 @@ func (p *processor) parseTxErr(tx *abiEncodedTx, contractAbi abi.ABI) (*string,
 			p.logger.Info("encountered likely old-style reverted transaction", "revert reason", tx.TxRevertReason, "tx hash", tx.TxHash, "err", err)
 			return nil, nil
 		}
+		// Return a default error message if no revert reason was provided.
+		if len(txrr) == 0 {
+			return common.Ptr(runtime.DefaultTxRevertErrMsg), nil
```
mitjat, line 207: Nit: Per docstring, we're returning the error _name_ (with no args) here. IIRC that's because we abuse the error name field to encode unstructured errors' (= old-style) messages. Edit: yeah, documented in [the DB](https://github.com/oasisprotocol/nexus/blob/b102297b96db6cebc68023a66e38e984affe2075/storage/migrations/01_runtimes.up.sql#L81-L84) and [the API](https://github.com/oasisprotocol/nexus/blob/b102297b96db6cebc68023a66e38e984affe2075/api/spec/v1.yaml#L1541-L1541). But other code maintainers won't know the lore.
Maybe the function docstring here can also say "parse ... into the error name (or plaintext message; see chain.runtime_transactions.error_message in DB)"?

mitjat (commented):
Thank you, looks good!

Let's follow our belated New Year's resolution though: unit tests.
They don't need to be big, but we should get into the habit more. You can use white-box tests: For example, `evm_abi_backfill_test.go` will start with `package evmabibackfill` and can exercise the private parseTxErr directly.
We have a few ABIs lying around already if you need them for testing, e.g. the erc20 one.



I'm surprised there are no regression test changes. Actually, of the last 10_000_000 runtime txs, none reverted with `"reverted: "` :thinking: . But there are some interesting ones (only those rows preserved below)
```
select *, count(*) cnt from (select error_message from chain.runtime_transactions limit 10000000) x where error_message like 'reverted%' group by 1 order by cnt desc;
                                    error_message                                    |  cnt  
-------------------------------------------------------------------------------------+-------
 reverted: unknown                                                                   | 99353
 reverted: no revert reason                                                          |   725
 reverted: -                                                                         |    62
```

"reverted: unknown" is also the most common of all revert reasons. Maybe the "unknown" is an artifact of our own parsing? Please check what the unit tests return if your code diff is not in place, I'm curious. Hm or maybe it's from the olden runtimes that tried to decode the revert reason themselves, always assuming the ABI is `Error(string)`. When I look at modern txs, I see no instances of that revert reason.

My guess is that the other two "non-reasons" in the table above are manual strings by smart contract authors.

mitjat:
Oh haha I see we do have a unit test for extract.go already, and it's failing in CI because it wants a `nil` output :)

Andrew7234:
> Let's follow our belated New Year's resolution though: unit tests.

Good call; I'll add unit tests to the abi analyzer!

> "reverted: unknown" is also the most common of all revert reasons... maybe it's from the olden runtimes

Looks like the `reverted: unknown` was indeed from the cobalt days (damask started at 1003298). We _could_ also insert the default error message when we see this case, but given that those txs were from so long ago I'm leaning against it. This also avoids another special-case we'd have to think about going forward.

```
oasisindexer=> select runtime, round, error_message_raw from chain.runtime_transactions where error_message='reverted: unknown' order by round desc limit 100;
 runtime |  round  | error_message_raw
---------+---------+-------------------
 emerald | 1003293 | reverted: unknown
```

analyzer/evmabibackfill/evm_abi_backfill.go
Andrew7234: Added!

analyzer/runtime/extract_test.go
```diff
@@ -76,9 +76,14 @@ func TestUnknownError(t *testing.T) {
...
-	msg := tryParseErrorMessage(input.module, input.code, input.message)
-	require.Nil(t, msg, "tryParseErrorMessage extracted a message when it should not have")
+	inputs := map[string]mockError{
+		DefaultTxRevertErrMsg: {module: "evm", code: 8, message: "reverted: "},
+	}
+	for parsed, input := range inputs {
+		msg := tryParseErrorMessage(input.module, input.code, input.message)
+		require.NotNil(t, msg, "failed to extract default error message")
+		require.Equal(t, parsed, *msg, "tryParseErrorMessage extracted a message that differed from the expected default error message")
+	}
```
mitjat, line 86: We don't need a `map` and a `for` loop here, nor do we expect to need it in the future (I think?). So let's keep the simpler, more readable structure without loops.

analyzer/evmabibackfill/evm_abi_backfill_test.go
```diff
@@ -0,0 +1,298 @@
...
+	}
+
+	name1 := "Transfer"
+	sig1 := ethCommon.HexToHash("0xddf252ad1be2c89b69c2b068fc378daa952ba7f163c4a11628f55a4df523b3ef")
+	name2 := "Withdrawal"
+	sig2 := ethCommon.HexToHash("0x7fcf532c15f0a6db0bd6d0e038bea71d30d808c7d98cb3bf7268a95bf5081b65")
+	name3 := "Deposit"
+	sig3 := ethCommon.HexToHash("0xe1fffcc4923d04b559f4d29a8bfc6cda04eb5b0d3c460751c2402c5c5cc9109c")
+	name4 := "Approval"
+	sig4 := ethCommon.HexToHash("0x8c5be1e5ebec7d5bd14f71427d1e84f3dd0314c0f7b2291e5b200ac8c7c3b925")
```
mitjat, line 73: These are not reused anywhere, so let's inline them (with `common.Ptr()`, if needed).

In general, the rule of thumb with unit test is to keep control flows as simple as possible, inline things as much as possible, reduce the number of helper funcs, etc, all with the goal of making the test linearly readable. Tests of that kind are easier to fix and (IMO even more importantly) read, so they can serve as pretty good documentation on how a function is expected to be used, without having to wade through 100 abstraction layers in the actual non-test program.

analyzer/evmabibackfill/evm_abi_backfill_test.go
```diff
@@ -0,0 +1,298 @@
...
+}
+
+func TestParseUnknownEvent(t *testing.T) {
+	abi := evmabi.WROSE
+	p := &processor{
+		runtime: common.RuntimeSapphire,
+		target:  nil,
+		logger:  log.NewDefaultLogger("testing"),
+	}
+	// Non-WROSE event.
```
mitjat, line 166:
```suggestion
	// Non-WROSE event; it doesn't fit the `abi` and should fail to parse.
```

analyzer/evmabibackfill/evm_abi_backfill_test.go
```diff
@@ -0,0 +1,298 @@
...
+
+func TestParseTransactionCall(t *testing.T) {
+	abi := evmabi.WROSE
+	p := &processor{
+		runtime: common.RuntimeSapphire,
+		target:  nil,
+		logger:  log.NewDefaultLogger("testing"),
+	}
+	txs := []*abiEncodedTx{
+		// unmarshalEvmTx(t, "omRib2R5o2Jwa1gg6KR1JehYdjBZyYOH06CycJIxAuTmmY22C6nLtybyPTlkZGF0YVgbjk8cKK1D/PuiREnFTOkHm4YosIhcDHtRzPkPZW5vbmNlT95YOkkjR3eTTH7QR/O8tmZmb3JtYXQB", nil), // The WROSE abi in the nexus repository is out of date/incomplete.
```
mitjat, line 185: What does this base64 represent? Unless it covers an important new aspect of ABI parsing, let's just remove it from here.

analyzer/evmabibackfill/evm_abi_backfill_test.go
```diff
@@ -0,0 +1,298 @@
...
+}
+
+func TestParseEvent(t *testing.T) {
+	abi := evmabi.WROSE
+	p := &processor{
+		runtime: common.RuntimeSapphire,
+		target:  nil,
+		logger:  log.NewDefaultLogger("testing"),
+	}
+	evs := []*abiEncodedEvent{
```
mitjat, line 59: Do these test cases cover different code paths inside `p.parseEvent()`? If not, we can prune this down to just one case. The regression tests have more of a more-is-better philosophy; here, it's nice to have fewer tests so that 1) the intent behind each one can be easier to understand and 2) there's less to fix/maintain/update if the code changes.

OTOH if it does make sense to keep multiple test cases, it would be really nice to keep the inputs (= this array) next to the expected outputs. Basically, the `mockParsedEvent` type would gain a new field `Input` or sth. This [short section 10](https://www.practical-go-lessons.com/chap-19-unit-tests#how-to-write-a-table-test) gives a nice example and motivation, and also explains how to use `t.Run()` so we don't get failure messages from `evm_abi_backfill_test:155` (inside the test's final for loop) without knowing which iteration of the loop failed.

mitjat (approved):
Thank you for the added tests! I was trying to request just a test for what this PR fixes (= parsing of `reverted: `), but of course this is even nicer :). I see the `reverted: ` test is still TODO but that should be easy now that you have all the infrastructure set up in the new file.

No good deed goes unpuninshed, so with more tests, I have more feedback/tips. Nothing major though, and thanks again!

analyzer/evmabibackfill/evm_abi_backfill_test.go
Andrew7234: It's a more complex tx call from the actual wrose abi; but it shouldn't test any new codepaths so I'll remove it. Can revisit if/when we update the wrose abi in nexus.

analyzer/evmabibackfill/evm_abi_backfill_test.go
Andrew7234: Got it, pruned down to a `Transfer` event!

PR #651, opened 2024-03-01
analyzer/consensus: Do not bump nonce for consensus.Meta txs

mitjat:
The `consensus.Meta` tx is special, and calling it does not increase the nonce. This PR fixes nexus's dead-reckoning of the nonce accordingly.

Implementation details in oasis-core:
- `consensus.Meta` is special-handled [here](https://github.com/oasisprotocol/oasis-core/blob/fba5c6cc0b2038860333eb09a4c7ccf560150c70/go/consensus/cometbft/abci/transaction.go#L59-L59)
- The rest of txs bump the nonce in `txAuthHandler.AuthenticateTx()`, same file, line 77.
  - There is only one non-test implementation of `AuthenticateTx()`. It is set for "the abci app" (= _all_ of oasis, as it presents itself to cometbft), not just for staking, [here](https://github.com/oasisprotocol/oasis-core/blob/fba5c6cc0b2038860333eb09a4c7ccf560150c70/go/consensus/cometbft/staking/staking.go#L429-L429), and increases the nonce [here](https://github.com/oasisprotocol/oasis-core/blob/fba5c6cc0b2038860333eb09a4c7ccf560150c70/go/consensus/cometbft/apps/staking/state/gas.go#L116-L116)
  
The PR also improves the documentation around nonces.

Testing: None post-fix. I discovered the bug when inspecting statecheck output.


pro-wh (approved):
as discussed

Andrew7234 (approved):

PR #652, opened 2024-03-01
statecheck: Use BigInt, not uint64, for balances

mitjat:
Consensus statecheck was failing in part because it was unable to pull large (>uint64) balances from the db into internal uin64 variables:
```
      Error:          Received unexpected error:
                      can't scan into dest[1]: cannot scan 70039999999999963480000000 into *uint64
```

**This PR:**
- fixes that bug in statecheck/consensus
- avoids some less critical unneccessary BigInt->uint64 conversions in statecheck/runtime
- tightens the test a little; we now do NOT ignore accounts with 0 balance but non-0 nonce
- (no-op: renames some cryptically-named variables)

**Testing:** Deployed a statecheck manually, verified in the output that type errors were gone.

### DIFF
--- a/common/types.go
+++ b/common/types.go
@@ -34,6 +34,10 @@ func (b BigInt) Times(other BigInt) BigInt {
 	return BigInt{result}
 }
 
+func (b BigInt) IsZero() bool {
+	return b.Sign() == 0
+}
+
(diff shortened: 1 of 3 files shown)


pro-wh (approved):
changes lg

Andrew7234 (approved):

PR #656, opened 2024-03-08
[consensus] add height/timestamp to governance votes

Andrew7234:
Resolves https://app.clickup.com/t/8693prg5a

This PR adds two additional fields, `height` and `timestamp`, to the `ProposalVote` object in the api. 

As Mitja pointed out in the ticket, the "api fields will be optional because the votes might have been observed via the genesis doc rather than via events, in which case the voting height [is not available](https://github.com/oasisprotocol/nexus/blob/f6af89d1101fd6c983ade51755a44dee0b6d0597/analyzer/consensus/genesis.go#L571-L572)."

This is the case for the e2e_regression tests, which query for a proposal that was observed via genesis doc. 

Testing: e2e_regression tests, also I inserted some fake round data into the `votes` table and the resulting query looks like 
```
» curl "http://localhost:8008/v1/consensus/proposals/2/votes" | jq                                 andrewlow@Beleriand
100 11956    0 11956    0     0   249k      0 --:--:-- --:--:-- --:--:--  277k
{
  "proposal_id": 2,
  "total_count": 90,
  "votes": [
    {
      "address": "oasis1qp0j5v5mkxk3eg4kxfdsk8tj6p22g4685qk76fw6",
      "height": 16818000,
      "timestamp": "2023-11-29T03:31:10-08:00",
      "vote": "yes"
    },
    {
      "address": "oasis1qp0xuvw2a93w4yp8jwthfz93gxy87u7hes9eu2ev",
      "height": 16818000,
      "timestamp": "2023-11-29T03:31:10-08:00",
      "vote": "yes"
    },
```

### DIFF
--- a/storage/client/queries/queries.go
+++ b/storage/client/queries/queries.go
@@ -246,10 +246,11 @@ const (
 			WHERE id = $1::bigint`
 
 	ProposalVotes = `
-		SELECT voter, vote
-			FROM chain.votes
+		SELECT votes.voter, votes.vote, votes.height, blocks.time
+			FROM chain.votes as votes
+			LEFT JOIN chain.blocks as blocks ON votes.height = blocks.height
 			WHERE proposal = $1::bigint
-		ORDER BY proposal DESC
+		ORDER BY height DESC, voter ASC
 		LIMIT $2::bigint
 		OFFSET $3::bigint`
 
--- a/storage/migrations/11_governance_votes.up.sql
+++ b/storage/migrations/11_governance_votes.up.sql
@@ -1,0 +1,6 @@
+BEGIN;
+
+ALTER TABLE chain.votes
+    ADD COLUMN height UINT63; -- Can be NULL; when the vote comes from a genesis file, height is unknown.
+
+COMMIT;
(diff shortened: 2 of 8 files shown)


storage/client/queries/queries.go
```diff
@@ -246,10 +246,11 @@ const (
...
 
 	ProposalVotes = `
-		SELECT voter, vote
-			FROM chain.votes
+		SELECT votes.voter, votes.vote, votes.height, blocks.time
+			FROM chain.votes as votes
+			LEFT JOIN chain.blocks as blocks ON votes.height = blocks.height
 			WHERE proposal = $1::bigint
-		ORDER BY proposal DESC
+		ORDER BY height DESC, voter ASC
```
Andrew7234, line 253: proposal is required in the query; changed it here for regression test output consistency

storage/migrations/11_governance_votes.up.sql
```diff
@@ -0,0 +1,6 @@
+BEGIN;
+
+ALTER TABLE chain.votes
+    ADD COLUMN height UINT63;
```
mitjat, line 4:
```suggestion
    ADD COLUMN height UINT63; -- Can be NULL; when the vote comes from a genesis file, height is unknown.
```

mitjat (approved):
Thank you!

PR #657, opened 2024-03-08
Andrew7234/workflow nits

Andrew7234:
do not merge (yet)

This PR has misc tweaks/fixes for things that were broken in my local dev setup (zsh on Mac OS). I'm not sure if all of it should be merged; just wanted to start a discussion. I tried to separate different issues by commit


tests/e2e_regression/ensure_consistent_config.sh
```diff
@@ -19,7 +19,7 @@ shopt -s expand_aliases
 alias yaml2json="python3 -c 'import sys,yaml,json; print(json.dumps(yaml.safe_load(str(sys.stdin.read()))))'" 
 
 # Compare
-for range in $(find "$directory" -type d); do
+for range in $(find "tests/e2e_regression" -maxdepth 1 -mindepth 1 -type d); do
```
Andrew7234, line 22: `./tests/e2e_regression/ensure_consistent_config.sh: line 26: directory: unbound variable`

Makefile
```diff
@@ -110,14 +110,14 @@ accept-e2e-regression-suite:
...
 	rm -rf ./tests/e2e_regression/$(SUITE)/expected; \
-	# Copy the actual outputs to the expected outputs. \
+	# Copy the actual outputs to the expected outputs.
 	cp -r  ./tests/e2e_regression/$(SUITE)/{actual,expected}; \
-	# The result of the "spec" test is a special case. It should always match the \
-	# current openAPI spec file, so we symlink it to avoid having to update the expected \
-	# output every time the spec changes. \
+	# The result of the "spec" test is a special case. It should always match the
+	# current openAPI spec file, so we symlink it to avoid having to update the expected
+	# output every time the spec changes.
```
Andrew7234, line 120: I think the `\` at the end of commented lines caused subsequent non-comment lines to be treated as a comment and skipped

tests/e2e_regression/ensure_consistent_config.sh
pro-wh: woa how did this used to work

Makefile
pro-wh: what was the intention with these? run the whole thing in one shell instead of several?

Andrew7234: hm I'm not sure.. maybe @mitjat  knows? I'm also curious as to whether it works locally for y'all as-is

pro-wh: just tried on linux with GNU bash, version 5.2.21(1)-release and GNU Make 4.3, the lower lines seem to run normally

tests/e2e_regression/ensure_consistent_config.sh
mitjat: :raised_eyebrow: yeah the old code makes no sense

Makefile
mitjat: Interesting. I don't remember why the `\` are there, unfortunately. If this works locally and in CI, LGTM.

Although I'd say:
- If we're doing this, let's remove all `\` in this recipe so we're not using a weird mix.
- Prepend every `#` with `@` or else `make` will echo those comments to the console every time we run the recipe.

tests/e2e_regression/run.sh
```diff
@@ -89,7 +89,7 @@ for ((i = 0; i < nCases; i++)); do
       >/tmp/pretty 2>/dev/null &&
       cp /tmp/pretty "$outDir/$name.body" || true
     # Sanitize the current timestamp out of the response header so that diffs are stable
-    sed -i -E 's/^(Date|Content-Length|Last-Modified): .*/\1: UNINTERESTING/g' "$outDir/$name.headers"
+    sed -i '' -E 's/^(Date|Content-Length|Last-Modified): .*/\1: UNINTERESTING/g' "$outDir/$name.headers"
```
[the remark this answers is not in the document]
mitjat: uhh I remember we've had back-and-forth with `sed -i ''` before :(
This doesn't work on GNU sed (which is what I have locally, and what CI has). It interprets the `''` as the sed script to run. You can try it yourself: `docker run -it bash`, then `echo foo >/tmp/x; sed -i '' 's/o/a/' /tmp/x`.

You could try instead:
- using `echo foo >/tmp/x; sed --in-place='' 's/o/a/' /tmp/x` (though I think that's not supported on MacOS?)
- creating a little helper function that reimplements `sed -i` without using `-i` (*), or using `-i` based on some OS-specific check
- leaving this as-is and using GNU sed on your machine?

(*) something like
```
sed_i() {
    local file="${@: -1}"  # Get the last argument (file path)
    local script="${@: -2:1}" # Get the penultimate argument (sed script)
    local flags="${@:1:$#-2}"  # Get flags (all arguments except the last two)

    local temp_file="$(mktemp)"
    sed "${flags[@]}" "$script" "$file" > "$temp_file"
    mv "$temp_file" "$file"  # Overwrites the original file
}
```

mitjat (approved):
Nice, thank you for all of these! Approving because LGTM (pending comments), but happy to look at it again if you wanted to add more to the same PR.

tests/e2e_regression/run.sh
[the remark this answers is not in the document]
mitjat: @Andrew7234 I think you said you're using GNU sed on your machine currently? So that resolves this thread for now (remember to undo the change), and you can submit the rest of the PR at least. It's useful stuff, would be a shame if it rotted away.

tests/e2e_regression/ensure_consistent_config.sh
pro-wh: changed in #665, this script now takes the name of a suite as a parameter instead of searching around

Makefile
pro-wh: this is moved to a new script e2e_regression/accept.sh, which makes it no longer necessary to use `\`s. it also should quiet down the logging of every line

ptrus:
Not relevant anymore.

PR #653, opened 2024-03-07
[api] return evm addresses in event body when available

Andrew7234:
Part of https://app.clickup.com/t/8693up8ft

This PR updates the events query to augment the body of non-`evm.log` events with evm-style addresses when available. 

Todo:
- `evm.log` events always use evm-style addresses; is there a demand for returning the oasis-style address for those events as well? edit: not yet; will hold off here

Testing: Only manual and e2e_regression tests for now as this is an api-only change. Performance benchmarks on production showed that the new query is slower but still within acceptable bounds. 

Note: 
- The current impl does not scale well as the number of event body fields that can hold an oasis-style address increases, although I don't think it will increase much, if at all. An alternative impl would be to extract all oasis-style addresses from events and do a subsequent db query to fetch the mappings to the eth-style addresses, and then update the event body(s) accordingly. I tried this originally but it got messy quickly, but in theory it should avoid the aforementioned scaling issue and avoid query bloat. 

Sample result:
```
» curl "http://localhost:8008/v1/sapphire/events?type=consensus_accounts.deposit&limit=2" | jq                andrewlow@Beleriand
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
100   739  100   739    0     0  43727      0 --:--:-- --:--:-- --:--:-- 61583
{
  "events": [
    {
      "body": {
        "amount": {
          "Amount": "100000000000000000",
          "Denomination": ""
        },
        "from": "oasis1qry5d5zxs58w9deu5ara5zaamhkzqtg99s00ls90",
        "nonce": 27405,
        "to": "oasis1qquzek9y9guk6plys32s7zvf4g2f8rvn9slu8uye",
        "to_eth": "0x8b0b51A92B0FD6d7A9595f057EdF3D85C76E571c"
      },
      "round": 158613,
      "timestamp": "2023-02-21T09:06:32-08:00",
      "type": "consensus_accounts.deposit"
    },
    {
      "body": {
        "amount": {
          "Amount": "100000000000000000",
          "Denomination": ""
        },
        "from": "oasis1qry5d5zxs58w9deu5ara5zaamhkzqtg99s00ls90",
        "nonce": 27404,
        "to": "oasis1qquzek9y9guk6plys32s7zvf4g2f8rvn9slu8uye",
        "to_eth": "0x8b0b51A92B0FD6d7A9595f057EdF3D85C76E571c"
      },
      "round": 158606,
[...]

PR #658, opened 2024-03-08
[api] shorten query

Andrew7234:
Per https://github.com/oasisprotocol/nexus/pull/653#discussion_r1516723205, this PR shortens the query since the address context/identifier are checked later on anyway. If the check fails, the corresponding address field is nil as expected, so there is no change to api behavior.


pro-wh (approved):

pro-wh:
fewer sources of truth for the eth-compatible context+version :+1:

mitjat (approved):

PR #661, opened 2024-03-13
analyzer/consensus: enumerate all tx body types

Andrew7234:
@mitjat reported that the testnet deployment was generating lots of error logs about being unable to cbor-decode the consensus tx body. Further investigation revealed that there was a change to the `roothash.ExecutorCommit.Commits.Header` struct between damask/eden that was not backwards-compatible but this change slipped through the cracks in #655 . 

The first commit of this PR fixes the known problem. The second commit adds adds all the types just in case. It's verbose but might be easier to maintain in the future.

Todo: test on production testnet [x] also production mainnet [x]

Todo: investigate why it didn't fail on the eden test range, which should have used the latest eden type of roothash.ExecutorCommit edit*: our eden test range doesn't have any of that type of tx

### DIFF
--- a/analyzer/consensus/consensus_test.go
+++ b/analyzer/consensus/consensus_test.go
@@ -78,7 +78,7 @@ func TestMalformedTxBody(t *testing.T) {
 	parsed, err := unpackTxBody(&tx)
 	require.Nil(t, parsed)
 	require.NotNil(t, err)
-	require.Equal(t, "unable to cbor-decode consensus tx body: cbor: invalid additional information 30 for type tag, body: deadbeef", err.Error())
+	require.Equal(t, "unable to cbor-decode consensus tx body: cbor: invalid additional information 30 for type tag, method: staking.Allow, body: deadbeef", err.Error())
 }
 
 func TestMalformedTxMethod(t *testing.T) {
@@ -97,5 +97,5 @@ func TestMalformedTxMethod(t *testing.T) {
 	parsed, err := unpackTxBody(&tx)
 	require.Nil(t, parsed)
 	require.NotNil(t, err)
-	require.Equal(t, "unknown tx method: not_a_valid_method", err.Error())
+	require.Equal(t, "unable to cbor-decode consensus tx body: unknown tx method, method: not_a_valid_method, body: deadbeef", err.Error())
 }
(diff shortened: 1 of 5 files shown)


mitjat:
Everything looks good so far in the last ~1M blocks of Eden, right?

> investigate why it didn't fail on the eden test range, which should have used the latest eden type of roothash.ExecutorCommit

Any new thoughts on this?

Andrew7234:
Turns out there aren't any `roothash.ExecutorCommit` transactions in our eden test range: 
```
indexer=# select *, count(*) as cnt from (select method from chain.transactions) as x group by 1 order by cnt desc;
        method         | cnt
-----------------------+------
 consensus.Meta        | 1000
 registry.RegisterNode |  459
 beacon.VRFProve       |  115
 staking.Transfer      |   18
 staking.ReclaimEscrow |   10
 staking.AddEscrow     |    6
```

analyzer/consensus/consensus_test.go
```diff
@@ -97,5 +97,5 @@ func TestMalformedTxMethod(t *testing.T) {
 	parsed, err := unpackTxBody(&tx)
 	require.Nil(t, parsed)
 	require.NotNil(t, err)
-	require.Equal(t, "unknown tx method: not_a_valid_method", err.Error())
+	require.Equal(t, "unable to cbor-decode consensus tx body: %!w(<nil>), method: not_a_valid_method, body: deadbeef", err.Error())
```
mitjat, line 100: the "<nil>" suggests we should initialize `error` in `consensus.go:382` to `fmt.Errorf("unknown tx method")`?

coreapi/v23.0/keymanager/api/secret.go
```diff
@@ -31,6 +34,21 @@ type EncryptedSecret struct {
 	Ciphertexts map[x25519.PublicKey][]byte `json:"ciphertexts"`
 }
 
+// We implement MarshalJSON here because the the encoding/json library
```
mitjat, line 37: "We" is ambiguous

```suggestion
// XXX: Nexus-specific addition/hack.
// We implement MarshalJSON here because the the encoding/json library
```

mitjat (approved):
Thank you!

PR #662, opened 2024-03-13
Tweak ci-lint workflow

buberdds:
- prevent issues with dependabot PRs
- limit towncrier check to pull_request event to fix an issue with failing workflow for merge commits (push event)
- enable gitlint

### DIFF
--- a/.github/workflows/ci-lint.yaml
+++ b/.github/workflows/ci-lint.yaml
@@ -54,6 +54,9 @@ jobs:
           towncrier check --compare-with origin/${BASE_BRANCH}
         env:
           BASE_BRANCH: ${{ github.base_ref }}
+        if: |
+          github.event_name == 'pull_request' &&
+          github.actor != 'dependabot[bot]'
       - name: Lint Change Log fragments
         run: |
           make lint-changelog
--- a/.github/workflows/ci-lint.yaml
+++ b/.github/workflows/ci-lint.yaml
@@ -39,6 +39,9 @@ jobs:
         with:
           python-version: '3.x'
         # Needed for Towncrier fork to work with 3.12 and above
+      - name: Install gitlint
+        run: |
+          python -m pip install gitlint
       - name: Install setuptools
         run: |
           python -m pip install setuptools
@@ -57,6 +60,11 @@ jobs:
         if: |
           github.event_name == 'pull_request' &&
           github.actor != 'dependabot[bot]'
+      - name: Lint git commits
+        run: |
+          make lint-git
+        # Always run this step so that all linting errors can be seen at once, but skip for dependabot
+        if: always() && github.actor != 'dependabot[bot]'
       - name: Lint Change Log fragments
         run: |
           make lint-changelog
--- a/.changelog/662.process.md
+++ b/.changelog/662.process.md
@@ -1,0 +1,6 @@
+Tweak ci-lint workflow
+
+- limit towncrier check to pull_request event to fix an issue with failing
+  workflow for merge commits triggered by push event
+- enable gitlint
+- prevent dependabot issues related to gitlint and towncrier checks
--- a/Makefile
+++ b/Makefile
@@ -128,12 +128,15 @@ fmt:
 	@goimports -w -local github.com/oasisprotocol/nexus .
 
 # Lint code, commits and documentation.
-lint-targets := lint-go lint-go-mod-tidy lint-changelog
+lint-targets := lint-go lint-go-mod-tidy lint-changelog lint-git
 
 lint-go: codegen-go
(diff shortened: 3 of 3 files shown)


.github/workflows/ci-lint.yaml
```diff
@@ -48,6 +51,14 @@ jobs:
           towncrier check --compare-with origin/${BASE_BRANCH}
         env:
           BASE_BRANCH: ${{ github.base_ref }}
+        if: |
+          github.event_name == 'pull_request' &&
+          github.actor != 'dependabot[bot]'
+      - name: Lint git commits
+        run: |
+          make lint-git
```
mitjat, line 59: I hate lint-git :cry: , it's pedantic for very little payoff. Mostly beneficial for people reading patches in emails in Mutt, Linus style. </rant>

Do we have an autoformat tool, ideally in conjunction with commit hooks? This is not a request for one, just trying to see if there's existing tooling support.

.github/workflows/ci-lint.yaml
```diff
@@ -48,6 +51,14 @@ jobs:
...
+        # Always run this step so that all linting errors can be seen at once, but skip for dependabot
+        if: always() && github.actor != 'dependabot[bot]'
```
mitjat, line 61: We probably need this for some other CI jobs too? See https://github.com/oasisprotocol/nexus/pull/663 and all the jobs that fail there. But if we skip them, I guess we'll have the same problem that I was solving in https://github.com/oasisprotocol/nexus/pull/660 because required jobs can't be skipped? I hope there's a better solution to that problem than my hack in #660.

mitjat (commented):
Thank you!
I think we'll need to put in more effort for dependabot support. Happy to punt it to a later PR though if you prefer.

.github/workflows/ci-lint.yaml
buberdds: (relates to lint workflow only) We are skipping steps in a job not a workflow, so we won't hit #660 problem here.

I think build issues will have to be handle by someone more familiar with Nexus.

.github/workflows/ci-lint.yaml
buberdds: I hate incoming `lint-docs` even more. Anyway, I will double check if this is a hard requirement.

None of a repo I am familiar with is triggering commit hooks to check gitlint. Tbh, I hit that error once this year, so commit hook would be more painful.

buberdds: Org level decision is: gitlint and markdownlint should be the same. Golint is subject for per repo config

.github/workflows/ci-lint.yaml
mitjat: Ah the build issues in #663 actually look genuine; looks like dependabot was not fully successful with its patch.

But there's also an issue in #663 where it complains of a missing changelog fragment. That's what I meant we should create an exemption for.

.github/workflows/ci-lint.yaml
mitjat: Yeah I resigned to having the org-wide configs a long time ago.
FWIW, my personal philosophy is that autoformat is good (brings consistency at the minor cost of odd formatting choices sometimes), whereas enforced syntax-only lint without autoformat is bad (brings consistency but requires masochism). But this is not the place for that discussion :)

Anyway, in my experience they can be automated away pretty well with autoformatting. I might invest some time into that, proportional to how much the new lints annoy me. Thanks for the info on what already exists (which is not much, sounds like.)

mitjat: This looks potentially helpful for markdownlint, IDK if you're using it already: https://github.com/DavidAnson/vscode-markdownlint/blob/main/README.md#fix

buberdds: yes it is helpful, but markdownlint cannot fix MD013 which is the most common issue https://github.com/DavidAnson/markdownlint/issues/535

.github/workflows/ci-lint.yaml
buberdds: towncrier is handled in L~54

.github/workflows/ci-lint.yaml
mitjat: I'm preparing autoformatting with `prettier` in #666. I don't know if it does a perfect job of breaking lines (though I do think it has markdown awareness), but almost certainly it's good enough for the great majority of these simple short texts: commit messages and changelog fragments.

.github/workflows/ci-lint.yaml
mitjat: Ah I missed that, thank you!

mitjat (approved):
Thanks Michal!

PR #668, opened 2024-03-18
Use Punch tool for tracking and bumping project's version

buberdds:
Migration of Punch tool. Looking at repos there is an assumption that everyone has `Punch` tool installed locally `pip install punch.py`.

### DIFF
--- a/.changelog/668.internal.md
+++ b/.changelog/668.internal.md
@@ -1,0 +1,3 @@
+Use [Punch] tool for tracking and bumping project's version
+
+[Punch]: https://github.com/lgiordani/punch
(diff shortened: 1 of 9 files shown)


mitjat (approved):
Ah uniformity ... So much overkill for this particular project :smiling_face_with_tear: 
But also it'll be nice to be all buttoned up.

Thank you for working on this.

Andrew7234 (approved):
So when making a release, iiuc we now run `make changelog` beforehand. Is there anything else?

Also side note - why did we choose `punch`? (just curious)

buberdds:
> So when making a release, iiuc we now run `make changelog` beforehand. Is there anything else?
`make changelog` (create assemble/release PR)  `make releae-tag`

> Also side note - why did we choose `punch`? (just curious)
In Nexus - to introduce the same flow/tools for release process across our backend repos.
If you are asking about org level I have no idea. It was selected ~2 years before I have joined Oasis.

PR #669, opened 2024-03-18
storage: retain debond end time when converting genesis

pro-wh:
we've been entering them as all zero


Andrew7234 (approved):

mitjat (approved):
:pray:

pro-wh:
oh do I need to rebuild the caches again

pro-wh:
but why did the analysis succeed if it would have been missing the right cache?

pro-wh:
I guess that's just nexus's style, if one analyzer fails, the rest continues fine, even if the rest is 0 other analyzers

PR #670, opened 2024-03-21
query latest acct balance from node

Andrew7234:
Resolves https://app.clickup.com/t/8694204n5

Currently, the Explorer [fetches account balances directly ](https://github.com/oasisprotocol/explorer/blob/master/src/app/utils/rpc-utils.ts#L68) from the node when displaying the runtime account details page. This is because of known irregularities with the dead-reckoned balance tracking in Nexus. We'd prefer to make this request on the Nexus backend in order to improve visibility and enable querying our nonpublic nodes if necessary.

This PR now dispatches a concurrent request to the node with a timeout of 500ms to fetch the sdk balance of a given account at the latest indexed height. If the node balance differs from the dead-reckoned balance fetched from the Nexus db, the node balance supersedes the other. 

Edit** the below points are left for informational purposes; we now fetch the balance at the latest indexed height so that Nexus returns consistent information.

Note: The PR originally fetched the node balance at the latest nexus-indexed height, but this was removed in favor of fetching the balance at the latest node height for several reasons:
* the node takes >5s to respond to this query for blocks in the earlier parts of eden
* older cobalt/damask archive nodes have missing runtime data/do not support historical balance queries
* The existing explorer code [fetches the balance at the latest height](https://web3js.readthedocs.io/en/v1.2.9/web3-eth.html#getbalance)
* Displaying old/historical balances on the explorer will likely cause trouble, since users expect to see their current balances.
* (minor) fetching the latest indexed height added another database round trip. this could be optimized into an existing query if necessary.

Todo:
* Currently the api will query the node for all `/account/{}` requests in sapphire/emerald, or not at all (depending on config file). Do we want to make this set of runtimes configurable? We could also use some form of lazy connection to prevent unsupported runtime errors on api initialization. Edit*: We'll support all runtimes by default (emerald/sapphire/pontusx), but if a runtime isn't configured it no longer causes an error on api initialization.


storage/client/client.go
```diff
@@ -1568,6 +1592,17 @@ func (c *StorageClient) RuntimeAccount(ctx context.Context, address staking.Addr
 		return nil, wrapError(err)
 	}
 
+	// Reconcile sdk balance from node with the balance fetched from the db.
+	nodeBalance := <-ch
+	if nodeBalance != nil {
+		if len(a.Balances) != 1 {
+			c.logger.Warn("multiple sdk balance denominations detected; skipping node balance check", "address", address)
```
lukaw3d, line 1599: How come

you can get all balances from sdk (https://github.com/oasisprotocol/explorer/pull/1318#discussion_r1531434363)

Andrew7234: Oh thanks for the pointer; didn't realize that. Updating now

cmd/api/api.go
```diff
@@ -106,14 +110,30 @@ type Service struct {
...
-	client, err := storage.NewStorageClient(cfg.ChainName, backing, logger)
+	runtimeClients := make(map[common.Runtime]storage.RuntimeClient)
+	if cfg.Node != nil {
+		apiRuntimes := []common.Runtime{common.RuntimeEmerald, common.RuntimeSapphire}
+		for _, runtime := range apiRuntimes {
+			client, err := source.NewRuntimeClient(ctx, cfg.Node, runtime)
+			if err != nil {
+				return nil, err
+			}
+			runtimeClients[runtime] = storage.RuntimeClient{
```
mitjat, line 129: Open to debate, but: I think it's needlessly layered to introduce a wrapper around RuntimeApiLite just to store the NativeTokenSymbol. Because that's all that the new RuntimeClient does, right? Encapsulation is nice, but lots of layers can also make it harder to read and navigate code.

An alternative approach, which I'd like to suggest here, is what analyzers already do (and mimicking their structure is a win on its own): Pass `cfg.Node` into NewStorageClient(), and then StorageClient can [implement](https://github.com/oasisprotocol/nexus/blob/b789dca343c4e81ce336b4354242fbe1ac3520dc/analyzer/runtime/runtime.go#L68-L71) `nativeTokenSymbol(runtime)` itself, basically reusing your one-liner from here.

storage/client/client.go
```diff
@@ -1425,13 +1437,59 @@ func (c *StorageClient) RuntimeEvents(ctx context.Context, p apiTypes.GetRuntime
...
 }
 
+func (c *StorageClient) fetchAccountBalanceFromNode(ctx context.Context, ch chan *RuntimeSdkBalance, runtime common.Runtime, address staking.Address) {
+	runtimeApi, ok := c.runtimeClients[runtime]
+	if !ok {
+		c.logger.Warn("no runtime api configured to fetch account balances", "runtime", runtime)
+		close(ch)
+		return
+	}
+	balances, err := runtimeApi.Client.GetBalances(ctx, roothash.RoundLatest, address)
```
mitjat, line 1447: I still think we shouldn't fetch at RoundLatest. You explained the motivation for it in the PR description nicely, but most of those arguments only apply when nexus is reindexing, and I don't think we should care much about API robustness or responsiveness during that time.

My primary concern is that by fetching RoundLatest, the DB balance will quite frequently lag behind the node-fetched balance (because nexus has to first notice that a new block arrived, then fully analyze it, then write it to DB). So we'll log some non-negligible amount of warnings about mismatched balances between nexus and node, and we'll never know if it's a genuine mismatch or if the two data sources were just out of sync. That makes logs much much less actionable.

Having good logs here is one of the two main motivations (or maybe even _the_ main motivation) for this PR, i.e. querying the node in nexus and not on the frontend. The other one is that nexus will have less incorrect data, which is nice for non-Explorer users.

Another small benefit of not querying at RoundLatest is that the output will be internally consistent. (Mostly. I guess we should ideally be doing all the DB queries inside a DB tx.)

storage/client/client.go
```diff
@@ -1425,13 +1437,59 @@ func (c *StorageClient) RuntimeEvents(ctx context.Context, p apiTypes.GetRuntime
 	return &es, nil
 }
 
+func (c *StorageClient) fetchAccountBalanceFromNode(ctx context.Context, ch chan *RuntimeSdkBalance, runtime common.Runtime, address staking.Address) {
```
mitjat, line 1440: Nice use of channels :+1:

mitjat (commented):
Thank you, Andy!
Overall looks good, but I have some thoughts on lower-level stuff (and one high-level question), have a look and let me know.

cmd/api/api.go
Andrew7234: Valid points, updated! Note that the set of `nativeTokenSymbols` is decoupled from the set of `runtimeClients` but the former should be a superset of the latter so it should not be an issue.

config/local-dev.yml
```diff
@@ -29,6 +29,12 @@ server:
   storage:
     endpoint: postgresql://rwuser:password@localhost:5432/indexer?sslmode=disable
     backend: postgres
+  node:
```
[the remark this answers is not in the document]
Andrew7234: Updated, but added the anchor one level down at `SourceConfig.nodes` since the analyzer `SourceConfig` has fields (eg cache) that aren't needed for the server source config.

storage/client/client.go
Andrew7234: Hm I'd be okay with fetching at the latest indexed round if the frontend (explorer) will continue to fetch the latest balances from the node and overwrite any stale data Nexus might return. Stale balances would likely cause end users to panic, and our sapphire/emerald nodes have had less than perfect uptime. I assumed that the main goal of this PR was to have the explorer stop querying the node on their end though. 

Maybe we could query at the latest indexed round for now as an experiment to find any stale balances in the logs, and once those are fixed we'll start querying at `RoundLatest`?

config/config.go
```diff
@@ -512,6 +519,9 @@ type ServerConfig struct {
 	Endpoint string `koanf:"endpoint"`
 
 	Storage *StorageConfig `koanf:"storage"`
+
+	// Node is the configuration for accessing oasis-node(s).
+	Node *SourceConfig `koanf:"node"`
```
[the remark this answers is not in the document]
pro-wh: could move this stuff up so that it's not within either of server config or analyzer config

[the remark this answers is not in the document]
pro-wh: nvm I see action is already taken on this

cmd/api/api.go
```diff
@@ -118,15 +118,19 @@ func NewService(cfg *config.ServerConfig) (*Service, error) {
...
 		for _, runtime := range apiRuntimes {
 			client, err := source.NewRuntimeClient(ctx, cfg.Node, runtime)
 			if err != nil {
 				return nil, err
 			}
-			runtimeClients[runtime] = client
+			runtimeClients[runtime] = storage.RuntimeClient{
+				Runtime:           runtime,
+				Client:            client,
+				NativeTokenSymbol: cfg.Node.SDKNetwork().ParaTimes.All[string(runtime)].Denominations[sdkConfig.NativeDenominationKey].Symbol,
```
pro-wh: could we save the whole ParaTimes.All[...] paratime structure here, then we'd have data about the decimals later too

storage/client/client.go
```diff
@@ -1440,20 +1449,45 @@ func (c *StorageClient) RuntimeEvents(ctx context.Context, p apiTypes.GetRuntime
...
 	}
-	ch <- balance
+	foundNativeBalance := false
+	for denom, amount := range balances {
+		balance := RuntimeSdkBalance{
+			Balance: amount,
+			// HACK: 18 is accurate for Emerald and Sapphire, but Cipher has 9.
+			// Once we add a non-18-decimals runtime, we'll need to query the runtime for this
+			// at analysis time and store it in a table, similar to how we store the EVM token metadata.
+			TokenDecimals: 18,
```
pro-wh, line 1478: here...

config/local-dev.yml
```diff
@@ -29,6 +29,12 @@ server:
   storage:
     endpoint: postgresql://rwuser:password@localhost:5432/indexer?sslmode=disable
     backend: postgres
+  source:
+    chain_name: mainnet
+    nodes: *src
+      # eden:
+      #   default:
+      #     rpc: unix:/tmp/node.sock #unix:/node/data/internal.sock
```
pro-wh, line 37: can remove the commented out part, there's a copy in &src

storage/client/client.go
```diff
@@ -1437,13 +1451,63 @@ func (c *StorageClient) RuntimeEvents(ctx context.Context, p apiTypes.GetRuntime
...
+			TokenSymbol:   denom.String(),
+		}
+		if denom.IsNative() {
+			foundNativeBalance = true
```
pro-wh, line 1482: imo send the whole balances map back, we might as well offer all this latest data in the api

storage/client/client.go
```diff
@@ -1580,9 +1644,36 @@ func (c *StorageClient) RuntimeAccount(ctx context.Context, address staking.Addr
...
+	for nb := range ch {
+		found := false
+		// If the node balance differs from the balance in Nexus, defer to the node balance.
+		for i, b := range acct.Balances {
+			if b.TokenSymbol == nb.TokenSymbol {
+				found = true
+				if !b.Balance.Eq(nb.Balance) {
+					c.logger.Warn("sdk balance from node differed from dead-reckoned balance; defaulting to node balance", "address", acct.Address, "height", roothash.RoundLatest, "node_balance", nb.Balance.String(), "db_balance", b.Balance.String())
+					bPtr := &acct.Balances[i]
+					bPtr.Balance = nb.Balance
```
pro-wh, line 1665: ok I don't know how for...range loops work in go, but is it that `b.Balance = nb.Balance` doesn't really update acct?

storage/client/client.go
```diff
@@ -1580,9 +1644,36 @@ func (c *StorageClient) RuntimeAccount(ctx context.Context, address staking.Addr
...
+
 	return &a, nil
 }
 
+// Reads node sdk balances from ch and upserts them into acct.Balances, logging a
+// warning if the balances are mismatched.
+func (c *StorageClient) upsertBalances(ch chan *RuntimeSdkBalance, acct *RuntimeAccount) {
+	for nb := range ch {
+		found := false
+		// If the node balance differs from the balance in Nexus, defer to the node balance.
```
pro-wh, line 1658: possibly unpopular opinion, but could we include both in the API :grimacing: 

I think it'd be clearer if the explorer could show a note that it's using the latest balance when it uses the answer from the node. because I see the explanation that we're querying at latest, which can be inconsistent with the latest indexed height

storage/client/client.go
Andrew7234: yep :\

storage/client/client.go
Andrew7234: Discussed further in product sync today, and we decided to query for the balance at the current indexed height (not RoundLatest). If Nexus is behind/reindexing, the explorer will warn users appropriately.

cmd/api/api.go
Andrew7234: Good idea, updated!

storage/client/client.go
Andrew7234: Discussed in product sync; decision was to only query at the latest indexed height and have the explorer just warn the user if it's falling behind.

storage/client/client.go
Andrew7234: consistency is good 👍 

querying at latest indexed height now.

storage/client/client.go
```diff
@@ -1437,13 +1462,85 @@ func (c *StorageClient) RuntimeEvents(ctx context.Context, p apiTypes.GetRuntime
...
 }
 
+// Fetch account balances from the node and pass them back via the channel. Note that if the input context
+// times out or we encounter an error, we explicitly close the channel and return. This enables the calling
+// function to create a context with a timeout and then listen to the channel with the expectation that the
+// channel will return values and/or close approximately within the timeout.
+func (c *StorageClient) fetchAccountBalanceFromNode(ctx context.Context, ch chan *RuntimeSdkBalance, runtime common.Runtime, address staking.Address) {
+	var latestIndexedHeight uint64
+	// Query latest block.
+	err := c.db.QueryRow(
```
pro-wh, line 1472: does our library allow concurrent queries from one shared db object? might want to query this outside the goroutine if we're not sure

storage/client/client.go
```diff
@@ -131,14 +135,35 @@ func NewStorageClient(chainName common.ChainName, db storage.TargetStorage, l *l
...
 }
 
+// Returns the native token symbol of the specified runtime in the network
+// specified by the networkConfig.
+func (c *StorageClient) nativeTokenSymbol(runtime common.Runtime) string {
+	if c.networkConfig == nil {
+		c.logger.Warn("no network config available; unable to determine native token symbol", "runtime", runtime)
+		return ""
+	}
+	return c.networkConfig.ParaTimes.All[string(runtime)].Denominations[oasisConfig.NativeDenominationKey].Symbol
```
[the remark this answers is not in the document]
Andrew7234: Good point; added check for missing paratime. `Denomination` is a bit less clear; all paratimes are currently configured to have a native denomination but I'm not sure whether this is required for future paratimes. Will add the check just in case.

storage/client/client.go
```diff
@@ -1437,13 +1462,85 @@ func (c *StorageClient) RuntimeEvents(ctx context.Context, p apiTypes.GetRuntime
...
+	err := c.db.QueryRow(
+		ctx,
+		queries.Status,
+		runtime,
+	).Scan(&latestIndexedHeight, nil)
+	switch err {
+	case nil:
+	case pgx.ErrNoRows:
+		// No runtime blocks indexed yet.
+		latestIndexedHeight = roothash.RoundLatest
```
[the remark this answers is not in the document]
Andrew7234: added!

storage/client/client.go
Andrew7234: think so; `http.Server` is already serving queries concurrently using the same db object, and the analyzers share a db object as well.

pro-wh (approved):
sounds good

tests/e2e_regression/damask/expected/emerald_contract_account.body
```diff
@@ -1,6 +1,12 @@
 {
   "address": "oasis1qz2rynvcmrkwd57v00298uc2vtzgatde3cjpy72f",
-  "balances": [],
+  "balances": [
+    {
+      "balance": "0",
+      "token_decimals": 0,
+      "token_symbol": ""
```
Andrew7234, line 7: we don't index any emerald blocks in the damask e2e range; nexus defaults to returning a 0 balance if no blocks have been indexed. (in order to be consistent with nexus' internal state)

Issue #638, opened 2024-02-11
`evm_tokens` doesn't accept Ethereum formatted addresses


CedarMist:
##### SUMMARY

The `http://nexus.oasis.io/v1/{runtime}/evm_tokens/{address}` endpoint requires `oasis` addresses.

However, it's unusual to have to use `oasis` encoded addresses to query information about EVM related things.

##### STEPS TO REPRODUCE

It returns:

    Invalid format for parameter address: error unmarshalling '{address}' text as *api.Address: address: decoding from bech32 failed: decoding bech32 failed: invalid separator index -1

or

    Invalid format for parameter address: error unmarshalling '0x0245254C9Fb58A6c820a90f0c06694F1a512983D' text as *api.Address: address: decoding from bech32 failed: decoding bech32 failed: string not all lowercase or all uppercase

##### EXPECTED RESULTS

For example on https://explorer.oasis.io/mainnet/sapphire/address/0x4ebA1A83d00C0CCCBE680ecC730a5b1Dc2189F86/tokens/erc-20#tokens

I should be able to use that address in the API URI:

    https://nexus.oasis.io/v1/sapphire/evm_tokens/0x4ebA1A83d00C0CCCBE680ecC730a5b1Dc2189F86

And it should return the list of tokens held by that account.

PR #675, opened 2024-03-25
runtime api: query by eth address 

Andrew7234:
Resolves https://github.com/oasisprotocol/nexus/issues/638

This PR now allows querying runtime api endpoints by both ethereum and oasis addresses. This applies to both path params and query params.

```
» curl "http://localhost:8008/v1/sapphire/events?contract_address=39d22B78A7651A76Ffbde2aaAB5FD92666Aca520" | jq                                                                                                                                                         andrewlow@Beleriand
100 79573    0 79573    0     0  1158k      0 --:--:-- --:--:-- --:--:-- 1253k
{
  "events": [
    {
      "body": {
        "address": "OdIr...
```


analyzer/evmverifier/evmverifier.go
```diff
@@ -93,12 +93,9 @@ type oasisAddress string
 
 // A smart contract, and info on verification progress.
 type contract struct {
-	Addr                  oasisAddress
-	AddrContextIdentifier string
-	AddrContextVersion    int
-	AddrData              []byte
```
Andrew7234, line 99: unrelated nit: these fields were unused

storage/client/client.go
```diff
@@ -1621,7 +1634,7 @@ func (c *StorageClient) RuntimeTokens(ctx context.Context, p apiTypes.GetRuntime
 	return &ts, nil
 }
 
-func (c *StorageClient) RuntimeTokenHolders(ctx context.Context, p apiTypes.GetRuntimeEvmTokensAddressHoldersParams, address staking.Address) (*TokenHolderList, error) {
+func (c *StorageClient) RuntimeTokenHolders(ctx context.Context, p apiTypes.GetRuntimeEvmTokensAddressHoldersParams, address *staking.Address) (*TokenHolderList, error) {
```
Andrew7234, line 1636: changed for consistency with the rest of the api; if this was intentionally not a pointer please lmk

tests/e2e_regression/test_cases.sh
```diff
@@ -9,107 +9,111 @@
...
-  'validators                     /v1/consensus/validators'
-  'emerald_blocks                 /v1/emerald/blocks'
-  'emerald_txs                    /v1/emerald/transactions'
-  'emerald_events                 /v1/emerald/events'
-  'emerald_events_by_type         /v1/emerald/events?type=sdf'
-  'emerald_tokens                 /v1/emerald/evm_tokens'
-  'emerald_token                  /v1/emerald/evm_tokens/oasis1qpgcp5jzlgk4hcenaj2x82rqk8rrve2keyuc8aaf'
-  'emerald_token_holders          /v1/emerald/evm_tokens/oasis1qpgcp5jzlgk4hcenaj2x82rqk8rrve2keyuc8aaf/holders'
-  'emerald_account_with_rose      /v1/emerald/accounts/oasis1qrt0sv2s2x2lkt9e7kmr2mzxgme8m0pzauwztprl'
-  'emerald_account_with_evm_token /v1/emerald/accounts/oasis1qpwx3ptmvcceqkd4syjmqf9jmdlf90xmuuy0f6y9'
```
Andrew7234, line 74: the main change here was that these test cases were moved to the eden-specific range since they aren't valid in damask. Note that no suitable damask equivalent exists because the damask test range is very early and has no evm_tokens/accounts

storage/client/client.go
pro-wh: wait everything else uses pointer?

pro-wh: oh yeah most runtime ones use address

api/v1/types/util.go
```diff
@@ -63,3 +69,23 @@ func (p *GetLayerStatsActiveAccountsParams) Validate() error {
...
+// conforms to either an oasis bech32-encoded address or an ethereum
+// address, but we double check just to be sure.
+func UnmarshalToOcAddress(raw *string) (*StakingAddress, error) {
+	if raw == nil {
+		return nil, nil
+	}
+	if ethCommon.IsHexAddress(*raw) {
+		ea := ethCommon.HexToAddress(*raw)
+		ctx := sdkTypes.AddressV0Secp256k1EthContext
+		oasisAddr := StakingAddress(address.NewAddress(ctx, ea[:]))
```
pro-wh, line 83: is this cast right?

pro-wh (commented):
we want this to simplify how sourcify accesses contract creation info

tests/e2e_regression/test_cases.sh
pro-wh: I'm gonna have to transfer these for #665

storage/client/client.go
Andrew7234: oh the address param is required for this endpoint; which is why some endpoints use pointers. reverting this.

api/v1/types/util.go
Andrew7234: Think so; they're the same underlying type but `StakingAddress` is the type used in this `api` package. Unless I'm missing something?

tests/e2e_regression/test_cases.sh
Andrew7234: Ah I'll wait for 665 to go in first :)

api/v1/types/util.go
pro-wh: leme just try this, I thought maybe this would cast the bytes-like address into the string-like address via treating it like UTF-8, without going through the bech32 stuff

pro-wh: ok you're right, StakingAddress is not the string-like thing

pro-wh: for my information:

- address.NewAddress here is from the oasis-core/go/common/crypto/address package where `type Address [Size]byte`
- StakingAddress is from the generated API types where `type StakingAddress = staking.Address` (alias declaration)
  - `staking` here is our vendored Damask staking/api package where `type Address = original.Address` (alias declaration)
    - `original` here is oasis-core/go/staking/api where `type Address address.Address` (new type definition)
      - `address` here is oasis-core/go/common/crypto/address

api/spec/v1.yaml
```diff
@@ -1235,6 +1232,14 @@ components:
...
       example: "oasis1qpg2xuz46g53737343r20yxeddhlvc2ldqsjh70p"
       description: An Oasis-style (bech32) address.
+    EthOrOasisAddress:
+      type: string
+      pattern: '^oasis1[a-z0-9]{40}$|^(0x)?[0-9a-fA-F]{40}$'
+      examples:
+        eth: 
+          value: "0x0cAf4748697AE0751728824527044D8c583342c6"
+        oasis:
+          value: "oasis1qpg2xuz46g53737343r20yxeddhlvc2ldqsjh70p"
```
mitjat, line 1242: TIL about `examples` vs `example`. Nice.

api/v1/strict_server.go
```diff
@@ -337,15 +353,27 @@ func (srv *StrictServerImpl) GetRuntimeEvents(ctx context.Context, request apiTy
 }
 
 func (srv *StrictServerImpl) GetRuntimeAccountsAddress(ctx context.Context, request apiTypes.GetRuntimeAccountsAddressRequestObject) (apiTypes.GetRuntimeAccountsAddressResponseObject, error) {
-	account, err := srv.dbClient.RuntimeAccount(ctx, request.Address)
+	ocAddr, err := apiTypes.UnmarshalToOcAddress(&request.Address)
+	if err != nil {
+		return nil, err
+	}
+	account, err := srv.dbClient.RuntimeAccount(ctx, *ocAddr)
```
mitjat, line 360: `UnmarshalToOcAddress` can return `nil, nil`, but here we dereference `ocAddr`

api/v1/types/util.go
```diff
@@ -63,3 +69,23 @@ func (p *GetLayerStatsActiveAccountsParams) Validate() error {
 	}
 	return &InvalidParamFormatError{ParamName: "window_step_seconds", Err: fmt.Errorf("invalid value: %d", *p.WindowStepSeconds)}
 }
+
+// The autogenerated server performs type validation that the input string
+// conforms to either an oasis bech32-encoded address or an ethereum
+// address, but we double check just to be sure.
```
mitjat, line 75: This is a comment on why we have this func. It's a good extra, but please prepend it with a description of what this func does.

Also, semi-nit: Openapi specifies a regex pattern, but the autogenerated code ignores that pattern. So this func is _the_ validity check. Example: https://nexus.oasis.io/v1/sapphire/accounts/foobar

tests/e2e_regression/test_cases.sh
pro-wh: leme try rebasing that earlier then

api/v1/types/util.go
```diff
@@ -63,3 +69,23 @@ func (p *GetLayerStatsActiveAccountsParams) Validate() error {
...
+func UnmarshalToOcAddress(raw *string) (*StakingAddress, error) {
```
mitjat, line 76: Nit: when speaking of addresses, `raw` to me implies unencoded bytes. The type on this line is clear, but it makes for stilted reading later on. Consider renaming to `text` (my favorite, for parallels with `UnmarshalText()`) or `encoded` or `serialized` or similar.

tests/e2e_regression/eden/expected/emerald_token_nfts_eth.headers
```diff
@@ -1,4 +1,4 @@
-HTTP/1.1 404 Not Found
+HTTP/1.1 200 OK
```
mitjat, line 1: Note to self/readers: This went 404->200 because even though it's the same endpoint, we now run it in eden, but we used to run it in damask: https://github.com/oasisprotocol/nexus/pull/675#discussion_r1538516256

mitjat (approved):
Nice, thank you!

tests/e2e_regression/test_cases.sh
pro-wh: thanks, merged. continue removing this from here. it's now renamed to common_test_cases.sh. edenTestCases is moved to eden/test_cases.sh, so add them there instead

api/v1/types/util.go
Andrew7234: Good idea, updated!

api/v1/types/util.go
Andrew7234: Oh I thought openapi was doing input validation somewhere deep down; that's good to know. Updated with comment!

api/v1/strict_server.go
Andrew7234: Hrm technically it should never happen because the `request.Address` is required to be non-nil. Added an explicit check+comment anyway for maintainability though.

api/spec/v1.yaml
Andrew7234: our openapi linter doesn't like it 💀

mitjat: You nerd-sniped me. That took some careful spec reading :/

The [Parameter](https://swagger.io/specification/#parameter-object) object takes `examples`, but [Schema](https://swagger.io/specification/) only takes `example`, and notes that "The example property has been deprecated in favor of the JSON Schema examples keyword."

From what I can tell, "the JSON Schema examples keyword" is simply the one at Parameter level.

So this works:
```yaml
        - in: query
          name: rel
          schema:
            allOf: [$ref: '#/components/schemas/EthOrOasisAddress']
          examples:
            eth: 
              value: "0x0cAf4748697AE0751728824527044D8c583342c6"
            oasis:
              value: "oasis1qpg2xuz46g53737343r20yxeddhlvc2ldqsjh70p"
```

Unfortunately that's pretty verbose and repetitive: Every time we specify the `EthOrOasisAddress` schema, we also have to specify the examples. I shortened it a little more by reusing example objects, and pushed a commit because it's easier than explaining. Feel free to tweak further.

Here's how it renders: 
image

Andrew7234: Thank you for delving into that rabbit hole and adding the fix! Looks great; merging asap

PR #676, opened 2024-03-26
api: log binned request latencies

Andrew7234:
## Task

The API layer logs requests but does not log query params: [https://github.com/oasisprotocol/nexus/blob/908a841427c146cd81fdfac00bfebf771f885d1a/api/middleware.go#L85-L85](https://github.com/oasisprotocol/nexus/blob/908a841427c146cd81fdfac00bfebf771f885d1a/api/middleware.go#L85-L85)
  
Add those, and add the duration of query, and add a field that categorizes the duration (e.g. <100ms, 100-300ms, 300-1000ms, >1000ms) so we can find expensive queries easily from the logs.

## This PR
Adds query params and binned request latency to api logging. Also renames the existing `time` field to `latency`.

### DIFF
--- a/api/middleware.go
+++ b/api/middleware.go
@@ -82,10 +82,13 @@ func MetricsMiddleware(m metrics.RequestMetrics, logger log.Logger) func(next ht
 				// like we're doing here if we don't want to use it.
 				timer.ObserveDuration()
 			}
+			latency := time.Since(t)
 			logger.Info("ending request",
-				"endpoint", r.URL.Path,
+				"query_path", r.URL.Path,
+				"query_params", r.URL.RawQuery,
 				"request_id", requestID,
-				"time", time.Since(t),
+				"latency", latency,
+				"latency_bin", binQueryLatency(latency),
 				"status_code", httpStatus,
 			)
 
@@ -101,6 +104,23 @@ func MetricsMiddleware(m metrics.RequestMetrics, logger log.Logger) func(next ht
 			}
 			m.RequestCounts(metricName, statusTxt).Inc()
 		})
+	}
+}
+
+// Bin request durations to make it easier to search
+// for slow queries in Nexus logs.
+func binQueryLatency(t time.Duration) string {
+	switch {
+	case t < 100*time.Millisecond:
+		return "<100ms"
+	case t < 300*time.Millisecond:
+		return "100-300ms"
+	case t < 500*time.Millisecond:
+		return "300-500ms"
+	case t < 1000*time.Millisecond:
+		return "500-1000ms"
+	default:
+		return ">1000ms"
 	}
 }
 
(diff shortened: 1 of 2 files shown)


pro-wh (approved):
as described

cc @lukaw3d @lubej we're going to be logging query parameters

api/middleware.go
```diff
@@ -104,6 +107,23 @@ func MetricsMiddleware(m metrics.RequestMetrics, logger log.Logger) func(next ht
 	}
 }
 
+// Bin request durations to make it easier to search
+// for slow queries in Nexus logs.
```
mitjat, line 111: FWIW, GCP Logs Explorer does allow arithmetic comparisons, see [example](https://cloudlogging.app.goo.gl/iJQ585LPytqCJW1u9). And of course we slice and dice in whatever way we want if we download the logs to our local machine. So I don't see the benefit of this field, but it also doesn't hurt me :)

api/middleware.go
```diff
@@ -82,10 +82,13 @@ func MetricsMiddleware(m metrics.RequestMetrics, logger log.Logger) func(next ht
 				// like we're doing here if we don't want to use it.
 				timer.ObserveDuration()
 			}
+			latency := time.Since(t)
 			logger.Info("ending request",
 				"endpoint", r.URL.Path,
+				"query_params", r.URL.RawQuery,
 				"request_id", requestID,
-				"time", time.Since(t),
```
mitjat, line 88: :astonished:  It was really unexpected to me that GCP just swallows the "time" field. Nice find. Just leaving a comment for posterity :)

api/middleware.go
```diff
@@ -82,10 +82,13 @@ func MetricsMiddleware(m metrics.RequestMetrics, logger log.Logger) func(next ht
 				// like we're doing here if we don't want to use it.
 				timer.ObserveDuration()
 			}
+			latency := time.Since(t)
 			logger.Info("ending request",
 				"endpoint", r.URL.Path,
+				"query_params", r.URL.RawQuery,
```
mitjat, line 88: Kinda silly, but: Since these two fields have such different names, I fear that a casual / less informed reader of these logs (including us in a few months :)) might miss one or the other. Can we go with names that share a common prefix? Or just replace both of these with a single field that logs `r.URL.RequestURI()` ([link](https://pkg.go.dev/net/url#URL.RequestURI)). If we ever want to filter by just the path (unlikely, IMO, because the params are so important), there's always regex search.

mitjat (approved):
Thank you!

api/middleware.go
Andrew7234: Agreed, it's also a fallback in case gcp keeps eating the `time`/`latency` field for whatever reason.

PR #677, opened 2024-03-27
analyzer/consensus: Skip nonexistent mainnet block 8048955

mitjat:
On Oasis mainnet, consensus block 8048955 does not exist: 8048954 is the last block of Cobalt, and 8048956 is the first block of Damask. This was an unintended side effect of the dump-restore Damask upgrade.

Nexus was unaware of this, and would get stuck on that block when reindexing the entire history. The only way to "unstick" it was (small) manual DB surgery. This PR hardcodes the block as an exception, and makes Nexus skip it.

**Testing:**
One-off manual test. Ran mainnet analysis with the following config:
```yaml
    consensus:
      from: 8_048_946  # Damask genesis was at 8_048_956; probe slightly into Cobalt
      to:   8_048_966
```
and verified that blocks are analyzed as expected.

### DIFF
--- a/analyzer/consensus/consensus.go
+++ b/analyzer/consensus/consensus.go
@@ -66,7 +66,6 @@ func OpenSignedTxNoVerify(signedTx *transaction.SignedTransaction) (*transaction
 
 // processor is the block processor for the consensus layer.
 type processor struct {
-	chain   common.ChainName
 	mode    analyzer.BlockAnalysisMode
 	history config.History
 	source  nodeapi.ConsensusApiLite
@@ -79,9 +78,8 @@ type processor struct {
 var _ block.BlockProcessor = (*processor)(nil)
 
 // NewAnalyzer returns a new analyzer for the consensus layer.
-func NewAnalyzer(chain common.ChainName, blockRange config.BlockRange, batchSize uint64, mode analyzer.BlockAnalysisMode, history config.History, source nodeapi.ConsensusApiLite, network sdkConfig.Network, target storage.TargetStorage, logger *log.Logger) (analyzer.Analyzer, error) {
+func NewAnalyzer(blockRange config.BlockRange, batchSize uint64, mode analyzer.BlockAnalysisMode, history config.History, source nodeapi.ConsensusApiLite, network sdkConfig.Network, target storage.TargetStorage, logger *log.Logger) (analyzer.Analyzer, error) {
 	processor := &processor{
-		chain:   chain,
 		mode:    mode,
 		history: history,
 		source:  source,
@@ -309,9 +307,7 @@ func (m *processor) ProcessBlock(ctx context.Context, uheight uint64) error {
 	height := int64(uheight)
 	batch := &storage.QueryBatch{}
 
-	isBlockAbsent := m.chain == common.ChainNameMainnet && height == 8048955 // Block does not exist due to mishap during upgrade to Damask.
-
-	if !isBlockAbsent {
+	if _, isBlockAbsent := m.history.MissingBlocks[uheight]; !isBlockAbsent {
 		// Fetch all data.
 		fetchTimer := m.metrics.BlockFetchLatencies()
 		data, err := fetchAllData(ctx, m.source, m.network, height, m.mode == analyzer.FastSyncMode)
--- a/analyzer/consensus/consensus.go
+++ b/analyzer/consensus/consensus.go
@@ -66,6 +66,7 @@ func OpenSignedTxNoVerify(signedTx *transaction.SignedTransaction) (*transaction
 
 // processor is the block processor for the consensus layer.
 type processor struct {
+	chain   common.ChainName
 	mode    analyzer.BlockAnalysisMode
 	history config.History
 	source  nodeapi.ConsensusApiLite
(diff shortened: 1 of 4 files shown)


analyzer/consensus/consensus.go
```diff
@@ -299,25 +301,29 @@ func (m *processor) ProcessBlock(ctx context.Context, uheight uint64) error {
...
+	batch := &storage.QueryBatch{}
 
-	// Fetch all data.
-	fetchTimer := m.metrics.BlockFetchLatencies()
-	data, err := fetchAllData(ctx, m.source, m.network, height, m.mode == analyzer.FastSyncMode)
-	fetchTimer.ObserveDuration()
-	if err != nil {
-		if strings.Contains(err.Error(), fmt.Sprintf("%d must be less than or equal to the current blockchain height", height)) {
-			return analyzer.ErrOutOfRange
+	isBlockAbsent := m.chain == common.ChainNameMainnet && height == 8048955
```
pro-wh, line 306: I feel like this could go as some kind of addition to the History config. this is the height just below when mainnet damask started. maybe we could put something in the mainnet cobalt record to say height 8048955 was absent

mitjat: That was my first instinct too. But Jernej was saying how this was definitely a bug/mistake when doing the Damask dump-restore. In the future, a) dump-restores are unlikely, b) if they do happen, we're unlikely to make the same mistake twice. So it's IMO an overgeneralization to bloat the history config with this, and make the consuming code (= here) more complex and general and indirect. Let's expand History once we have more than one such block in the future.

I also thought about at least referencing the Damask genesis height from History config here. So instead of comparing to 8048955, I could compare to (genesis_height - 1). But then I decided that's no more informative, and just harder to grep for.

analyzer/consensus/consensus.go
```diff
@@ -299,25 +301,29 @@ func (m *processor) ProcessBlock(ctx context.Context, uheight uint64) error {
...
+	batch := &storage.QueryBatch{}
 
-	// Fetch all data.
-	fetchTimer := m.metrics.BlockFetchLatencies()
-	data, err := fetchAllData(ctx, m.source, m.network, height, m.mode == analyzer.FastSyncMode)
-	fetchTimer.ObserveDuration()
-	if err != nil {
-		if strings.Contains(err.Error(), fmt.Sprintf("%d must be less than or equal to the current blockchain height", height)) {
-			return analyzer.ErrOutOfRange
+	isBlockAbsent := m.chain == common.ChainNameMainnet && height == 8048955 // Block does not exist due to mishap during upgrade to Damask.
```
Andrew7234, line 306: just curious - what do you think of something like 
```suggestion
	if m.chain == common.ChainNameMainnet && height == 8048955 { // Block does not exist due to mishap during upgrade to Damask.
	    return nil
	}
```

analyzer/consensus/consensus.go
```diff
@@ -77,8 +78,9 @@ type processor struct {
 var _ block.BlockProcessor = (*processor)(nil)
 
 // NewAnalyzer returns a new analyzer for the consensus layer.
-func NewAnalyzer(blockRange config.BlockRange, batchSize uint64, mode analyzer.BlockAnalysisMode, history config.History, source nodeapi.ConsensusApiLite, network sdkConfig.Network, target storage.TargetStorage, logger *log.Logger) (analyzer.Analyzer, error) {
+func NewAnalyzer(chain common.ChainName, blockRange config.BlockRange, batchSize uint64, mode analyzer.BlockAnalysisMode, history config.History, source nodeapi.ConsensusApiLite, network sdkConfig.Network, target storage.TargetStorage, logger *log.Logger) (analyzer.Analyzer, error) {
```
Andrew7234, line 81: Hm the list of args is getting quite long, let's maybe group them into a `AnalyzerParams` struct (can punt to a future PR tho)

Andrew7234 (approved):

analyzer/consensus/consensus.go
pro-wh: opposite opinion from the same principles:
- it'll take at least a little code change no matter what
- it happened once and we anticipate not again, so this only exists as a historical artifact
- better to shove the details into the "history" data rather than bloat the code with special (name, height) coordinates

ok on not optimizing generality or auto computing the height though

pro-wh: this would prevent us from having to add this special purpose chain name arg throughout

pro-wh: extra thing: I'm against doing special behavior for named default chains that a custom history config can't do

analyzer/consensus/consensus.go
Andrew7234: nvm we still have to update the indexing progress

analyzer/consensus/consensus.go
```diff
@@ -66,6 +66,7 @@ func OpenSignedTxNoVerify(signedTx *transaction.SignedTransaction) (*transaction
 
 // processor is the block processor for the consensus layer.
 type processor struct {
+	chain   common.ChainName
```
pro-wh, line 69: leftover

Andrew7234:
Ok updated to address comments. Tried testing on the same range as the PR description but it's having issues connecting to the production cobalt archive node.

config/history.go
```diff
@@ -23,7 +23,8 @@ type Record struct {
 }
 
 type History struct {
-	Records []*Record `koanf:"records"`
+	Records       []*Record `koanf:"records"`
+	MissingBlocks map[uint64]struct{}
```
pro-wh, line 27: can you put a `koanf` tag here :pray:

pro-wh (approved):
thanks for the changes, still looks very clean

config/history.go
Andrew7234: added!

PR #679, opened 2024-03-27
analyzer/runtime: limit total_{received,sent} to native denomination

mitjat:
pontusx is the first runtime where this makes a difference because it uses two separate SDK denominations. Before this PR, their stats got conflated in the total_received and total_sent metrics (i.e. the metric was a sum of the two SDK tokens, EUROe and TEST).

This PR is a no-op for other existing runtimes because the only SDK denomination they use is the native denomination.

### DIFF
--- a/scripts/git-hooks/pre-commit
+++ b/scripts/git-hooks/pre-commit
@@ -14,7 +14,7 @@ staged_fragments="$(git diff --cached --name-only -- .changelog/*.md)"
 # Re-wraps file $1 to 80 characters per line, using prettier, in-place.
 rewrap() {
   # Suppress the output of prettier, as it's overly verbose. Show only on errors.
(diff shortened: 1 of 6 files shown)


scripts/git-hooks/pre-commit
```diff
@@ -14,7 +14,7 @@ staged_fragments="$(git diff --cached --name-only -- .changelog/*.md)"
 # Re-wraps file $1 to 80 characters per line, using prettier, in-place.
 rewrap() {
   # Suppress the output of prettier, as it's overly verbose. Show only on errors.
-  prettier --no-config --no-editorconfig --prose-wrap always --print-width 80 --write "$f" >/tmp/prettier.out \
+  prettier --no-config --no-editorconfig --prose-wrap always --print-width 78 --write "$f" >/tmp/prettier.out \
```
mitjat, line 17: unrelated, brings the autoformatter in line with the linter config

Andrew7234 (approved):

PR #680, opened 2024-04-03
analyzer: make abi_backfill analyzer robust to malformed abis

Andrew7234:
After enabling the abi analyzer on mainnet, it started complaining because [this contract](https://explorer.dev.oasis.io/mainnet/sapphire/address/0x32F7551057EBc3BF34E61FEd38A2D54143454639), despite being fully verified, had a malformed abi that go-eth refused to parse. This blocked the analyzer from proceeding.

This PR makes the analyzer robust to malformed abis. It will now log a warning and mark any associated txs/events as parsed.


analyzer/runtime/abiparse/abiparse.go
```diff
@@ -87,7 +87,7 @@ func ParseEvent(topics [][]byte, data []byte, contractABI *abi.ABI) (*abi.Event,
 	}
 	event, err := contractABI.EventByID(topicsEC[0])
 	if err != nil {
-		return nil, nil, fmt.Errorf("contract ABI EventByID: %w", err)
+		return nil, nil, fmt.Errorf("contract ABI EventByID: %w, eventID: %s", err, topicsEC[0].Hex())
```
Andrew7234, line 90: the [go-eth EventByID](https://github.com/ethereum/go-ethereum/blob/v1.13.14/accounts/abi/abi.go#L223) mangles the event signature so we explicitly log it here

pro-wh: > mangles

> 	return nil, fmt.Errorf("no event with id: %#x", topic.Hex())

what does it do?

pro-wh (approved):
ok

analyzer/runtime/abiparse/abiparse.go
Andrew7234: `topic.Hex()` returns a hex `string` which is then interpreted as utf-8 bytes and converted again to hex bytes (`%#x`). tldr the topic becomes this ~140 char long hex string.

pro-wh: their bug huh, rip

PR #683, opened 2024-04-18
bump oasis-sdk version to v0.8.2

Andrew7234:
As part of https://github.com/oasisprotocol/nexus/pull/670, we want to pull the runtime denomination decimal info from the oasis-sdk. This requires bumping the sdk to the latest version, which includes the pontusx information.


pro-wh (approved):
ya

PR #685, opened 2024-04-19
Andrew7234/api entity address

Andrew7234:
This PR makes consensus entities query-able by oasis address instead of entity pubkey. This affects multiple endpoints. 

Notes:
* this is a *breaking* change. We _could_ continue support for pubkey querying, but the implementation would be a bit messier (akin to the recent query by EthOrOasisAddress)

Open question:
* nodes are still only query-able by pubkey (eg `node_id`) as we do not currently track their addresses. Is there a demand for querying the `/consensus/entities/{address}/nodes/{node_id}` endpoint by node address instead?

### DIFF
--- a/.changelog/685.breaking.md
+++ b/.changelog/685.breaking.md
@@ -1,0 +1,1 @@
+api: query entities by address instead of pubkey
--- a/api/spec/v1.yaml
+++ b/api/spec/v1.yaml
@@ -410,16 +410,16 @@ paths:
                 $ref: '#/components/schemas/EntityList'
         <<: *common_error_responses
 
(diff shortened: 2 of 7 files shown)


pro-wh (approved):
ya this seems fine. did someone request this?

in the future we can hash a given public key, as you point out similar to how we query by ethereum address

could also start tracking node addresses in the future for consistency

Andrew7234:
> did someone request this?

Yeah explorer folks requested it as part of the Validator support: https://app.clickup.com/t/86941rbrt

> in the future we can hash a given public key

Yeah; I'll confirm with explorer team that they have no use for the existing pubkey endpoint

PR #686, opened 2024-04-30
Andrew7234/validator api

Andrew7234:
This PR will add support for the following fields to the `/consensus/validators` and `/consensus/validators/{address}` endpoints. It also adds an optional `name` filter to that endpoint.

* data for balance distribution chart 
* data to render "Voting Power" snapshot box
* voting power
* total_voting_power
* rank
* num_delegators
* start_date
* active (changed from status, semantics differ as well)
* in_validator_set

Validator details umbrella ticket: https://app.clickup.com/t/86941rbrt


api/spec/v1.yaml
```diff
@@ -1874,8 +1874,14 @@ components:
           description: The public key identifying this Validator's node.
           example: *node_id_1
         escrow:
+          allOf: [$ref: '#/components/schemas/Escrow']
+          description: The escrow account data for this validator.
+        voting_power:
           allOf: [$ref: '#/components/schemas/TextBigInt']
-          description: The amount staked.
+          description: The voting power of this Validator.
+        voting_power_total:
```
Andrew7234, line 1882: this (and `active_balance_total` below) seems a bit out of place here; but i'm not sure whether it's worth creating a separate endpoint to return this network-level info.

pro-wh: ya I don't know what to do about this

pro-wh: oh clarification on this and active_balance_total below: "active" refers to all validators, i.e. in the current committee? e.g. if a validator is down mid-epoch, it's still counted in these right?

api/spec/v1.yaml
```diff
@@ -1893,6 +1899,28 @@ components:
...
 
+    Escrow:
+      type: object
+      properties:
+        active_balance:
+          allOf: [$ref: '#/components/schemas/TextBigInt']
+          description: The amount of token actively delegated to this account.
+        active_shares:
+          allOf: [$ref: '#/components/schemas/TextBigInt']
+          description: The shares of tokens delegated to this account.
```
pro-wh, line 1910: a call out to our team, this might need more explanation than "the shares of ..."

api/spec/v1.yaml
Andrew7234: > "active" refers to all validators, i.e. in the current committee? e.g. if a validator is down mid-epoch, it's still counted in these right?

Yep

analyzer/queries/queries.go
```diff
@@ -278,10 +278,18 @@ var (
...
 	ConsensusEntityUpsert = `
-    INSERT INTO chain.entities (id, address) VALUES ($1, $2)
-      ON CONFLICT (id) DO
-      UPDATE SET
-        address = excluded.address`
+    INSERT INTO chain.entities (id, address, start_date) 
+    SELECT
+      $1,
+      $2,
+      (SELECT MIN(time) FROM chain.blocks as blocks
```
Andrew7234, line 285: if a `registerEntity` tx occurs in the same block, we rely on the tx/block `insert` to occur before this one (which is currently the case), but it introduces some code smell. If the order is reversed this column is `null`. 

alternatives considered:
- trigger function: also not ideal since it introduces dependent logic that is hard for devs to find/grok.
- lookup at query time: the join over the txs table is way too slow

storage/client/queries/queries.go
```diff
@@ -286,30 +286,67 @@ const (
...
+				FROM chain.delegations
+				WHERE delegator = delegatee
+				GROUP BY delegatee
+		),
+		delegators_count AS (
+			SELECT delegatee AS address, COUNT(*) AS count
+				FROM chain.delegations
+				GROUP BY delegatee
+		),
+		active_nodes AS (
```
ptrus, line 300: Maybe rename `active_` prefix or add a comment, since IMO the "active" term could be a bit ambiguous. Does it refer to nodes that are currently registered as a validator, or does it refer only to nodes that are actually validating current epoch (are actually active). 

Some background:
The network supports a limited number of validators (consensus parameter, currently 120 on mainnet). If there are more validator nodes registered, then the top 120 (by stake) are picked at the start of the epoch and are selected as being the validator for the full duration of the epoch.

Looking at the DB setup I believe that `chain.nodes` contains all currently registered nodes and nodes with `voting_power > 0` are the actually active validators.


So I believe this query currently returns all registered validator nodes. I think this is fine, and what we actually want to return, but maybe just rename the `active_nodes` or add a clarifying comment.

storage/client/queries/queries.go
```diff
@@ -286,30 +286,67 @@ const (
...
+			chain.entities.id AS entity_id,
+			chain.entities.address AS entity_address,
+			chain.nodes.id AS node_address,
+			chain.accounts.escrow_balance_active AS active_balance,
+			chain.accounts.escrow_total_shares_active AS active_shares,
+			chain.accounts.escrow_balance_debonding AS debonding_balance,
+			chain.accounts.escrow_total_shares_debonding AS debonding_shares,
+			COALESCE (
+				ROUND(COALESCE(self_delegations.shares, 0) * chain.accounts.escrow_balance_active / NULLIF(chain.accounts.escrow_total_shares_active, 0))
+			, 0) AS self_delegation_balance,
```
ptrus, line 318: Could also consider skipping this and have frontend compute this.  Since we return `active_balance`, `active_shares` and `self_delegation_shares` this can be trivially comptued.

But I guess it is also totally fine if we do it here.

storage/client/queries/queries.go
```diff
@@ -286,30 +286,67 @@ const (
...
+				delegators_count.count
+			, 0) AS num_delegators,
+			active_nodes.voting_power,
+			COALESCE (
+				(SELECT SUM(voting_power)
+				FROM active_nodes)
+			, 0) AS total_voting_power,
+			COALESCE(chain.commissions.schedule, '{}'::JSONB) AS commissions_schedule,
+			chain.entities.start_date AS start_date,
+			ROW_NUMBER() OVER (ORDER BY escrow_balance_active DESC) AS rank,
```
ptrus, line 332: Can we use `active_balance` instead of  `escrow_balance_active` here (since we have `chain.accounts.escrow_balance_active AS active_balance` above), or does that not work?

I think it would be a bit clearer.

storage/client/queries/queries.go
```diff
@@ -286,30 +286,67 @@ const (
...
+		LEFT JOIN chain.commissions ON chain.entities.address = chain.commissions.address
+		LEFT JOIN self_delegations ON chain.entities.address = self_delegations.address
+		LEFT JOIN delegators_count ON chain.entities.address = delegators_count.address
+		JOIN active_nodes ON active_nodes.address = entities.address
+		JOIN chain.nodes ON chain.entities.id = chain.nodes.entity_id
+			AND chain.nodes.roles like '%validator%'
+			AND chain.nodes.voting_power = active_nodes.voting_power
+		WHERE ($1::text IS NULL OR chain.entities.address = $1::text) AND
+				($2::text IS NULL OR chain.entities.meta->>'name' LIKE '%' || $2::text || '%')
 		ORDER BY escrow_balance_active DESC
```
ptrus, line 347: Can we use `active_balance` here instead.

storage/client/queries/queries.go
```diff
@@ -286,30 +286,67 @@ const (
...
+			active_nodes.voting_power,
+			COALESCE (
+				(SELECT SUM(voting_power)
+				FROM active_nodes)
+			, 0) AS total_voting_power,
+			COALESCE(chain.commissions.schedule, '{}'::JSONB) AS commissions_schedule,
+			chain.entities.start_date AS start_date,
+			ROW_NUMBER() OVER (ORDER BY escrow_balance_active DESC) AS rank,
+			EXISTS(SELECT NULL FROM chain.nodes WHERE chain.entities.id = chain.nodes.entity_id AND voting_power > 0) AS active,
+			EXISTS(SELECT NULL FROM chain.nodes WHERE chain.entities.id = chain.nodes.entity_id AND chain.nodes.roles like '%validator%') AS status,
```
ptrus, line 334: Could this re-use `active_nodes`?

storage/client/queries/queries.go
```diff
@@ -286,30 +286,67 @@ const (
...
+			chain.entities.meta AS meta
+		FROM chain.entities
+		JOIN chain.accounts ON chain.entities.address = chain.accounts.address
+		LEFT JOIN chain.commissions ON chain.entities.address = chain.commissions.address
+		LEFT JOIN self_delegations ON chain.entities.address = self_delegations.address
+		LEFT JOIN delegators_count ON chain.entities.address = delegators_count.address
+		JOIN active_nodes ON active_nodes.address = entities.address
```
ptrus, line 341: I believe this query currently returns all `chain.entities` even those that were never registered as validators (and don't intend to be).

I think this matches what oasiscan returns and I think it does actually make sense to return this. That way if some validator goes offline, you will still see it in the list of inactive validators.

I would just add a clarification to the API documentation for validator endpoints. Maybe reword it to:

> Returns a list of consensus validators (the list includes all registered entities even those without currently active validator node).

api/spec/v1.yaml
ptrus: > but i'm not sure whether it's worth creating a separate endpoint to return this network-level info

Instead of introducing a new endpoint we could also update `consensus/validators` response type which would include `NetworkInfo` next to `ValidatorList`. 

But it is probably also fine if we keep it as is, if that works better for frontend.

ptrus (commented):
Thanks, looks mostly good. Some minor comments!

storage/client/queries/queries.go
```diff
@@ -286,30 +286,67 @@ const (
 			ORDER BY id DESC`
 
 	ValidatorsData = `
+		WITH self_delegations AS (
```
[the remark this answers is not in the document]
ptrus: Nevermind, not true. The endpoint returns all entities so we should compute this for all (https://github.com/oasisprotocol/nexus/pull/686#discussion_r1637813210).

storage/client/queries/queries.go
```diff
@@ -286,30 +286,67 @@ const (
...
+			SELECT delegatee AS address, SUM(shares) AS shares
+				FROM chain.delegations
+				WHERE delegator = delegatee
+				GROUP BY delegatee
+		),
+		delegators_count AS (
```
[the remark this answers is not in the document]
ptrus: Nevermind, not true. The endpoint returns all entities so we should compute this for all (https://github.com/oasisprotocol/nexus/pull/686#discussion_r1637813210).

api/spec/v1.yaml
```diff
@@ -1874,11 +1881,28 @@ components:
...
+        start_date:
+          type: string
+          format: date-time
+          description: The second-granular consensus time.
+          example: *iso_timestamp_3
+        rank:
+          type: integer
+          format: uint64
+          description: The rank of the Validator, determined by voting power.
         status:
```
mitjat, line 1906: Unrelated to PR, but quick fix: Can we rename this pls? The name is completely uninformative in combination with the bool value. Maybe `is_active`, in line with Peter's https://github.com/oasisprotocol/nexus/pull/686/files#r1637773395 ?

mitjat (commented):
Nice overall! Only a few non-nits from my side.

analyzer/queries/queries.go
```diff
@@ -278,10 +278,18 @@ var (
...
+      $1,
+      $2::text,
+      (SELECT MIN(time) FROM chain.blocks as blocks
+        JOIN chain.transactions as txs ON
+        blocks.height = txs.block AND
+        txs.sender = $2::text AND
+        txs.method = 'registry.RegisterEntity')
+    ON CONFLICT (id) DO
+    UPDATE SET
+      address = excluded.address`
```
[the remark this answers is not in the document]
ptrus: Indeed address of an entity cannot change.

storage/client/queries/queries.go
```diff
@@ -286,30 +286,67 @@ const (
 			ORDER BY id DESC`
 
 	ValidatorsData = `
+		WITH self_delegations AS (
+			SELECT delegatee AS address, SUM(shares) AS shares
```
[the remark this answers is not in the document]
Andrew7234: updated!

api/spec/v1.yaml
```diff
@@ -1893,6 +1917,32 @@ components:
...
+          description: The amount of token currently being undelegated (debonded) from this account.
+        debonding_shares:
+          allOf: [$ref: '#/components/schemas/TextBigInt']
+          description: The shares of tokens being undelegated from this account.
+        self_delegation_balance:
+          allOf: [$ref: '#/components/schemas/TextBigInt']
+          description: The amount of token this validator has delegated to itself.
+        self_delegation_shares:
+          allOf: [$ref: '#/components/schemas/TextBigInt']
+          description: The shares of tokens the validator has delegated to itself.
```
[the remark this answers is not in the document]
Andrew7234: Just active for now; updated the descriptions

api/spec/v1.yaml
```diff
@@ -1893,6 +1917,32 @@ components:
...
+      properties:
+        active_balance:
+          allOf: [$ref: '#/components/schemas/TextBigInt']
+          description: The amount of token actively delegated to this account.
+        active_shares:
+          allOf: [$ref: '#/components/schemas/TextBigInt']
+          description: The shares of tokens delegated to this account.
+        debonding_balance:
+          allOf: [$ref: '#/components/schemas/TextBigInt']
+          description: The amount of token currently being undelegated (debonded) from this account.
```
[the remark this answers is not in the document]
Andrew7234: Thanks, updated!

storage/client/queries/queries.go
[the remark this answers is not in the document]
Andrew7234: Thanks for the detailed explanation! I ended up doing the subquery for now, since it might be hard for users to judge what % of voting power is significant or not. Will keep an eye on query performance and eliminate this if necessary though.

api/spec/v1.yaml
Andrew7234: I started by changing `status` to `is_eligible` instead since we have an `active` field that denotes whether the validator is actively participating (ie whether it's in the top 120). I wanted `is_eligible` to convey whether the entity has an eligible validator node at all, but upon further investigation it turns out that we only have 124 entities with a validator node, which is less than the [226](https://www.oasisscan.com/validators) displayed on oasisscan. I suspect that oasisscan is displaying every entity that ever had a validator node, but I'll need to do more research on what constitutes active/waiting/inactive for a validator. 

Perhaps we only have 4 'waiting' validators and the rest can be considered inactive. At any rate I'll need to update the validators query because it makes some incorrect assumptions about the presence of a node.

I'll likely rename `is_eligible` to `is_waiting` or something similar, but I'll hold off until I'm sure the terminology is correct.

analyzer/queries/queries.go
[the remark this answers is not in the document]
Andrew7234: Fixed!

storage/client/queries/queries.go
Andrew7234: Good call, updated!

api/spec/v1.yaml
```diff
@@ -1874,11 +1881,28 @@ components:
...
+          description: The total voting power across all Validators.
         active:
           type: boolean
           description: Whether the entity is part of validator set (top <scheduler.params.max_validators> by stake).
+        start_date:
+          type: string
+          format: date-time
+          description: The second-granular consensus time.
+          example: *iso_timestamp_3
+        rank:
```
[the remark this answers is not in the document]
Andrew7234: It's cheap to compute and potentially useful for users, so I'll leave it in for now. Will discuss further at the product sync though.

analyzer/queries/queries.go
Andrew7234: Good idea, updated to use height instead!

api/spec/v1.yaml
ptrus: I think there re 3 cases
- active (registered and running)
- "waiting" (registered and running but not in top 120 by stake, or it can also happen that the node just registered and is among top 120, but is not yet eligible - validator needs to be online for 2 epochs before it's picked up in the validator set)
- inactive (an entity that was a validator in past, but currently is not)

So, I think `active` makes sense when discriminating between validators who are online and those who are not. The second one should maybe something in line of `in_committe`/`is_validating`/`in_validator_set`, which would represent whether the validator is in the current validator set committee or not.

Andrew7234: Got it, thanks for the explanation! Updated to `active` and `in_validator_set`. One complication - we don't have a great way currently to distinguish between entities that are inactive validators and entities that have never been a validator since we don't yet store validator history. Once #703 is in, we can use that history to find inactive validators but for now I'll leave this endpoint as-is to only return active validators. Will file a ticket for the follow-up work there.

storage/client/queries/queries.go
Andrew7234: Updated!

storage/client/queries/queries.go
```diff
@@ -286,31 +286,82 @@ const (
...
+		validator_nodes AS (
+			SELECT entities.address, COALESCE(MAX(nodes.voting_power), 0) AS voting_power
+				FROM chain.entities AS entities
+				JOIN chain.nodes AS nodes ON
+					entities.id = nodes.entity_id AND
+					nodes.roles LIKE '%validator%'
+				GROUP BY entities.address
+		),
+		-- Compute validator rank by escrow balance active in a subquery to support querying a validator by address.
+		validator_rank AS (
```
ptrus, line 313: Should we sort/rank this list similarly as Oasis scan does by first listing all validators that are currently validating, and then those that don't (and within each of these two groups sorted by active escrow).

So something like:
```
order by is_validating, chain.accounts.escrow_balance_active DESC
```

storage/client/queries/queries.go
```diff
@@ -286,31 +286,82 @@ const (
...
 	ValidatorsData = `
+		WITH
+		-- Find all self-delegations for all accounts with active delegations.
+		self_delegations AS (
+			SELECT delegatee AS address, shares AS shares
+				FROM chain.delegations
+				WHERE delegator = delegatee
+				GROUP BY delegator, delegatee
+		),
+		-- Compute the number of delegators for each validator.
```
ptrus, line 297:
```suggestion
		-- Compute the number of delegators for each delegatee.
```

Minor nit, since this computes number of delegators for every account, not just validators.

storage/client/queries/queries.go
```diff
@@ -286,31 +286,82 @@ const (
...
+		delegators_count AS (
+			SELECT delegatee AS address, COUNT(*) AS count
+				FROM chain.delegations
+				GROUP BY delegatee
+		),
+		-- Each entity can have multiple nodes. For each entity, find the validator node with the highest voting power.
```
ptrus, line 303: An additional note here would be nice: only one of the entity nodes will have non-zero voting power.

api/spec/v1.yaml
ptrus: > One complication - we don't have a great way currently to distinguish between entities that are inactive validators and entities that have never been a validator

👍  I think that's not that problematic/fine for now.

ptrus (commented):
Thanks, this looks mostly good to me, only a couple of minor remaining comments.

api/spec/v1.yaml
```diff
@@ -1900,14 +1907,31 @@ components:
...
+          type: integer
+          format: uint64
+          description: The rank of the Validator, determined by voting power.
+        in_validator_set:
+          type: boolean
+          description: Whether the entity is part of the validator set (top <scheduler.params.max_validators> by stake).
```
ptrus, line 1934:
```suggestion
          description: Whether the entity is part of the validator set (active and top <scheduler.params.max_validators> by stake among active entities).
```

storage/migrations/22_validator_start.up.sql
```diff
@@ -0,0 +1,22 @@
+BEGIN;
+
+ALTER TABLE chain.entities
+    ADD start_block UINT63;
```
ptrus, line 4: Maybe also add a commented out field to `00_consensus.up.sql`

storage/client/queries/queries.go
Andrew7234: Good call, updated!

api/spec/v1.yaml
Andrew7234: Filed a follow-up ticket: https://app.clickup.com/t/86952ac6h

ptrus (approved):

storage/client/queries/queries.go
ptrus: Sorry, I forgot about this before, but maybe also active before non-active? 

```
ORDER BY is_validating DESC, active DESC, chain.accounts.escrow_balance_active DESC
```

That way entities with potentially a lot of stake that don't even intend to be validators (for whatever reason) will be after entities that are registered for validators but just don't have enough stake to be among top X.

Andrew7234: Ah good catch, agreed and updated!

storage/client/queries/queries.go
lukaw3d: Note: API still exposes `active` field on validators
```
        active:
          type: boolean
          description: Whether the entity has a node that is registered for being a validator, node is up to date, and has successfully registered itself. It may or may not be part of validator set.
```

Should it be renamed?

And frontend https://github.com/oasisprotocol/explorer/pull/1478/files#diff-3d672171124172cffa1ca415897063fa7c5fb37391d024b4febf7ab4e76406e2R20-R21 labels them Active + Waiting. Should we change

Andrew7234: Yep this is intentional; the old `active` field has been renamed to `in_validator_set`. Apologies for the messiness :\ 

I could change the api fields so that we have `active` and `is_eligible`? In this case `active==in_validator_set`, which matches the description on the frontend. `is_eligible` would then be true if the validator node is registered and running. Wdyt?

The changes in the linked PR look good, the Explorer should continue to show Active/Waiting/Inactive.

ptrus: I'd keep `in_validator_set` and `active`, since these make the most sense to me. Also frontend PR seems to be updated to work correctly with these two, so let's keep it as it is?

Andrew7234: Agreed, I'll leave as-is and we can revisit in a follow-up if necessary.

PR #687, opened 2024-04-30
Add data about known addresses

csillag:
We have decided to temporarily store the metadata about known addresses in the nexus repo, so that the frontends can pull this directly from github.

Later, hopefully, Nexus will serve this data via some API.

See https://app.clickup.com/t/8693rg7nz for some details.

### DIFF
--- a/.changelog/2663.internal.md
+++ b/.changelog/2663.internal.md
@@ -1,0 +1,1 @@
+Add data about known addresses
--- a/account-names/mainnet_consensus.json
+++ b/account-names/mainnet_consensus.json
@@ -1,0 +1,37 @@
+[
+    {
+        "Name": "Common Pool",
+        "Address": "oasis1qrmufhkkyyf79s5za2r8yga9gnk4t446dcy3a5zm",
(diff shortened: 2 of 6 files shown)


pro-wh:
looking good. can we go without the .idea directory?

csillag:
> can we go without the .idea directory?

Ahh sorry, I have no idea how did it manage to sneak on board. Fixed.

pro-wh (approved):
thanks

.changelog/2663.internal.md
```diff
@@ -0,0 +1 @@
+Add data about known addresses
```
buberdds, line 1: Filename is incorrect. Requires tweak after running make changelog.

pro-wh: oh shoot was it wrong number?

buberdds: yeah looks like it was created manually with some random number

csillag: Sorry about that, I have no idea where that number came from.

pro-wh: I think we can open a PR to rename it

PR #689, opened 2024-05-07
rename changelog entry

pro-wh:
#687 changelog entry had the wrong number in the filename


pro-wh:
o crud the mock e2e regression test workflow has the wrong names

Andrew7234 (approved):

PR #691, opened 2024-05-11
ERC-20 token prices

pro-wh:
this adds a "relative" token price estimate and total value estimate to the tokens API.

this data comes from on-chain swap contracts. for contracts that follow the Uniswap v2 interface, we index the swap contracts in a new `chain.evm_swap_pairs` table. the pair contracts emit a "Sync" event, which reports a contract's reserves, and the contract trades its two tokens in proportion to its reserves of the two tokens.

we add a new configuration that specifies a collection of swap contracts identified by their "factory" contract plus a "reference" token. the relative token price estimate for a token is available when there is a swap pair contract for that token and the reference token or if the token is the reference token itself.

in this PR, we select illumineX swap and Wrapped ROSE (0x8Bc2B030b299964eEfb5e1e0b36991352E56D2D3) for Sapphire mainnet.


pro-wh:
this adds gpl'd code to our repo

storage/client/queries/queries.go
```diff
@@ -578,20 +578,47 @@ const (
...
 		FROM chain.evm_tokens AS tokens
 		JOIN chain.address_preimages AS preimages ON (token_address = preimages.address AND preimages.context_identifier = 'oasis-runtime-sdk/address: secp256k1eth' AND preimages.context_version = 0)
 		LEFT JOIN holders USING (token_address)
 		LEFT JOIN chain.evm_contracts as contracts ON (tokens.runtime = contracts.runtime AND tokens.token_address = contracts.contract_address)
+		LEFT JOIN chain.evm_swap_pairs AS ref_swap_pairs ON
+		    ref_swap_pairs.runtime = $1 AND
+		    ref_swap_pairs.factory_address = $4 AND
+		    (
+		    	(tokens.token_address < $5 AND ref_swap_pairs.token0_address = tokens.token_address AND ref_swap_pairs.token1_address = $5) OR
+		    	(tokens.token_address > $5 AND ref_swap_pairs.token0_address = $5 AND ref_swap_pairs.token1_address = tokens.token_address)
```
pro-wh, line 605: indentation here is wrong

api/spec/v1.yaml
```diff
@@ -2943,6 +2943,46 @@ components:
...
+          type: string
+        ref_swap_pair_address_eth:
+          type: string
+        ref_swap_factory_address:
+          type: string
+        ref_swap_factory_address_eth:
+          type: string
+        ref_swap_token0_address:
+          type: string
+        ref_swap_token0_address_eth:
```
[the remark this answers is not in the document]
Andrew7234: Oh I see the `/evm_tokens` endpoints need to be updated with the additional params

config/reference_swaps.go
```diff
@@ -0,0 +1,24 @@
+package config
+
+import (
+	apiTypes "github.com/oasisprotocol/nexus/api/v1/types"
+	"github.com/oasisprotocol/nexus/common"
+)
+
+type ReferenceSwap struct {
+	FactoryAddr        apiTypes.Address `koanf:"factory_addr"`
+	ReferenceTokenAddr apiTypes.Address `koanf:"reference_token_addr"`
```
Andrew7234, line 10: should these be configurable in the config file as well?

Andrew7234 (commented):
Thanks for putting this all together! Looks good so far

analyzer/queries/queries.go
```diff
@@ -956,6 +956,25 @@ var (
       token_address = $2 AND
       account_address = $3`
 
+	RuntimeEVMSwapPairUpsertCreated = `
+    INSERT INTO chain.evm_swap_pairs (runtime, pair_address, factory_address, token0_address, token1_address, create_round)
+    VALUES ($1, $2, $3, $4, $5, $6)
+    ON CONFLICT (runtime, pair_address) DO UPDATE
+    SET
+      factory_address = excluded.factory_address,
```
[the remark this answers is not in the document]
pro-wh: this is meant to "change" it from NULL, in the case where out-of-order processing causes us to see a Sync from the address before we find out what factory it was from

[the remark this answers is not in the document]
pro-wh: huh but the security might not be good, how do we prevent any arbitrary contract from claiming a pair by emitting a pair create event?

storage/client/queries/queries.go
```diff
@@ -578,20 +578,43 @@ const (
...
+			ref_swap_pairs.last_sync_round AS ref_swap_last_sync_round,
+			swap_last_sync_blocks.timestamp AS ref_swap_last_sync_block_timestamp,
 			contracts.verification_level
 		FROM chain.evm_tokens AS tokens
 		JOIN chain.address_preimages AS preimages ON (token_address = preimages.address AND preimages.context_identifier = 'oasis-runtime-sdk/address: secp256k1eth' AND preimages.context_version = 0)
 		LEFT JOIN holders USING (token_address)
 		LEFT JOIN chain.evm_contracts as contracts ON (tokens.runtime = contracts.runtime AND tokens.token_address = contracts.contract_address)
+		LEFT JOIN chain.evm_swap_pairs AS ref_swap_pairs ON
+			ref_swap_pairs.runtime = $1 AND
+			ref_swap_pairs.factory_address = $4 AND
```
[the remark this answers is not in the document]
pro-wh: not aware of any performance difference, we can do it that way instead

api/spec/v1.yaml
```diff
@@ -2943,6 +2943,46 @@ components:
...
+          type: string
+        ref_swap_token1_address:
+          type: string
+        ref_swap_token1_address_eth:
+          type: string
+        ref_swap_create_round:
+          type: integer
+          format: int64
+        ref_swap_create_block_timestamp:
```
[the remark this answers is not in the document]
pro-wh: don't know if the mocks call for timestamp. I just thought I'd like to see it in the API. we can go for the simpler query without the block though

api/spec/v1.yaml
```diff
@@ -2943,6 +2943,46 @@ components:
...
+          type: integer
+          format: int64
+        ref_swap_last_sync_block_timestamp:
+          type: string
+          format: date-time
+        relative_token_address:
+          type: string
+        relative_price:
+          type: number
+          format: double
```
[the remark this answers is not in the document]
pro-wh: internally it's calculated as a double. these are approximate anyway, so rounding errors shouldn't break things. use cases that need accurate info should use the reserve numbers and real uniswap math

api/spec/v1.yaml
```diff
@@ -2943,6 +2943,46 @@ components:
           description: |
             The number of addresses that have a nonzero balance of this token.
           example: 123
+        ref_swap_pair_address:
```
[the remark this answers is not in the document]
pro-wh: they are conceptually together...

api/spec/v1.yaml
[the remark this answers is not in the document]
pro-wh: no, the idea in this PR is to have a single "reference" token and this relative value functionality only looks at swapping against that reference token

config/reference_swaps.go
pro-wh: if desired :shrug:

analyzer/queries/queries.go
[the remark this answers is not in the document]
pro-wh: let's make it a many-factory-to-one-pair table. any contract can claim it's the factory of a pair, and we only check that our reference factory has an entry in that table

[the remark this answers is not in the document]
pro-wh: ok multi-factory-claim issue fixed. now there's a separate table for PairCreated records

storage/client/queries/queries.go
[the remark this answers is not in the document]
pro-wh: changed to do it that way

api/spec/v1.yaml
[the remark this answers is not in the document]
pro-wh: ok I'll separate the ref_swap stuff into a reference swap object

api/spec/v1.yaml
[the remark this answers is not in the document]
pro-wh: gonna leave this as it is

api/spec/v1.yaml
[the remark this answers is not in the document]
pro-wh: separated

analyzer/runtime/extract.go
```diff
@@ -1037,6 +1056,69 @@ func extractEvents(blockData *BlockData, relatedAccountAddresses map[apiTypes.Ad
 					}
 					return nil
 				},
+				IUniswapV2FactoryPairCreated: func(token0ECAddr ethCommon.Address, token1ECAddr ethCommon.Address, pairECAddr ethCommon.Address, allPairsLength *big.Int) error {
```
Andrew7234, line 1059: Is the plan to populate the `eventData.{RelatedAddresses, EvmLog{Name,Signature,Params}}` in this PR? I imagine we'll want to show this info in the explorer eventually

api/spec/v1.yaml
```diff
@@ -3082,6 +3093,37 @@ components:
...
+        token0_address_eth:
+          type: string
+        token1_address:
+          type: string
+        token1_address_eth:
+          type: string
+        create_round:
+          type: integer
+          format: int64
+        reserve0:
```
Andrew7234, line 3119: does this correspond to `token0`? If so can we name it something like `token0_reserve`?

config/reference_swaps.go
Andrew7234: Hm it'd be nice to have in the config file; given how infrequently we cut releases 😅 

> sorting by value will make this less easy to change

Hm would this require everything to be recalculated, eg in a reindex? 

Also how is the reference token used? Is it the base unit to express all other prices? Eg 1 wrapped eth would be 10^5 WROSE?

analyzer/runtime/extract.go
pro-wh: oh have we been doing EvmLog ones?

related addresses we should do

api/spec/v1.yaml
pro-wh: corresponds to token0. but I prefer the uniswapv2 name reserve0

config/reference_swaps.go
pro-wh: we don't have the sorting yet, so let's move this to a config. it might not even be a problem later e.g. if we can sort at query time

analyzer/runtime/extract.go
pro-wh: oh I see about the evm log stuff. I'll try doing it similar to the surrounding code

pro-wh: added

analyzer/queries/queries.go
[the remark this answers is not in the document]
pro-wh: ok so in conclusion, this is now in a table that lets any contract claim any pair, and which the "factory" in that relation won't change. resolving

api/spec/v1.yaml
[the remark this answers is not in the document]
pro-wh: gonna resolve this, I've removed them. they were nice though imo

api/spec/v1.yaml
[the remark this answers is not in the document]
pro-wh: oh I forgot to answer one of the questions: will it work if a coin is part of multiple pair contracts -> yes, but it'll only look at one pair of the coin with the chosen reference coin. other pairs shouldn't interfere

analyzer/runtime/extract.go
pro-wh: gonna resolve

config/reference_swaps.go
pro-wh: ok it's put in awkwardly in CustomChainConfig, so do

```yaml
server:
  chain_name: mainnet # irrelevant
  source:
    chain_name: mainnet # use defaults for a known named network
```

or

```yaml
server:
  source:
    custom_chain:
      reference_swaps:
        sapphire:
          factory_addr: oasis1...
          reference_token: oasis1...
```

pro-wh: leaving this open for a reaction

pro-wh: also leaving server.source blank---which seems to be legal---causes no reference swap to be chosen, which will make the API never return relative price info and reference swap info

api/spec/v1.yaml
pro-wh: keep these as is

config/reference_swaps.go
pro-wh: ok so summary of the current stuff: changing the reference swap info won't need a reindex. we index all events that match the signatures, and the API selects based on the configuration

analyzer/runtime/extract.go
```diff
@@ -1037,6 +1056,209 @@ func extractEvents(blockData *BlockData, relatedAccountAddresses map[apiTypes.Ad
...
+						{
+							Name:    "reserve0",
+							EvmType: "uint112",
+							// JSON supports encoding big integers, but many clients (javascript, jq, etc.)
+							// will incorrectly parse them as floats. So we encode uint256 as a string instead.
+							Value: reserve0.String(),
+						},
+						{
+							Name:    "reserve1",
+							EvmType: "uint112",
```
Andrew7234, line 1254: TIL there's a `uint112`

analyzer/runtime/extract.go
```diff
@@ -1037,6 +1056,209 @@ func extractEvents(blockData *BlockData, relatedAccountAddresses map[apiTypes.Ad
...
+					eventData.RelatedAddresses[token1Addr] = struct{}{}
+					pairAddr, err := addresses.RegisterRelatedEthAddress(blockData.AddressPreimages, relatedAccountAddresses, pairECAddr.Bytes())
+					if err != nil {
+						return fmt.Errorf("pair: %w", err)
+					}
+					eventData.RelatedAddresses[pairAddr] = struct{}{}
+					blockData.SwapCreations[SwapCreationKey{
+						Factory: eventAddr,
+						Token0:  token0Addr,
+						Token1:  token1Addr,
```
Andrew7234, line 1078: Does uniswap enforce some ordering on token0/token1? Like is it possible to create a usdt/wrose pair and then create a wrose/usdt pair that's distinct from it? (and if uniswap combines the pairs, would nexus do the same?)

api/spec/v1.yaml
```diff
@@ -3082,6 +3106,61 @@ components:
...
+          description: |
+            The round when this swap pair was created.
+        reserve0:
+          allOf: [$ref: '#/components/schemas/TextBigInt']
+          description: |
+            The swap's liquidity pool of the first token, in that token's base units.
+        reserve1:
+          allOf: [$ref: '#/components/schemas/TextBigInt']
+          description: |
+            The swap's liquidity pool of the second token, in that token's base units.
```
Andrew7234, line 3157: Is there any decimal information available? Or would the explorer just display in base units

config/reference_swaps.go
Andrew7234: Wdyt of moving `reference_swaps` to be a child of `server` instead? Or is that too repetitive? I'm a bit hesitant to have to use `custom_chain` to specify additional details on a known default network. Also it'd be nice to use the default networks to keep our mainnet/testnet deploy config easy to maintain.

Andrew7234 (commented):
Thanks! The api descriptions are looking good; left a couple questions too.

analyzer/runtime/extract.go
pro-wh: my reading of the actual contract is they have to have the lower token address in token0

api/spec/v1.yaml
pro-wh: not from here, you'd have to get that from the token info

config/reference_swaps.go
Andrew7234: oh I think I misunderstood; the config lgtm 👍

api/spec/v1.yaml
```diff
@@ -3070,6 +3070,30 @@ components:
           description: |
             The number of addresses that have a nonzero balance of this token.
           example: 123
+        ref_swap:
+          allOf: [$ref: '#/components/schemas/EvmTokenSwap']
+          description: |
+            Information about a swap contract between this token and a
+            reference token. The relative price and relative total value of
+            this token are estimated based on this swap contract.
+
```
Andrew7234, line 3079:
```suggestion
```

api/spec/v1.yaml
Andrew7234: Hm would it be possible to add it? (on a best-effort approach, since the `evm_tokens` analyzer might not have processed it or even run at all). The explorer will probably need this info, and would have to dispatch the second token info query afterwards which might be slow.

Andrew7234 (approved):
Thank you! Looks good, left a minor comment

api/spec/v1.yaml
pro-wh: ya I'll try joining this in

pro-wh: added

api/spec/v1.yaml
pro-wh: should be removed now

pro-wh:
had to disable the gocyclo on visit evm event

PR #693, opened 2024-05-16
node stats analyzer: handle nonexistent runtimes gracefully

Andrew7234:
the node stats analyzer errored after redeploying on mainnet. This PR cleans up the analyzer and makes it more robust to unsupported runtimes.


pro-wh (approved):
oo this new map thing looks good

PR #694, opened 2024-05-16
Andrew7234/evm tokens bugfix

Andrew7234:
nexus logs show lots of errors when trying to download a token; turns out we were using the wrong error in some checks

```
"downloading new token oasis1qrez2td5r99j9v5ezv85ts647e909xuvygvd5fmg: download token ERC-721: calling name: %!w(<nil>)"
```

### DIFF
--- a/.changelog/694.bugfix.md
+++ b/.changelog/694.bugfix.md
@@ -1,0 +1,1 @@
+evm_tokens: use right error struct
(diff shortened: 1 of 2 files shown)


pro-wh (approved):
o shoot yeah
tsk go error handling

PR #695, opened 2024-05-20
Fix regression tests

vitrvvivs:
It must be pulling data from somewhere, because  the tests went from passing to failing 4 days ago, on multiple branches, without any changes.

https://github.com/oasisprotocol/nexus/actions/runs/9105766749/attempts/1
https://github.com/oasisprotocol/nexus/actions/runs/9105766749/attempts/2


Andrew7234 (approved):
thanks!

Andrew7234:
thanks for digging into that, fix looks good

PR #696, opened 2024-05-21
feature: add pontusx_{test, dev} support

Andrew7234:
Adds support for the incoming pontusx_dev paratime and the rename of the existing pontusx paratime -> pontusx_test

Testing:
- Ran the current main branch on testnet pontusx for 120 blocks, then switched to this feature branch and ran Nexus on the existing db to confirm that the migrations work as expected.
- Was able to query blocks from both pontusx runtimes successfully, but didn't find any txs in the first couple hundred blocks. Need to figure out where the first txs are.

todo:
- bump oasis client-sdk version once a new release is cut [x]
- confirm whether new pontusxdev runtime will require a new upgrade/deployment of testnet nodes. does it start at 0? edit: yes and yes.
- db migration to move existing pontusx stuff to pontusx-dev? or could just reindex [x]
  - added migration.
- test once pontusx dev is available on oasis-node [x]
Screenshot 2024-05-27 at 2 42 24 PM


common/types.go
```diff
@@ -224,21 +224,23 @@ const (
...
-	LayerEmerald   Layer = "emerald"
-	LayerCipher    Layer = "cipher"
-	LayerSapphire  Layer = "sapphire"
-	LayerPontusx   Layer = "pontusx"
+	LayerConsensus  Layer = "consensus"
+	LayerEmerald    Layer = "emerald"
+	LayerCipher     Layer = "cipher"
+	LayerSapphire   Layer = "sapphire"
+	LayerPontusx    Layer = "pontusx"
+	LayerPontusxDev Layer = "pontusx_dev"
```
pro-wh, line 232: is it weird that we use underscore here but no underscore in the routes?

config/config.go
```diff
@@ -91,6 +91,16 @@ func (cfg *AnalysisConfig) Validate() error {
 			return err
 		}
 	}
+	if cfg.Analyzers.Pontusx != nil {
+		if err := cfg.Analyzers.Pontusx.Validate(); err != nil {
```
pro-wh, line 95: huh, did we omit this last time

pro-wh (approved):
looks ok, fix anything if we find a mistake later

config/config.go
Andrew7234: yeah

common/types.go
Andrew7234: hm we could change the routes to have the underscore but it seems a bit strange. we do have to keep the underscore here though to align with the client-sdk config.

PR #697, opened 2024-05-21
augment ix_accounts_related_transactions_address_block index

Andrew7234:
@csillag reported that the `/consensus/transactions?rel={}` endpoint was very slow. Further investigation showed that the existing index did not contain enough information, which is fixed in this PR.

testing: Already enabled in testnet production; sample query is now performant: https://nexus.prd.oasis.io/v1/consensus/transactions?limit=10&offset=0&rel=oasis1qp65laz8zsa9a305wxeslpnkh9x4dv2h2qhjz0ec

### DIFF
--- a/.changelog/697.bugfix.md
+++ b/.changelog/697.bugfix.md
@@ -1,0 +1,1 @@
+db: augment accounts_related_transactions_address index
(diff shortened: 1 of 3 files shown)


pro-wh (approved):
ya, so it can get recent transactions given an account

PR #698, opened 2024-05-22
Andrew7234/consensus account stats

Andrew7234:
Explorer folks requested account-level stats for consensus. This PR adds `num_txns` stats for this purpose. Note that we do not track total sent/received currently and so those fields are now optional.

### DIFF
--- a/tests/e2e_regression/damask/expected/account.body
+++ b/tests/e2e_regression/damask/expected/account.body
@@ -6,5 +6,8 @@
   "debonding_delegations_balance": "0",
   "delegations_balance": "0",
   "escrow": "0",
-  "nonce": 5
+  "nonce": 5,
+  "stats": {
+    "num_txns": 0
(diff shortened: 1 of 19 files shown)


tests/e2e_regression/damask/expected/account.body
```diff
@@ -6,5 +6,8 @@
   "debonding_delegations_balance": "0",
   "delegations_balance": "0",
   "escrow": "0",
-  "nonce": 5
+  "nonce": 5,
+  "stats": {
+    "num_txns": 0
```
pro-wh, line 11: is this from transactions outside of the block scanning range? either before or after

Andrew7234: yeah; the txs are from before the block scanning range.

pro-wh (approved):
yeah

PR #701, opened 2024-05-28
pontusx/pontusx_dev post-launch bugfixes

Andrew7234:
Previously, nexus would directly use the paratime name in the sdk-config, which is inconsistent with our internal `runtime` enum in the database. We are planning to [change the sdk-config](https://github.com/oasisprotocol/oasis-sdk/pull/1797), so this PR is intended as a patch until the upstream fix goes in.

Also contains several bugfixes for the pontusx api server. Already deployed in staging/production.

### DIFF
--- a/analyzer/node_stats/node_stats.go
+++ b/analyzer/node_stats/node_stats.go
@@ -44,9 +44,9 @@ func NewAnalyzer(
 		cfg.Interval = 3 * time.Second
 	}
 	logger = logger.With("analyzer", nodeStatsAnalyzerName)
-	// Default to [consensus, emerald, sapphire, pontusx] if layers is not specified.
+	// Default to [consensus, emerald, sapphire, pontusx_test, pontusx_dev] if layers is not specified.
 	if len(layers) == 0 {
-		layers = []common.Layer{common.LayerConsensus, common.LayerEmerald, common.LayerSapphire, common.LayerPontusx}
+		layers = []common.Layer{common.LayerConsensus, common.LayerEmerald, common.LayerSapphire, common.LayerPontusxTest, common.LayerPontusxDev}
 	}
 	p := &processor{
 		layers:          layers,
--- a/api/middleware.go
+++ b/api/middleware.go
@@ -279,8 +279,8 @@ func RuntimeFromURLMiddleware(baseURL string) func(next http.Handler) http.Handl
 				runtime = common.RuntimeSapphire
 			case strings.HasPrefix(path, "/cipher/"):
 				runtime = common.RuntimeCipher
-			case strings.HasPrefix(path, "/pontusx/"):
-				runtime = common.RuntimePontusx
+			case strings.HasPrefix(path, "/pontusxtest/"):
+				runtime = common.RuntimePontusxTest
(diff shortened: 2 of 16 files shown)


ptrus:
This is deployed on testnet right? But once https://github.com/oasisprotocol/oasis-sdk/pull/1797 goes in, we don't need this patch?

Lets try to get the SDK PR merged soon and open a PR in Nexus that updates the SDK version.
That way we can avoid having to merge this PR and then reverting it in a short-term.

ptrus:
I guess the SDK PR won't get merged. So I guess this should be refactored in a proper solution.

storage/client/client.go
```diff
@@ -68,14 +68,35 @@ func translateTokenType(tokenType common.TokenType) apiTypes.EvmTokenType {
 	}
 }
 
+// The apiTypes Layers may be named differently from Nexus-internal Layers
+// to make the api more ergonomic. Currently the only differing layer is
```
mitjat, line 72: If we make the API less ergonomic (that's arguable anyway), does that make this PR much simpler? I'm sure the frontend can adapt.

api/middleware.go
```diff
@@ -279,8 +279,8 @@ func RuntimeFromURLMiddleware(baseURL string) func(next http.Handler) http.Handl
 				runtime = common.RuntimeSapphire
 			case strings.HasPrefix(path, "/cipher/"):
 				runtime = common.RuntimeCipher
-			case strings.HasPrefix(path, "/pontusx/"):
-				runtime = common.RuntimePontusx
+			case strings.HasPrefix(path, "/pontusxtest/"):
+				runtime = common.RuntimePontusxTest
```
Andrew7234, line 283: Note: Explorer folks will have to update this since pontusx data on testnet will now be served under the `pontusxtest` prefix. If/when pontusx on mainnet is enabled, this section will need to be updated as well.

storage/client/client.go
Andrew7234: Hm not by much, the bulk of this PR was for translating `pontusx_test` in the sdk to `pontusx` in Nexus. I updated the PR to now just use `pontusx_test` everywhere, so now the `pontusx_test` data is served under `/pontusxtest/`, see comment below.

analyzer/node_stats/node_stats.go
```diff
@@ -44,9 +44,9 @@ func NewAnalyzer(
 		cfg.Interval = 3 * time.Second
 	}
 	logger = logger.With("analyzer", nodeStatsAnalyzerName)
-	// Default to [consensus, emerald, sapphire, pontusx] if layers is not specified.
+	// Default to [consensus, emerald, sapphire, pontusx_dev] if layers is not specified.
```
ptrus, line 47: Why dont we also default to pontusx_test?

ptrus (approved):
Thanks!

analyzer/node_stats/node_stats.go
Andrew7234: I hesitated because `pontusx_test` is only on testnet, but then decided to add it anyway. At worst, this default config will lead to a bit of additional error spam.

PR #708, opened 2024-06-14
storage/consensus: Add index for transaction type

ptrus:
Adds an index on consensus transactions method.

Additionally removes `minFee`, `maxFee` and `code` query filters from `/consensus/transactions`. These filters don't seem all that useful (I have confirmed with frontend that these are not needed) and don't have any indexes set, so these aren't usable in practice anyway.


storage/migrations/16_consensus_tx_index.up.sql
```diff
@@ -0,0 +1,5 @@
+BEGIN;
+
+CREATE INDEX ix_transactions_method ON chain.transactions (method);
```
pro-wh, line 3: this kind of index usually ends up needing `height` at the end, otherwise the api can't get "recent transactions of x method" without walking this index for all transactions of a method and sorting them or ignoring this index and scanning chronologically and filtering by method

pro-wh:
ok on removing fee and code filters :shrug: seems like they would have been an easy way to DoS

storage/migrations/16_consensus_tx_index.up.sql
ptrus: Good point, will update 👍

ptrus: Do you think It might make sense to update the `ix_transactions_sender` index in the same way?

Most of the accounts probably have a relatively low number of transactions which are fine, but there are likely some accounts with huge amount of transactions which could cause problems for the "recent transactions of x sender" query, since it would need to sort the filtered list of transactions which could be large?

Likely way less problematic, I should probably find the accounts with most transactions and test the queries on a live nexus deployment.

ptrus: Adding the `ix_transactions_sender_block` index fixes the above examples to be quick.

storage/migrations/17_consensus_tx_index.up.sql
```diff
@@ -0,0 +1,8 @@
+BEGIN;
```
ptrus, line 1: Added these indexes might take non-trivial time. Do we have any special requirements or handling of such migrations that aren't super quick? 

The specific changes in this migration are not breaking and could actually be made whenever (don't need to happen at deployment), so these can be handled that way, but i'm asking more for any general guidelines.

.changelog/708.breaking.md
mitjat, line 1: I forget, when did we say we were going to call a change "breaking"? Is it whenever it introduces a backwards-incompatible DB change? I _think_ this is not breaking because:
- We're breaking the API, but it's the consensus API which is not officially released yet, so we're OK to muck with it. Even if this were an "official" breaking API change, I guess it would mean that we need to introduce api/spec/v2.yaml (and do some thinking about how we'll manage multiple versions), but nexus itself would stay at the same major version? This philosophy is untested :)
- We're not breaking nexus: A current nexus operator can deploy this PR over an existing DB, and it will keep working thanks to migrations. An example of a breaking change here is #574 which requires an operator to be very careful and hands-on when deploying.

mitjat (approved):

.changelog/708.breaking.md
ptrus: I noticed that https://github.com/oasisprotocol/nexus/pull/685 was marked as a breaking change and it is only an API change.

But i agree that the state (consensus part of) nexus is currently in, then most changes would be breaking. I'll make it non breaking, imo it doesn't really matter that much as there's not many external uses of nexus i think.

storage/migrations/17_consensus_tx_index.up.sql
ptrus: Thanks for the info. For now I updated just update the migration to do `CREATE INDEX IF NOT EXISTS` so that we are more flexible on how to handle this deployment.

storage/migrations/17_consensus_tx_index.up.sql
```diff
@@ -0,0 +1,8 @@
...
+
+CREATE INDEX ix_transactions_method_height ON chain.transactions (method, block);
```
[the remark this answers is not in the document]
ptrus: 👍 good idea

PR #709, opened 2024-06-17
api/epochs: Skip 'end_height' for latest epoch

ptrus:
The `/consensus/epochs/<latest_epoch>` endpoint for the currently active epoch returns the ever increasing 'end_height'.
This is inconsistent with the API spec and also with the `/consensus/epochs`. Skip returning the `end_height` for the latest epoch.

For `/consensus/epochs` we already do this.

### DIFF
--- a/storage/client/queries/queries.go
+++ b/storage/client/queries/queries.go
@@ -252,14 +252,10 @@ const (
 		SELECT id, start_height,
 			(CASE id WHEN (SELECT max(id) FROM chain.epochs) THEN NULL ELSE end_height END) AS end_height
 			FROM chain.epochs
+		WHERE ($1::bigint IS NULL OR id = $1::bigint)
 		ORDER BY id DESC
-		LIMIT $1::bigint
-		OFFSET $2::bigint`
-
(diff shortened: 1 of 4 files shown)


storage/client/queries/queries.go
```diff
@@ -257,7 +257,8 @@ const (
 		OFFSET $2::bigint`
 
 	Epoch = `
-		SELECT id, start_height, end_height
+		SELECT id, start_height,
+			(CASE id WHEN (SELECT max(id) FROM chain.epochs) THEN NULL ELSE end_height END) AS end_height
```
mitjat, line 261: Optional:
To simplify the code and bring it more in line with other similar queries, we could abandon the `Epoch` SQL query and instead expand `Epochs` with `WHERE ($1::bigint IS NULL OR $1::bigint = id)`.

mitjat (approved):
Nice, thanks!

storage/client/queries/queries.go
ptrus: Updated 👍

PR #710, opened 2024-06-18
api/proposals: decode parameters_change for proposals

ptrus:
Frontend deosn't want to deal with cbor. So let's decode it for them and return a JSON instead.

Example response:
```
curl localhost:8099/v1/consensus/proposals/9
{"closes_at":30705,"created_at":30693,"deposit":"10000000000000","id":9,"invalid_votes":"0","parameters_change":{"upgrade_cancel_min_epoch_diff":15},"parameters_change_module":"governance","state":"passed","submitter":"oasis1qrx85mv85k708ylww597rd42enlzhdmeu56wqj72","target":{}}
```


analyzer/consensus/genesis.go
```diff
@@ -296,7 +295,7 @@ func (mg *GenesisProcessor) addGovernanceBackendMigrations(batch *storage.QueryB
 				proposal.State.String(),
 				proposal.Deposit.ToBigInt(),
 				proposal.Content.ChangeParameters.Module,
-				hex.EncodeToString(proposal.Content.ChangeParameters.Changes),
+				proposal.Content.ChangeParameters.Changes,
```
ptrus, line 298: This was actually broken since https://github.com/oasisprotocol/nexus/pull/630 

The old version had to use `hex.EncodeToString` with `decode('%s', 'hex')` since it was working with strings. https://github.com/oasisprotocol/nexus/pull/630/files#diff-dad00025ae7b123664532df3771612a4339161128cb7de77fcc18af2bfa122faL513-L521

Probably none of the deployments had a reindex done which since then, so it is probably ok on all live deployments.

storage/client/proposals.go
```diff
@@ -0,0 +1,98 @@
...
+
+	"github.com/oasisprotocol/oasis-core/go/common/cbor"
+
+	governanceV22 "github.com/oasisprotocol/nexus/coreapi/v22.2.11/governance/api"
+	registryV22 "github.com/oasisprotocol/nexus/coreapi/v22.2.11/registry/api"
+	roothashV22 "github.com/oasisprotocol/nexus/coreapi/v22.2.11/roothash/api"
+	schedulerV22 "github.com/oasisprotocol/nexus/coreapi/v22.2.11/scheduler/api"
+	stakingV22 "github.com/oasisprotocol/nexus/coreapi/v22.2.11/staking/api"
+	governanceV23 "github.com/oasisprotocol/nexus/coreapi/v23.0/governance/api"
+	keymanagerV23 "github.com/oasisprotocol/nexus/coreapi/v23.0/keymanager/api"
```
Andrew7234, line 15: Just checking - is `keymanagerV22` not present because the struct doesn't exist in damask?

Also, do we need to include the cobalt types here? [convert_tx.go](https://github.com/oasisprotocol/nexus/blob/main/analyzer/consensus/convert_tx.go) might be a helpful reference.

analyzer/consensus/genesis.go
Andrew7234, line 298: Wdyt of converting the cbor to json here in the analyzer and then storing it as a `JSONB` column in the database? It would be nice to do as much of the processing in the analyzers as possible to keep the api server fast/simple.

Andrew7234 (commented):
Looking good overall! Left a couple questions

analyzer/consensus/genesis.go
ptrus: Hm i think I would slightly prefer having the raw field in the DB, since that matches what was actually proposed. But maybe we could store both the raw cbor and the parsed JSONB.

Maybe we can keep this as a potential followup in future, the number of proposals is really low (4 on mainnet and 11 on testnet) so I don't think this will be problematic anytime soon.
(and even out of those 15 proposals, in total 3 are of the change-parameter types, for which this parsing is relevant).

storage/client/proposals.go
ptrus: There's no such types for Cobalt, since `ChangeParameters` proposals were only introduced with Damask. 

But v24 does introduce a new consensus app - Vault, which also has `ChangeProposals`. We need to add v24 types. I opened: https://app.clickup.com/t/8694wjwh2

v24 is generally non-breaking (only adds stuff) so I think we should be able to add the new stuff into the v23.0?

ptrus: Done in https://github.com/oasisprotocol/nexus/pull/716 will add `vault.ChangeParameters` here, once that PR is merged and this one is rebased.

ptrus: Added

analyzer/consensus/genesis.go
Andrew7234: ah makes sense, keeping it as a follow-up sgtm

Andrew7234 (approved):

PR #711, opened 2024-06-18
delegations: remove delegations with 0 shares

ptrus:
This will ensure that the delegations with 0 shares get cleaned up/removed from the table.

### DIFF
--- a/analyzer/consensus/consensus.go
+++ b/analyzer/consensus/consensus.go
@@ -921,6 +921,33 @@ func (m *processor) queueEscrows(batch *storage.QueryBatch, data *stakingData) e
 			e.Owner.String(),
 			e.ActiveShares.String(),
 		)
+		// Ideally this would be merged with the ConsensusDebondingStartDelegationsUpdate query above,
+		// something like:
+		//
+		// ```
+		//   WITH updated AS (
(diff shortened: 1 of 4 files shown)


analyzer/queries/queries.go
```diff
@@ -338,15 +338,6 @@ var (
         escrow_balance_active = chain.accounts.escrow_balance_active + $2,
         escrow_total_shares_active = chain.accounts.escrow_total_shares_active + $3`
 
-	ConsensusAddDelegationsUpsertReplace = `
```
ptrus, line 341: This was unused.

mitjat (approved):
Thank you!

analyzer/consensus/consensus.go
```diff
@@ -921,6 +921,33 @@ func (m *processor) queueEscrows(batch *storage.QueryBatch, data *stakingData) e
 			e.Owner.String(),
 			e.ActiveShares.String(),
 		)
+		// Ideally this would be merged with the ConsensusDebondingStartDelegationsUpdate query above,
```
ptrus, line 924: @mitjat note that the previous version didn't actually work, i had to refactor the CTE into two separate queries.

PR #712, opened 2024-06-21
debonding_delegations: remove id from the table

ptrus:
Fix handling of multiple debonding-delegations ending in the same epoch. 

**Scenario:**
A person undelegates twice within the same epoch, first X, then Y shares. This ends up in two `DebondingStartEscrowEvent` events with the same `DebondEndTime`. 

_Note that this is likely a common scenario since people tend to first issue a small reclaim escrow as a test, and then soon after a larger reclaim escrow for the rest of the shares._

**Oasis-node handling:** 
Oasis-node [internally merges this case](https://github.com/oasisprotocol/oasis-core/blob/774316020d821c99307f444a4de21775fc01f048/go/consensus/cometbft/apps/staking/transactions.go#L488-L492) into a single debonding-delegation, therefore once the stake is debondend, only a single `ReclaimEscrowEvent` event is returned from oasis-node, with the sum of shares `X+Y`.

**Nexus handling:**
Nexus inserts two debonding delegatinos into the DB (uses a "virtual" `id` field to avoid conflicts) and expects two `ReclaimEscrowEvents` one for `X` shares and one for `Y` shares. Since these events are not receieved the debonding delegations do not get cleaned up from the nexus db.

**Solution:**
Remove the "id" field from `chain.debonding_delegation` and handle such debonding delegations ending in the same epoch by merging them.

---

Related mini-fix:  There is an additional issue where some debonidng delegations with `debond_end=0`  are present in the DB:
https://nexus.stg.oasis.io/v1/consensus/accounts/oasis1qpqz2ut6a6prfcjm64xnpnjhsnqny6jqfyav829v/debonding_delegations_to 
I believe these might be there because in very past (pre-damask) `DebondingStartEscrowEvent` did not include End time (https://github.com/oasisprotocol/oasis-core/pull/4437). I did not confirm this as a fact, but this seems the most likely explanation and the only reasonable i could think off. So as a mini fix we also handle that case in `ConsensusDeleteDebondingDelegations`.

This issue might not be as problematic in practice, because `fast_sync`-ing over old blocks mitigates this.


analyzer/queries/queries.go
```diff
@@ -387,16 +389,14 @@ var (
...
 	ConsensusDeleteDebondingDelegations = `
     DELETE FROM chain.debonding_delegations
-      WHERE id = (
-        SELECT id
-        FROM chain.debonding_delegations
-        WHERE
-          delegator = $1 AND delegatee = $2 AND shares = $3 AND debond_end IN ($4::bigint, $4::bigint - 1)
-        LIMIT 1
-      )`
+      WHERE delegator = $1 AND delegatee = $2 AND shares = $3 AND debond_end IN ($4::bigint, $4::bigint - 1, 0)`
```
ptrus, line 399: We could also do `debond_end <= $4`, but IMO let's keep it as is, so that any additional issues get noticed

ptrus:
Please review when you have some time @Andrew7234 @pro-wh @jberci

storage/migrations/20_consensus_debonding_delegations_id_remove.up.sql
```diff
@@ -0,0 +1,48 @@
...
+DO $$
+DECLARE
+  current_epoch UINT63;
+BEGIN
+  SELECT id INTO current_epoch
+  FROM chain.epochs
+  ORDER BY id DESC
+  LIMIT 1;
+
+  -- Step 3: Insert data from the old table into the new table, merging shares on conflict
```
pro-wh, line 24: other steps unnumbered?

ptrus:
> did anyone in the explorer team say anything about wanting these to stay separate?

No, and I also don't see what the benefit would be. I believe it would be confusing since it doesn't match the consensus state where these are merged.

PR #713, opened 2024-06-21
delegations: do not track rewards as delegations

ptrus:
One more source of chain.delegations with `0` shares.

See relevant oasis-core code: https://github.com/oasisprotocol/oasis-core/blob/774316020d821c99307f444a4de21775fc01f048/go/consensus/cometbft/apps/staking/state/state.go#L996-L1001


Andrew7234 (approved):
Good catch, thanks for the fix!

PR #716, opened 2024-06-24
oasis core v24.0

ptrus:
Vendor v24.0. This release contains only additions (no breaking removals from v23) so we remove v23.0 and use v24.0 now for Eden types.


coreapi/v23.0/staking/api/address.go
```diff
@@ -1,7 +0,0 @@
-package api
-
-import (
-	original "github.com/oasisprotocol/oasis-core/go/staking/api"
-)
-
-type Address = original.Address
```
pro-wh, line 7: hm github doesn't detect this as moved

pro-wh (approved):
thanks. if it compiles

coreapi/v23.0/staking/api/address.go
pro-wh: isn't it in though?

coreapi/v24.0/staking/api/address.go
```diff
@@ -0,0 +1,7 @@
+package api
+
+import (
+        original "github.com/oasisprotocol/oasis-core/go/staking/api"
+)
+
+type Address = original.Address
```
pro-wh, line 7: here

coreapi/v23.0/staking/api/address.go
ptrus: Oh, ok even better :) I didn't double-check.

PR #717, opened 2024-06-25
consensus/accounts: optimize query and include delegated balances

ptrus:
Uses a materialized view to periodically compute the view needed for consensus/accounts endpoint.

Related fix:
- remove unneeded filters for consensus/accounts endpoint which are missing relevant indexes anyway


storage/client/queries/queries.go
```diff
@@ -147,46 +147,62 @@ const (
...
+		filtered_accounts AS (
+			SELECT
+				address,
+				COALESCE(nonce, 0),
+				COALESCE(general_balance, 0),
+				COALESCE(escrow_balance_active, 0),
+				COALESCE(escrow_balance_debonding, 0)
+			FROM chain.accounts
+			WHERE ($1::text IS NULL OR address = $1::text)
+			ORDER BY address -- TODO: maybe order by account balance?
```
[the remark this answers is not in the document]
ptrus: Based on a quick test running the above query for all accounts on mainnet (~300_000 accounts) completes in around 20 seconds. So if we setup a periodic view for this we could compute it every 5/10 minutes or so.

storage/client/queries/queries.go
```diff
@@ -147,46 +147,62 @@ const (
...
-				WHERE delegator = $1::text AND escrow_total_shares_debonding != 0)
-			, 0) AS debonding_delegations_balance
-		FROM chain.accounts
-		WHERE address = $1::text`
+			fa.address,
+			fa.nonce,
+			fa.general_balance,
+			fa.escrow_balance_active,
+			fa.escrow_balance_debonding,
+			COALESCE(ROUND(SUM(ad.shares * ad.escrow_balance_active / NULLIF(ad.escrow_total_shares_active, 0))), 0) AS delegations_balance,
```
pro-wh, line 195: interesting, I also wonder about pushing this aggregation into the delegations subquery

ptrus: Hm yeah i think doing that makes sense

ptrus: Done. This query was now moved into the meterialized view.

storage/migrations/21_accounts_list.up.sql
```diff
@@ -0,0 +1,47 @@
+BEGIN;
+
+CREATE SCHEMA IF NOT EXISTS views;
+GRANT USAGE ON SCHEMA views TO PUBLIC;
+
+CREATE MATERIALIZED VIEW views.accounts_list AS
```
Andrew7234, line 6: 🙏 very neat, looks good

analyzer/consensus/consensus.go
```diff
@@ -36,6 +38,12 @@ import (
 
 const (
 	consensusAnalyzerName = "consensus"
+
+	// The query takes ~20 second to run on mainnet at the time of implementing this.
+	accountListViewRefreshInterval = 2 * time.Minute
```
Andrew7234, line 43: Callout to explorer folks that the `consensus/accounts` endpoint might return slightly stale (<2min) info, but we don't expect this to be an issue since the `consensus/account/{address}` endpoint used for account details view always returns the latest info. @lukaw3d @csillag @lubej

api/spec/v1.yaml
```diff
@@ -513,46 +513,6 @@ paths:
       parameters:
```
Andrew7234, line 513:
```suggestion
            Note that for performance reasons, the info returned by this endpoint may be slightly stale (<2 minutes). For the most up-to-date account state, please query the single-account endpoint.
      parameters:
```

Andrew7234: Feel free to reword; just wanted to update the `summary` line above.

Andrew7234 (commented):
Thanks Peter! Looks good overall; mostly left minor comments.

config/config.go
```diff
@@ -401,6 +401,13 @@ func (r BlockRange) Validate() error {
 	return nil
 }
 
+type ConsensusAnalyzerConfig struct {
+	BlockBasedAnalyzerConfig `koanf:",squash"`
+
+	// AccountListRefresh is the config for the account list refresh analyzer.
+	AccountListRefresh ItemBasedAnalyzerConfig `koanf:"account_list_refresh"`
```
ptrus, line 408: I have "stuffed" this into the Consensus analyzer config instead of making a standalone top-level config for the account list analyzer. Also the analyzer doesn't need to be separately enabled, but is enabled automatically if consensus analyzer is enabled. This makes sense to me since this analyzer is tightly coupled with the consensus analyzer.

Let me know if you agree @Andrew7234

analyzer/consensus/consensus.go
```diff
@@ -36,6 +38,12 @@ import (
...
+	accountsListViewRefreshTimeout = 2 * time.Minute
+	accountListViewRefreshQuery    = `REFRESH MATERIALIZED VIEW CONCURRENTLY views.accounts_list`
+	accountListOpName              = "consensus_account_list_refresh"
```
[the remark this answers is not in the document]
Andrew7234: Looks great, that was a quick refactor 😄 

> hurdle was ensuring the analyzer was run once after the (slow) consensus block-based analyzer 

Ah so we actually ran into this issue before for other item based analyzers, so we made the e2e_regression tests  run analyzers in 2 separate stages denoted in `e2e_config_1.yml` and `e2e_config_2.yml`. The first stage has all the block analyzers, and the second stage runs the item-based analyzers that do subsequent processing of the data ingested by the first stage. The `accounts_list` analyzer probably belongs in the second stage, which should simplify the `computed_height`/queueLength.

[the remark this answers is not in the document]
ptrus: Oh yeah that makes sense. I'll move it into the second stage. Will probably keep the the computed_height - since it might end up being useful for debugging purposes.

Andrew7234 (approved):
Looks great!

PR #719, opened 2024-07-01
storage/client: don't crash on 0 delegated shares

ptrus:
Due to bugs, we can end up with invalid delegations in DB with 0 shares.

I believe most of these cases are fixed in: https://github.com/oasisprotocol/nexus/pull/713 https://github.com/oasisprotocol/nexus/pull/711 https://github.com/oasisprotocol/nexus/pull/712
but still, don't panic since this is part of the public API; instead, log an error and report 0 shares.

### DIFF
--- a/storage/client/client.go
+++ b/storage/client/client.go
@@ -844,10 +844,19 @@ func (c *StorageClient) Account(ctx context.Context, address staking.Address) (*
 }
 
 // Computes shares worth given total shares and total balance.
-func amountFromShares(shares common.BigInt, totalShares common.BigInt, totalBalance common.BigInt) common.BigInt {
+func amountFromShares(shares common.BigInt, totalShares common.BigInt, totalBalance common.BigInt) (common.BigInt, error) {
+	if shares.IsZero() {
+		return common.NewBigInt(0), nil
+	}
+	if totalShares.IsZero() {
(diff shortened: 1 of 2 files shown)


storage/client/client.go
```diff
@@ -844,10 +844,19 @@ func (c *StorageClient) Account(ctx context.Context, address staking.Address) (*
 }
 
 // Computes shares worth given total shares and total balance.
-func amountFromShares(shares common.BigInt, totalShares common.BigInt, totalBalance common.BigInt) common.BigInt {
+func amountFromShares(shares common.BigInt, totalShares common.BigInt, totalBalance common.BigInt) (common.BigInt, error) {
+	if shares.IsZero() {
```
pro-wh, line 848: :information_source: db has some leftover cases where shares=0 and totalShares=0. this prevents us from complaining in that case

(is that right? in my understanding from the PR description and related PRs)

pro-wh (approved):
as described

storage/client/client.go
ptrus: > is that right? in my understanding from the PR description and related PRs

Yes, although I'm not 100% sure if shares=0 and totalShares=0 will still exist after all PRs are deployed and we do a reindex.

ptrus: Will recheck in future

PR #721, opened 2024-07-04
consensus/blocks: various updates

ptrus:
Adds gas_limit, epoch, state_root to consensus/blocks endpoint.


api/spec/v1.yaml
```diff
@@ -1350,6 +1352,18 @@ components:
           format: int32
           description: Number of transactions in the block.
           example: 17
+        gas_limit:
+          allOf: [$ref: '#/components/schemas/TextBigInt']
+          description: The gas limit for the block.
```
Andrew7234, line 1357:
```suggestion
          description: The gas limit for the block. A gas limit of 0 means that the max block gas was unlimited. 
          Blocks from Cobalt and Damask always have a gas limit of 0 because the consensus parameters were static during those periods.
```

tests/e2e_regression/eden/expected/blocks.body
```diff
@@ -1,603 +1,903 @@
 {
   "blocks": [
     {
+      "epoch": 28018,
+      "gas_limit": "0",
```
Andrew7234, line 5: Just checking - is this expected? Is the gas limit set dynamically, but in practice it's not set often?

storage/oasis/nodeapi/damask/node.go
```diff
@@ -63,6 +63,20 @@ func (c *ConsensusApiLite) StateToGenesis(ctx context.Context, height int64) (*g
 	return &rsp, nil
 }
 
+func (c *ConsensusApiLite) GetConsensusParameters(ctx context.Context, height int64) (*nodeapi.ConsensusParameters, error) {
+	/*
+		var rsp consensus.Parameters
+		if err := c.grpcConn.Invoke(ctx, "/oasis-core.Consensus/GetParameters", nil, &rsp); err != nil {
+			return nil, fmt.Errorf("GetParameters(damask): %w", err)
+		}
+	*/
```
Andrew7234, line 72: Can we remove this commented out block? I think the example in `eden/node.go` should suffice, but open to debate. Same applies for `cobalt/node.go`.

Andrew7234 (approved):
Thanks! Looks good overall, left a couple comments

tests/e2e_regression/eden/expected/blocks.body
ptrus: Block gas limit is still 0 for now on mainnet and will be updated soon with an upcoming governance proposal: https://github.com/oasisprotocol/oasis-core/blob/66af60621e63fc250d00bffe7e6f70d5075c9f69/go/upgrade/migrations/consensus_240.go#L59

api/spec/v1.yaml
ptrus: > Blocks from Cobalt, Damask, and early Eden had no gas limit as their sizes were only restricted by byte size until a mid-Eden upgrade introduced a gas limit.

^ this is a bit more factually correct.

PR #724, opened 2024-07-10
consensus/api: add block size_limit

ptrus:
Displaying block gas limit only might be misleading for blocks where it's 0, since it could lead to believe that blocks were unlimited in size. There is also block size limit (in bytes) so include that one in API as well.


Andrew7234 (approved):
Thanks, LGTM!

PR #725, opened 2024-07-10
nodeapi: introduce genesis document struct and include proposal metadata

ptrus:
`v24.x` introduced proposal metadata, which we now support.

Before this PR, the `v22.x` version of `Proposals` was used as the Nexus-internal type for `governance.Proposals`. Now we also need fields added in `v24.x`, so we switch to the `v24.x` struct for `governance.Proposals`. One additional complication was that `governance.Proposals` is also used/needed within the `genesis.Document` struct, which is also using the `v22.x` struct.  To avoid migrating the whole genesis data to `v24.x`, we introduce a Nexus-internal `GenesisDocument` type, which uses the same structs as before, with only `governance.Proposal` updated.

There was also a TODO in code to do this since we currently used a semi-populated genesis document
```
// TODO: Introduce internal, stripped-down version of `genesis.Document`.
```

TODO:
- [x] re-do e2e caches (updated genesis key)


storage/oasis/nodeapi/api.go
```diff
@@ -16,13 +16,12 @@ import (
...
 	consensusTransaction "github.com/oasisprotocol/nexus/coreapi/v22.2.11/consensus/api/transaction"
 	consensusResults "github.com/oasisprotocol/nexus/coreapi/v22.2.11/consensus/api/transaction/results"
-	genesis "github.com/oasisprotocol/nexus/coreapi/v22.2.11/genesis/api"
-	governance "github.com/oasisprotocol/nexus/coreapi/v22.2.11/governance/api"
 	registry "github.com/oasisprotocol/nexus/coreapi/v22.2.11/registry/api"
 	roothash "github.com/oasisprotocol/nexus/coreapi/v22.2.11/roothash/api"
 	"github.com/oasisprotocol/nexus/coreapi/v22.2.11/roothash/api/message"
 	scheduler "github.com/oasisprotocol/nexus/coreapi/v22.2.11/scheduler/api"
 	staking "github.com/oasisprotocol/nexus/coreapi/v22.2.11/staking/api"
+	governance "github.com/oasisprotocol/nexus/coreapi/v24.0/governance/api"
```
Andrew7234, line 24: Mind adding a doc-level comment explaining how these types work? Perhaps something similar to the one in `compat_types.go`:

``` The nodeapi package provides a low-level interface to the Oasis node API.
Due to the (sometimes breaking) changes in the types used by different 
evolutions of oasis-core over the Cobalt, Damask, and Eden upgrades, 
Nexus defines its own set of internal types here. 

The types that it exposes are mostly simplified versions of the types exposed by
oasis-core Damask: The top-level type structs are defined in this file, and
the types of their fields are almost universally directly the types exposed
by oasis-core Damask. The reason is that as oasis-core evolves, Damask types
are mostly able to represent all the information from Cobalt, plus some. However, 
there are a few exceptions where we use oasis-core Eden types instead, e.g. in the 
`governance` module.

When a new type is added, devs should try to convert it to the oasis-core Damask 
type first. If that is not possible, devs can either define a custom internal type or
repackage the latest (Eden) type as the Nexus-internal type. The first option is 
mainly used when we want to pare down a given type to only retain the fields 
Nexus needs.
```

Feel free to reword; that last paragraph may be unnecessary.

Andrew7234 (approved):
Looks good to me! If we start to add more eden types in the future we should consider making the vendored eden types the default (as opposed to damask currently). I think it'd be premature to do it now though.

storage/oasis/nodeapi/api.go
ptrus: Thanks, mainly added what you wrote. I skipped the last paragraph and added a sentence that over time, the types should be migrated to newer versions as needed.

storage/oasis/nodeapi/cobalt/convert.go
```diff
@@ -153,18 +153,24 @@ func ConvertGenesis(d genesisCobalt.Document) *genesis.Document {
...
+		Time:      d.Time,
+		ChainID:   d.ChainID,
+		BaseEpoch: uint64(d.Beacon.Base),
 		Governance: governance.Genesis{
 			Proposals:   proposals,
 			VoteEntries: voteEntries,
 		},
 		Registry: registry.Genesis{
 			Entities:          d.Registry.Entities,
-			Runtimes:          []*registry.Runtime{},
```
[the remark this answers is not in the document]
ptrus: Yeah, luckily, I noticed this since I used the cobalt/convert version as a baseline for the added eden/convert. And since we have e2e regression tests for eden, it caught this error.

PR #727, opened 2024-07-13
consensus accounts first activity

ptrus:


tests/e2e_regression/run.sh
```diff
@@ -102,11 +102,18 @@ for ((i = 0; i < nCases; i++)); do
     curl --silent --show-error --dump-header "$outDir/$name.headers" "$url" >"$outDir/$name.body"
     # Try to pretty-print and normalize (for stable diffs) the output.
     # If `jq` fails, output was probably not JSON; leave it as-is.
+    #
+    # .first_activity is ignored, because its not deterministic between fast-sync and slow-sync.
```
ptrus, line 106: Unfortunately, I had to do this. Because of the height at which we process genesis changes the results of first_activity.

Since first_activity is not the most important field here I prefer this solution to the alternative, which would have separate tests/expected results for `eden` and `edenfast`.

storage/migrations/22_consensus_accounts_first_activity.up.sql
```diff
@@ -0,0 +1,6 @@
+BEGIN;
```
ptrus, line 1: I swapped accounts_list migration and this new one, so that this one comes first. That way we can use `first_activity` in the account_lists without writing one more migration.

This should be fine since account-lists migration was not yet released/deployed anywhere.

storage/migrations/22_consensus_accounts_first_activity.up.sql
```diff
@@ -0,0 +1,6 @@
...
+
+ALTER TABLE chain.accounts
+	ADD COLUMN first_activity TIMESTAMP WITH TIME ZONE;
```
Andrew7234, line 4: Would it work to store the `first_activity_block` instead and `join` it with the blocks table for the timestamp?

ptrus: I almost had it done like that already but it doesnt work nicely with the static accounts. Since static accounts would need to insert fist activity at block height that wont even exist in the indexer db.

Andrew7234: Ah right, ok that makes sense

analyzer/consensus/consensus.go
```diff
@@ -66,27 +67,29 @@ func OpenSignedTxNoVerify(signedTx *transaction.SignedTransaction) (*transaction
...
+func NewAnalyzer(chainName common.ChainName, blockRange config.BlockRange, batchSize uint64, mode analyzer.BlockAnalysisMode, history config.History, source nodeapi.ConsensusApiLite, network sdkConfig.Network, target storage.TargetStorage, logger *log.Logger) (analyzer.Analyzer, error) {
 	processor := &processor{
-		mode:    mode,
-		history: history,
-		source:  source,
-		network: network,
-		target:  target,
-		logger:  logger.With("analyzer", consensusAnalyzerName),
-		metrics: metrics.NewDefaultAnalysisMetrics(consensusAnalyzerName),
+		chainName: chainName,
```
pro-wh, line 85: take this from a field in history instead of pulling chainName from the config. this is so that we can unset chainName and use a custom history if we need to adjust something (i.e. something unrelated to the static accounts set) at runtime.

having a "mainnet" and "testnet" known static account sets is fine, and the new history field can take on the values "mainnet" and "testnet."

storage/migrations/22_consensus_accounts_first_activity.up.sql
pro-wh: has accounts_list never been deployed?

tests/e2e_regression/run.sh
pro-wh: why does it end up different?

ptrus: Some addresses are only visible when fetching state-to-genesis and processing it. Therefore, the earliest first_activity timestamp we have for those addresses is when this process occurs. This is because we do not process the entire chain. Otherwise, each address would either be among the static list of addresses or be seen in a transaction.

Specifically, for 'eden' and 'edenfast', the state-to-genesis occurs at different heights - since we only perform it at the beginning of slow-sync.

storage/migrations/22_consensus_accounts_first_activity.up.sql
ptrus: I believe so yes, it has been merged only last week.

tests/e2e_regression/run.sh
pro-wh: does this mean real genesis plus events gives a different set of accounts from state-to-genesis?

storage/migrations/22_consensus_accounts_first_activity.up.sql
pro-wh: ohh if it's never been in a released version that's even better

tests/e2e_regression/run.sh
ptrus: > does this mean real genesis plus events gives a different set of accounts from state-to-genesis?

I dont follow 100%. I dont think we fetch and process the real genesis in eden/edenfast tests, no? 

We only do StateToGenesis. In eden test, we do it at the first specified height; and in edenfast we do it at the first slow-sync height.

pro-wh: oh it's different due to using `document.Time.UTC()`

thanks

analyzer/consensus/consensus.go
ptrus: updated, check again please.

pro-wh: lg thanks

Andrew7234 (approved):

csillag:
Clarifying question: Is it possible that when checking an account (using `/v1/<runtime>/accounts/<address>`), in the results, `first_activity` is missing, but some other value (like `available`, `debonding_delegations_balance`, `delegations_balance`, `escrow`, `total`, `nonce` or `stats.num_txns`) is still non-zero?

PR #728, opened 2024-07-18
Andrew7234/consolidate migrations

Andrew7234:
Resolves https://app.clickup.com/t/86953vvm2

Consolidates our rather long list of migrations from ~32 -> 5. Although this PR is intended to be deployed on a fresh database, operators can redeploy with this version and then manually update their existing deployments by going into the db and running `update schema_migrations set version = 5`. This will allow future migrations (which will now start at `06_...` to properly apply to the database. 

Functionally, this PR is mostly a no-op, with two exceptions: 
- `chain.blocks.metadata` was removed because it was not being used. It was likely an artifact from years ago.
- There was a missing permission for the new `history.staking_events` table; it has been granted properly as of this PR.

Validation:
Comparison of the dumped database schema (using `pg_dump -s`) for main branch / feature branch:
```
> diff main.sql new.sql

Explanation: chain.blocks.metadata was removed; gas_limit, size_limit, and epoch were reordered.
354a355,357
>     gas_limit public.uint_numeric DEFAULT 0 NOT NULL,
>     size_limit public.uint_numeric DEFAULT 0 NOT NULL,
>     epoch public.uint63 DEFAULT 0 NOT NULL,
358,361d360
<     metadata jsonb,
<     gas_limit public.uint_numeric DEFAULT 0 NOT NULL,
<     epoch public.uint63 DEFAULT 0 NOT NULL,
<     size_limit public.uint_numeric DEFAULT 0 NOT NULL,

Explanation: reordering of columns
638a638,639
>     title text,
>     description text,
649,651c650
<     invalid_votes public.uint_numeric DEFAULT 0 NOT NULL,
<     title text,
<     description text
---
>     invalid_votes public.uint_numeric DEFAULT 0 NOT NULL
806a806,808
>     fee_symbol text DEFAULT ''::text NOT NULL,
>     fee_proxy_module text,
>     fee_proxy_id bytea,
813a816
>     amount_symbol text,
821a825,830
>     oasis_encrypted_format public.call_format,
>     oasis_encrypted_public_key bytea,
>     oasis_encrypted_data_nonce bytea,
>     oasis_encrypted_data_data bytea,
>     oasis_encrypted_result_nonce bytea,
>     oasis_encrypted_result_data bytea,
829,838d837
<     amount_symbol text,
<     fee_symbol text DEFAULT ''::text NOT NULL,
<     fee_proxy_module text,
<     fee_proxy_id bytea,
<     oasis_encrypted_format public.call_format,
[...]

PR #732, opened 2024-07-24
Andrew7234/validator staking api

Andrew7234:
Resolves https://app.clickup.com/t/8694zgh42

Adds the corresponding api for https://github.com/oasisprotocol/nexus/pull/728. 

Also cleans up some inconsistencies in the spec.

### DIFF
--- a/api/spec/v1.yaml
+++ b/api/spec/v1.yaml
@@ -92,8 +92,8 @@ x-examples:
   parameters-change:
     - &parameters_change_1 '{"min_validators":null,"max_validators":"120","voting_power_distribution":null}'
   epoch:
-    - &epoch_1 8048956
-    - &epoch_2 8048966
+    - &epoch_1 13402
+    - &epoch_2 13403
   event-type:
     - &event_type_1 'staking.escrow.take'
   roothash-message-type:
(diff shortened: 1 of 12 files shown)


api/spec/v1.yaml
```diff
@@ -1933,6 +1970,61 @@ components:
...
           description: The shares of tokens this validator has delegated to itself, and are NOT in the process of debonding.
+        active_balance_24:
+          allOf: [$ref: '#/components/schemas/TextBigInt']
+          description: The active_balance of this validator account 24 hours ago.
+        num_delegators:
+          type: integer
+          format: uint64
+          description: The number of accounts that have delegated token to this account.
+    
+    EscrowHistory:
```
Andrew7234, line 1981: Was also considering naming this `ValidatorHistory` and `ValidatorHistoryPoint` below; open to suggestions

api/spec/v1.yaml
```diff
@@ -1933,6 +1970,61 @@ components:
         self_delegation_shares:
           allOf: [$ref: '#/components/schemas/TextBigInt']
           description: The shares of tokens this validator has delegated to itself, and are NOT in the process of debonding.
+        active_balance_24:
```
Andrew7234, line 1973: Note: This attempts to find the balance 24 epochs ago, which isn't exactly 24 hours but it's accurate enough. Also open to field name suggestions here.

Issue #734, opened 2024-08-09
Transaction fetch fails for account on Sapphire Testnet


kostko:
##### SUMMARY

Requesting transactions for account `oasis1qp66ryj9caek77kewkxxvjvkzypljhsdgvm5q34d` on Sapphire Testnet returns an error.

##### ISSUE TYPE
- Bug Report

##### STEPS TO REPRODUCE

Fetch URL:

https://testnet.nexus.oasis.io/v1/sapphire/transactions?limit=10&offset=0&rel=oasis1qp66ryj9caek77kewkxxvjvkzypljhsdgvm5q34d

##### ACTUAL RESULTS

```json
{"msg":"storage error: can't scan into dest[5]: cannot scan NULL into *string"}
```

##### EXPECTED RESULTS

Actual relevant transactions are returned.

kostko:
It is likely that the recently introduced ROFL-related transactions are the problem. They look like the following (decoding via Oasis CLI):

```
=== Transaction 0 ===
Kind: oasis
Hash: 5b0829b2e9ff2337b31045fbe61cc3c107bff891b2e8555a8adddf174c083556
Signer(s):
  1. A80H67H9SkwA+gjSUVHObr6zUWC5qKwzUpccMfLODlHQ
     (signature: MEUCIQCuoX1SLvNHFVxB8qLtNGsS0uPvVSmDKNEK/LIlwullSgIgZ0yc5c/OktaVLRYNmNDCnEa7pxB0Ff+UY1BKVcJlk3I=)
Content:
  Format: plain
  Method: evm.Call
  Body:
    {
      "address": "2mVLuIfSkTyU1ECc/bAoZacGyTA=",
      "value": "AAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAA=",
      "data": "2uHuHwAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAq"
    }
  Authorized signer(s):
    1. A80H67H9SkwA+gjSUVHObr6zUWC5qKwzUpccMfLODlHQ (secp256k1eth)
       Nonce: 25
  Fee:
    Amount: 0.0001706 TEST
    Gas limit: 1706
    (gas price: 0.0000001 TEST per gas unit)
    Fee proxy:
      Module: rofl
      ID:     00694cb01f85408d624ea267f657bf285787a61db3

=== Result of transaction 0 ===
Status:  failed
Module:  evm
[...]

PR #735, opened 2024-08-09
Andrew7234/rofl tx support

Andrew7234:
Resolves https://github.com/oasisprotocol/nexus/issues/734

The root cause of the issue was that oasis-sdk was outdated and Nexus threw an error when trying to cbor-decode the transaction to the old sdk `Transaction` type. The first commit in this PR bumps the sdk version to the latest @ `v0.10.1`.

This PR also adds support for tracking rofl fee proxy information in runtime transactions.


pro-wh (approved):
new fields

PR #738, opened 2024-08-12
api: add validator agg stats

Andrew7234:
Moves `voting_power_total` to a sibling struct (ValidatorsStats) in ValidatorList. Also adds the total number of delegators and the total staked amount in the network to the new ValidatorsStats.

The total number of delegators/staked amount is required for the dashboard view in the mocks.

### DIFF
--- a/storage/client/queries/queries.go
+++ b/storage/client/queries/queries.go
@@ -278,10 +278,26 @@ const (
 		LIMIT $2::bigint
 		OFFSET $3::bigint`
 
-	Validators = `
+	LatestEpochStart = `
 		SELECT id, start_height
 			FROM chain.epochs
 			ORDER BY id DESC`
+
+	ValidatorsAggStats = `
+		SELECT
+			COALESCE (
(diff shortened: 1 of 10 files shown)


storage/client/queries/queries.go
```diff
@@ -278,11 +278,31 @@ const (
...
+			SELECT delegator FROM chain.delegations
+			GROUP BY delegator
+		)
+		SELECT
+			COALESCE (
+				(SELECT SUM(voting_power)
+				FROM chain.nodes)
+			, 0) AS total_voting_power,
+			COALESCE (
+				(SELECT COUNT(*)
```
ptrus, line 297: Could use `SELECT COUNT(distinct delegator) from chain.delegations`?

It might be a bit clearer than using a CTE?

ptrus (approved):

storage/client/queries/queries.go
Andrew7234: good idea; updated!

PR #746, opened 2024-08-30
storage: nil out transaction .Error on unknown result

### LINKED ISSUE #704: runtime tx unexpected fields
cc @matevz @csillag 

https://explorer.oasis.io/mainnet/sapphire/tx/7224d59e6b74d9b2caf55ab40dfe0757cef4f71c66d3a7aba3de1d267d2ea409

this tx has `error` set while the status is unknown. api should clear out the error

also its method is blank, which isn't right. allegedly it's supposed to be accounts.Transfer

pro-wh:
Unknown -> .Success = nil -> .Error = nil
Ok -> .Success = common.Ptr(true) -> .Error = nil
Failed -> .Success = common.Ptr(false) -> .Error not nil

is my understanding for today

fixes #704, unexpeced non-nil .Error

### DIFF
--- a/storage/client/client.go
+++ b/storage/client/client.go
@@ -1531,7 +1531,11 @@ func (c *StorageClient) RuntimeTransactions(ctx context.Context, p apiTypes.GetR
 		); err != nil {
 			return nil, wrapError(err)
 		}
-		if t.Success != nil && *t.Success {
+		// If success field is unset (i.e. encrypted "Unknown" result) or
+		// successful, some database versions have non-null error module/code
+		// from when the analyzer would insert ""/0 instead. There's no error
+		// information, so empty this stuff out.
+		if t.Success == nil || *t.Success {
 			t.Error = nil
 		}
 		if encryptionEnvelopeFormat != nil { // a rudimentary check to determine if the tx was encrypted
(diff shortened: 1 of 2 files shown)


storage/client/client.go
```diff
@@ -1531,7 +1531,7 @@ func (c *StorageClient) RuntimeTransactions(ctx context.Context, p apiTypes.GetR
 		); err != nil {
 			return nil, wrapError(err)
 		}
-		if t.Success != nil && *t.Success {
+		if t.Success == nil || *t.Success {
```
Andrew7234, line 1534:
```suggestion
               // If the transaction was encrypted or was successful, no error information is available.
		if t.Success == nil || *t.Success {
```

Andrew7234 (approved):
Thanks! Mind also updating the [runtime analyzer](https://github.com/oasisprotocol/nexus/blob/main/analyzer/runtime/runtime.go#L302) to use pointers for `error_code` and `error_module` as well?

pro-wh:
k let's update analyzer too

pro-wh:
separate PR since this is already reviewed tho

storage/client/client.go
pro-wh: added

PR #747, opened 2024-08-30
Oasis-style encryption envelope

pro-wh:
cross reference #407, getting the EVM-style encryption envelope

cross reference #704, confusion about why method is not known

sample output for the tx in #704, https://explorer.oasis.io/mainnet/sapphire/tx/7224d59e6b74d9b2caf55ab40dfe0757cef4f71c66d3a7aba3de1d267d2ea409

```json
{
  "total_count": 1,
  "transactions": [
    {
      "amount": "0",
      "charged_fee": "1238900000000000",
      "fee": "1238900000000000",
      "fee_symbol": "ROSE",
      "gas_limit": 12389,
      "gas_used": 12389,
      "hash": "7224d59e6b74d9b2caf55ab40dfe0757cef4f71c66d3a7aba3de1d267d2ea409",
      "index": 4,
      "method": "",
      "nonce_0": 27,
      "oasis_encryption_envelope": {
        "data": "M+EtPEsK0Jx6YBWovQqDc8Vfmwy86L0EE66kv1++2RYxbXYj2zQsPmHPJLhC1VELHKVJQ/WKF+lTYhV3XhPdjTl1CiY8HiY7BcbL+t4nHB77mrGAeDBZV3nAJt4=",
        "data_nonce": "p8fnoq24sBge89XBVnvG",
        "format": "encrypted/x25519-deoxysii",
        "public_key": "ZDSfM+5QCZez+7fEjVOGmY2UlF82Q/KKR7Void3UHUQ=",
        "result": "noRd8cBSdvefZ7LS+OtovFFkOKRz",
        "result_nonce": "AAAAAAA89jEAAAAEAAAA"
      },
      "round": 3995186,
      "sender_0": "oasis1qqugfeelqtp3qaxhdrwfv9nhpvwckjxr7yk369p7",
      "sender_0_eth": "0x4f1DAE99883E44c924Db37e4B537Da87A7459029",
      "size": 389,
      "timestamp": "2024-06-03T02:21:26-07:00"
    }
  ]
}
```


analyzer/runtime/oasis_encryption.go
```diff
@@ -0,0 +1,65 @@
...
+)
+
+type OasisEncryptedData struct {
+	Format      common.CallFormat
+	PublicKey   []byte
+	DataNonce   []byte
+	DataData    []byte
+	ResultNonce []byte
+	ResultData  []byte
+}
```
Andrew7234, line 19: The evmEncrypted and oasisEncrypted tx data share almost all of the same fields; would it make more sense to factor this into something like `common.EncryptedTxData` and use it in both cases?

storage/migrations/31_runtime_txs_oasis_encryption.up.sql
```diff
@@ -0,0 +1,12 @@
+BEGIN;
+
+ALTER TABLE chain.runtime_transactions
+    -- Encrypted data in encrypted Oasis-format transactions.
+    ADD COLUMN oasis_encrypted_format call_format,
+    ADD COLUMN oasis_encrypted_public_key BYTEA,
+    ADD COLUMN oasis_encrypted_data_nonce BYTEA,
+    ADD COLUMN oasis_encrypted_data_data BYTEA,
+    ADD COLUMN oasis_encrypted_result_nonce BYTEA,
+    ADD COLUMN oasis_encrypted_result_data BYTEA;
```
Andrew7234, line 10: Would it make sense to rename the existing `evm_encrypted_*` fields to drop the `evm_` prefix and reuse those fields for the encrypted tx data here? And then have some flag to indicate whether the encrypted tx was on the sdk level or in evm?

Andrew7234 (commented):
Thank you for looking into this!

analyzer/runtime/oasis_encryption.go
pro-wh: oh yeah we can make a struct just for envelope stuff. but I'm in favor of storing these separately. gonna leave a comment on why

storage/migrations/31_runtime_txs_oasis_encryption.up.sql
pro-wh: what's the benefit, narrower table? dunno if that's worth the effort to deal with the extra logic to combine them when inserting and separate them when selecting

pro-wh:
the logic for handling these uses these two sets of encrypted stuff differently, so it's fine and/or convenient to have these be separate tx properties. the logic is roughly this:

distinctions, btw:
- tx Call vs evm data Call
- outer Call vs inner Call
- (and similarly for CallResult)

1. have outer tx Call
2. if outer tx Call format is encrypted
   i. parse encryption envelope <-- oasis_encrypted_*
   ii. display encryption envelope
   iii. if we can decrypt inner tx Call (i.e. the runtime does this with the runtime's key, explorer will maybe let a user provide the cleartext)
      a. parse inner tx Call
   iv. otherwise we have no further info, exit
3. otherwise outer tx Call is unencrypted, and inner tx Call is the same structure
4. have inner tx Call
5. display method name
6. if it's an evm method
   i. display to address (for calls), value, gas, etc.
   ii. if data parses as outer evm data Call:
      a. parse encryption envelope <-- evm_encrypted_*
      b. display encryption envelope
      c. if we can decrypt inner evm data Call
         (1) parse inner evm data Call
      d. otherwise we have no further info, exit
      e. have inner evm data Call
      f. extract data from inner evm data Call
   iii. otherwise data is the actual evm data
   iv. have evm data
   v. (further parsing or display of evm data)

storage/migrations/31_runtime_txs_oasis_encryption.up.sql
Andrew7234: Makes sense, let's go with the current form then 👍

Andrew7234 (approved):
Thanks for the explanation! Lgtm

pro-wh:
unified the EncryptedData structure

PR #748, opened 2024-09-03
runtime: store NULL error module/code when no error

pro-wh:
in fact the database allows NULL in these columns already, but a part of the analyzer fills in ""/0 

- allow inserting NULL
- no corrections to existing records
- rename variables to follow variable naming style
- no change to api.....................

### DIFF
--- a/.changelog/748.feature.md
+++ b/.changelog/748.feature.md
@@ -1,0 +1,1 @@
+runtime: store NULL error module/code when no error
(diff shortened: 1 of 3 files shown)


pro-wh:
oop doesn't even build

pro-wh:
ok I am informed by CI that consensus also uses the TxError struct. additionally,  in consensus nexus land, it really is code NOT NULL and message NULL. with a special code = errors.CodeNoError = 0 for success :thinking:

pro-wh:
welp let's undo most of these changes lol

Andrew7234 (approved):

pro-wh:
huh did this no affect any of those e2e regression db dumps?

pro-wh:
indeed, we don't dump those columns

PR #749, opened 2024-09-04
consensus/api: Handle duplicated consensus transaction hashes

ptrus:


storage/client/client.go
```diff
@@ -53,7 +53,6 @@ type StorageClient struct {
 	networkConfig  *oasisConfig.Network
 
 	blockCache *ristretto.Cache
```
ptrus, line 55: txCache is removed because the current trivial implementation is wrong due to the possibility of duplicate transaction hashes.

Also, this has been implemented very long ago, and maybe a bit prematurely, since, e.g., for runtime txes, we don't have any such caches.

I propose to drop it at this point, and we can revisit in the future if this query ends up problematic.

Andrew7234 (approved):
Thanks! Looks good

PR #755, opened 2024-09-27
 storage: sort tokens by total value 

pro-wh:
we do some logic similar to storage/client.fillInPrice

### DIFF
--- a/storage/client/queries/queries.go
+++ b/storage/client/queries/queries.go
@@ -680,8 +680,8 @@ const (
 			ref_swap_pairs.runtime = tokens.runtime AND
 			ref_swap_pairs.pair_address = ref_swap_pair_creations.pair_address
 		LEFT JOIN chain.evm_tokens AS ref_tokens ON
-		    ref_tokens.runtime = tokens.runtime AND
-		    ref_tokens.token_address = $5
+			ref_tokens.runtime = tokens.runtime AND
+			ref_tokens.token_address = $5
 		LEFT JOIN chain.evm_contracts as contracts ON (tokens.runtime = contracts.runtime AND tokens.token_address = contracts.contract_address)
 		WHERE
 			(tokens.runtime = $1) AND
--- a/storage/client/queries/queries.go
+++ b/storage/client/queries/queries.go
@@ -689,7 +689,36 @@ const (
 			($3::text IS NULL OR tokens.token_name ILIKE '%' || $3 || '%' OR tokens.symbol ILIKE '%' || $3 || '%') AND
 			tokens.token_type IS NOT NULL AND -- exclude token _candidates_ that we haven't inspected yet
 			tokens.token_type != 0 -- exclude unknown-type tokens; they're often just contracts that emitted Transfer events but don't expose the token ticker, name, balance etc.
-		ORDER BY num_holders DESC, contract_addr
+		ORDER BY
+			(
+				CASE
+					-- For the reference token itself, it is 1:1 in value with, you know, itself.
+					WHEN
+						tokens.token_address = $5
+					THEN 1.0
+					-- The pool keeps a proportion of reserves so that reserve0 of token0 is worth about as much as reserve1 of token1.
+					-- When token0 is the reference token, more reserve0 means token1 is worth more than the reference token.
+					WHEN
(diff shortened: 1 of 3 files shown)


storage/client/queries/queries.go
```diff
@@ -680,16 +680,21 @@ const (
...
 			($2::oasis_addr IS NULL OR tokens.token_address = $2::oasis_addr) AND
 			($3::text IS NULL OR tokens.token_name ILIKE '%' || $3 || '%' OR tokens.symbol ILIKE '%' || $3 || '%') AND
 			tokens.token_type IS NOT NULL AND -- exclude token _candidates_ that we haven't inspected yet
 			tokens.token_type != 0 -- exclude unknown-type tokens; they're often just contracts that emitted Transfer events but don't expose the token ticker, name, balance etc.
-		ORDER BY num_holders DESC, contract_addr
+		ORDER BY CASE
+			WHEN tokens.token_address = $4 THEN 1.0
+			WHEN ref_swap_pair_creations.token0_address = $4 AND ref_swap_pairs.reserve0 IS NOT NULL THEN ref_swap_pairs.reserve0::REAL / ref_swap_pairs.reserve1::REAL
+			WHEN ref_swap_pair_creations.token1_address = $4 AND ref_swap_pairs.reserve1 IS NOT NULL THEN ref_swap_pairs.reserve1::REAL / ref_swap_pairs.reserve0::REAL
+			ELSE 0.0
```
Andrew7234, line 696: Could you add a comment here explaining the sort order/calculation? If I'm reading it right, this sorts by the highest token price, not market cap? If that's the case I'd prefer to keep num_holders as the primary sort column since price could be manipulated.

pro-wh: :+1: on adding comment. it is meant to be some cap though, it's multiplied by total supply lower. probably better to put these in named SELECT things. could also reduce some logic on the go side

pro-wh: adding the comment

Andrew7234: oh ok nice, makes sense to me. thanks for the comment

storage/client/queries/queries.go
```diff
@@ -680,16 +680,25 @@ const (
...
 			tokens.token_type != 0 -- exclude unknown-type tokens; they're often just contracts that emitted Transfer events but don't expose the token ticker, name, balance etc.
-		ORDER BY num_holders DESC, contract_addr
+		ORDER BY CASE
+			-- For the reference token itself, it is 1:1 in value with, you know, itself.
+			WHEN tokens.token_address = $4 THEN 1.0
+			-- The pool keeps a proportion of reserves so that reserve0 of token0 is worth about as much as reserve1 of token1.
+			-- When token0 is the reference token, more reserve0 means token1 is worth more than the reference token.
+			WHEN ref_swap_pair_creations.token0_address = $4 AND ref_swap_pairs.reserve0 IS NOT NULL THEN ref_swap_pairs.reserve0::REAL / ref_swap_pairs.reserve1::REAL
+			-- When token1 is the reference token, more reserve1 means token0 is worth more than the reference token.
+			WHEN ref_swap_pair_creations.token1_address = $4 AND ref_swap_pairs.reserve1 IS NOT NULL THEN ref_swap_pairs.reserve1::REAL / ref_swap_pairs.reserve0::REAL
```
Andrew7234, line 699: just confirming - we don't need to check if `reserve0` is not null here?

Andrew7234 (approved):
looks good!

storage/client/queries/queries.go
pro-wh: let's add those

pro-wh: :facepalm:  this whole time the query was wrong, $4 is the factory address, we want $5, which is the reference token address

pro-wh: gonna change the indentation a little to make it clearer about the times-total-supply at the bottom

oh should also check about reserves not being zero >_>

pro-wh: new finding: lint fixes from last token price PR actually broke the api. gotta fix that now

pro-wh: ok so yeah, added a lot more checks, including reserve0 is not null

PR #759, opened 2024-10-01
storage: bump consensus RegistryEvents cache key

pro-wh:
reindex encountered a legacy `RegistryRuntimeRegistered` at height 3027847


Andrew7234 (approved):
Thanks!

pro-wh:
oh this needs new e2e data

PR #761, opened 2024-10-02
Post-reindex hotfixes

Andrew7234:
This branch has been pushed to dockerhub with the image tag `v0.4.0-andy`, and is currently running in production testnet/mainnet. ~~It is a live branch and will be updated as necessary.~~ Ready for review


storage/migrations/00_consensus.up.sql
```diff
@@ -375,7 +375,7 @@ CREATE TABLE history.escrow_events
   epoch UINT63 NOT NULL,
   type TEXT NOT NULL,
   delegatee oasis_addr NOT NULL,
-  delegator oasis_addr NOT NULL,
+  delegator oasis_addr NOT NULL, -- NULL in 06_escrow_history_delegator.up.sql
```
pro-wh, line 378: what's the case where it's null?

pro-wh (approved):
ok

storage/migrations/00_consensus.up.sql
Andrew7234: slashing events

pro-wh: oh ok

PR #763, opened 2024-10-04
storage: omitempty nodeapi.Event alternates

pro-wh:
nodeapi.Event, used in many consensus source storage APIs, get serialized to cache with all these fields having explicit nulls. that increases the sensitivity to field changes. for example, many events cached from the cobalt upgrade had `RegistryRuntimeRegistered` = null, which causes the deserialization to fail, on events that aren't even related to the registry.

### DIFF
--- a/storage/oasis/nodeapi/api.go
+++ b/storage/oasis/nodeapi/api.go
@@ -131,29 +131,29 @@ type Event struct {
 	// Called "Kind" in oasis-core but "Type" in Nexus APIs and DBs.
 	Type apiTypes.ConsensusEventType
 
-	StakingTransfer             *TransferEvent
-	StakingBurn                 *BurnEvent
-	StakingAddEscrow            *AddEscrowEvent
-	StakingTakeEscrow           *TakeEscrowEvent
-	StakingEscrowDebondingStart *DebondingStartEscrowEvent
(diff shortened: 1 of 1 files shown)


storage/oasis/nodeapi/api.go
```diff
@@ -131,29 +131,29 @@ type Event struct {
...
-
-	RoothashMisc              *RoothashEvent
-	RoothashExecutorCommitted *ExecutorCommittedEvent
-	RoothashMessage           *MessageEvent // Available only in Cobalt.
-
-	GovernanceProposalSubmitted *ProposalSubmittedEvent
-	GovernanceProposalExecuted  *ProposalExecutedEvent
-	GovernanceProposalFinalized *ProposalFinalizedEvent
-	GovernanceVote              *VoteEvent
+	StakingTransfer             *TransferEvent             `json:",omitempty"`
```
ptrus, line 134: It looks like we removed a lot of these in https://github.com/oasisprotocol/nexus/commit/02df003e352aaa58b5faabf6c4a6d9394e7107b9 (by mistake?).

pro-wh: @Andrew7234 how did those json annotations come up in the api tests PR?

ptrus: Good question

PR #766, opened 2024-10-14
config/server: clean-up chainName and reuse source config

ptrus:
The server config had a (legacy?) `chain_name` field, which was only used in one place - to determine the runtime ID from the runtime name in the `API/runtime/status` endpoint. This was a bit of a kludge and didn't support custom networks.

The server config already contains the `SourceConfig` section, which supports both Default and Custom networks, so let's use that to determine the runtime ID.

Note needed change to configs is the following:

Old:
```yaml
server:
  chain_name: mainnet
```

New:
```yaml
server:
  source:
    chain_name: mainnet
```

If the deployment already had the `server.source` configured, then just removing the existing `server.chain_name` is ok.

### DIFF
--- a/.changelog/766.feature.md
+++ b/.changelog/766.feature.md
@@ -1,0 +1,18 @@
+config/server: clean-up chainName and reuse source config
+
+Required config change:
+
+Old:
+
+```yaml
+server:
+  chain_name: mainnet
+```
+
+New:
+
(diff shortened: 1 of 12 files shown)


pro-wh (approved):
ya I think this works. server.source becomes required, server.chain_name removed

PR #767, opened 2024-10-16
Add validators addresses to named accounts

buberdds:
We have a few endpoints that return only validator address (sample https://testnet.nexus.oasis.io/v1/consensus/accounts/oasis1qzgn6ygt2kpmhe0gappadrsj7l9u3xcxkyp3qxt6/delegations?limit=5&offset=0) without `ValidatorMedia`. In the UI, we should display the name wherever possible. To achieve this, we need to extend named accounts to include validator addresses. This approach allows us to show names without needing to add `ValidatorMedia` to multiple API responses

dump from 
https://nexus.oasis.io/v1/consensus/validators
https://testnet.nexus.oasis.io/v1/consensus/validators

### DIFF
--- a/.changelog/767.trivial.md
+++ b/.changelog/767.trivial.md
@@ -1,0 +1,1 @@
+Add validators addresses to named accounts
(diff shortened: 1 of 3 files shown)


buberdds:
will discuss on Slack how we want to handle this

PR #768, opened 2024-10-17
update evmverifier to use current sourcify api

Andrew7234:
The [sourcify endpoint](https://sourcify.dev/server/api-docs/#/Repository/get_files_contracts__chain_) we were using is now deprecated in favor of several paginated endpoints. We now first fetch the [fully verified contracts](https://sourcify.dev/server/api-docs/#/Repository/get_files_contracts_full__chain_) and then[ all the contracts](https://sourcify.dev/server/api-docs/#/Repository/get_files_contracts_any__chain_) and take the diff to find the partially verified ones.

```
jsonPayload: {
analyzer: "evm_contract_verifier_sapphire"
caller: "item.go:215"
err: "error fetching work items: failed to get verified contract addresses: failed to fetch verified contract addresses: sourcify API call failure: 404 Not Found (https://sourcify.dev/server/files/contracts/23295)"
level: "error"
module: "analysis_service"
msg: "error processing batch"
ts: "2024-10-17T17:32:53.129850035Z"
}
```

todo: 

- [x] update sourcify client tests


analyzer/evmverifier/sourcify/client.go
```diff
@@ -86,35 +86,78 @@ func (s *SourcifyClient) callAPI(ctx context.Context, method string, url *url.UR
...
+	return addresses, nil
+}
+
+func (s *SourcifyClient) GetAllAddressPages(ctx context.Context, u *url.URL) ([]ethCommon.Address, error) {
+	page := 0
+	addresses := []ethCommon.Address{}
+	for {
+		// https://sourcify.dev/server/api-docs/#/Repository/get_files_contracts_any__chain_
+		q := u.Query()
+		q.Set("page", string(page))
```
pro-wh, line 125: lint says this doesn't give the string of the number :skull:

analyzer/evmverifier/sourcify/client.go
```diff
@@ -86,35 +86,78 @@ func (s *SourcifyClient) callAPI(ctx context.Context, method string, url *url.UR
...
+		u.RawQuery = q.Encode()
+
+		body, err := s.callAPI(ctx, http.MethodGet, u)
+		if err != nil {
+			return nil, fmt.Errorf("failed to fetch fully verified contract addresses: %w (%s)", err, u.String())
```
pro-wh, line 130: imo put the url before the %w

pro-wh (approved):
ok this is a really simple way to do it

problems:

1. they have to list fully verified contracts twice, one in /full, one in /any -> that's just how the API is
2. if contracts get added while we're paginating, we'll miss it -> this analyzer will run again and those contracts will have another chance
3. are we really going to scan every single page, every single time period? -> yeah, as the API has no way to look for new verifications. plus, the current list is only one page
4. a single failure of a page will cause us to fail the entire list -> this could be improved, but the current list is only one page
5. we read the whole list into memory only to scan through linearly -> at least it's simple this way, and the current list is only one page

analyzer/evmverifier/sourcify/client.go
Andrew7234: ya i guess it's more like `chr()`; updated to use `fmt.Sprintf` tho

analyzer/evmverifier/sourcify/client.go
Andrew7234: updated!

pro-wh:
oh do we need to update some cache for the e2e regression tests?

Andrew7234:
yeah i didn't realize we were caching sourcify too

cache/httpproxy/caching_http_proxy.go
```diff
@@ -37,15 +37,16 @@ type CacheableResponse struct {
...
 func (p cachingHttpProxy) ServeHTTP(w http.ResponseWriter, req *http.Request) {
 	// Rewrite the request URL to new host
-	targetURL, _ := url.JoinPath(p.targetHost, req.URL.RequestURI())
-	req, err := http.NewRequestWithContext(req.Context(), req.Method, targetURL, req.Body)
+	targetURL, _ := url.JoinPath(p.targetHost, req.URL.Path)
+	proxyReq, err := http.NewRequestWithContext(req.Context(), req.Method, targetURL, req.Body)
 	if err != nil {
 		panic(fmt.Sprintf("error rewriting request to %s %s: %v", req.Method, req.URL.String(), err))
 	}
+	proxyReq.URL.RawQuery = req.URL.Query().Encode()
```
pro-wh, line 45: what is this, we update the url after making the request?

cache/httpproxy/caching_http_proxy.go
```diff
@@ -37,15 +37,16 @@ type CacheableResponse struct {
 
 func (p cachingHttpProxy) ServeHTTP(w http.ResponseWriter, req *http.Request) {
 	// Rewrite the request URL to new host
-	targetURL, _ := url.JoinPath(p.targetHost, req.URL.RequestURI())
```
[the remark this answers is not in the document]
Andrew7234: `req.URL.RequestURI()` returns the escaped url which means the query is displayed as `/files/contracts/any/42262%3Fpage=0`, which when used as input to `url.JoinPath` causes the query param to be melded into the chain and then sourcify complains that there's no chain called `42262?page=0`

cache/httpproxy/caching_http_proxy.go
Andrew7234: It just adds the query params properly to the new proxied request. The proxy request is then queried below in `GetFromCacheOrCall`

tests/e2e_regression/damask/expected/emerald_contract_account.body
```diff
@@ -783,45 +783,45 @@
...
-          "name": "IERC20.sol",
-          "path": "/home/app/repository/contracts/partial_match/42262/0xa6E864266b6707a6eA9529D83d8e0499E660f566/sources/contracts/uniswapv2/interfaces/IERC20.sol"
+          "path": "contracts/partial_match/42262/0xa6E864266b6707a6eA9529D83d8e0499E660f566/sources/contracts/uniswapv2/UniswapV2Pair.sol"
```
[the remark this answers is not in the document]
Andrew7234: Seems like sourcify updated the file schema

cache/httpproxy/caching_http_proxy.go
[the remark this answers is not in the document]
pro-wh: weird, https://pkg.go.dev/net/url#URL.RequestURI that's not documented to be escaped. was it actually %3F in the original URL?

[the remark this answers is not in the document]
Andrew7234: Huh that's strange. I think the issue is elsewhere though since the proxy server constructs a new `url` object, perhaps the query params aren't being properly parsed there somehow.

> was it actually %3F in the original URL?
In the sourcify client it's encoded properly as `&` and the params themselves are added via `url.RawQuery` but when it gets to the proxy server the `&` was escaped with `%3F`.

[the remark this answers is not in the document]
pro-wh: `%3F` is `?`, `%26` is `&`

PR #769, opened 2024-10-17
Andrew7234/add validator names

Andrew7234:
Per request from explorer folks, we now include the validator human-readable name in `/*delegations*` endpoints when available. Note that the e2e test updates do not reflect this change because the `metadata_registry` analyzer isn't enabled for e2e_regression tests. However, I did enable it manually and found a couple outdated tests which are also fixed in this PR. 

Screenshot 2024-10-17 at 5 17 09 PM

### DIFF
--- a/api/spec/v1.yaml
+++ b/api/spec/v1.yaml
@@ -1511,6 +1511,9 @@ components:
           type: string
           description: The delegatee (validator) address.
           example: *staking_address_1
+        validator_name:
+          type: string
+          description: The name of the validator entity.
         delegator:
           type: string
           description: The delegator address.
@@ -1545,6 +1548,9 @@ components:
           type: string
           description: The delegatee (validator) address.
           example: *staking_address_1
+        validator_name:
+          type: string
+          description: The name of the validator entity.
         delegator:
           type: string
           description: The delegator address.
--- a/.changelog/769.feature.md
+++ b/.changelog/769.feature.md
@@ -1,0 +1,1 @@
+api: include validator names in /delegations endpoint responses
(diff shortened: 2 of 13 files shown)


api/spec/v1.yaml
```diff
@@ -1511,6 +1511,9 @@ components:
           type: string
           description: The delegatee (validator) address.
           example: *staking_address_1
+        validator_name:
```
ptrus, line 1514: The API's approach to handling validator metadata has become somewhat inconsistent.

Currently, we return `validator_name` for this endpoints, `EntityInfo` for block proposers and signers, and `ValidatorMedia` for validator-specific endpoints. Although these represent the same underlying data, the choice of response fields across different endpoints feels somewhat arbitrary.

To improve this, I suggest either:

1) Unify the responses so all relevant endpoints return the same metadata structure, or
2) (my preference, but apparently not the forntend's): Keep metadata to be returned only by the validator endpoint. The frontend can cache and reuse this information across different parts of the app.

---

Main Reasoning for Option 2:
- If a delegator is a validator, the frontend would likely want to show the validator name. With the current approach, we'd need to introduce additional fields like `delegator_name` to the delegation endpoint
- Consider other endpoints, like `consensus/account/{address}`. When querying account information for a validator, the frontend would want to show the validator's name. Following the current setup, this would mean adding `ValidatorMedia` to the account endpoint response.
- Similarly, if a consensus transaction was submitted by an address that belongs to a validator (or if the recipient is a validator), the frontend would likely want to display the validator's name. This would mean adding the validator metadata to yet another endpoint, resulting in more redundancy.

buberdds: > (my preference, but apparently not the forntend's): Keep metadata to be returned only by the validator endpoint. 
We have discussed this in https://app.clickup.com/t/869683ayv and I was under impression that we agreed to do it on frontend side (PR was created last week https://github.com/oasisprotocol/explorer/pull/1576) because we need names in more views.

ptrus:
Closing since we decided against this https://github.com/oasisprotocol/nexus/pull/769#discussion_r1810207279

PR #796, opened 2024-10-31
add cpu/memory profiling draft

Andrew7234:
For debugging purposes

### DIFF
--- a/cmd/root.go
+++ b/cmd/root.go
@@ -55,6 +55,22 @@ func rootMain(cmd *cobra.Command, args []string) {
 	}
 	logger := common.RootLogger()
 
+	// Initialize cpu profiling.
+	if cfg.Metrics.CpuProfile != "" {
+		logger.Warn("CPU profiling is enabled, this will impact performance (negatively)", "dest", cfg.Metrics.CpuProfile)
+		cpuF, err := os.Create(cfg.Metrics.CpuProfile)
+		if err != nil {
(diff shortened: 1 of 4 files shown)


cmd/root.go
```diff
@@ -55,6 +56,22 @@ func rootMain(cmd *cobra.Command, args []string) {
 	}
 	logger := common.RootLogger()
 
+	// Initialize cpu profiling.
+	if cfg.Metrics.CpuProfile != "" {
+		logger.Warn("CPU profiling is enabled, this will impact performance (negatively)", "dest", cfg.Metrics.CpuProfile)
+		cpuF, err := os.Create(cfg.Metrics.CpuProfile)
```
ptrus, line 62: Can we instead add `pprof.address` (or similar) config parameter. Then if set, we start the `pprof` server on the provided address?

That is more generic and covers more profiling cases, since the user has more control over the profiling setup. That's the way we have it done in other projects, for example [web3-gateway](https://github.com/oasisprotocol/oasis-web3-gateway/pull/497) and [oasis-core](https://github.com/oasisprotocol/oasis-core/blob/6ffc7355616ac3b46747da653b5c16b9a90a25bc/go/oasis-node/cmd/common/pprof/pprof.go#L51-L87), and it tends to work pretty well in practice.

ptrus:
Superseded by https://github.com/oasisprotocol/nexus/pull/1021

PR #798, opened 2024-11-12
Add ERC721 and ERC20 to hardcoded named accounts

buberdds:
In Explorer we want to show token names instead of contract addresses in multiple places

- moved hardcoded addresses to `named-accounts` 
- split paratime files into runtime specific

### DIFF
--- a/named-addresses/mainnet_emerald.json
+++ b/named-addresses/mainnet_emerald.json
@@ -1,0 +1,37 @@
+[
+    {
+        "Name": "Common Pool",
+        "Address": "oasis1qz78phkdan64g040cvqvqpwkplfqf6tj6uwcsh30",
+        "Description": "ParaTime account for rewarding token stakers."
+    },
+    {
+        "Name": "Fee Accumulator",
+        "Address": "oasis1qp3r8hgsnphajmfzfuaa8fhjag7e0yt35cjxq0u4",
+        "Description": "This account collects transaction fees of each ParaTime block that are distributed among the validators."
+    },
+    {
+        "Name": "Rewards Account",
+        "Address": "oasis1qp7x0q9qahahhjas0xde8w0v04ctp4pqzu5mhjav",
+        "Description": "ParaTime account for rewarding the compute nodes."
+    },
+    {
+        "Name": "Pending Withdrawal Account",
+        "Address": "oasis1qr677rv0dcnh7ys4yanlynysvnjtk9gnsyhvm6ln",
+        "Description": "Withdrawal from the ParaTime is performed in two steps. This account holds the tokens that are still waiting for the consensus transaction to be confirmed."
+    },
+    {
+        "Name": "Pending Delegation Account",
+        "Address": "oasis1qzcdegtf7aunxr5n5pw7n5xs3u7cmzlz9gwmq49r",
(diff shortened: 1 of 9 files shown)


named-addresses/mainnet_emerald.json
```diff
@@ -0,0 +1,127 @@
...
+        "Address": "oasis1qpal7x2mn0d0jqa7wtyh7wf3klftp6aufcgzjekt",
+        "Description": "ERC20"
+    },
+    {
+        "Name": "Wrapped ROSE",
+        "Address": "oasis1qpgcp5jzlgk4hcenaj2x82rqk8rrve2keyuc8aaf",
+        "Description": "ERC20"
+    },
+    {
+        "Name": "YuzuSwap LP Token",
```
ptrus, line 63: Why do we have multiple "YuzuSwap LP Token" 's/"Wrapped ROSE"... here? I thought we would just have "trusted"/"official" addresses, not the list of all verified tokens. 

If we have a list of all verified contracts, can't we get that via the existing API? (https://github.com/oasisprotocol/nexus/blob/1f050d70544a0c226f70984ccf8725a84175e9ed/api/spec/v1.yaml#L1062). This seems just a duplicate of that one, with the downside that this one needs to be maintained manually.

named-addresses/mainnet_sapphire.json
```diff
@@ -0,0 +1,187 @@
...
+        "Address": "oasis1qpaeymv9ly8gqweg2qvpnd2ufcj33vgdyqqlruqw",
+        "Description": "ERC20"
+    },
+    {
+        "Name": "wstROSE",
+        "Address": "oasis1qr2ta7ussnps40fnmp58p04cwn7yvr56eqepa7xh",
+        "Description": "ERC20"
+    },
+    {
+        "Name": "Sapphire Inu",
```
ptrus, line 93: Same question here; I thought we would just have a couple of known/trusted addresses here, not all contracts. 
Some of the token names are quite apparently non-serious/spam and don't belong in any "official list" like this one.

buberdds: agree, lists need more removals. Will try to find someone who can decide which tokens are "trusted" for us.

buberdds: based on LJU Summit need to get feedback from Ed first

buberdds: https://github.com/oasisprotocol/nexus/pull/798/commits/d4d5513766923040167ba290e039d04ca2fae25f commit has a list filtered by Ed

named-addresses/readme.txt
```diff
@@ -0,0 +1,5 @@
+This directory contains lists of data for known addresses, to be used by Explorer (and other other frontends) to display names and such.
+
+Contains:
+- Reserved Addresses (https://docs.google.com/spreadsheets/d/1F9CuAqjo4BT8W8JMvogZk-20SO5TgFd8SojgZr2zFCE/edit?usp=sharing)
+- ERC-20 and ERC-721 tokens (verified contracts)
```
ptrus, line 5:
```suggestion
- Known and reputable ERC-20 and ERC-721 tokens
```
Or something like that IMO.

PR #800, opened 2024-11-18
Add recent uptimes to validator metadata in the client api

jberci:


storage/client/client.go
```diff
@@ -44,6 +44,12 @@ const (
 	maxTotalCount = 1000
 )
 
+var (
+	// These two should be kept the same as in the views.validator_uptimes materialized view.
```
ptrus, line 48: These two can be made constants instead of variables.

ptrus (commented):
:+1: Just a couple of minor comments.

api/spec/v1.yaml
```diff
@@ -2085,8 +2085,33 @@ components:
...
+      type: object
+      properties:
+        window_length:
+          type: integer
+          format: uint64
+          description: The length of the window this object describes, in blocks.
+        partial_length:
+          type: integer
+          format: uint64
+          description: The length of the partial windows within window_length, in blocks.
```
Andrew7234, line 2104:
```suggestion
        segment_length:
          type: integer
          format: uint64
          description: The length of the window segment, in blocks. We subdivide the window into segments of equal length and aggregate the uptime of each segment into `segment_uptimes`. Currently defaults to 1200 blocks, which is approximately 2 hours.
```

api/spec/v1.yaml
```diff
@@ -2085,8 +2085,33 @@ components:
...
-        An validator registered at the consensus layer.
+        A validator registered at the consensus layer.
+
+    ValidatorUptime:
+      type: object
+      properties:
+        window_length:
+          type: integer
+          format: uint64
+          description: The length of the window this object describes, in blocks.
```
Andrew7234, line 2100:
```suggestion
          description: The length of the historical window for which this object provides uptime information, in blocks. Currently defaults to 14400 blocks, or approximately 24 hours.
```

api/spec/v1.yaml
Andrew7234: I'm not very enthusiastic about `segment`, I considered using `chunk` or `subwindow` too. Feel free to change

api/spec/v1.yaml
```diff
@@ -2085,8 +2085,33 @@ components:
...
+          type: integer
+          format: uint64
+          description: The length of the partial windows within window_length, in blocks.
+        overall_uptime:
+          type: integer
+          format: uint64
+          description: The number of blocks signed by the validator out of the last window_length blocks.
+        partial_uptimes:
+          type: array
+          description: An array showing the signed block counts for each partial slot within window_length.
```
Andrew7234, line 2111:
```suggestion
        segment_uptimes:
          type: array
          description: An array showing the signed block counts for each sub-segment within window_length. The segments are in reverse-chronological order; ie the first element represents the most recent segment of blocks. 
```

analyzer/consensus/consensus.go
```diff
@@ -316,6 +318,10 @@ func (m *processor) queueDbUpdates(batch *storage.QueryBatch, data allData) erro
 		}
 	}
 
+	if m.mode == analyzer.SlowSyncMode {
+		batch.Queue(validatorUptimesViewRefreshQuery)
```
Andrew7234, line 322: Let's refresh this in a new `ItemBasedAnalyzer`, see `accounts_list.go` for an example. Alternatively, feel free to refactor that analyzer to refresh both views; I think we'll soon have more views that need refreshing, and it'd be good to have them all in one spot so we can tweak how often they get recalculated.

Andrew7234 (commented):
Thank you! Nice job with the sql-fu to calculate the partial windows, left some comments as well. Lmk if you have any questions!

ptrus (approved):

Issue #804, opened 2024-11-27
Querying transfers for a specific token doesn't use an index


ptrus:
On the [tokens page](https://explorer.oasis.io/mainnet/sapphire/token/0x08Fe02Da45720f754e6FCA338eC1286e860d2d2f), querying all the transfers uses the following API call: https://nexus.oasis.io/v1/sapphire/events?offset=0&limit=10&type=evm.log&evm_log_signature=ddf252ad1be2c89b69c2b068fc378daa952ba7f163c4a11628f55a4df523b3ef&rel=0x08Fe02Da45720f754e6FCA338eC1286e860d2d2f

This query doesn't use an index because we don't have an index covering all of the filtered columns. The combined index doesn't include the related accounts, while the related accounts index doesn't include other columns:
https://github.com/oasisprotocol/nexus/blob/7170c28a9dc3762410bbcc20352b0af56ea373eb/storage/migrations/01_runtimes.up.sql#L172-L183

Actually, thinking about it a bit more, the above query is not actually the correct one, as it queries all `ERC20Transfer` events, where related account is the contract address. This would also include any other token transfers which would be sent to the contract address. The frontend should use the following query instead: https://nexus.oasis.io/v1/sapphire/events?offset=0&limit=10&type=evm.log&evm_log_signature=ddf252ad1be2c89b69c2b068fc378daa952ba7f163c4a11628f55a4df523b3ef&contract_address=0x08Fe02Da45720f754e6FCA338eC1286e860d2d2f (note`rel` replaced with `contract_address`). 

However, the indexing problem remains, we need an index for evm log events, something like:

```
 CREATE INDEX ix_runtime_events_evm_events ON chain.runtime_events (runtime,  (body ->> 'address'), evm_log_signature, round) 
     WHERE 
         type = 'evm.log'
```

---

However, on the Accounts page, token-transfers sub-page (e.g. https://explorer.oasis.io/mainnet/sapphire/address/0x1265997E14Eb5e801eA46C360D964203d05C1821/token-transfers) , we also use the events query with same filters: https://nexus.oasis.io/v1/sapphire/events?offset=0&limit=10&type=evm.log&evm_log_signature=ddf252ad1be2c89b69c2b068fc378daa952ba7f163c4a11628f55a4df523b3ef&rel=0x1265997E14Eb5e801eA46C360D964203d05C1821
In this case, using `rel` is correct, so we should update the indexes to also support this case.

ptrus:
[...]

PR #808, opened 2024-11-29
storage: add index for efficient fetching of events emitted by a specific contract

ptrus:
Part of: https://github.com/oasisprotocol/nexus/issues/804

This will enable efficient querying of all transfers of a specific ERC20 token. This is needed for the Transfer tab in the tokens details page (https://explorer.oasis.io/mainnet/sapphire/token/0x6665a6Cae3F52959f0f653E3D04270D54e6f13d8). See issue description for some more details.

Creating the index on the live instance might take some time, so ideally, the index is created CONCURENTLY sometime before deployment to prevent the migration from blocking the deployment.

Note: This doesn't resolve the inefficient accounts-related query, which is also mentioned in the issue. That one will require a separate (likely more significant) change.

TODO: 
- [x] add a test case for the newly supported query
- [x] I applied the index on labs staging mainnet deployment and confirmed the queries are now fast, e.g. (https://nexus.oasis.io/v1/sapphire/events?offset=0&limit=10&type=evm.log&evm_log_signature=ddf252ad1be2c89b69c2b068fc378daa952ba7f163c4a11628f55a4df523b3ef&contract_events=0x6665a6Cae3F52959f0f653E3D04270D54e6f13d8)
- [x] Also applied the index to labs production mainnet 
- [x] prod testnet
- [x] staging testnet

### DIFF
--- a/.changelog/808.feature.md
+++ b/.changelog/808.feature.md
@@ -1,0 +1,1 @@
+storage: add index for fetching events emitted by a specific contract
--- a/storage/client/client.go
+++ b/storage/client/client.go
@@ -1582,9 +1582,17 @@ func (c *StorageClient) RuntimeEvents(ctx context.Context, p apiTypes.GetRuntime
 		h := ethCommon.HexToHash(*p.EvmLogSignature)
 		evmLogSignature = &h
 	}
-	if p.NftId != nil && p.ContractAddress == nil {
-		return nil, fmt.Errorf("must specify contract_address with nft_id")
+
+	// Validate query parameter constraints.
+	// Due to DB indexes setup, other query combinations are inefficient and not supported.
+	switch {
+	case p.NftId != nil && p.ContractAddress == nil:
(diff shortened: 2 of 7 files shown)


pro-wh (approved):
index looks ok, so I'll just approve it. but this won't be the last we hear from this category of problems.

for #804, I think the original intention of the db design was to get a good-enough scan of the events_related_accounts index and then filter specifically token transfers emitted by the contract. but from what I'm hearing, the events_related_accounts index is not viable, I'm guessing due to it not being sorted by round. we need to do something about that, which can be separate from this PR, but still ugh

ptrus:
> events_related_accounts index is not viable, I'm guessing due to it not being sorted by round

Yeah exactly. It works ok for contracts/accounts that don't have a lot of transactions. Probably when it cannot do the sort in-memory anymore. I started tackling that one in https://github.com/oasisprotocol/nexus/pull/809 but it's still in progress.

However, this PR solves a slightly orthogonal problem since the related_accounts query was actually the wrong query for this specific case.

Issue #786, opened 2024-10-23
Track event indexes in block


pro-wh:
<div class="ql-block" data-block-id="block-46237e28-88bd-47a6-8f39-826580054f8b">In both consensus and runtimes, an event can be uniquely identified by the (block height, consecutive idx) pair. We don't track the latter, but we should.</div><div class="ql-block" data-block-id="block-8dbe62d6-7332-4217-9d28-91506054635a"><br></div><div class="ql-block" data-block-id="block-d48917d1-1a42-47c6-8c5b-99fd5411d518"><strong>Task:</strong></div><ol><li class="ql-list-item" data-list="bullet" data-block-id="block-380ec57c-33a6-4aef-adae-7435cf8e072d" style="display: block;"><span tabindex="0" class="ql-ui" contenteditable="false" style="opacity: 1;"></span>add event index to the DB (NOTE - requires DB reindex, hence high-ish priority)</li><li class="ql-list-item" data-list="bullet" data-block-id="block-b1a6b5c3-db62-496b-b9a0-aaa0991dc6d0" style="display: block;"><span tabindex="0" class="ql-ui" contenteditable="false" style="opacity: 1;"></span>expose it in the API (? - optional, possibly not yet needed)</li></ol><div class="ql-block" data-block-id="block-de3db66d-19a4-468e-8d91-d1c6278d86a3"><br></div><div class="ql-block" data-block-id="block-e7edf8d3-38a6-48d4-a4be-4cec5c671691"><strong>Motivation:</strong></div><div class="ql-block" data-block-id="block-feb947dd-3614-47d3-99da-cb3155d4fc35">This will enable us to link to single events from the activity table. We'll either deep-link into the table of a block's events, or we'll create a new FE page that shows a single event.</div>

ptrus:
Fixed in #809

PR #809, opened 2024-11-30
storage: split runtime events related accounts into a separate table

ptrus:
Fixes: #804 #786

This splits `runtime_events.related_accounts` into a separate table, because otherwise we are unable to support efficient queries for event related accounts ordered by round. Some more context in https://github.com/oasisprotocol/nexus/issues/804

Additionally, we need to introduce an event index (solves https://github.com/oasisprotocol/nexus/issues/786) because otherwise the new table cannot reference the existing `runtime_events` table, which currently has no primary key, so rows cannot be referenced uniquely.


analyzer/util/addresses/registration.go
```diff
@@ -112,69 +112,3 @@ func registerRuntimeAddress(addressPreimages map[apiTypes.Address]*PreimageData,
...
-	return addr, nil
-}
-
-func RegisterRelatedEthAddress(addressPreimages map[apiTypes.Address]*PreimageData, relatedAddresses map[apiTypes.Address]struct{}, ethAddr []byte) (apiTypes.Address, error) {
-	addr, err := registerEthAddress(addressPreimages, ethAddr)
-	if err != nil {
-		return "", err
-	}
-
-	relatedAddresses[addr] = struct{}{}
```
pro-wh, line 166: what does these now

analyzer/runtime/visitors.go
```diff
@@ -130,63 +130,69 @@ func VisitCall(call *sdkTypes.Call, result *sdkTypes.CallResult, handler *CallHa
...
-			if err = handler.EVM(&evmEvents[i]); err != nil {
-				return fmt.Errorf("decoded event %d evm: %w", i, err)
+			if err = handler.EVM(&evmEvents[i], currIdx); err != nil {
+				return currIdx, fmt.Errorf("decoded event %d evm: %w", i, err)
 			}
+			currIdx++
 		}
 	}
-	return nil
+	return currIdx, nil
```
ptrus, line 188: This doesn't work ok with events that are currently not tracked and would be tracked in future. E.g. adding a handler for new event types would change the resulting indexes on reindex. 

The index should probably be made out of 2 values (index within GetEventsRaw, and then the index within the "DecodedEvents").

pro-wh: do we skip events?

ptrus: > do we skip events?

Yeah for example rofl events are currently skipped (added in https://github.com/oasisprotocol/nexus/pull/812)

> come to think of it, how do oasis-core and oasis-sdk architecturally represent events? I think they're put into a map[event_type]slice_of_events kind of structure per block or tx

Yeah that's how it is done in oasis-sdk. So maybe going with (`event_type`, `index`(within the type)) pair would be the best, to match how oasis-sdk represents it?

pro-wh: that sounds ok. I hope using a text column in the key is not too expensive

ptrus: Actually I went with the per block unique event indexes (regardless of type) since that was the most straightforward.

analyzer/util/addresses/registration.go
pro-wh: eh ok if we can void missing a spot

pro-wh: although is it better without these?

> where all the lines that update the eventData.RelatedAddresses map remain as they were
> parts of the code the functions were used, while in other parts (e.g. updating events related addresses) the maps were manually updated

inconsistency bad. among consistent options you prefer no-helper over helper?

analyzer/runtime/extract.go
```diff
@@ -344,7 +342,7 @@ func ExtractRound(blockHeader nodeapi.RuntimeBlockHeader, txrs []nodeapi.Runtime
 					}
 					// Set the 'Success' field to 'Pending' for deposits. This is because the outcome of the Deposit tx is only known in the next block.
 					blockTransactionData.Success = nil
-
+					blockTransactionData.RelatedAccountAddresses[to] = struct{}{}
```
pro-wh, line 345: oh I see. so here for example there's additional logic used in arriving at the "to" address that we want to apply before adding the "to" address to the related accounts

although we lucked out here because the signers are elsewhere added to the related accounts set

analyzer/util/addresses/registration.go
pro-wh: ok yeah I see later on in extractEvents it gets all fiddly with places where they used to go into two different related address sets so you end up having to do `...[] = struct{}{}` either way

storage/migrations/01_runtimes.up.sql
```diff
@@ -180,12 +184,27 @@ CREATE INDEX ix_runtime_events_nft_transfers ON chain.runtime_events (runtime, (
...
+-- Added in 13_runtime_events_related_accounts.up.sql.
+-- CREATE TABLE chain.runtime_events_related_accounts
+-- (
+--     runtime runtime NOT NULL,
+--     round UINT63 NOT NULL,
+--     event_index UINT31 NOT NULL,
+
+--     tx_index UINT31,
+--     type TEXT NOT NULL,
+--     account_address oasis_addr NOT NULL,
```
pro-wh, line 203: this is a little out of sync with 13_....up.sql, missing e.g. the PRIMARY KEY

storage/migrations/14_events_related_accounts.up.sql
```diff
@@ -0,0 +1,155 @@
...
+    -- This is a runtime identifier, which is a binary "namespace," e.g.
+    -- `000000000000000000000000000000000000000000000000f80306c9858e7279` for
+    -- Sapphire on Mainnet. This is taken from the event from the node, and it
+    -- is set even for runtimes that nexus isn't configured to analyze.
+    roothash_runtime_id HEX64,
+    roothash_runtime runtime,
+    roothash_runtime_round UINT63
+);
+
+CREATE TABLE chain.events_related_accounts
```
pro-wh, line 34: not replicated as a comment in 00_consensus.up.sql. should we? I noticed you did for the runtimes schema

pro-wh (approved):
big change but the underlying concept is simple: moving an array out to its own table so that we can index it in order of blocks.

also revamping the events tables so that they have a fully defined order, based on how oasis-core does it.

there's some changes to the extractor code, thanks for wading through, sorry that the "register related" system ended up not helping

storage/migrations/01_runtimes.up.sql
ptrus: Updated, thanks for noticing!

storage/migrations/14_events_related_accounts.up.sql
ptrus: Updated, thanks!

Issue #803, opened 2024-11-26
Do not run initializing queries in all parallel workers


ptrus:
Block analyzer initialization requires some somewhat expensive queries https://github.com/oasisprotocol/nexus/blob/7170c28a9dc3762410bbcc20352b0af56ea373eb/analyzer/block/block.go#L319-L327 

This can be problematic if you are running fast-sync with height parallelisation since you have A LOT of contesting queries on startup. Especially running the `softEnqueueGapsInProcessedBlocks` in parallel might cause a lot of contention.

These queries don't need to be run for all parallel workers, it would be enough to run these queries just once, before starting the N parallel analyzers. 

---

This was noticed by @uscinski when experimenting on the OPF deployment, where each restart took more than one hour actually to start indexing.

PR #810, opened 2024-11-30
fix: Run PreWork once, even for parallel analyzers

ptrus:
Fixes: https://github.com/oasisprotocol/nexus/issues/803

### DIFF
--- a/analyzer/block/block.go
+++ b/analyzer/block/block.go
@@ -39,7 +39,9 @@ const (
 // block based analyzer.
 type BlockProcessor interface {
 	// PreWork performs tasks that need to be done before the main processing loop starts.
+	// In parallel mode, this method is called by a single instance.
 	PreWork(ctx context.Context) error
+
 	// ProcessBlock processes the provided block, retrieving all required information
 	// from source storage and committing an atomically-executed batch of queries
 	// to target storage.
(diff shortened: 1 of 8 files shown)


analyzer/block/block.go
```diff
@@ -39,7 +39,9 @@ const (
 // block based analyzer.
 type BlockProcessor interface {
 	// PreWork performs tasks that need to be done before the main processing loop starts.
+	// When running in parallel mode, this method will only be called once.
```
pro-wh, line 42: more info: when running in parallel mode with multiple analyzer instances, only one of those instances will have PreWork called. so don't use it to initialize any fields of the processor. only use it for shared state, such as fixing up the database

pro-wh (approved):
good, I've wanted this

pro-wh:
o sorry I gotta ask a question first

analyzer/block/block.go
```diff
@@ -316,16 +333,6 @@ func (b *blockBasedAnalyzer) Start(ctx context.Context) {
 		to = b.blockRange.To
```
pro-wh, line 333: does any prework depend on this

pro-wh (approved):
ok nvm it calls PreWork on both one fast sync analyzer and the slow sync analyzer. seems fine as the PreWork internally branches on b.slowSync. everybody just keep that in mind

analyzer/block/block.go
ptrus: Nope

cmd/analyzer/analyzer.go
```diff
@@ -694,16 +694,31 @@ func (a *Service) Start() {
...
 			fastSyncWg[an.SyncTag] = wg
+
+			// This is the first fast-sync analyzer with this SyncTag, so PreWork should be run.
+			// This relies on the assumption that each distinct 'type' of fast-sync analyzer uses a unique `SyncTag`.
+			// This assumption holds true at the moment.
+			runPreWork = true
 		}
 		wg.Add(1)
-		go func(an SyncedAnalyzer) {
+		go func(an SyncedAnalyzer, runPreWork bool) {
```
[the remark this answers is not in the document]
ptrus: Thanks, updated for consistency with `wg`. Although, for me the explicit goroutine param seems a bit cleaner to read.

analyzer/block/block.go
pro-wh: kool

cmd/analyzer/analyzer.go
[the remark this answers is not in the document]
pro-wh: oh ok, thanks

Issue #818, opened 2024-12-03
Higher level status of consensus-accounts runtime transactions


ptrus:
Consensus-accounts runtime transactions (`consensus.Deposit`, `consensus.Withdraw`...) can technically succeed, but the intended actions (which happen in the next round) fail. For example, if one submits a structurally valid `consensus.Deposit` transaction and doesn't have the necessary `Allowance` set on the account, the actual despoit fails. This is signaled by the `DelegateEvent` emitted event in the next round.

Nexus should report these transactions as pending until the event is received, and once the result is known, mark them as either failed or successful.

Likely the "cleanest" implementation is to:
- rename the `success` field to `status` (with three possible values: success, fail, pending) in the `runtime_transactions` table (or keep it named `success` for now?)
- re-use the existing Error related columns in the `runtime_transactions` table 
- update the transaction processing, so that the consensus-account transactions are first marked as pending
- add event-processing for consensus-account events and update the transaction status accordingly
  - this cannot be done in fast-sync, since we don't process blocks sequentially. We will likely need to handle this in the "finalize-fast-sync" step.

Related explorer issue: https://github.com/oasisprotocol/explorer/issues/528

PR #820, opened 2024-12-04
runtime: Handle status of consensus-accounts txs

ptrus:
Fixes: #818

Tested on localnet, works. The transaction status is listed as unknown until it is then either marked as successful or failed. We should probably also update the icon/text on the frontend specifically for statuses for these transactions.

Screenshot 2024-12-04 at 16 52 55

(ignore the outdated text)


storage/migrations/01_runtimes.up.sql
```diff
@@ -86,19 +86,19 @@ CREATE TABLE chain.runtime_transactions
   oasis_encrypted_result_data BYTEA,
 
   -- Error information.
-  success       BOOLEAN,  -- NULL means success is unknown (can happen in confidential runtimes)
+  success       BOOLEAN,  -- NULL means success is unknown (can happen in confidential runtimes, or for 'consensusaccounts' transactions which whose action is known only in the next round)
```
ptrus, line 89: Not convinced if it's better to re-use the existing `success` field, or introduce a new additional `status` field. With the new field, we would be able to differentiate between "transaction was successful, but the event in the next round failed", but I don't think we need that. One can get that info by looking at events.

Unless there's good reasoning against it, I would just reuse the `status` field.

Andrew7234: Agreed 👍

analyzer/queries/queries.go
```diff
@@ -1147,4 +1147,107 @@ var (
...
+    WHERE
+      rt.runtime = matched_tx.runtime AND
+      rt.round = matched_tx.round AND
+      rt.tx_index = matched_tx.tx_index`
+
+	RuntimeConsensusAccountTransactionStatusUpdateFastSync = `
+    INSERT INTO todo_updates.transaction_status_updates (runtime, round, method, sender, nonce, success, error_module, error_code, error_message)
+    VALUES ($1, $2, $3, $4, $5, $6, $7, $8, $9)`
+
+	RuntimeTransactionStatusUpdatesRecompute = `
```
Andrew7234, line 1198: This looks good since it's similar to the query above, but if you have a chance could you double check whether the fast-sync eden e2e regression test has one of the consensus-accounts txs and that it's being properly processed? It'd be nice to have another data point confirming that finalizeFastSync will work as expected. Thanks!

Andrew7234 (approved):
Thank you! This lgtm overall, though @pro-wh is more familiar with this area so I'd be curious to hear his feedback too

analyzer/queries/queries.go
```diff
@@ -1147,4 +1147,107 @@ var (
...
+        rt.runtime = rts.runtime AND
+        rt.round = rts.round AND
+        rt.tx_index = rts.tx_index
+      WHERE
+        rt.runtime = $1 AND
+        rt.round = (SELECT round FROM tx_round) AND
+        rt.method = $3 AND
+        rts.signer_address = $4 AND
+        rts.signer_index = 0 AND
+        rts.nonce = $5
```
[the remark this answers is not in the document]
ptrus: > Does the (signer, nonce) here uniquely identify the corresponding transaction if there are two consensus.Deposit txs in the same block?

Yes, since there can be only one transaction per signer in a block.

> the confirmation event is guaranteed to be emitted in block n+1 and not any future block, correct?

Close, but no. The event will be in the first next _successfully processed_ block. In practice this will be the next block in most cases, but we cannot rely on it. We currently don't track the block status (TODO: will add an issue for this), so in `tx_round` subquery we search for the previous block with at least one transaction (that should be the block where the corresponding transaction was sent).

analyzer/queries/queries.go
ptrus: Yes, for example the `consensus.Withdraw`'s which are before round 8059839 in the https://github.com/oasisprotocol/nexus/blob/main/tests/e2e_regression/eden/expected/emerald_txs_by_method.body test

PR #821, opened 2024-12-05
server: cancel the request context on timeout

ptrus:
Added `ContextTimeoutMiddleware` to enforce request context timeout and ensure proper cancellation of timed-out requests. This ensures that the context for requests exceeding the defined timeout limit are properly canceled.

I noticed cases like the one below, where request takes 15 seconds:
```json
{"caller":"middleware.go:77","latency":"14.69731168s","latency_bin":">1000ms","level":"info","module":"api","msg":"ending request","query_params":"limit=10&offset=0","query_path":"/v1/sapphire/evm_tokens","request_id":"3fecdf6b-ad74-42f0-8cfc-52ba0787f05a","status_code":200,"ts":"2024-12-05T08:37:32.086384228Z"}
```

Note that the server logs show a 200 OK status, but the response is not delivered to the client. Instead, the client receives an EOF since the [configured server timeout is reached](https://github.com/oasisprotocol/nexus/blob/ca4ef6dc4be3f322762b7b59b1df14531514a92f/cmd/api/api.go#L202-L203), while the server continues processing the request. 

Ref upstream issue: https://github.com/golang/go/issues/59602

### DIFF
--- a/api/errors.go
+++ b/api/errors.go
@@ -8,6 +8,8 @@ import (
 	"reflect"
 
 	apiTypes "github.com/oasisprotocol/nexus/api/v1/types"
+	"github.com/oasisprotocol/nexus/common"
+	"github.com/oasisprotocol/nexus/log"
 )
 
 var (
@@ -64,16 +66,29 @@ func HttpCodeForError(err error) int {
 	}
 }
 
-// A simple error handler that renders any error as human-readable JSON to
+// A simple error handler that logs and renders any error as human-readable JSON to
 // the HTTP response stream `w`.
-func HumanReadableJsonErrorHandler(w http.ResponseWriter, r *http.Request, err error) {
-	w.Header().Set("content-type", "application/json; charset=utf-8")
-	w.Header().Set("x-content-type-options", "nosniff")
-	w.WriteHeader(HttpCodeForError(err))
+func HumanReadableJsonErrorHandler(logger log.Logger) func(http.ResponseWriter, *http.Request, error) {
+	return func(w http.ResponseWriter, r *http.Request, err error) {
+		logger.Debug("request failed, handling human readable error",
+			"err", err,
+			"request_id", r.Context().Value(common.RequestIDContextKey),
+			"ctx_err", r.Context().Err(),
+		)
 
-	// Wrap the error into a trivial JSON object as specified in the OpenAPI spec.
-	msg := err.Error()
-	errStruct := apiTypes.HumanReadableError{Msg: msg}
+		// If request context is closed, don't bother writing a response.
+		if r.Context().Err() != nil {
+			return
+		}
 
-	_ = json.NewEncoder(w).Encode(errStruct)
+		w.Header().Set("content-type", "application/json; charset=utf-8")
+		w.Header().Set("x-content-type-options", "nosniff")
+		w.WriteHeader(HttpCodeForError(err))
+
+		// Wrap the error into a trivial JSON object as specified in the OpenAPI spec.
+		msg := err.Error()
+		errStruct := apiTypes.HumanReadableError{Msg: msg}
+
+		_ = json.NewEncoder(w).Encode(errStruct)
+	}
 }
--- a/cmd/api/api.go
+++ b/cmd/api/api.go
@@ -30,6 +30,8 @@ const (
 	moduleName = "api"
 	// The path portion with which all v1 API endpoints start.
 	v1BaseURL = "/v1"
+
+	defaultRequestHandleTimeout = 10 * time.Second
 )
 
 var (
@@ -108,6 +110,8 @@ type Service struct {
 	address string
 	target  *storage.StorageClient
 	logger  *log.Logger
+
+	requestTimeout time.Duration
 }
 
 // NewService creates a new API service.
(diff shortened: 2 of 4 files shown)


api/errors.go
```diff
@@ -64,16 +66,29 @@ func HttpCodeForError(err error) int {
...
-// A simple error handler that renders any error as human-readable JSON to
+// A simple error handler that logs and renders any error as human-readable JSON to
 // the HTTP response stream `w`.
-func HumanReadableJsonErrorHandler(w http.ResponseWriter, r *http.Request, err error) {
-	w.Header().Set("content-type", "application/json; charset=utf-8")
-	w.Header().Set("x-content-type-options", "nosniff")
-	w.WriteHeader(HttpCodeForError(err))
+func HumanReadableJsonErrorHandler(logger log.Logger) func(http.ResponseWriter, *http.Request, error) {
+	return func(w http.ResponseWriter, r *http.Request, err error) {
+		logger.Debug("request failed, handling human readable error",
```
ptrus, line 73: The openapi-codegen generanted code can return an error and call `ResponseErrorHandlerFunc` in some places:

```go
        // server.gen.go:5111
	response, err := handler(r.Context(), w, r, request)

	if err != nil {
		sh.options.ResponseErrorHandlerFunc(w, r, err)
	} else if validResponse, ok := response.(GetRuntimeEvmTokensResponseObject); ok {
		if err := validResponse.VisitGetRuntimeEvmTokensResponse(w); err != nil {
			sh.options.ResponseErrorHandlerFunc(w, r, err)
		}
	} else if response != nil {
		sh.options.ResponseErrorHandlerFunc(w, r, fmt.Errorf("Unexpected response type: %T", response))
	}
```

This is the only place where we can log these errors, so let's do it.

pro-wh: good, lots of our other products log request errors too

api/errors.go
```diff
@@ -64,16 +66,29 @@ func HttpCodeForError(err error) int {
...
+			"err", err,
+			"request_id", r.Context().Value(common.RequestIDContextKey),
+			"ctx_err", r.Context().Err(),
+		)
 
-	// Wrap the error into a trivial JSON object as specified in the OpenAPI spec.
-	msg := err.Error()
-	errStruct := apiTypes.HumanReadableError{Msg: msg}
+		// If request context is closed, don't bother writing a response.
+		if r.Context().Err() != nil {
```
pro-wh, line 80: does this mean we leave the request hanging if our handling times out?

cmd/api/api.go
```diff
@@ -187,20 +189,24 @@ func (s *Service) Start() {
...
 	// HandlerWithOptions() above does not apply it to some requests (404 URLs, requests with bad params, etc.).
 	handler = api.MetricsMiddleware(metrics.NewDefaultRequestMetrics(moduleName), *s.logger)(handler)
 
 	server := &http.Server{
 		Addr:           s.address,
 		Handler:        handler,
-		ReadTimeout:    10 * time.Second,
-		WriteTimeout:   10 * time.Second,
+		ReadTimeout:    requestsTimeout,
+		WriteTimeout:   requestsTimeout,
```
pro-wh, line 209: ReadTimeout and WriteTimeout are for request I/O and not handling right? these don't have to be the same as the requestsTimeout

api/middleware_test.go
```diff
@@ -0,0 +1,52 @@
...
+	"time"
+
+	"github.com/stretchr/testify/require"
+)
+
+func TestTimeoutMiddleware(t *testing.T) {
+	withMiddleware := false
+	baseHandler := http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		// Wait some time, so that the request timeout (10ms) should be reached.
+		<-time.After(2 * time.Second)
```
pro-wh, line 18: actually does this work? we don't set up any blocked writes

cmd/api/api.go
ptrus: No, WriteTimeout is not just the IO for writing the response, it includes the server processing.

pro-wh: oh, should that be some amount higher than the processing time limit then?

api/errors.go
ptrus: If the context is canceled the client already received an EOF (due to the WriteTimeout being reached), so doing anything here will have no effect.


But maybe instead of using our own 'ContextTimeoutHandler' which only cancels the context, we should use upstream [TimeoutHandler](https://cs.opensource.google/go/go/+/refs/tags/go1.23.4:src/net/http/server.go;l=3620). That one also handles context cancellation, but it also returns a 503, instead of just closing the connection as it is done now. Thats probably nicer.

cmd/api/api.go
ptrus: Good idea, i think we should; and then by also using the upstream Timeout handler (i mentioned in the other comment) the client will get a 503 in case the processing timeout is reached.

api/errors.go
ptrus: Now that we use `http.TimeoutHandler`, the client receives a 503 and the connection is closed when the timeout is reached. So like before, if the context is canceled there is nothing to do. Even if we didn't return here, the Writes below would fail with `ErrHandlerTimeout`.

api/middleware_test.go
ptrus: It just tested that the request context was canceled, but now that we use the upstream TimeoutHandler, there's no more need for testing the timeout middleware anymore, so I removed the test.

cmd/api/api.go
ptrus: Updated.

ptrus:
I'm merging since we want to do a new release today, can apply potential requestsed changes in a future PR.

PR #822, opened 2024-12-09
analyer/aggregate_stats: fix errors check

ptrus:
Minifix extracted from: https://github.com/oasisprotocol/nexus/pull/809


pro-wh (approved):
oh yeah, docs do say `err` is first

PR #823, opened 2024-12-09
api/consensus/events: Include timestamp

ptrus:
Having events timestamps is useful; it also unifies the API with runtime events, which already contain timestamps.

TODO:
- [x] update E2E tests


pro-wh (approved):
I'm in favor

PR #832, opened 2024-12-18
storage: Mark postgresql helpers functions as STABLE

ptrus:
Marking these functions as STABLE allows them to be used in index expressions and enables PostgreSQL to optimize queries by avoiding unnecessary recomputations of the function during query execution.

Without the STABLE attribute, PostgreSQL treats the functions as VOLATILE, forcing it to recompute the function every time it is called within a query, even if the input is static.

More details: https://www.postgresql.org/docs/current/xfunc-volatility.html#XFUNC-VOLATILITY

Changes already applied to nexus.prd and testnet.nexus.prd.

### DIFF
--- a/.changelog/832.feature.md
+++ b/.changelog/832.feature.md
@@ -1,0 +1,5 @@
+storage: Mark postgresql address helpers functions as STABLE
+
+Marking these functions as STABLE allows them to be used in index expressions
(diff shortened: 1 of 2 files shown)


ptrus:
Merging since this has already been applied.

PR #836, opened 2024-12-18
storage: Update block proposer index to include height

ptrus:
We always order by height, so make the index more efficient by including the height.

Already applied.


ptrus:
Merging since this has already been applied.

PR #837, opened 2024-12-18
analyzer: configurable MaxBackoffTimeout

ptrus:
This adds a configurable `MaxBackoffTime` to analyzers. This is useful for analyzers such as `validatorstakinghistory`, where once caught up, we only expect new work on epoch transitions and don't need to query every 6 seconds.

Additional note: the queries for fetching `validatorstakinghistory` work are not the most efficient (take ~2 seconds). I noticed this because these queries are currently using a lot of the database's CPUs—even if there's no work to be done. The idea is that with the `MaxBackoffTime` change, we don't need to bloat the DB with additional indexes and just be smarter about how frequently we do the queries.

Alternatives considered:
- We could manually update the deployment configuration for `validatorstakinghistory` to use a "fixed-interval" once caught up. But if the analyzer falls behind (for whatever reason. e.g. node downtime etc) we would need manual intervention again to update the config. With implemented solution it should recover automatically and use a shorter timeout if there's more work to do.

- "Epoch"-aware Item analyzer. If the ItemAnalyzer would be aware of epochs, it could then know when to call the analyzer if it is caught up. However this solution would require more (non-trivial) work to implement, which doesn't seem worth for this specific case.

### DIFF
--- a/.changelog/837.feature.md
+++ b/.changelog/837.feature.md
@@ -1,0 +1,1 @@
+analyzers: add configurable `MaxBackoffTimeout`
--- a/analyzer/item/item.go
+++ b/analyzer/item/item.go
@@ -36,6 +36,7 @@ type itemBasedAnalyzer[Item any] struct {
 	stopIfQueueEmptyFor time.Duration
 	fixedInterval       time.Duration
 	interItemDelay      time.Duration
+	maxBackoffTime      time.Duration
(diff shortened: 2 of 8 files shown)


ptrus:
I'm merging since we want to do a new release today. I can apply the potential changes requested in a future PR.

Issue #884, opened 2025-01-16
Runtime API expose all signers for runtime transactions


ptrus:
Runtime transactions can have multiple signers. The runtime API currently only exposes the first signer of the transactions. 
The DB has all the needed data, just the API/queries need to be updated.

PR #885, opened 2025-01-17
api/runtime/transactions: Add signers field to response

ptrus:
Fixes https://github.com/oasisprotocol/nexus/issues/884

The existing sender_0, sender_0_eth, and nonce_0 are deprecated and will be removed in the future.

### DIFF
--- a/api/spec/v1.yaml
+++ b/api/spec/v1.yaml
@@ -2914,7 +2914,7 @@ components:
     RuntimeTransaction:
       type: object
       # NOTE: Not guaranteed to be present: eth_hash, to, amount.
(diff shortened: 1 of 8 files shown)


api/spec/v1.yaml
```diff
@@ -2942,24 +2942,34 @@ components:
...
             Absent for non-Ethereum-format transactions.
           example: 9e6a5837c6366d4a7e477c71ffe32d40915cdef7ef209792259e5ee70caf2705
+        signers:
+          type: array
+          items:
+            allOf: [$ref: '#/components/schemas/RuntimeTransactionSigner']
+          description: The signers of this transaction.
         sender_0:
           allOf: [$ref: '#/components/schemas/Address']
+          deprecated: true
```
csillag, line 2952: Nice  touch, keeping the old stuff for a smooth transition

Issue #886, opened 2025-01-18
API: Consensus block proposal data is often (always?) missing


csillag:
On the Nexus API, consensus blocks have a data field, called `proposer`, defined here:

https://github.com/oasisprotocol/nexus/blob/c711d6af51c48c42934aec40722802669aff40d5/api/spec/v1.yaml#L1459-L1461

```
    proposer:
      allOf: [$ref: '#/components/schemas/EntityInfo']
      description: The entity that proposed this block.
```

However, looking at actual data ([example](https://nexus.oasis.io/v1/consensus/blocks/22875632)), this field is often missing.

Technically speaking, this doesn't make the data invalid, since on the API, this field has been defined optional, but looking at the reality, it doesn't make much sense, since every block has a proposer.

We know we have the data, since on the validator API endpoints the proposed/signed blocks seem to be displayed correctly.

So I assume it's just the API code that is not tunneling this piece of data? In any case, as discussed elsewhere, this should be fixed.

ptrus:
This was only problematic when querying a single block by height via (`/consensus/blocks/{height}`) endpoint. The `/consensus/blocks` works correctly.   Fixed in [#888](https://github.com/oasisprotocol/nexus/pull/888)

PR #888, opened 2025-01-18
api/consensus/block: Fix response when querying a single block

ptrus:
Fixes https://github.com/oasisprotocol/nexus/issues/886

API endpoint returning a single block by height was using a separate code path to fetch the blocks which lead to the regression mentioned in the above issue.

This is now fixed so that a single code path is used when fetching either a single or multiple consensus blocks. This also resolves the issue.


ptrus:
Yes for `proposer` we can, for `signers` we cannot, since we only get the signers of block X, when block X+1 is processed,so the latest block will always be missing signers.

Will update

csillag:
Yes I meant the proposer. Thanks.

Issue #896, opened 2025-01-22
Support efficient query for related account with method filter


ptrus:
We currently lack an efficient index for the "related_account" + "method type" filter combination when fetching transactions. To handle this efficiently, we would need to de-normalize and duplicate the "method type" field in the "runtime_related_transactions" table (and similarly for consensus tables).

This would allow us to create an index that includes both the "related_account" and "method type" fields.

Since this is a very common and useful filtering combination, this de-normalization is justified for this case, but we should aim to limit such changes where possible.

PR #897, opened 2025-01-23
Support efficient query for related account with method filter

ptrus:
Fixes: https://github.com/oasisprotocol/nexus/issues/896

- Promote 'likely_native_transfer' to a first-class database field
- Denormalize 'method' and 'likely_native_transfer' to create efficient indexes for account filtering

Note: The migrations will likely take multiple hours Before this is released and deployed, it would be good to have "Nexus Blue " ready so that we can switch Nexus backend explorer uses during the update.


storage/client/queries/queries.go
```diff
@@ -544,10 +550,10 @@ const (
 						TRUE
 					-- Special case to return are 'likely to be native transfers'.
 					WHEN $5::text = 'native_transfers' THEN
-						(txs.method = 'accounts.Transfer' OR (txs.method = 'evm.Call' AND (txs.body ->> 'data') = ''))
+						(txs.likely_native_transfer = TRUE)
```
pro-wh, line 553: boolean_field = TRUE

is that our style?

storage/migrations/13_related_transactions_method_denorm.up.sql
```diff
@@ -4,13 +4,52 @@ BEGIN;
...
-	FROM chain.transactions AS t
-	WHERE art.tx_block = t.block AND art.tx_index = t.tx_index;
+DO $$
+DECLARE
+    block_batch_size INT := 50000;
+    start_block INT;
+    end_block   INT;
+    max_block   INT;
+    affected_rows INT;
+BEGIN
```
pro-wh, line 14: some of this stuff seems to be in the wrong commit?

pro-wh (commented):
wait where does this go relative to the related accounts change?

ptrus:
> where does this go relative to the related accounts change?

The related accounts change (#809) was only for the events tables. Transactions tables already had related accounts split out in a separate table, so no changes were needed in #809. 

This PR only deals with transaction tables. We denomralize some fields so that we can have more efficient indexes for the combination of related accounts + method filters.

storage/client/queries/queries.go
ptrus: Fixed 🙈

storage/migrations/13_related_transactions_method_denorm.up.sql
ptrus: Fixed!

storage/migrations/18_related_runtime_transactions_method_denorm.up.sql
```diff
@@ -0,0 +1,89 @@
...
+
+-- We commit here, to ensure index creation below works.
+COMMIT;
+
+BEGIN;
+
+CREATE INDEX ix_runtime_related_transactions_address_method_round ON chain.runtime_related_transactions (runtime, account_address, method, tx_round, tx_index);
+CREATE INDEX ix_runtime_related_transactions_address_native_transfer_round ON chain.runtime_related_transactions (runtime, account_address, likely_native_transfer, tx_round, tx_index);
+-- This combination is needed, since explorer needs to obtain evm.Call's which are not native transfers.
+CREATE INDEX ix_runtime_related_transactions_address_method_native_transfer_round ON chain.runtime_related_transactions (runtime, account_address, method, likely_native_transfer, tx_round, tx_index);
```
Andrew7234, line 87: do we want `tx_round DESC` here?

pro-wh: should be equivalent either way, with the preceding fields only filtered by equality

Issue #900, opened 2025-01-29
Curated ERC20 token ordering


ptrus:
We should support some curated token ordering for the evm_tokens endpoint.

PR #903, opened 2025-02-13
runtime/evm_tokens: Support (partially) curated ordering

ptrus:
Fixes: https://github.com/oasisprotocol/nexus/issues/900


pro-wh (approved):
:+1: on impl

- new config field evm_tokens_custom_ordering (see doc comment for structure)
- parses it into core address array and rank number array
- sends arrays to db inside the query, query joins and sorts
- tokens table is small enough that db can do this however
- tests

Issue #878, opened 2025-01-12
Add endpoints for total and circulating supplies


kostko:
We should add an endpoint that shows the total supply and an endpoint that shows the current circulating supply, computed by taking the total supply and subtracting the balances of some specific addresses (eg. the common pool and some others - should be configurable via config).

Ideally the endpoint also supports raw output without json etc (for external services that will fetch this).

ptrus:
@jberci LMK if you have the time for this one in this/next sprint

PR #904, opened 2025-02-14
Add endpoints for total and circulating supplies

ptrus:
Fixes: https://github.com/oasisprotocol/nexus/issues/878


tests/e2e_regression/damask/e2e_config_1.yml
```diff
@@ -12,7 +12,7 @@ analysis:
     nodes:
       # Node will be accessed only when the filling up the cache.
       # In CI, the cache is expected to have all the node responses as the node is not reachable.
-      damask: { default: { rpc: unix:/tmp/node.sock } }
+      damask: { default: { rpc: internal-mainnet-archive-oasis-3.prd.oasis.io } }
```
pro-wh, line 15: double checking, do we want to check this in?

tests/e2e_regression/edenfast/e2e_config_1.yml
```diff
@@ -52,6 +52,10 @@ server:
   storage:
     endpoint: postgresql://api:password@localhost:5432/indexer?sslmode=disable
     backend: postgres
+  # Set some reserved addresses, so that circulating supply functionality is tested.
+  circulating_supply_reserved_addresses:
+    - oasis1qp3r8hgsnphajmfzfuaa8fhjag7e0yt35cjxq0u4
```
pro-wh, line 57: are we allowed to make the actual list public

api/spec/v1.yaml
```diff
@@ -149,6 +149,30 @@ paths:
...
         <<: *common_error_responses
 
+  /consensus/total_supply_raw:
+    get:
+      summary: Returns the total supply of the consensus layer token.
+      responses:
+        '200':
+          description: A string containing the total supply of consensus layer token.
+          content:
+            text/plain:
```
pro-wh, line 159: what's the story behind this, did someone want text/plain?

pro-wh (commented):
- get total supply over staking grpc
- subtract out a set of accounts from the config (implicitly incl. common pool)

question about api being text/plain. avoids the json quotes :shrug:

tests/e2e_regression/edenfast/e2e_config_1.yml
ptrus: Not sure, but these are "random" accounts I picked from the chain

api/spec/v1.yaml
ptrus: Yeah the referenced issue

> Ideally the endpoint also supports raw output without json etc (for external services that will fetch this).

And since this is currently the only use case, I only implemented the "raw" text/plain endpoint.

tests/e2e_regression/damask/e2e_config_1.yml
ptrus: :+1: will remove. This ednpoint is not public (yet), firewalled for now.

api/spec/v1.yaml
pro-wh: oh huh interesting

tests/e2e_regression/edenfast/e2e_config_1.yml
pro-wh: eh ok

pro-wh (approved):
text/plain is as requested in the feature request

> Ideally the endpoint also supports raw output without json etc

:shrug:

Issue #905, opened 2025-02-17
Validator history endpoint sometimes times-out


ptrus:
e.g. https://nexus-blue-testnet.prd0-gke.oasis.io/v1/consensus/validators/oasis1qqv25adrld8jjquzxzg769689lgf9jxvwgjs8tha/history?limit=1

PR #906, opened 2025-02-17
consensus/validator_history: Optimize validator history query

ptrus:
Fixes: https://github.com/oasisprotocol/nexus/issues/905

### DIFF
--- a/.changelog/906.bugfix.md
+++ b/.changelog/906.bugfix.md
@@ -1,0 +1,1 @@
+consensus/validator_history: Optimize validator history query
--- a/storage/client/queries/queries.go
+++ b/storage/client/queries/queries.go
@@ -448,6 +448,13 @@ const (
 		OFFSET $4::bigint`
 
 	ValidatorHistory = `
+		-- This would ideally use the address preimages table, but it looks like the address preimages table was not populated for consensus data.
+		-- https://github.com/oasisprotocol/nexus/issues/907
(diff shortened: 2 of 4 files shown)


pro-wh (approved):
fine, but this is a loss for address_preimage generality

ptrus:
> but this is a loss for address_preimage generality

Can you elaborate?

pro-wh:
the address preimages table has (should?) these mappings from oasis address to the thing that was hashed into the address, e.g. entity public key. but here we're adding an index to the entities table.

ptrus:
> the address preimages table has (should?) these mappings from oasis address to the thing that was hashed into the address, e.g. entity public key. but here we're adding an index to the entities table.

Oh I see. Yeah we can omit adding the index and rewrite the query as:

```
WITH entity AS (
                        SELECT address_data
                        FROM chain.address_preimages
                        WHERE address = $1::text
                )
                SELECT
                        v.epoch,
                        v.escrow_balance_active,
                        v.escrow_total_shares_active,
                        v.escrow_balance_debonding,
                        v.escrow_total_shares_debonding,
                        v.num_delegators
                FROM history.validators v
                WHERE v.id = (SELECT encode(address_data, 'base64') FROM entity)
                ORDER BY v.epoch DESC
                LIMIT 1000;
```

Which should work.

pro-wh:
o dang, good fine

Issue #908, opened 2025-02-19
Extend evm_tokens API endpoint with a type filter


buberdds:
In Explorer we would like to filter token list by a type:
![Image](https://github.com/user-attachments/assets/09362870-ea8b-4a84-9c54-0fe51a66785e)

API endpoint: `https://nexus.oasis.io/v1/{runtime}/evm_tokens`
filter values: '', 'ERC20', 'ERC721'

PR #914, opened 2025-02-27
api/runtime/evm_tokens: Support token-type filters

ptrus:
Fixes: https://github.com/oasisprotocol/nexus/issues/908


api/spec/v1.yaml
```diff
@@ -1073,6 +1073,13 @@ paths:
           description: |
             The field to sort the tokens by.
             If unset, the tokens will be sorted by number of holders.
+        - in: query
+          name: type
+          schema:
+            type: string
```
buberdds, line 1079: Can we use EvmTokenType here ?

ptrus: Good idea, updated!

PR #921, opened 2025-03-03
Andrew7234/staking rewards api

Andrew7234:
Adds an endpoint to fetch the staking rewards for a given account in a given calendar year.

Todo:
[] performance testing, especially on a large db with a full year's worth of data

Alternatives considered:
- injecting synthetic events for each staking reward (delegator x delegatee x epoch). I didn't like this because it bloats the events table. At the start of 2025 there were ~62300 unique delegations, which over the course of a year would lead to 62300 * 365 * 24 = ~545 million synthetic events per year, not including prior history or new delegations that are created. Open to revisiting this approach or exploring alternatives based on the results of the performance tests though.


analyzer/consensus/static/accounts.go
```diff
@@ -17,16 +17,32 @@ import (
 //go:embed accounts
 var genesisAccountsFs embed.FS
 
+//go:embed delegations
+var genesisDelegationsFs embed.FS
```
ptrus, line 21: would it make sense to move the new code into the `consnesus/static/delegations.go` file

analyzer/consensus/static/accounts.go
```diff
@@ -107,3 +165,27 @@ func QueueConsensusAccountsFirstActivity(batch *storage.QueryBatch, chainName co
 
 	return nil
 }
+
+func QueueConsensusDelegationsSnapshots(batch *storage.QueryBatch, chainName common.ChainName, logger *log.Logger) error {
```
ptrus, line 169: Can you please add a delegation/README.md file that explains how the delegation files were obtained

api/spec/v1.yaml
```diff
@@ -718,6 +718,38 @@ paths:
...
+          schema:
+            allOf: [$ref: '#/components/schemas/StakingAddress']
+          description: The staking address of the that is being delegated to.
+        - in: path
+          name: year
+          required: true
+          schema:
+            type: integer
+            format: uint64
+          description: The staking address of the that is being delegated to.
```
ptrus, line 742: Maybe mention that 2021 is the first supported year.

storage/client/client.go
```diff
@@ -1121,6 +1121,179 @@ func (c *StorageClient) DebondingDelegationsTo(ctx context.Context, address stak
 	return &ds, nil
 }
 
+// StakingRewards returns a list of consensus staking rewards for a given account.
+func (c *StorageClient) StakingRewards(ctx context.Context, address staking.Address, year uint64, p apiTypes.GetConsensusAccountsAddressStakingRewardsYearParams) (*StakingRewardList, error) {
+	// Validate year, find starting snapshot of account's delegations at year start
+	// Fetch all escrow events where delegator==address from that year, this includes staking reward disbursements which have newShares == 0
```
ptrus, line 1127: Seems like a lot of work needs to be done on this query. Could we compute this once for each past year and every account and and store in the DB? Via some background job? And then for the most recent year we can compute it daily and update the results.

PR #929, opened 2025-03-20
storage/db: Configuring max connection to the database

ptrus:
By default `pgxpool` uses a very conservative limit for maximum number of connections (`max(num_cpu's, 4)`). We make this configurable, and also bump the default to 10.

In production we probably want to bump this to something like 50 or 100.

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -63,6 +63,7 @@ type AnalysisConfig struct {
 	// They are instantiated if and only if analyzers are instantiated.
 	Helpers HelperList `koanf:"helpers"`
 
+	// Storage is the storage configuration for the analyzers.
 	Storage *StorageConfig `koanf:"storage"`
 }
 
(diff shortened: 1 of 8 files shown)


config/config.go
```diff
@@ -705,6 +707,15 @@ type StorageConfig struct {
...
 	WipeStorage bool `koanf:"DANGER__WIPE_STORAGE_ON_STARTUP"`
+
+	// Postgres is the postgres specific configuration to use.
+	Postgres *PostgresConfig `koanf:"postgres"`
+}
+
+// PostgresConfig is the postgres specific configuration to use when using the postgres backend.
+type PostgresConfig struct {
+	// MaxConnections is the maximum number of connections in the database connection pool.
+	MaxConnections *int32 `koanf:"max_connections"`
```
uscinski, line 718: To my understanding, this parameter can be configured via a connection string (specified in either key/value or URI format), such as:
`user=jack password=secret host=pg.example.com port=5432 dbname=mydb sslmode=verify-ca pool_max_conns=10`

Therefore, there's no need to expose it specifically via the Nexus' config.

See here for details:
- https://www.postgresql.org/docs/current/libpq-connect.html#LIBPQ-CONNSTRING
- https://github.com/jackc/pgx/blob/v4.18.3/pgxpool/pool.go#L305

ptrus: Hm you're right. Will remove it in a future release, to avoid duplication of possible configuration.

Issue #932, opened 2025-03-27
Env config provider does not parse list of addresses correctly


uscinski:
Nexus version v0.6.0+ added new configuration options to API Server that allow excluding a list of consensus circulating supply addresses. If this option is given through the configuration entry `server.consensus_circulating_supply_exclusions`, it works fine.

Due to potentially confidential nature of these addresses, on k8s deployment we might want to pass them instead via the respective environment variable that is read from a SealedSecret. This works when only one address is given, e.g.:

```bash
SERVER__CONSENSUS_CIRCULATING_SUPPLY_EXCLUSIONS=oasis1qrq7hgvv26003hy89klcmy3mnedrmyd7lvf0k6qn
```

But when given a list of addresses, it fails:

```bash
SERVER__CONSENSUS_CIRCULATING_SUPPLY_EXCLUSIONS=oasis1qrq7hgvv26003hy89klcmy3mnedrmyd7lvf0k6qn,oasis1qz9x0zpja6n25hc5242k2e60l6a7ke2zsq9cqrqz
```

API server errors out with the following log message:

```json
{"caller":"api.go:108","error":"server: invalid circulating supply reserved address: oasis1qrq7hgvv26003hy89klcmy3mnedrmyd7lvf0k6qn,oasis1qz9x0zpja6n25hc5242k2e60l6a7ke2zsq9cqrqz","level":"error","module":"init","msg":"init failed","ts":"2025-03-27T08:49:55.19273176Z"}
```

The reason is that unlike in the file-entry configuration, which is parsed with a YAML parser, the env provider in the `koanf` library parses everything as a string. Upon unmarshaling the configuration, it is coerced to type `[]string`, meaning that it contains just one element. There is no split on `,` or any other character.

The `koanf` docs state:

> You can also use the `env.ProviderWithValue` with a callback that supports mutating both the key and value to return types other than a string.

This implies that for any other type than string we should use `env.ProviderWithValue` and implement how both, keys and values are parsed.

PR #933, opened 2025-03-27
config: add support for array values via environment variables

ptrus:
Fixes: https://github.com/oasisprotocol/nexus/issues/932

This should be good enough since we only use env variables for a couple of config settings (none of those should have commas unless providing a list of values).

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -770,9 +770,16 @@ func initConfig(p koanf.Provider) (*Config, error) {
 	}
 
 	// Load environment variables and merge into the loaded config.
-	if err := k.Load(env.Provider("", ".", func(s string) string {
+	if err := k.Load(env.ProviderWithValue("", ".", func(key string, value string) (string, interface{}) {
 		// `__` is used as a hierarchy delimiter.
-		return strings.ReplaceAll(strings.ToLower(s), "__", ".")
+		key = strings.ReplaceAll(strings.ToLower(key), "__", ".")
+
+		// Support comma separated list for specific keys.
+		if key == "server.consensus_circulating_supply_exclusions" {
+			return key, strings.Split(value, ",")
+		}
+
+		return key, value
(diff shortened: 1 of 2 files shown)


config/config.go
```diff
@@ -770,9 +770,15 @@ func initConfig(p koanf.Provider) (*Config, error) {
 	}
 
 	// Load environment variables and merge into the loaded config.
-	if err := k.Load(env.Provider("", ".", func(s string) string {
+	if err := k.Load(env.ProviderWithValue("", ".", func(key string, value string) (string, interface{}) {
 		// `__` is used as a hierarchy delimiter.
-		return strings.ReplaceAll(strings.ToLower(s), "__", ".")
+		key = strings.ReplaceAll(strings.ToLower(key), "__", ".")
+
+		// If vlaue contains a comma, split it into a list.
```
uscinski, line 777:
```suggestion
		// If value contains a comma, split it into a list.
```

config/config.go
```diff
@@ -770,9 +770,15 @@ func initConfig(p koanf.Provider) (*Config, error) {
...
+		if strings.Contains(value, ",") {
```
uscinski, line 778: It would be safer to explicitly check for the key and only split, if the key is the one that should result in `[]string`.

It is not just theoretically unsafe. The other configuration option that we pass via env, is the database connection string. This one can contain commas, but should remain a string, not `[]string`, e.g.:

```text
postgresql://host1:123,host2:456/somedb?target_session_attrs=any&application_name=myapp
```

uscinski (commented):
While generic, its unsafe. See the comments.

config/config.go
ptrus: Updated

uscinski:
Tested with Docker image tag `latest-2025-03-27-gita620dadb`. It now works as expected.

PR #956, opened 2025-04-02
openapi: Clarify account escrow and debonding fields (for validators)

lukaw3d:
This still causes confusion. Thoughts on this wording?

### DIFF
--- a/api/spec/v1.yaml
+++ b/api/spec/v1.yaml
@@ -2562,10 +2562,10 @@ components:
           description: The available balance, in base units.
         escrow:
           allOf: [$ref: '#/components/schemas/TextBigInt']
-          description: The active escrow balance, in base units.
+          description: Amount delegated to this validator, in base units.
         debonding:
           allOf: [$ref: '#/components/schemas/TextBigInt']
-          description: The debonding escrow balance, in base units.
+          description: Amount debonding from this validator, in base units.
         delegations_balance:
           allOf: [$ref: '#/components/schemas/TextBigInt']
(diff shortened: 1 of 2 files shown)


api/spec/v1.yaml
```diff
@@ -2562,10 +2562,12 @@ components:
           description: The available balance, in base units.
         escrow:
           allOf: [$ref: '#/components/schemas/TextBigInt']
-          description: The active escrow balance, in base units.
+          description: |
+            Effectively only useful on validator accounts: active escrow balance, in base units.
```
ptrus, line 2566: Maybe 

> Amount delegated to the account, in base units. Effectively only useful on validator accounts.

and

> Debonding amount delegated to the account, in base units. Effectively only useful on validator accounts.

But what you did is fine as well.

ptrus (approved):
LGTM

Btw I know about the CI failures (started happening today)

>  batch response: This repository exceeded its LFS budget. The account responsible for the budget should increase it to restore access.

maybe we should switch away from git LFS for test artifacts, will consider fixes in the following days.

I can force-merge your PR.

ptrus:
Please rebase on main, I think the CI should start working then.

api/spec/v1.yaml
lukaw3d: How about:
Amount delegated to this validator, in base units.
Amount debonding from this validator, in base units.

ptrus: 👍

PR #972, opened 2025-04-11
storage/accounts: Move tx_count into materilized view

ptrus:


ptrus:
fixed in https://github.com/oasisprotocol/nexus/pull/975

PR #1073, opened 2025-06-27
analyzer/consensus: Record timing only for successful requests

ptrus:
This makes it consistent with other (runtime) block analyzers.

### DIFF
--- a/.changelog/1073.bugfix.md
+++ b/.changelog/1073.bugfix.md
@@ -1,0 +1,1 @@
+analyzer/consensus: Record timing only for successful requests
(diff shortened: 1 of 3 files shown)


uscinski (approved):
This looks good and more consistent now.

I think we should also have a counter of all requests, including successful and failed ones, if there isn't already one.

ptrus:
> I think we should also have a counter of all requests, including successful and failed ones, if there isn't already one.


Opened https://github.com/oasisprotocol/nexus/issues/1074

PR #1082, opened 2025-07-03
analyer/consensus: Support new BlockMeta version

ptrus:


coreapi/v24.0/consensus/cometbft/api/api.go
```diff
@@ -0,0 +1,70 @@
...
+//
+// It first tries to unmarshal into V1, and if that fails, it tries to
+// unmarshal into V2.
+//
+// We only try to unmarshal into the version of the metadata structure
+// starting at Oasis-Core Eden (V1), and into V2 (starting at #6235).
+// This may fail on blocks from an incompatible earlier version.
+func (b *BlockMeta) TryUnmarshal(data []byte) error {
+	// Try to unmarshal into V1 first.
+	if err := cbor.Unmarshal(data, &b); err == nil {
```
peternose, line 29: I would use `select` here as reader might not see `==`.

ptrus: Updated

PR #1167, opened 2025-09-25
Add Prometheus metrics for consensus block fetch statuses

### LINKED ISSUE #1074: Add prometheus metrics for block fetching success/failures
I think we should also have a counter of all requests, including successful and failed ones, if there isn't already one.

_Originally posted by @uscinski in https://github.com/oasisprotocol/nexus/pull/1073#pullrequestreview-2965784336_

jberci:
Closes #1074

### DIFF
--- a/.changelog/1074.internal.md
+++ b/.changelog/1074.internal.md
@@ -1,0 +1,1 @@
+Add consensus analyzer Prometheus metrics for block fetch statuses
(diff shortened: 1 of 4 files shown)


ptrus (commented):
Looks good. Let's also add this metric to analyzer/runtime/runtime.go

PR #1217, opened 2025-12-04
docs: Generate docs using @scalar/api-reference instead of redoc-cli

lukaw3d:


ptrus (approved):
api/README.md still references the removed `make docs-api` command. 

Thanks!

